PR #195, opened 2023-09-06
Unify logic for building liquidation orders

jayy04:


protocol/x/clob/keeper/process_operations_liquidations.go
```diff
@@ -11,6 +11,56 @@ import (
...
+// ValidateLiquidationOrderAgainstProposedMatch performs stateless validation of a liquidation order
+// against a proposed match.
+// An error is returned when
+//   - The CLOB pair IDs of the order and proposed match do not match.
+//   - The perpetual IDs of the order and proposed match do not match.
+//   - The size of the order is less than the size of the proposed match.
+//   - The side of the order and proposed match do not match.
+func (k Keeper) ValidateLiquidationOrderAgainstProposedMatch(
+	ctx sdk.Context,
+	order *types.LiquidationOrder,
```
jonfung-dydx, line 23: call validate on all the subaccount ids in the liquidation order and the liquidation match?

protocol/x/clob/keeper/process_operations_liquidations.go
```diff
@@ -11,6 +11,56 @@ import (
...
+	if order.MustGetLiquidatedPerpetualId() != proposedMatch.GetPerpetualId() {
+		return sdkerrors.Wrapf(
+			types.ErrClobPairAndPerpetualDoNotMatch,
+			"Order Perpetual ID: %v, Match Perpetual ID: %v",
+			order.MustGetLiquidatedPerpetualId(),
+			proposedMatch.GetPerpetualId(),
+		)
+	}
+
+	if order.GetBaseQuantums() != satypes.BaseQuantums(proposedMatch.TotalSize) {
```
jonfung-dydx, line 44: I think this is wrong right? this is checking equality instead of ` The size of the order is less than the size of the proposed match.`?

protocol/x/clob/keeper/process_operations_liquidations.go
jayy04: This should be done in validate basic, right? 

we do call this for all subaccount ids except for match.Liquidated. added

protocol/x/clob/keeper/process_operations_liquidations.go
jayy04: Comment is wrong - updated. let me know if that doesn't make sense

protocol/x/clob/keeper/process_operations.go
```diff
@@ -366,11 +366,16 @@ func (k Keeper) PersistMatchLiquidationToState(
 		)
 	}
 
-	takerOrder, err := k.ConstructTakerOrderFromMatchPerpetualLiquidation(ctx, matchLiquidation)
+	takerOrder, err := k.MaybeGetLiquidationOrder(ctx, matchLiquidation.Liquidated)
 	if err != nil {
 		return err
 	}
 
+	// Perform some validation on the liquidation order.
```
jonfung-dydx, line 374:
```suggestion
	// Perform stateless validation on the liquidation order.
```

jonfung-dydx (approved):

protocol/x/clob/keeper/liquidations.go
```diff
@@ -490,9 +490,19 @@ func (k Keeper) GetPerpetualPositionToLiquidate(
...
+	subaccountLiquidationInfo := k.GetSubaccountLiquidationInfo(ctx, subaccountId)
+
+	var perpetualPosition *satypes.PerpetualPosition
+
+	for _, position := range subaccount.PerpetualPositions {
+		if !subaccountLiquidationInfo.HasPerpetualBeenLiquidatedForSubaccount(position.PerpetualId) {
+			perpetualPosition = position
+			break
+		}
+	}
```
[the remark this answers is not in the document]
lucas-dydx: Mind noting this in a comment?

Also might be worth noting in the comment that the protocol always liquidates positions in the orders that they're opened.

protocol/x/clob/keeper/process_operations_liquidations.go
```diff
@@ -11,6 +11,56 @@ import (
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+// ValidateLiquidationOrderAgainstProposedMatch performs stateless validation of a liquidation order
+// against a proposed match.
```
lucas-dydx, line 15: Small suggestion: I think this function name is slightly clearer (when I first read this it sounded like the validation was being performed on the maker fills):

```suggestion
// ValidateLiquidationOrderAgainstProposedLiquidation performs stateless validation of a liquidation order
// against a proposed liquidation.
```

lucas-dydx (approved):

protocol/x/clob/abci.go
```diff
@@ -193,12 +194,18 @@ func PrepareCheckState(
...
 		if err != nil {
+			// Subaccount might not always be liquidatable since liquidation daemon runs
+			// in a separate goroutine and is not always in sync with the application.
+			// Therefore, if subaccount is not liquidatable, continue.
+			if errors.Is(err, types.ErrSubaccountNotLiquidatable) {
+				continue
+			}
+
+			// Panic on unexpected errors.
 			panic(err)
```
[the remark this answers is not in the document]
jayy04: Double checked the code quickly. I think the only errors that can be returned as part of this flow is `ErrFokOrderCouldNotBeFullyFilled` and `ErrPostOnlyWouldCrossMakerOrder`, both of which are not applicable for liquidations. I think panic'ing here is fine given the current implementation. Let me know if I am missing anything. 

I think whether or not we should halt the chain after unsuccessful liquidation/deleveraging should be addressed as part of emergency powers (M6+)?

protocol/x/clob/abci.go
```diff
@@ -193,12 +194,18 @@ func PrepareCheckState(
 		// If attempting to liquidate a subaccount returns an error, panic.
 		liquidationOrder, err := keeper.MaybeGetLiquidationOrder(ctx, subaccountId)
 		if err != nil {
+			// Subaccount might not always be liquidatable since liquidation daemon runs
+			// in a separate goroutine and is not always in sync with the application.
+			// Therefore, if subaccount is not liquidatable, continue.
+			if errors.Is(err, types.ErrSubaccountNotLiquidatable) {
+				continue
```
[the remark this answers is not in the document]
jayy04: we already have that as part of `MaybeGetLiquidationOrder`. I think it makes more sense here so moved it here.

protocol/x/clob/keeper/liquidations.go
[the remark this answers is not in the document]
jayy04: Added comment. 

note that positions are always sorted by id, so the position with smallest id will get picked first (not in the orders that they 'are opened).

PR #198, opened 2023-09-06
[CORE-322] Escalate log level for missing smoothed prices, index prices, ~10s after protocol start

clemire:
Missing index and smoothed prices are expected for the first few blocks due to the cold start issue with the price daemon. However, we want to escalate this log level after some time has passed in order to capture legitimate error conditions, such as a nonoperational daemon.

Add a map of market to creation timestamp and log these events as errors after a 10s startup grace period for the protocol.

Note: most of this PR is updating unit tests to mock the price keeper's mock time provider response because market creation now requires adding the market creation timestamp to the keeper's timestamp map.


protocol/x/prices/keeper/update_price.go
```diff
@@ -63,8 +63,13 @@ func (k Keeper) GetValidMarketPriceUpdates(
 		// Skip proposal logic in the event of invalid inputs, which is only likely to occur around network genesis.
 		if !indexPriceExists {
 			metrics.IncrCountMetricWithLabels(types.ModuleName, metrics.IndexPriceDoesNotExist, marketMetricsLabel)
-			// TODO(CORE-322): Conditionally escalate log level to error 5s after genesis/restart.
-			k.Logger(ctx).Info(fmt.Sprintf("Index price for market (%v) does not exist", marketId))
+			// Conditionally escalate log level to error 5s after genesis/restart. We expect that it may take a few
```
roy-dydx, line 66: `5s` I think this comment is outdated

protocol/x/prices/keeper/update_price.go
```diff
@@ -80,8 +85,14 @@ func (k Keeper) GetValidMarketPriceUpdates(
 		// We generally expect to have a smoothed price history for each market, except during the first few blocks
 		// after network genesis or a network restart. In this scenario, we use the index price as the smoothed price.
 		if len(historicalSmoothedPrices) == 0 {
-			// TODO(CORE-322): Conditionally escalate log level to error 5s after genesis/restart.
-			k.Logger(ctx).Info(fmt.Sprintf("Smoothed price for market (%v) does not exist", marketId))
+			// Conditionally escalate log level to error 5s after genesis/restart. We expect that there will be a delay
```
roy-dydx, line 88: same here

roy-dydx (approved):

PR #201, opened 2023-09-07
remove redundant validation

jayy04:


protocol/x/clob/keeper/liquidations.go
```diff
@@ -35,19 +35,21 @@ func (k Keeper) MaybeGetLiquidationOrder(
 	}
 
 	// The subaccount is liquidatable. Get the perpetual position and position size to liquidate.
-	clobPair, positionSizeBig, err := k.GetPerpetualPositionToLiquidate(ctx, subaccountId)
```
jayy04, line 38: Unrelated to this PR - I always felt like returning clob pair here is weird, since the function is called `GetPerpetual...`.

I think longer term, we can deprecate clob pair id in the match liquidation proto (it was added before we have the in memory mapping between clob pair and perpetual). so updating here to make our lives easier in the future

protocol/x/clob/types/liquidations_keeper.go
```diff
@@ -71,7 +71,7 @@ type LiquidationsKeeper interface {
 		ctx sdk.Context,
 		subaccountId satypes.SubaccountId,
 	) (
-		clobPair ClobPair,
+		perpetualID uint32,
```
jonfung-dydx, line 74:
```suggestion
		perpetualId uint32,
```

protocol/x/clob/keeper/liquidations.go
```diff
@@ -35,19 +35,21 @@ func (k Keeper) MaybeGetLiquidationOrder(
...
-	perpetualId := clobPair.GetPerpetualClobMetadata().PerpetualId
 
 	// Get the fillable price of the liquidation order in subticks.
 	deltaQuantumsBig := positionSizeBig.Neg(positionSizeBig)
 	fillablePriceRat, err := k.GetFillablePrice(ctx, subaccountId, perpetualId, deltaQuantumsBig)
 	if err != nil {
 		return nil, err
 	}
+
+	// Calculaet the fillable price.
```
jonfung-dydx, line 50:
```suggestion
	// Calculate the fillable price.
```

jakob-dydx (approved):

jonfung-dydx (approved):

PR #211, opened 2023-09-07
update bridge contract address

tqin7:
This new contract interacts with the mock token contract [here](https://sepolia.etherscan.io/address/0x6d5bb505a4f85c10b122ccc36e30f57e2b86a291#code) to simulate the real workflow. Previous bridge contract does not actually reduce user balance.


protocol/testing/genesis.sh
```diff
@@ -19,7 +19,7 @@ NATIVE_TOKEN="dv4tnt" # public testnet token
 DEFAULT_SUBACCOUNT_QUOTE_BALANCE=100000000000000000
 DEFAULT_SUBACCOUNT_QUOTE_BALANCE_FAUCET=900000000000000000
 ETH_CHAIN_ID=11155111 # sepolia
-ETH_BRIDGE_ADDRESS="0xEf01c3A30eB57c91c40C52E996d29c202ae72193"
+ETH_BRIDGE_ADDRESS="0xcca9D5f0a3c58b6f02BD0985fC7F9420EA24C1f0"
```
teddyding, line 22: Unrelated to this PR, could you put `TODO(GENESIS)` over any genesis setup that needs to be updated before this is used for a real network. Similar to https://github.com/dydxprotocol/v4-chain/blob/ea5657e73fe537416f30bf6804232d301dec85ad/protocol/cmd/dydxprotocold/cmd/config.go#L15

teddyding: Could you add block explorer link to the contract?

teddyding (approved):

PR #212, opened 2023-09-07
[CLOB-827] introudce weighted randomness to liquidation order selection

jayy04:


protocol/lib/random.go
```diff
@@ -4,8 +4,43 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+
+	exprand "golang.org/x/exp/rand"
```
jayy04, line 8: unfortunately `gonum` uses a different `rand` (not math/rand)

protocol/lib/random_test.go
```diff
@@ -81,3 +82,70 @@ func TestRandomBytesBetween_InvalidInputs(t *testing.T) {
...
+		require.NoError(t, err)
+		freq[sample[0]]++
+	}
+
+	exp := []float64{100_000, 200_000, 300_000, 400_000}
+
+	// Check that this is within statistical expectations.
+	require.Less(
+		t,
+		chi2(freq, exp),
```
jayy04, line 102: not sure if we need to do chi2 here, or just compare with the actual expected values. let me know if anyone has a preference

PR #213, opened 2023-09-08
remove redundant interface registration calls in codec.go

jakob-dydx:
- removes redundant calls to `RegisterImplementations` from module.go for many modules. RegisterImplementations is called for every route configured on the MsgServer with `msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)`, so the manual RegisterImplementations calls are redundant
- removes unneeded `RegisterCodec` logic for unused legacy codec


protocol/x/bridge/module_test.go
```diff
@@ -118,7 +98,7 @@ func TestAppModuleBasic_RegisterInterfaces(t *testing.T) {
 	mockRegistry.On("RegisterImplementations", (*sdk.Msg)(nil), mock.Anything).Return()
 	mockRegistry.On("RegisterImplementations", (*tx.MsgResponse)(nil), mock.Anything).Return()
 	am.RegisterInterfaces(mockRegistry)
-	mockRegistry.AssertNumberOfCalls(t, "RegisterImplementations", 15)
```
lcwik, line 121: Can we move away from the `mockRegistry` and instead using `EnsureRegistered()` or one of the other equivalent methods to make sure that all the types we want to have registered are registered here and elsewhere?

This would allow us to remove the `InterfaceRegistry` from `mock-gen` as well.

In Cosmos 0.50.0 you can't mock the registry anymore since it has a private field that is accessed outside of the type.

Feel free to defer doing this.

protocol/x/assets/types/codec.go
```diff
@@ -3,21 +3,12 @@ package types
...
-}
+func RegisterCodec(cdc *codec.LegacyAmino) {}
 
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
-	// this line is used by starport scaffolding # 3
-
 	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
 }
-
-var (
```
ttl33, line 20: Hmm, this might be required when trying to use `x/authz` later (which we prev discussed might be of use). So I don't think we should remove these just yet

See [example](https://github.com/osmosis-labs/osmosis/blob/013cd96fdd40d0ddcff3351240492bbc3a3c10ed/x/lockup/types/codec.go#L38).

protocol/x/clob/types/codec.go
```diff
@@ -3,29 +3,12 @@ package types
...
-func RegisterCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&MsgProposedOperations{}, "clob/ProposedOperations", nil)
-	cdc.RegisterConcrete(&MsgPlaceOrder{}, "clob/PlaceOrder", nil)
-	cdc.RegisterConcrete(&MsgCancelOrder{}, "clob/CancelOrder", nil)
-	// this line is used by starport scaffolding # 2
-}
+func RegisterCodec(cdc *codec.LegacyAmino) {}
 
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
-	registry.RegisterImplementations((*sdk.Msg)(nil),
```
ttl33, line 18: Dumb question: I am having a hard time following why removing `RegisterImplementations` is okay. Can you point me to where in `RegisterMsgServiceDesc` we register the module msg types?

protocol/x/clob/types/codec.go
```diff
@@ -3,29 +3,12 @@ package types
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/msgservice"
 )
 
-func RegisterCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&MsgProposedOperations{}, "clob/ProposedOperations", nil)
```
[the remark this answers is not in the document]
jakob-dydx: From my understanding they do something very similar, but RegisterConcrete is for the legacy amino codec which appears to be unused in our application

protocol/x/clob/types/codec.go
jakob-dydx: https://github.com/dydxprotocol/cosmos-sdk/blob/6d0ef15b0a45078584f629ba14cb38ebe0fd0b81/types/msgservice/msg_service.go#L53

We use this method to iterate over the proto generated descriptor of the msg server and it automatically registers all of the messages for us (both the request and response messages)

protocol/x/assets/types/codec.go
jakob-dydx: Hmm, trying to dig into this. I don't currently understand where Amino is or is not used. Seems like it isn't used in our current application, but that certain modules might still rely on their amino codec being configured?

ttl33: Yah, I am not 100% sure if Amino is fully unused either. Based on this comment [here](https://github.com/dydxprotocol/v4-chain/blob/68cea103835050b218a3672d02bf6389c331e23f/protocol/lib/encoding/encoding.go#L29), there might still be some use case. 

This feels like a P3 issue/cleanup task that can be deferred to M6+. Can we just create a ticket to track this rather than spending more time on this now?

jakob-dydx: cut an issue to track https://linear.app/dydx/issue/CORE-556/determine-if-legacy-amino-codec-is-required-for-xmodules

protocol/x/bridge/module_test.go
jakob-dydx: will capture in this issue: https://linear.app/dydx/issue/CORE-557/remove-mockregistry-from-module-testgo

protocol/x/bridge/types/codec.go
```diff
@@ -3,35 +3,12 @@ package types
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/msgservice"
 )
 
-func RegisterCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&MsgAcknowledgeBridges{}, "bridge/AcknowledgeBridges", nil)
```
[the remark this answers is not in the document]
jakob-dydx: the RegisterConcrete calls that we have are incomplete. Some modules have them, some don't. The ones that do only call RegisterConcrete for some of the Msgs because they were not updated when new Msgs were added. Don't think it is really important to keep these tbh since they are in a weird incomplete state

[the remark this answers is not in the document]
ttl33: Okay, can you share which modules + msgs are missing the `RegisterConcrete`? 

Trying to better understand if they are not registered intentionally.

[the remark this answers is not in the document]
jakob-dydx: modules missing RegisterConcrete for message on `main`

clob: missing all of the new messages we have added for governance
vest: missing `MsgDeleteVestEntry`
prices: missing `MsgCreateOracleMarket`
blocktime: missing `MsgUpdateDowntimeParams` and `MsgIsDelayedBlock`
perpetuals: missing `MsgCreatePerpetual`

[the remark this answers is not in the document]
ttl33: I see. I don't see an intentional pattern. For example, `vest` should register both `MsgDeleteVestEntry` + `MsgSetVestEntry`.

I agree that we can delete these for now and add them back later

protocol/x/bridge/module_test.go
```diff
@@ -78,16 +78,6 @@ func TestAppModuleBasic_RegisterCodec(t *testing.T) {
 	var buf bytes.Buffer
 	err := cdc.Amino.PrintTypes(&buf)
 	require.NoError(t, err)
```
ttl33, line 80: For these tests, should we validate that `buf.String()` is empty or something? Not sure if `buf.String()` has other stuff

protocol/x/prices/types/codec_test.go
ttl33, line 1: did we mean to delete this entire file?

jakob-dydx: Yes, these duplicate test coverage from module_test.go

protocol/x/bridge/module_test.go
jakob-dydx: it isn't empty, it is formatted like this when it is empty
```
| Type | Name | Prefix | Length | Notes |
| ---- | ---- | ------ | ----- | ------ |
```

ttl33 (approved):

PR #214, opened 2023-09-08
[CORE-542] - Update delaymsg module to execute messages with a cached context

clemire:
[This issue](https://informalsystems.atlassian.net/wiki/spaces/DQ2/pages/107937920/Dispatching+of+delayed+messages+should+be+done+over+cached+context) came up as a critical issue in our audit discussions, and I agree, so here is a proposed fix.

Summary: delaymsg module message execution happens in the EndBlocker context. If message execution fails, partial state transitions are captured and persisted to state. This is inconsistent with how governance and other solutions, like the pattern Mirel suggested for the fix, works now.

Solution:
- construct a cachedCtx, execute each message independently with it's own cached ctx, and write the context to state if the message succeeds.


protocol/lib/abci/util.go
```diff
@@ -0,0 +1,18 @@
+package abci
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// RunCached wraps a function with a cache context and writes the cache to the context if the
+// function call succeeds. If the function call fails, the cache is discarded.
+func RunCached(c sdk.Context, f func(sdk.Context) error) error {
```
teddyding, line 9: Should we also do the following, similar to Axelar [RunCached](https://github.com/axelarnetwork/axelar-core/blob/e9287fd765f7e501238ab85395ba874d5d89b52c/utils/abci.go#L19) function:
- Add recovery. Since  x/delay-msg `DispatchMessage` happens in the `EndBlocker`, a panic in `f(ctx)`, if not recovered leads to consensus failure. We should also log error in this case, similar to the behavior when `runTx` panic is recoverred

clemire: I added panic recovery and check in the test cases that events are added to the context's event manager only when the function succeeeds. PTAL!

protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -256,12 +272,17 @@ func TestDispatchMessageForBlock_Mixed(t *testing.T) {
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
 			k := &mocks.DelayMsgKeeper{}
-			ctx := sdktest.NewContextWithBlockHeightAndTime(0, time.Now())
```
teddyding, line 259: Nit: this function name should be `TestDispatchMessagesForBlock_Mixed`? (Missing an `s`)

protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -53,7 +52,8 @@ func TestDispatchMessagesForBlock(t *testing.T) {
 	require.Equal(t, []uint32{0, 1, 2}, blockMessageIds.Ids)
 
 	// Mock the bridge keeper methods called by the bridge msg server.
-	bridgeKeeper.On("CompleteBridge", ctx, mock.Anything).Return(nil).Times(len(constants.AllMsgs))
+	bridgeKeeper.On("CompleteBridge", mock.AnythingOfType("types.Context"), mock.Anything).
```
teddyding, line 55: Do we have a test somewhere in this file that tests of panic (e.g. `CompleteBridge`) panics?

protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -256,12 +272,17 @@ func TestDispatchMessageForBlock_Mixed(t *testing.T) {
...
+
 			tc.setupMocks(t, ctx, k)
 
 			keeper.DispatchMessagesForBlock(k, ctx)
 
-			k.AssertExpectations(t)
+			mock.AssertExpectationsForObjects(t, k, ms, cms)
 		})
 	}
 }
```
teddyding, line 288: Non-blocking for this PR: I was looking for some way where we can verify that the messages are executed atomically (e.g. state change 1 succeeds, state change 2 fails, no state changes are commited). But it doesn't look like our current test setup support this, given `DispatchMessagesForBlocks` are done through mocks

Can we create a ticket for this? Also do we have any integration test plan that can test this?

protocol/x/delaymsg/keeper/dispatch_test.go
clemire: No, but there is a unit tests for RunCached that validates that panics are not propagated.

protocol/x/delaymsg/keeper/dispatch_test.go
clemire: That's not a part of the delaymsg contract at this time. we allow for messages to come in from any module and promise only that we will dispatch it - not that messages for a block will be executed in chunks or all-or-nothing per block. Each block could have multiple message authors contributing. Supporting this feature would be a significant re-design of the module itself.

protocol/lib/abci/util_test.go
```diff
@@ -0,0 +1,70 @@
...
+			},
+		},
+		"failure": {
+			f: func(ctx sdk.Context) error {
+				ctx.EventManager().EmitEvent(testEvent)
+				return fmt.Errorf("failure")
+			},
+			expectedError: fmt.Errorf("failure"),
+		},
+		"panic": {
```
clemire, line 33: @teddyding please see this unit test to validate that panics do not propagate upwards.

protocol/x/delaymsg/keeper/dispatch_test.go
clemire: Added one just in case!

protocol/x/delaymsg/keeper/dispatch_test.go
clemire: My understanding is that this is part of the contract for cachedCtx itself - that it commits all-or-nothing and does not affect the underlying store unless the write occurs. I can see the value of a test, but I assumed that this would be done on the cachedCtx functionality itself, or perhaps we could add coverage to `RunCached` - it's not really a delaymsg feature, so to me, mocking at the delay-msg level and verifying more specific behavior up the callstack is fine.

Is there something specific that you had in mind or a reason you were worried about coverage here?

clemire: https://linear.app/dydx/issue/CORE-570/delay-msg-testing-for-message-execution-atomicity

clemire: No current integration test plan for this. I'm unsure of any specific examples of messages dispatched by delay-msg that we could finagle to fail on a partial state transition, and may need to work with the rest of the team to come up with a good example.

clemire: I think that I could write an e2e test that includes multiple complete bridge events scheduled for the same block in delaymsg genesis and verifies that the bridge module account balance only decrements by the amounts included in the valid messages?

teddyding: Do you mean multiple `MsgCompleteBridge` messages? That still wouldn't test that each atomic message is executed atomically right?
Do we have any plan for supporting a list of `sdk.Msg` in a single `MsgDelayMessage` (similar to DeliverTx and governance executing multiple messages atomically)? That may be a good time to run some test for this.

clemire: No, we don't currently have a plan for that, as far as I'm aware.

clemire: Filed [this](https://linear.app/dydx/issue/CORE-576/evaluate-delay-msg-upgrade-to-handle-multiple-msgs) and slacked it to Tian.

teddyding (approved):

PR #220, opened 2023-09-08
[CLOB-867] make unfilled liquidations not count towards MaxLiquidationOrdersPerBlock

jayy04:


protocol/x/clob/abci.go
```diff
@@ -220,18 +220,24 @@ func PrepareCheckState(
...
+					liquidationOrders[i],
 					err,
 				),
 			)
 			panic(err)
 		}
+
+		// Keep a count of partially and fully filled liquidations for this block.
+		if optimisticallyFilledQuantums > 0 {
+			numFilledLiquidations++
```
lucas-dydx, line 239: Do we stat this anywhere? Should we stat it?

lucas-dydx: Should we also stat number of unfilled liquidations?

jonfung-dydx (approved):

jakob-dydx (approved):

PR #222, opened 2023-09-08
[CORE-550] reduce complete bridge delay in non-prod environments

tqin7:
for testing in local, dev, and staging envs, reduced delay to
- 30 blocks in localnet
- 600 blocks in dev and staging

### DIFF
--- a/protocol/testing/genesis.sh
+++ b/protocol/testing/genesis.sh
@@ -1483,7 +1483,7 @@ function update_genesis_use_test_volatile_market() {
 }
 
 # Modify the genesis file with reduced complete bridge delay (for testing in non-prod envs).
-update_genesis_reduced_complete_bridge_delay() {
+update_genesis_complete_bridge_delay() {
 	GENESIS=$1/genesis.json
 
 	# Reduce complete bridge delay to 600 blocks.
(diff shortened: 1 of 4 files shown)


protocol/testing/genesis.sh
```diff
@@ -1481,3 +1481,11 @@ function update_genesis_use_test_volatile_market() {
 	dasel put -t int -f "$GENESIS" '.app_state.clob.clob_pairs.last().subticks_per_tick' -v '1000000000'
 	dasel put -t int -f "$GENESIS" '.app_state.clob.clob_pairs.last().quantum_conversion_exponent' -v '-8'
 }
+
+# Modify the genesis file with reduced complete bridge delay (for testing in non-prod envs).
+update_genesis_reduced_complete_bridge_delay() {
```
clemire, line 1486: nit: this doesn't have to be a reduction, I think we could call it `update_genesis_complete_bridge_delay` or something.

clemire (approved):

protocol/testing/genesis.sh
tqin7: good call!

PR #230, opened 2023-09-11
[IND-391] Add restricted countries blocking logic.

vincentwschau:
Use Cloudflare headers to determine whether to return 403s in `comlink` and whether to terminate the connection request in `socks`.

### DIFF
--- a/indexer/services/comlink/src/config.ts
+++ b/indexer/services/comlink/src/config.ts
@@ -45,7 +45,7 @@ export const configSchema = {
 
   // Geo-blocking
   RESTRICTED_COUNTRIES: parseString({
-    default: '',
+    default: '', // comma de-limited
   }),
 };
 
--- a/indexer/services/comlink/src/config.ts
+++ b/indexer/services/comlink/src/config.ts
@@ -42,6 +42,11 @@ export const configSchema = {
   // point
   RATE_LIMIT_GET_POINTS: parseInteger({ default: 100 }),
   RATE_LIMIT_GET_DURATION_SECONDS: parseInteger({ default: 10 }), // 100 requests / 10 seconds
+
+  // Geo-blocking
+  RESTRICTED_COUNTRIES: parseString({
+    default: '',
+  }),
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/indexer/services/comlink/src/lib/restrict-countries.ts
+++ b/indexer/services/comlink/src/lib/restrict-countries.ts
@@ -1,0 +1,42 @@
+import { stats } from '@dydxprotocol-indexer/base';
+import express from 'express';
+
+import config from '../config';
+import { create4xxResponse } from './helpers';
+import { isRestrictedCountry } from './utils';
+
+/**
+ * Return an error code for users that access the API from a restricted country
+ */
+export function rejectRestrictedCountries(
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction,
+) {
+  const {
+    'cf-ipcountry': ipCountry,
+  } = req.headers as {
+    'cf-ipcountry'?: string,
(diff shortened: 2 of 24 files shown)


indexer/services/comlink/src/config.ts
```diff
@@ -42,6 +42,11 @@ export const configSchema = {
   // point
   RATE_LIMIT_GET_POINTS: parseInteger({ default: 100 }),
   RATE_LIMIT_GET_DURATION_SECONDS: parseInteger({ default: 10 }), // 100 requests / 10 seconds
+
+  // Geo-blocking
+  RESTRICTED_COUNTRIES: parseString({
```
Christopher-Li, line 47: How do you imagine this is stored? Comma separated list of strings? Can you include in comments

indexer/services/comlink/src/lib/restrict-countries.ts
```diff
@@ -0,0 +1,42 @@
...
+export function rejectRestrictedCountries(
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction,
+) {
+  const {
+    'cf-ipcountry': ipCountry,
+  } = req.headers as {
+    'cf-ipcountry'?: string,
+  };
```
Christopher-Li, line 20: this type definition is super weird, im assuming because js doesn't want hyphens in a field?

indexer/services/socks/__tests__/websocket/index.test.ts
```diff
@@ -60,9 +63,15 @@ describe('Index', () => {
     mockSub = new Subscriptions();
     invalidMsgHandlerSpy = jest.spyOn(InvalidMessageHandler.prototype, 'handleInvalidMessage');
     pingHandlerSpy = jest.spyOn(PingHandler.prototype, 'handlePing');
+    defaultRestricteCountries = config.RESTRICTED_COUNTRIES;
```
Christopher-Li, line 66: why do you redefine this at the start of each test? Seems unnecessary instead of defining it as a const and initializing to `config.RESTRICTED_COUNTRIES` on line 35

indexer/services/socks/src/config.ts
```diff
@@ -45,6 +45,11 @@ export const configSchema = {
   COMLINK_URL: parseString(),
   AXIOS_TIMEOUT_MS: parseInteger({ default: 5000 }), // 5 seconds
   INITIAL_GET_TIMEOUT_MS: parseInteger({ default: 20_000 }), // 20 seconds
+
+  // Restrictions
+  RESTRICTED_COUNTRIES: parseString({
```
Christopher-Li, line 50: why can't socks just depend on comlink for the RESTRICTED_COUNTRIES to be blocked there?

Christopher-Li (commented):
Why aren't we doing this blocking at the cloudflare level? I thought that's what we did in v3

indexer/services/comlink/src/lib/restrict-countries.ts
vincentwschau: Correct.

indexer/services/socks/src/config.ts
vincentwschau: `socks` queries `comlink` for initial data payload, and the origination of that request wouldn't be from a geoblocked region.

vincentwschau:
>Why aren't we doing this blocking at the cloudflare level? I thought that's what we did in v3

My understanding was we had middleware for blocking requests from restricted geography and used Cloudflare for blocking ofac sanctioned geography in v3.

indexer/services/socks/__tests__/websocket/index.test.ts
vincentwschau: Moved.

PR #231, opened 2023-09-11
[IND-372] Add mock /screen endpoint.

vincentwschau:
Add a mock /screen endpoint that returns `true` if the input argument ends in between 'a' to 'm', and false otherwise for testing purposes.

### DIFF
--- a/indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
@@ -22,7 +22,8 @@ class ComplianceController extends Controller {
     @Query() address: string,
   ): ComplianceResponse {
     // TODO(IND-372): Add logic to either use cached data or query provider
-    // Dummy logic for front-end testing
+    // Dummy logic for front-end testing, returns true if the address ends in a letter between
+    // 'a' and 'm'
     if (
       address.charCodeAt(address.length - 1) > 'a'.charCodeAt(0) &&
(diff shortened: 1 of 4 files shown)


indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
```diff
@@ -0,0 +1,82 @@
...
+  @Get('/')
+  screen(
+    @Query() address: string,
+  ): ComplianceResponse {
+    // TODO(IND-372): Add logic to either use cached data or query provider
+    // Dummy logic for front-end testing
+    if (
+      address.charCodeAt(address.length - 1) > 'a'.charCodeAt(0) &&
+      address.charCodeAt(address.length - 1) < 'm'.charCodeAt(0)
+    ) {
```
Christopher-Li, line 29: nit: logic is a little confusing, which you elaborated on in git commit message, could you include that in a comment here?

Christopher-Li (approved):

indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
```diff
@@ -0,0 +1,83 @@
...
+    } catch (error) {
+      return handleControllerError(
+        'ComplianceController GET /',
+        'Compliance error',
+        error,
+        res,
+      );
+    } finally {
+      stats.timing(
+        `${config.SERVICE_NAME}.${controllerName}.get_latest_block_height.timing`,
```
dydxwill, line 76: wrong stat name?

PR #234, opened 2023-09-12
remove extra logging file

BrendanChou:


ttl33 (approved):
I don't have a strong opinion on this. I think it's fine to keep things as is since we may need to re-use the `SourceModuleKey` in other places, but it can be broken up into a separate place later on.

Will defer to Crystal

clemire (approved):

PR #235, opened 2023-09-12
Make Median functions more generic

BrendanChou:


protocol/lib/math.go
```diff
@@ -63,14 +63,6 @@ func Min[T constraints.Ordered](x, y T) T {
 	return x
 }
 
-func MaxUint32(x, y uint32) uint32 {
-	if x < y {
-		return y
-	}
-
-	return x
-}
```
BrendanChou, line 72: removed this function as well since there is already a generic `Max` function

clemire (approved):

PR #236, opened 2023-09-12
remove medianizer

BrendanChou:
There is no need to define a `Medianizer` struct. Instead, we just pass around function definitions. In this case, pass a function that "resolves" the value. Usually this will be `lib.Median[uint64]` but we can pass in a different function during testing if needed. I have added that function (that returns an error value) to `testutil`.

Benefits:
- Reduces codebase size
- Redefines the medianizer/resolver type in the pricefeed daemon code instead of in the generic `lib/` code

Note: another possible name could be `reducer`


protocol/daemons/pricefeed/client/price_function/bitstamp/bitstamp_test.go
```diff
@@ -39,11 +36,11 @@ var (
 
 // Test response strings.
 var (
-	btcTicker = `{"timestamp": "1686600672", "open": "25940", "high": "26209", "low": "25634", "last": "25846", 
-		"volume": "1903.95560640", "vwap": "25868", "bid": "25841", "ask": "25842", "open_24": "26183", 
+	btcTicker = `{"timestamp": "1686600672", "open": "25940", "high": "26209", "low": "25634", "last": "25846",
```
clemire, line 39: 👍🏻

clemire (approved):

PR #238, opened 2023-09-12
[CORE-537] Panic and halt protocol on non-responsive daemons.

clemire:
This PR introduces an UpdateMonitor struct that attaches to the daemon server and monitors all running daemons to emit a response at a certain frequency. If the daemon does not meet the frequency requirement of the update, the updateMonitor panics and halts the protocol.

An alternative implementation was discussed with @lcwik to have the protocol throw errors in propose and process handlers, but Taehoon pointed out that this degrades network performance if a sufficient # of validators are experiencing daemon issues, so we're preferring the panic approach from the standpoint of removing validators that cannot contribute meaningfully to consensus, since daemon operation is required for proposing and validation (almost?) all blocks.

Attempting to clean up after the update monitor revealed that go routines spawned within `app.New` are not actually properly cleaned up after test case runs. In this case, because the implementation of this panic is done with a timer that invokes a function if not reset, we were seeing spurious panics at the end of test cases. In order to fix this, we added a global map that registers `Stoppable` services under an identifier unique to a particular test case. The uuid was decided to be that test's grpc address, since the network package generates unique grpc addresses for each test case, and this piece of unique per-test metadata is available in `app.New`, which has no references to an `sdk.Context`.

All cli tests that require daemons to operate have been updated to call a cleanup method that touches this global map. In addition, the default test config was updated to disable all daemons, and some tests that don't use daemons were updated to disable them.

Update:
- tested this on localnet, no issues.
- put a 1 minute sleep in the liquidations daemon on startup and tested this with a localnet build - the protocol correctly halts within a couple of minutes of network start.


protocol/app/stoppable/stoppable_test.go
```diff
@@ -0,0 +1,41 @@
+package stoppable
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+type MockStoppable struct {
```
lcwik, line 8: nit: `MockStoppable` -> `FakeStoppable` or `TestStoppable`. Lets reserve the word mock for things that are actually generated code that have the typical mock methods.

protocol/app/stoppable/stoppable_test.go
```diff
@@ -0,0 +1,41 @@
...
+	StopCalled bool
```
lcwik, line 9:
```suggestion
	stopCalled bool
```

protocol/app/stoppable/stoppable_test.go
```diff
@@ -0,0 +1,41 @@
+package stoppable
```
lcwik, line 1:
```suggestion
package stoppable_test
```

protocol/daemons/server/server.go
```diff
@@ -22,32 +27,73 @@ type Server struct {
...
 
+	updateMonitor *types.UpdateMonitor
+
 	BridgeServer
 	PriceFeedServer
 	LiquidationServer
 }
 
 // NewServer creates a single gRPC server that's shared across multiple daemons for communication.
+// cosmosGrpcServiceAddress is the address of the Cosmos gRPC query service. This address is used to uniquely
```
lcwik, line 38: mismatch between comment `cosmosGrpcServiceAddress` and parameter name `uniqueTestIdentifier`

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+
+// UpdateMonitor monitors the update frequency of daemon services. If a daemon service does not respond within
+// the maximum acceptable update delay set when the daemon is registered, the monitor will panic and halt the
+// protocol. This was judged to be the best solution for network performance because it prevents any validator from
+// interacting with the network at all if a daemon service is not responding.
+type UpdateMonitor struct {
+	// serviceToUpdateMetadata maps daemon service names to their update metadata.
+	serviceToUpdateMetadata map[string]updateMetadata
+	stopped                 bool
+	sync.Mutex
```
lcwik, line 22: I don't think we want to expose `Lock`/`Unlock` and other `sync.Mutex` methods on the API surface of `UpdateMonitor`

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+	defer ufm.Unlock()
+
+	// Don't register additional daemon services if the monitor has already been stopped.
+	// This could be a concern for short-running integration test cases, where the network
+	// stops before all daemon services have been registered.
+	if ufm.stopped {
+		return
+	}
+
+	ufm.serviceToUpdateMetadata[service] = updateMetadata{
```
lcwik, line 45: Check that service is not being double registered

protocol/daemons/server/types/update_monitor_test.go
```diff
@@ -0,0 +1,31 @@
...
+	require.NoError(t, ufm.RegisterValidResponse("test-service"))
+	time.Sleep(80 * time.Millisecond)
+	require.NoError(t, ufm.RegisterValidResponse("test-service"))
+	time.Sleep(80 * time.Millisecond)
+	require.NoError(t, ufm.RegisterValidResponse("test-service"))
+	time.Sleep(80 * time.Millisecond)
+	ufm.Stop()
+}
+
+// This test is disabled because the panic is not recoverable, since it's thrown
```
lcwik, line 24: You can make the code testable by creating a variant of `NewUpdateFrequencyMonitor()` which allows one to pass in a callback function on failure and the default implementation is the panic function with a string arg containing the service name.

Then you can separately test the panic function implementation by invoking it directly.

protocol/daemons/server/types/update_monitor_test.go
```diff
@@ -0,0 +1,31 @@
+package types
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
```
lcwik, line 8:
```suggestion

// TestUpdateFrequencyMonitor_Success may still intermittently fail on an overloaded system as it relies
// on `sleep` which isn't guaranteed to wake up after the specified amount of time.
```

protocol/x/sending/client/cli/sending_cli_test.go
```diff
@@ -50,17 +47,8 @@ func (s *SendingIntegrationTestSuite) SetupTest() {
...
 	// Configure test network.
-	s.cfg = network.DefaultConfig(&network.NetworkConfigOptions{
-		AppOptions: appOptions,
-		OnNewApp: func(val networktestutil.ValidatorI) {
-			// Disable the Bridge and Price daemons in the integration tests.
-			appOptions.Set(daemonflags.FlagPriceDaemonEnabled, false)
-			appOptions.Set(daemonflags.FlagBridgeDaemonEnabled, false)
-		},
-	})
+	s.cfg = network.DefaultConfig(nil)
```
lcwik, line 51: Even though we don't have to do the clean-up here should we still anyways since it is easier for all tests to _copy_ the same code for clean-up even if it is a no-op.

protocol/x/clob/client/cli/cancel_order_cli_test.go
```diff
@@ -58,15 +59,28 @@ func (s *CancelOrderIntegrationTestSuite) SetupTest() {
...
+			testval, ok := val.(networktestutil.Validator)
+			if !ok {
+				panic("incorrect validator type")
+			}
+
 			// Disable the Bridge and Price daemons in the integration tests.
 			appOptions.Set(daemonflags.FlagPriceDaemonEnabled, false)
 			appOptions.Set(daemonflags.FlagBridgeDaemonEnabled, false)
+
+			s.T().Log("setting up cancel order integration test GRPC ", testval.AppConfig.GRPC.Address)
```
lcwik, line 76:
```suggestion
```

lcwik (changes requested):

protocol/x/sending/client/cli/sending_cli_test.go
clemire: good point, thank you!

protocol/daemons/server/server.go
clemire: Sorry about that, I missed that in my refactor!

protocol/daemons/server/types/update_monitor.go
```diff
@@ -29,31 +32,64 @@ func NewUpdateFrequencyMonitor() *UpdateMonitor {
...
-	defer ufm.Unlock()
+// RegisterDaemonServiceWithCallback registers a new daemon service with the update frequency monitor. If the daemon
+// service fails to respond within the maximum acceptable update delay, the monitor will execute the callback function.
+// This method is synchronized. The method returns true if the daemon service was successfully registered,
+// and false if the daemon service was already registered or the monitor has already been stopped.
+func (ufm *UpdateMonitor) RegisterDaemonServiceWithCallback(
+	service string,
+	maximumAcceptableUpdateDelay time.Duration,
+	callback func(),
+) bool {
```
lcwik, line 43: use error here since that will cause the caller to have to check the response and handle it otherwise this `bool` will likely be ignored.

protocol/daemons/server/types/update_monitor_test.go
```diff
@@ -6,26 +6,135 @@ import (
...
+}
+
+func TestPanicServiceNotResponding(t *testing.T) {
+	panicFunc := PanicServiceNotResponding("test-service")
+	require.Panics(t, panicFunc)
+}
+
 // This test is disabled because the panic is not recoverable, since it's thrown
 // in a separate goroutine.
 //func TestUpdateFrequencyMonitor_Panics(t *testing.T) {
```
lcwik, line 135: You might as well and replace this and instead just have a simple test that invokes `PanicServiceNotResponding` with `require.Panics` which will give you all the coverage that you need.

protocol/daemons/server/types/update_monitor_test.go
```diff
@@ -6,26 +6,135 @@ import (
 	"time"
 )
 
-func TestUpdateFrequencyMonitor_Success(t *testing.T) {
+// The following tests may still intermittently fail on an overloaded system as they rely
+// on `time.Sleep`, which is not guaranteed to wake up after the specified amount of time.
+func TestRegisterDaemonService_Success(t *testing.T) {
```
lcwik, line 11:
```suggestion

func TestRegisterDaemonService_Success(t *testing.T) {
```

lcwik (approved):
This looks good to get the feature in.

protocol/daemons/server/types/update_monitor_test.go
clemire: That's the test above this one. I can remove this.

lucas-dydx (approved):

protocol/daemons/server/bridge.go
```diff
@@ -30,6 +39,7 @@ func (s *Server) AddBridgeEvents(
 	*api.AddBridgeEventsResponse,
 	error,
 ) {
+	_ = s.reportResponse(bridgeDaemonServiceName)
```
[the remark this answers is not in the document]
clemire: The daemons may start reporting responses before they are registered. In this case, we'd want to ignore the error.

protocol/daemons/server/server.go
```diff
@@ -22,32 +27,74 @@ type Server struct {
...
+func (server *Server) registerDaemon(
+	daemonKey string,
+	maximumAcceptableUpdateDelay time.Duration,
+) {
+	go func() {
+		// Give the protocol time to come up and start the gRPC query service. The daemons are unable
+		// to complete startup until this connection is established.
+		time.Sleep(DaemonStartupGracePeriod)
+		if err := server.updateMonitor.RegisterDaemonService(daemonKey, maximumAcceptableUpdateDelay); err != nil {
+			server.logger.Error("Failed to register daemon service with update monitor", "error", err)
```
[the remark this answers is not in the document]
clemire: The registration can fail in test cases where the test completes in less time than the daemon startup period.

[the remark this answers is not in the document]
lucas-dydx: Hmm can think deeper about a clean solution here, but somewhat uncomfortable allowing the daemon to not run here.. wouldn't this run into the same issue this PR is trying to prevent?

Understand this is an edge case so we can potentially create a ticket to fix this later. However ideally if the fix is simple, we can fix it now.

protocol/daemons/server/bridge.go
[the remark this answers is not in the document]
lucas-dydx: I strongly believe we should panic here if this happens more than the delay update multiple.

Is there any way we could update this logic to panic if a response isn't successfully recorded after multiple attempts? Maybe we could encode that logic in `RegisterValidResponse`?

protocol/daemons/server/server.go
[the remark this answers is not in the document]
clemire: Check out the fix!

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,123 @@
...
+	// This could be a concern for short-running integration test cases, where the network
+	// stops before all daemon services have been registered.
+	if ufm.stopped {
+		return fmt.Errorf("monitor has been stopped")
+	}
+
+	_, ok := ufm.serviceToUpdateMetadata[service]
+	if ok {
+		return fmt.Errorf("service already registered")
+	}
```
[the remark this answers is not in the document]
clemire: We do panic on this when we receive the error back from the daemon server.

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,123 @@
...
+	defer ufm.lock.Unlock()
+
+	// Don't bother to reset the timer if the monitor has already been stopped.
+	if ufm.stopped {
+		return nil
+	}
+
+	metadata, ok := ufm.serviceToUpdateMetadata[service]
+	if !ok {
+		return fmt.Errorf("service not registered")
```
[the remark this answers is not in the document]
clemire: I can move panics into the update monitor.

protocol/daemons/server/types/update_monitor.go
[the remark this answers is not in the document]
clemire: I see your point about panicking in the update_monitor. It simplifies the code, the alternative design gives the server the ability to de-escalate the server response to a daemon error for a non-critical daemon, not sure if that's useful.

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,150 @@
...
+	service string,
+	maximumAcceptableUpdateDelay time.Duration,
+	callback func(),
+) error {
+	ufm.lock.Lock()
+	defer ufm.lock.Unlock()
+
+	if maximumAcceptableUpdateDelay <= 0 {
+		return fmt.Errorf(
+			"registration failure for service %v: maximum acceptable update delay must be positive",
```
lucas-dydx, line 59: small nit: include `maximumAcceptableUpdateDelay` in the error as well

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// Don't register additional daemon services if the monitor has already been stopped.
+	// This could be a concern for short-running integration test cases, where the network
+	// stops before all daemon services have been registered.
+	if ufm.stopped {
+		return fmt.Errorf("registration failure for service %v: monitor has been stopped", service)
+	}
+
+	_, ok := ufm.serviceToUpdateMetadata[service]
+	if ok {
```
lucas-dydx, line 77: Suggestion, up to you:
```suggestion
	if _, ok := ufm.serviceToUpdateMetadata[service]; ok {
```

protocol/daemons/server/types/update_monitor.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("service already registered")
```
lucas-dydx, line 78: Suggestion: add `service` in the error (might not be a large deal since you add it downstream as part of the error handling)

protocol/daemons/server/types/util.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+import "time"
+
+// MaximumAcceptableUpdateDelay computes the maximum acceptable update delay for a daemon service as a
+// multiple of the loop delay.
+func MaximumAcceptableUpdateDelay(loopDelayMs uint32) time.Duration {
```
lucas-dydx, line 7: Small suggestion: wdyt about adding unit tests for this function? Even one should suffice

protocol/app/app.go
```diff
@@ -509,6 +509,7 @@ func New(
 		grpc.NewServer(),
 		&lib.FileHandlerImpl{},
 		daemonFlags.Shared.SocketAddress,
+		daemonFlags.Shared.GrpcServerAddress,
```
lucas-dydx, line 512: Small nit: can you give this daemon a custom name rather than the gRPC server address? Maybe `app-grpc-server`?

lucas-dydx (approved):

protocol/app/app.go
clemire: That's not a daemon. It's literally the grpc address that the daemons dial into to find the cosmos query server.

PR #240, opened 2023-09-12
[CLOB-85] - OrderRemoval DeliverTx validation

jakob-dydx:
branched off of Jonathan's [PR](https://github.com/dydxprotocol/v4-chain/pull/167)

Check linear ticket for details, but basically this PR implements as much DeliverTx validation as possible for stateful Order Removals operations with the current information provided in the order removal operations. e2e tests were favored over unit tests


protocol/x/clob/keeper/msg_server_proposed_operations.go
```diff
@@ -18,12 +19,10 @@ func (k msgServer) ProposedOperations(
...
-		panic(
-			errorsmod.Wrapf(
-				err,
-				"Block height: %d",
-				ctx.BlockHeight(),
-			),
+		return nil, errorsmod.Wrapf(
+			err,
+			"Block height: %d",
+			ctx.BlockHeight(),
```
jayy04, line 25: for my understanding - where are these errors being logged? 

previously this could be logged by the panic logging middleware. if we make this switch, do we need to update any of our alerts/monitors accordingly? if so, can you also own this?

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+// RemovalReason, and writes the removal to state.
+func (k Keeper) PersistOrderRemovalToState(
+	ctx sdk.Context,
+	orderRemoval *types.OrderRemoval,
+) error {
+	orderIdToRemove := orderRemoval.GetOrderId()
+	orderIdToRemove.MustBeStatefulOrder()
+
+	// Order removals are always for stateful orders that must exist.
+	orderToRemove, err := k.FetchOrderFromOrderId(ctx, orderIdToRemove, make(map[types.OrderId]types.Order, 0))
```
jayy04, line 227: `FetchOrderFromOrderId`  only fetches orders from long term store and *triggered* store. it does not fetch untriggered conditional orders. 

I think removals for untriggered conditional orders is still valid, for example due to collateralization reasons, so using `FetchOrderFromOrderId` is technically not always correct.

jayy04: I thought we have removal tests for untriggered conditional orders, but i could be wrong. if we don't can we add them

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+	if err != nil {
+		return err
+	}
+
+	// Statefully validate that the removal reason is valid.
+	switch removalReason := orderRemoval.RemovalReason; removalReason {
+	case types.OrderRemoval_REMOVAL_REASON_UNDERCOLLATERALIZED:
+		// For the collateralization check, use the remaining amount of the order that is resting on the book.
+		remainingAmount, hasRemainingAmount := k.MemClob.GetOrderRemainingAmount(ctx, orderToRemove)
```
jayy04, line 236: cleaner approach is to move `MemClob.GetOrderRemainingAmount` to the keeper. `GetOrderRemainingAmount` already calls keeper method anyways, so the logic to get remaining amount can also live on the keeper side.

Overall, we should try to avoid calling memclob methods if possible, since DeliverTx does not and should not modify memclob. Any modification should be delayed to prepare check state

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+		if !hasRemainingAmount {
+			panic(fmt.Sprintf("addOrderToOrderbookCollateralizationCheck: order has no remaining amount %v", orderToRemove))
```
jayy04, line 238: let's return an error here? according to the internal audit checklist, we should only panic when something is really really wrong.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+			subaccountOpenOrders,
+		)
+		if successPerSubaccountUpdate[orderIdToRemove.SubaccountId].IsSuccess() {
+			return errorsmod.Wrapf(
+				types.ErrInvalidOrderRemoval,
+				"Order Removal (%+v) invalid. Order passes collateralization check.",
+				*orderRemoval,
+			)
+		}
+	case types.OrderRemoval_REMOVAL_REASON_INVALID_REDUCE_ONLY:
```
jayy04, line 269: let's comment this out for now and add a TODO. RO is not yet enabled, so we shouldn't be supporting RO removal validation

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+		orderChangedSide := currentPositionSize.Sign()*newPositionSize.Sign() == -1
+		if !orderFillWouldIncreasePositionSize && !orderChangedSide {
+			return errorsmod.Wrapf(
+				types.ErrInvalidOrderRemoval,
+				"Order Removal (%+v) invalid. Order fill must increase position size or change side.",
+				*orderRemoval,
+			)
+		}
+	case types.OrderRemoval_REMOVAL_REASON_POST_ONLY_WOULD_CROSS_MAKER_ORDER:
+		// TODO
```
jayy04, line 298: can you add a ticket to this and other TODOs?

jayy04: for these and other removals that can not be verified without additional information. let's add a stat to keep track of the count (ideally with the proposer address in the labels)

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+		// this removal would have matched against this order.
+	case types.OrderRemoval_REMOVAL_REASON_CONDITIONAL_FOK_COULD_NOT_BE_FULLY_FILLED:
+		// TODO: The order must have been triggered in the last block
+
+		// The order should be in triggered state.
+		_, found := k.GetTriggeredConditionalOrderPlacement(ctx, orderIdToRemove)
+		if !found {
+			return errorsmod.Wrapf(
+				types.ErrInvalidOrderRemoval,
+				"Order Removal (%+v) invalid. Order is not in triggered state.",
```
jayy04, line 326: nit: instead of calling errorsmod.Wrapf for the errors within this function, simply return the error, e.g. types.ErrOrderNotTriggered, unless you have additional information to apply.

the caller of `PersistOrderRemovalToState` can do the Wrapf and provide additional details (in this case %+v order removal). This is the point of chained errors, so that each layer can provide some additional information

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +214,218 @@ func (k Keeper) PersistMatchToState(
...
+		// Temporarily construct the subaccountOpenOrders with a single PendingOpenOrder.
+		subaccountOpenOrders := map[satypes.SubaccountId][]types.PendingOpenOrder{
+			orderIdToRemove.SubaccountId: {
+				pendingOpenOrder,
+			},
+		}
+
+		// TODO(DEC-1896): AddOrderToOrderbookCollatCheck should accept a single PendingOpenOrder as a
+		// parameter rather than the subaccountOpenOrders map.
+		_, successPerSubaccountUpdate := k.AddOrderToOrderbookCollatCheck(
```
jayy04, line 257: hmm I think this check is for the taker order. but maker orders can also fail collat check. 

if we are always doing the most pessimistic check here, maybe this is okay?

protocol/x/clob/e2e/order_removal_test.go
```diff
@@ -361,6 +365,362 @@ func TestConditionalOrderRemoval(t *testing.T) {
...
+			msgProposedOperations: &clobtypes.MsgProposedOperations{
+				OperationsQueue: []clobtypes.OperationRaw{
+					clobtestutils.NewOrderRemovalOperationRaw(
+						constants.ConditionalOrder_Carl_Num0_Id0_Clob0_Buy05BTC_Price50000_GTBT10_SL_50003_FOK.OrderId,
+						clobtypes.OrderRemoval_REMOVAL_REASON_CONDITIONAL_FOK_COULD_NOT_BE_FULLY_FILLED,
+					),
+				},
+			},
+			expectedErrType: clobtypes.ErrStatefulOrderDoesNotExist,
+			expectedErr:     "does not exist in triggered conditional state",
```
jayy04, line 455: meganit: maybe just "conditional order is untriggered"? errors should be easy to understand even for someone without knowledge about the protocol

protocol/x/clob/e2e/order_removal_test.go
```diff
@@ -361,6 +365,362 @@ func TestConditionalOrderRemoval(t *testing.T) {
...
+					clobtestutils.NewOrderRemovalOperationRaw(
+						constants.ConditionalOrder_Carl_Num0_Id0_Clob0_Buy05BTC_Price50000_GTBT10_SL_50003_IOC.OrderId,
+						clobtypes.OrderRemoval_REMOVAL_REASON_CONDITIONAL_FOK_COULD_NOT_BE_FULLY_FILLED,
+					),
+				},
+			},
+			expectedErrType: clobtypes.ErrInvalidOrderRemoval,
+			expectedErr:     "Order is not fill-or-kill",
+		},
+		"invalid proposal: conditional fok order cannot be removed when fully filled": {
```
jayy04, line 480: not sure if we need a case for partially fill since it's technically invalid

protocol/x/clob/keeper/process_operations.go
jakob-dydx: How could we have Order Removals for untriggered conditional orders? Order Removals are triggered by the matching loop, but untriggered orders are untriggered and have never been placed. This means they can't be removed as a taker because that means they are triggered. They can't be removed as makers because they aren't on the book.

protocol/x/clob/e2e/order_removal_test.go
jakob-dydx: there isn't a test for partial fill here, this is testing to ensure we cannot include an Order Removal for a conditional fok order that was fully-filled. The order removal reason is to remove fok orders which were not fully filled, but in this test the order is fully-filled, so it throws an error

protocol/x/clob/keeper/msg_server_proposed_operations.go
jakob-dydx: will look into this. Intuitively it doesn't make sense to panic here but if we have some monitoring and alerts that are reliant on this I might leave it that way for now. This will change the DeliverTx process such that instead of panicking, the DeliverTx ABCI callback returns an error and I'm admittedly not totally sure what happens in terms of logging when that happens

protocol/x/clob/keeper/process_operations.go
jakob-dydx: I agree, I can update this in a different PR though. I started to make some changes but it started to blow up a bit and involves fixing a bunch of mocks in tests

jakob-dydx: https://linear.app/dydx/issue/CLOB-876/move-getorderremainingamount-into-clob-keeper

protocol/x/clob/e2e/order_removal_test.go
jakob-dydx: This error message is coming from the stateful validation jonathan just implemented where FetchOrderFromOrderId returns an error if the order cannot be found in triggered state. This function doesn't differentiate between: not in state and in state but untriggered, all it cares about is whether or not it is found in triggered state. Think that is most descriptive and accurate for now

protocol/x/clob/keeper/msg_server_proposed_operations.go
jakob-dydx: It seems like out of the box errors returned by `runMsgs` during `runTx` (the function that runs transactions during DeliverTx) are not logged. Options here would be to either
1: Add a line in `runTx` in our cosmos sdk fork to call `ctx.Logger().Error(err)` when an error is returned from `runMsgs`
2: just revert this and allow the panic recovery middleware to do the logging as it has been. Most of our message handlers don't seem to panic when an error is encountered, it seems that `ProposedOperations` is an exception here. Seems like the sole reason we are panicking is to make use of the logging in the recovery middleware. I can try to find out if any of our alarms rely on these logs. My thought would be that these logs contribute to the server error logs alarm we have set up, but nothing else

protocol/x/clob/keeper/process_operations.go
jakob-dydx: Yeah I think this is a fine way to do it, cc @lucas-dydx to confirm

protocol/x/clob/keeper/msg_server_proposed_operations.go
jakob-dydx: tbf though I don't think these errors should be surfacing in the error logs, as an error was encountered but it was not a critical application error, just an issue we gracefully handled. Think I'll keep this change as is

protocol/x/clob/memclob/memclob.go
```diff
@@ -1890,9 +1889,9 @@ func updateResultToOrderStatus(updateResult satypes.UpdateResult) types.OrderSta
 	return types.Undercollateralized
 }
 
-// getOrderRemainingAmount returns the remaining amount of an order (its size minus its filled amount).
+// GetOrderRemainingAmount returns the remaining amount of an order (its size minus its filled amount).
 // It also returns a boolean indicating whether the remaining amount is positive (true) or not (false).
-func (m *MemClobPriceTimePriority) getOrderRemainingAmount(
+func (m *MemClobPriceTimePriority) GetOrderRemainingAmount(
```
jonfung-dydx, line 1894: Technically, this function doesn't use any memclob functions/datastructures, it's on the clobKeeper. Maybe we should move it out in a future PR.

jonfung-dydx: nvm, I see the other comment.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +219,229 @@ func (k Keeper) PersistMatchToState(
...
+		// The order should be in triggered state.
+		_, found := k.GetTriggeredConditionalOrderPlacement(ctx, orderIdToRemove)
+		if !found {
+			return errorsmod.Wrap(
+				types.ErrConditionalOrderUntriggered,
+				"Order is not in triggered state.",
+			)
+		}
+
+		// The order should be FOK
```
jonfung-dydx, line 356: This check can be before the `GetTriggeredConditionalOrderPlacement` check, the first one would not return anything if it wasn't FOK

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +219,229 @@ func (k Keeper) PersistMatchToState(
...
+				"Order is not fill-or-kill.",
+			)
+		}
+
+		// The order should not be fully filled.
+		_, hasRemainingAmount := k.MemClob.GetOrderRemainingAmount(ctx, orderToRemove)
+		if !hasRemainingAmount {
+			return errorsmod.Wrap(
+				types.ErrOrderFullyFilled,
+				"Fill-or-kill order is fully filled.",
```
jonfung-dydx, line 369: Error messages here could be a bit confusing since it doesn't mention that it's an order removal. Perhaps `Order Removal Invalid: Fill Or Kill Order is Fully Filled`

jonfung-dydx: Ditto with other error messages in this stateful cehcking loop

protocol/x/clob/keeper/process_operations.go
jayy04: yeah previously I thought it can fail collat check. but I guess in that case, we should not even save that order to state in the first place. 

@jonfung-dydx can you confirm this is true?

protocol/x/clob/keeper/msg_server_proposed_operations.go
jayy04: we need some visibility into this so that we can debug issues. Proposed operations should never fail, unlike other messages like order placement, which can fail for a variety of reasons. so in a sense, it is a critical application error/logic bug.

imo we can just add the ctx.Logger().Error() here before we return. no need to modify cosmos sdk

jakob-dydx: Why is it that proposed operations should never fail? We don't do stateful validation until we reach DeliverTx and sometimes that validation fails. Reaching an error here just means that the block proposer proposed an invalid block, not that there was an application error or logic bug. In any case, to preserve the log I can add that `ctx.Logger().Error()` in this handler

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +219,229 @@ func (k Keeper) PersistMatchToState(
 	return nil
 }
 
+// statUnverifiedOrderRemoval stats the unverified order removal along with the block proposer. This
+// is used to track the number of unverified order removals a proposer has made, which may indicate malicious
+// behavior.
+func (k Keeper) statUnverifiedOrderRemoval(ctx sdk.Context, orderRemoval types.OrderRemoval) {
```
jayy04, line 225: ahhh I see you added all these logic to get the moniker (adding staking keeper which requires updates to all unit tests, etc). 

initially I thought we can just use the cons address as labels (just need to do something like `sdk.ConsAddress(block.Header().Propoer).String()`. Should be trivial to find the corresponding moniker manually I think.

jayy04: do you feel like we should leave this in? 

on one hand, it's nice to know the moniker so that it's more human readable. On the other hand, it does add some unnecessary complexity to CLOB (especially for unit tests and end to end tests) and we should generally keep it lean unless needed. Will let you make the call here 

Let's do the cleanup in a separate PR if you decide to revert this

protocol/x/clob/keeper/msg_server_proposed_operations.go
jayy04: A key difference between proposed operations vs. order placement/cancellation is that the operations are protocol injected. Since order placement/cancellation are user submitted, they can contain errors and arbitrary data that are invalid, and hence by rejected by the protocol.

On the other hand, there should be no reason for a honest proposer to inject invalid operations. Order matching should always use the latest state to make sure that we don't overfill orders, etc. This is kind of important since all proposed operations are in one transaction and is all-or-none. If proposers are constantly proposing operations that are rejected, the chain would be useless.

And yes, proposed operations can fail validation if it was proposed by a malicious proposer (dydx-research), but in this case, the operations queue has been tampered with so it's a different issue.

jayy04 (approved):

protocol/x/clob/keeper/process_operations.go
jonfung-dydx: I believe that when we place the stateful conditional order in DeliverTx, we run a collat check [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/orders.go#L208-L223). So I think it's ok if we don't check untriggered store. If we place it and it is collateralized, then it goes undercollateralized, when the order is triggered, we will move from untriggered state to triggered state in EndBlocker. Then, when we attempt to place the order via `PlaceConditionalOrdersTriggeredInLastBlock` in PrepareCheckState, we will always be generating removal reasons off of triggered conditional orders.

jonfung-dydx (approved):

protocol/x/clob/keeper/process_operations.go
jakob-dydx: ^ yep that sounds exactly right to me

protocol/x/clob/keeper/process_operations.go
jakob-dydx: I honestly don't even need the `GetTriggeredConditionalOrderPlacement` check. We already verify the order exists in state using FetchOrderFromOrderId and it requires conditional orders to be in triggered state, so this is redundant.

protocol/x/clob/keeper/process_operations.go
jakob-dydx: I've wrapped all of these errors in this block:
```
if err := k.PersistOrderRemovalToState(ctx, *orderRemoval); err != nil {
				return errorsmod.Wrapf(
					types.ErrInvalidOrderRemoval,
					"Order Removal (%+v) invalid. Error: %+v",
					*orderRemoval,
					err,
				)
			}
```

so we get the original error as well as the invalid order removal error which also logs the order removal

protocol/x/clob/keeper/process_operations.go
jakob-dydx: crap, yeah I just reverted this and updated the PR

protocol/x/clob/keeper/process_operations.go
```diff
@@ -248,6 +219,208 @@ func (k Keeper) PersistMatchToState(
...
+) {
+	proposerConsAddress := sdk.ConsAddress(ctx.BlockHeader().ProposerAddress)
+	telemetry.IncrCounterWithLabels(
+		[]string{types.ModuleName, metrics.ProcessOperations, metrics.UnverifiedStatefulOrderRemoval, metrics.Count},
+		1,
+		append(
+			orderRemoval.OrderId.GetOrderIdLabels(),
+			metrics.GetLabelForStringValue(metrics.RemovalReason, orderRemoval.GetRemovalReason().String()),
+			metrics.GetLabelForStringValue(metrics.Proposer, proposerConsAddress.String()),
+			metrics.GetLabelForIntValue(metrics.BaseQuantums, int(orderToRemove.GetBaseQuantums())),
```
jakob-dydx, line 237: @jayy04 does this stat look right?

PR #243, opened 2023-09-12
[CLOB-866] update sorting of liquidation orders

jayy04:


jakob-dydx (approved):

protocol/x/clob/keeper/liquidations.go
```diff
@@ -806,3 +810,67 @@ func (k Keeper) ConvertFillablePriceToSubticks(
...
+		xQuoteQuantums := k.getQuoteQuantumsForLiquidationOrder(ctx, x)
+		yQuoteQuantums := k.getQuoteQuantumsForLiquidationOrder(ctx, y)
+		if xQuoteQuantums.Cmp(yQuoteQuantums) != 0 {
+			return xQuoteQuantums.Cmp(yQuoteQuantums) == 1
+		}
+
+		// Sort by order hash by default.
+		xHash := x.GetOrderHash()
+		yHash := y.GetOrderHash()
+		return bytes.Compare(xHash[:], yHash[:]) == -1
```
jonfung-dydx, line 843: Do we need the [:] copy? also should this be 1 to keep in alignment with the first 2 cases? Both functions seem to return -1 if the first element is < the second element

jonfung-dydx (approved):

protocol/x/clob/keeper/liquidations.go
jayy04: yeah I just copied this from other sorting functions. IIRC this is to convert OrderHash to [32]byte?

sorting by order hash is arbitrary so either way is fine. I can update to +1 if you have a strong preference. let me know

jonfung-dydx: Looks like order hash is already a [32]byte.
```
// OrderHash is used to represent the SHA256 hash of an order.
type OrderHash [32]byte
```
Is it possible to just typecast it instead of copying the whole thing over? don't want to be nitty about runtime but seems like cheap improvements.

For the -1 vs 1, up to you, no biggie.

PR #248, opened 2023-09-12
Clean-up maps functions

BrendanChou:
**First Change:**
Remove some duplicate or unused functions in `lib/maps`

---

**Second Change:**
There were essentially two duplicate functions:
- `maps.GetSortedKeys`
- `lib.ConvertMapToSliceOfKeys` (would not sort the keys, but every time it was used, sorting was used afterwards)

So I combined these into a single function that is generically types (supports more than just `string` keys)


protocol/lib/slices.go
```diff
@@ -16,17 +19,17 @@ func ContainsDuplicates[V comparable](values []V) bool {
...
-// ConvertMapToSliceOfKeys converts a map to a slice of keys.
-// Note that this method returns the keys in a non-deterministic order.
-func ConvertMapToSliceOfKeys[K comparable, V any](m map[K]V) []K {
-	var slice = make([]K, len(m))
-	index := 0
-	for key := range m {
-		slice[index] = key
-		index += 1
+// GetSortedKeys returns the keys of the map in sorted order.
+func GetSortedKeys[R interface {
```
lcwik, line 23: Why do you want an explicit sort interface here since?

You could use [slices.SortFunc()](https://pkg.go.dev/golang.org/x/exp/slices#SortFunc) with a simple func that uses `<` since `k` is comparable?

BrendanChou: Confusing terminology but [comparable](https://go.dev/blog/comparable) means that you can use `==` to compare the type. I believe what you're thinking of is [constraints.Ordered](https://pkg.go.dev/golang.org/x/exp/constraints#Ordered) which means that `<` can be used.

BrendanChou: Since this function should be usable for arbitrary structs used as keys, then those types are `comparable` but not `Ordered`

lcwik, line 23: Should this be in `protocol/lib/maps` since it operates on a maps?

protocol/lib/slices.go
```diff
@@ -16,17 +19,17 @@ func ContainsDuplicates[V comparable](values []V) bool {
...
+	sort.Interface
+}, K comparable, V any](m map[K]V) R {
+	keys := make([]K, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
 	}
-
-	return slice
+	sort.Sort(R(keys))
+	return keys
```
lcwik, line 32: The tests are failing since we are returning `sort.StringSlice`

lcwik (changes requested):

protocol/lib/slices.go
BrendanChou: We might just remove `lib/maps` (TBD) since a lot of those are primarily used for testing

lcwik (approved):

lcwik (approved):

PR #249, opened 2023-09-12
[CLOB-852] move deleveraging out of liquidations

jayy04:


protocol/x/clob/keeper/liquidations_deleveraging.go
```diff
@@ -17,6 +18,33 @@ import (
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+// DeleverageSubaccount is the main entry point to deleverage a subaccount. It attempts to find positions
+// on the opposite side of deltaQuantums and use them to offset the liquidated subaccount's position at
+// the bankruptcy price of the liquidated position.
+func (k Keeper) DeleverageSubaccount(
```
jonfung-dydx, line 24:
```suggestion
func (k Keeper) MaybeDeleverageSubaccount(
```

Up to you

protocol/x/clob/memclob/memclob.go
```diff
@@ -676,53 +675,41 @@ func (m *MemClobPriceTimePriority) PlacePerpetualLiquidation(
...
+	return liquidationOrderStatus.OrderOptimisticallyFilledQuantums,
+		liquidationOrderStatus.OrderStatus,
+		offchainUpdates,
+		err
+}
 
-	fills, _ := m.clobKeeper.OffsetSubaccountPerpetualPosition(
+// DeleverageSubaccount will deleverage a subaccount by finding perpetual positions that can be used to offset
+// the offendign subaccount. All position will be closed at the bankruptcy price of the subaccount that is being
+// deleveraged.
```
jonfung-dydx, line 686:
```suggestion
// DeleverageSubaccount will deleverage a subaccount by finding perpetual positions that can be used to offset
// the offending subaccount. All position will be closed at the bankruptcy price of the subaccount that is being
// deleveraged.
```

Maybe good to also add a comment about this being optimistic and state transitions being persisted

protocol/x/clob/memclob/memclob.go
```diff
@@ -676,53 +675,41 @@ func (m *MemClobPriceTimePriority) PlacePerpetualLiquidation(
...
+func (m *MemClobPriceTimePriority) DeleverageSubaccount(
+	ctx sdk.Context,
+	subaccountId satypes.SubaccountId,
+	perpetualId uint32,
+	deltaQuantums *big.Int,
+) (
+	quantumsDeleveraged *big.Int,
+	err error,
```
jonfung-dydx, line 694: I think it's impossible for this function to return something that's not nil here?

protocol/x/clob/abci.go
```diff
@@ -238,12 +238,24 @@ func PrepareCheckState(
...
-			)
+			telemetry.IncrCounter(1, types.ModuleName, metrics.PrepareCheckState, metrics.UnfilledLiquidationOrders)
+
+			// The liquidation order was unfilled. Try to deleverage the subaccount.
+			subaccountId := liquidationOrders[i].GetSubaccountId()
+			perpetualId := liquidationOrders[i].MustGetLiquidatedPerpetualId()
+			deltaQuantums := liquidationOrders[i].GetDeltaQuantums()
+
+			_, err := keeper.DeleverageSubaccount(ctx, subaccountId, perpetualId, deltaQuantums)
+			if err != nil {
```
jonfung-dydx, line 249: Since this function cannot return nil. I think the only error that is possible to trigger this is when subaccount cannot be deleveraged (doesn't meet the two conditions). 

Should we stat here?

jonfung-dydx (approved):

jakob-dydx (approved):

protocol/x/clob/memclob/memclob.go
jayy04: yup - added err here for consistency with the liquidation methods.

I am planning to do some cleanup and potentially change some of the panics in this code path to error, so added err here in this PR

protocol/x/clob/abci.go
jayy04: when subaccounts cannot be deleveraged, error will be nil.

error will be non nil if `GetNetCollateralAndMarginRequirements` returns an error, and this should never happen.

protocol/x/clob/abci.go
```diff
@@ -238,12 +238,24 @@ func PrepareCheckState(
 		if optimisticallyFilledQuantums > 0 {
 			numFilledLiquidations++
 		} else {
-			telemetry.IncrCounter(
-				1,
-				types.ModuleName,
-				metrics.PrepareCheckState,
-				metrics.UnfilledLiquidationOrders,
-			)
+			telemetry.IncrCounter(1, types.ModuleName, metrics.PrepareCheckState, metrics.UnfilledLiquidationOrders)
```
lucas-dydx, line 241: Nit: isn't this the same as number of deleveraged subaccounts?

PR #251, opened 2023-09-12
[CORE-504] Implement MsgCreateOracleMarket handler

teddyding:


protocol/testutil/keeper/prices.go
```diff
@@ -258,3 +266,9 @@ func CreateTestPriceMarkets(
 		marketId++
 	}
 }
+
+func GetNumMarkets(t *testing.T, ctx sdk.Context, keeper *keeper.Keeper) uint32 {
```
teddyding, line 270: Moved this method from the keeper struct since protocol logic *should not* rely on number of markets of anything, now that we've removed consecutive ID requirement

protocol/x/prices/keeper/market.go
```diff
@@ -21,6 +21,14 @@ func (k Keeper) CreateMarket(
 	marketParam types.MarketParam,
 	marketPrice types.MarketPrice,
 ) (types.MarketParam, error) {
+	if _, exists := k.GetMarketParam(ctx, marketParam.Id); exists {
```
teddyding, line 24: Updated signature of `GetMarketParam` to be more generalized (before it throws an error if `id` is not found)

protocol/x/prices/keeper/msg_server_create_oracle_market.go
```diff
@@ -2,14 +2,37 @@ package keeper
...
+			msg.Authority,
+		)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Use zero oracle price to create the new market.
+	// Note that valid oracle price updates cannot be zero (checked in MsgUpdateMarketPrices.ValidateBasic),
+	// so a zero oracle price indicates that the oracle price has never been updated.
+	zeroMarketPrice := types.MarketPrice{
```
clemire, line 29: Do we have any tests that explicitly confirm that zero prices are ignored by clob, etc as unusable? Esp. integration / e2e.

Related question: do you have a test plan for adding a new market and seeing trading delayed until a price update occurs?

clemire (approved):

protocol/x/prices/keeper/msg_server_create_oracle_market.go
teddyding: Yes, we will cover e2e test in this [project](https://linear.app/dydx/project/[m5c-p0]-core-user-flow-test-coverage-5e4f1dc1cfef)! Will document test plan as part of this [ticket](https://linear.app/dydx/issue/CORE-571/e2e-test-for-gov-add-new-markets)

PR #253, opened 2023-09-13
[DEC-2104] move `time_provider.go` to `lib/time` package

ttl33:
* Simply moving the file to a different package

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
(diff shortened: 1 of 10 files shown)


protocol/app/app.go
```diff
@@ -95,7 +96,7 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/app/upgrades"
 
 	// Lib
-	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	libtime "github.com/dydxprotocol/v4-chain/protocol/lib/time"
```
BrendanChou, line 99: no need to rename this since there is no name conflict

BrendanChou (approved):

PR #232, opened 2023-09-12
[STAB-29] Pass in version/commit in dev/staging, shift app_info metrics to begin blocker

jonfung-dydx:

PR #259, opened 2023-09-13
[STAB-29] Ensure that version and commit are always defined. Also replace version/commit for dev/staging builds.

lcwik:


lcwik:
Fixed in https://github.com/dydxprotocol/v4-chain/pull/232

PR #261, opened 2023-09-13
Merge `lib/maps/` and `lib/slices.go` into `lib/collections.go`

BrendanChou:
Also remove a lot of unnecessary functions, either because they are not used, are covered by equivalent functions in `golang.org/x/exp`, or are redundant to how golang already works (e.g. panic'ing on an OOB)


protocol/lib/collections.go
```diff
@@ -32,17 +32,6 @@ func GetSortedKeys[R interface {
 	return keys
```
lcwik, line 32: good rename

lcwik (approved):
minor nit

PR #263, opened 2023-09-13
v0.2.2 hard fork to update clob pairs

jayy04:


protocol/app/upgrades.go
```diff
@@ -5,20 +5,19 @@ import (
...
 // setupUpgradeHandlers registers the upgrade handlers to perform custom upgrade
 // logic and state migrations for software upgrades.
 func (app *App) setupUpgradeHandlers() {
-	for _, upgrade := range Upgrades {
-		if app.UpgradeKeeper.HasHandler(upgrade.UpgradeName) {
-			panic(fmt.Sprintf("Cannot register duplicate upgrade handler '%s'", upgrade.UpgradeName))
-		}
-		app.UpgradeKeeper.SetUpgradeHandler(
-			upgrade.UpgradeName,
-			upgrade.CreateUpgradeHandler(app.ModuleManager, app.configurator),
```
jayy04, line 19: Probably need to update `CreateUpgradeHandler` so that it's more flexible to support different keepers. will think about that more

roy-dydx: Is passing in a reference to app itself a reasonable option?

roy-dydx (approved):

protocol/app/upgrades/v0.2.2/upgrades.go
```diff
@@ -0,0 +1,127 @@
...
+
+		offchainUpdates := clobKeeper.MemClob.PurgeInvalidMemclobState(
+			ctx,
+			[]clobtypes.OrderId{},
+			[]clobtypes.OrderId{},
+			[]clobtypes.OrderId{},
+			removedOrderIds,
+			clobtypes.NewOffchainUpdates(),
+		)
+		clobKeeper.SendOffchainMessages(offchainUpdates, nil, metrics.SendPrepareCheckStateOffchainUpdates)
```
jayy04, line 123: @lucas-dydx IIRC upgrade handlers are called in begin blocker, so I think we need to clear op queue and purge memclob. Let me know if this makes sense

protocol/app/upgrades.go
jayy04: yeah I think that's a good idea.

protocol/app/upgrades/v0.2.2/constants.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/dydxprotocol/v4-chain/protocol/app/upgrades"
+)
+
+const UpgradeName = "v0.2.2"
+
+var (
+	Fork = upgrades.Fork{
+		UpgradeName: UpgradeName,
+		// Target upgrade time is Sept 15, 2023, 1pm EST,
+		// estimated to occur on block 178500, assuming 1.5s block time.
```
teddyding, line 14: Meganit: current testnet block time around [1.4s](https://app.datadoghq.com/dashboard/wr9-2px-quy/v4?refresh_mode=sliding&tpl_var_env%5B0%5D=testnet&from_ts=1694636335033&to_ts=1694647135033&live=false&tile_focus=8932930888001242), up to you whether to upate height accordingly

protocol/app/app_test.go
```diff
@@ -150,11 +150,11 @@ func TestSimulationManager(t *testing.T) {
 }
 
 func TestUpgrades(t *testing.T) {
-	require.Len(t, app.Upgrades, 0, "Expected no Upgrades")
+	require.Len(t, app.Upgrades, 1, "Expected 1 Upgrades")
 }
 
 func TestForks(t *testing.T) {
-	require.Len(t, app.Forks, 0, "Expected no Forks")
```
jayy04, line 157: These are low value unit tests in my opinion and are just going to delay things in cases like this. 

I am going to remove them on main if there is no objection.

protocol/app/upgrades_test.go
```diff
@@ -1,12 +1,13 @@
...
 	"github.com/dydxprotocol/v4-chain/protocol/app"
 	"github.com/stretchr/testify/require"
-	"testing"
 )
 
 func TestDefaultUpgradesAndForks(t *testing.T) {
-	require.Empty(t, app.Upgrades, "Expected empty upgrades list")
-	require.Empty(t, app.Forks, "Expected empty forks list")
+	require.Len(t, app.Upgrades, 1, "Expected 1 upgrade")
+	require.Len(t, app.Forks, 1, "Expected 1 fork")
```
jayy04, line 12: also low value unit tests. Planning to remove unless theres objections

protocol/app/upgrades/v0.2.2/upgrades.go
```diff
@@ -0,0 +1,160 @@
...
+	clobKeeper clobmodulekeeper.Keeper,
+) upgradetypes.UpgradeHandler {
+	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		ctx.Logger().Info("Running v0.2.2 hard fork...")
+
+		// Update clob pairs.
+		clobPairs := clobKeeper.GetAllClobPairs(ctx)
+		for _, clobPair := range clobPairs {
+			switch clobPair.Id {
+			case 0:
```
ttl33, line 24: @jayy04 can you find someone who can double check these numbers?

protocol/app/upgrades/v0.2.2/upgrades.go
```diff
@@ -0,0 +1,160 @@
...
+		offchainUpdates := clobKeeper.MemClob.PurgeInvalidMemclobState(
+			ctx,
+			[]clobtypes.OrderId{},
+			[]clobtypes.OrderId{},
+			[]clobtypes.OrderId{},
+			removedOrderIds,
+			clobtypes.NewOffchainUpdates(),
+		)
+		clobKeeper.SendOffchainMessages(offchainUpdates, nil, metrics.SendPrepareCheckStateOffchainUpdates)
+
```
ttl33, line 157: @jayy04 @vincentwschau 

I think we also want to update the indexer. Can we try sending an Indexer event in the upgrade handler to see if this can trigger updates on the indexer side?

protocol/app/upgrades/v0.2.2/upgrades.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+	clobmodulekeeper "github.com/dydxprotocol/v4-chain/protocol/x/clob/keeper"
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+func CreateUpgradeHandler(
```
ttl33, line 12: For my own understanding, do we plan on deleting any of the migration/upgrade logic in the future? Logic including this handler and other dependent methods like `UnsafeSetClobPair`

protocol/app/upgrades/v0.2.2/upgrades.go
teddyding: Added logic to send UpdateClobPair events

protocol/app/upgrades/v0.2.2/upgrades.go
teddyding: [Confirmed](https://dydx-team.slack.com/archives/C03FP255RRU/p1694655414499069?thread_ts=1694644218.746589&channel=C03FP255RRU&message_ts=1694655414.499069) by Yu Jin

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -553,3 +553,26 @@ func (k Keeper) IsPerpetualClobPairActive(
...
+
+// UnsafeSetClobPair sets a specific `ClobPair` in the store from its index.
+func (k Keeper) UnsafeSetClobPair(ctx sdk.Context, clobPair types.ClobPair) {
+	b := k.cdc.MustMarshal(&clobPair)
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.ClobPairKeyPrefix))
+	// Write the `ClobPair` to state.
+	store.Set(types.ClobPairKey(clobPair.GetClobPairId()), b)
+
+	// Send UpdateClobPair to indexer.
+	k.GetIndexerEventManager().AddTxnEvent(
```
teddyding, line 565: Added TxnEvent for `UpdateClobPair`, @Christopher-Li PTAL

jakob-dydx: This is not actually related to this issue, but this indexer event should also be emitted during `keeper.UpdateClobPair` as well I believe. Think it was something we missed adding when working on the add new markets project (which relied on the new MsgUpdateClobPair) message

Christopher-Li: This looks good, I'm assuming that this should be one of the last events in the block? After all subaccountUpdateEvents and OrderFillEvents?

protocol/app/upgrades/v0.2.2/constants.go
```diff
@@ -0,0 +1,23 @@
+package v0_2_2
+
+import (
+	store "github.com/cosmos/cosmos-sdk/store/types"
+	"github.com/dydxprotocol/v4-chain/protocol/app/upgrades"
+)
+
+const UpgradeName = "v0.2.2"
```
ttl33, line 8: Nit / Non-blocking, I noticed that most teams use whole numbers for versions. Do we plan on using whole numbers for mainnet software or continue to use SemVer in the future?

@roy-dydx

protocol/x/clob/keeper/clob_pair.go
vincentwschau: Yes, this is currently called in the upgrade handler which should execute in the `EndBlock` ([reference](https://docs.cosmos.network/main/modules/upgrade))

vincentwschau: Non-blocking nit: Since this specific function is currently only run in the `EndBlocker`, we could change to using the correct event type.
e.g.
```
indexerEventManager.AddBlockEvent(
	ctx,
	indexerevents.SubtypeUpdateClobPair,
	indexer_manager.GetB64EncodedEventMessage(
		indexerevents.NewUpdateClobPairEvent(
			clobPair.Id,
			clobPair.Status,
			clobPair.QuantumConversionExponent,
			clobPair.SubticksPerTick,
			clobPair.StepBaseQuantums,
		),
	),
	indexer_manager.IndexerTendermintEvent_BLOCK_EVENT_END_BLOCK,
)
```

PR #265, opened 2023-09-14
add update clob pair indexer event to clob keeper UpdateClobPair

jakob-dydx:


protocol/indexer/events/update_clob_pair_create.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// NewUpdateClobPairEvent creates a UpdateClobPairEventV1
+// representing update of a clob pair.
+func NewUpdateClobPairEvent(
+	clobPairId uint32,
+	status types.ClobPair_Status,
+	quantumConversionExponent int32,
+	subticksPerTick uint32,
+	stepBaseQuantums uint64,
```
lucas-dydx, line 15: Possible to use `satypes.BaseQuantums` and `types.SubticksPerTick`, and `types.ClobPairId` for the params?

protocol/indexer/events/update_clob_pair_create.go
```diff
@@ -0,0 +1,24 @@
...
+) *UpdateClobPairEventV1 {
+	return &UpdateClobPairEventV1{
+		ClobPairId:                clobPairId,
+		Status:                    v1.ConvertToClobPairStatus(status),
```
lucas-dydx, line 19: Shouldn't the `ConvertToClobPairStatus` func panic if status is unspecified?

protocol/indexer/events/update_clob_pair_create.go
```diff
@@ -0,0 +1,24 @@
+package events
+
+import (
+	v1 "github.com/dydxprotocol/v4-chain/protocol/indexer/protocol/v1"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// NewUpdateClobPairEvent creates a UpdateClobPairEventV1
+// representing update of a clob pair.
```
lucas-dydx, line 9:
```suggestion
// representing an update of a clob pair.
```

protocol/indexer/events/update_clob_pair_create.go
```diff
@@ -0,0 +1,24 @@
...
+	quantumConversionExponent int32,
+	subticksPerTick uint32,
+	stepBaseQuantums uint64,
+) *UpdateClobPairEventV1 {
+	return &UpdateClobPairEventV1{
+		ClobPairId:                clobPairId,
+		Status:                    v1.ConvertToClobPairStatus(status),
+		QuantumConversionExponent: quantumConversionExponent,
+		SubticksPerTick:           subticksPerTick,
+		StepBaseQuantums:          stepBaseQuantums,
```
lucas-dydx, line 22: Worth adding a comment why you don't include updates to the `metadata` field?

lucas-dydx (approved):

protocol/indexer/events/update_clob_pair_create.go
Christopher-Li: yes

Christopher-Li (approved):

protocol/indexer/events/update_clob_pair_create.go
jakob-dydx: yeah sorry I just copied this from Teddy in the hard fork PR, will update

protocol/indexer/events/update_clob_pair_create.go
jakob-dydx: This was determined by the indexer `UpdateClobPairEventV1` proto, but I imagine it is because we don't support changing a clob pair's metadata because transitioning from a perpetual to an asset would break things

jakob-dydx: can add a comment though

PR #270, opened 2023-09-14
[CORE-559] implement governance message MsgSendFromModuleToAccount

tqin7:


protocol/x/sending/keeper/transfer_test.go
```diff
@@ -462,3 +463,111 @@ func TestProcessWithdrawFromSubaccount(t *testing.T) {
...
+		"Success - send to user account": {
+			initialModuleBalance: 1000,
+			balanceToSend:        100,
+			recipientAddress:     constants.AliceAccAddress.String(),
+		},
+		"Success - send to module account": {
+			initialModuleBalance: 1000,
+			balanceToSend:        100,
+			recipientAddress:     authtypes.NewModuleAddress("community_treasury").String(),
+		},
```
clemire, line 489: Can we add a test case for msg fails ValidateBasic?

clemire: Would it be possible to also add a test for sending from an nonexistent account with a valid formatted address?

clemire (dismissed):
Do you have forthcoming integration test plans for this? (Sorry if I've already asked on another PR!)

tqin7:
> Do you have forthcoming integration test plans for this? (Sorry if I've already asked on another PR!)

Yep! Will write integration tests for all governance messages by simulating governance workflow!

protocol/x/sending/keeper/transfer_test.go
tqin7: Will try to add both!

protocol/testutil/keeper/sending.go
```diff
@@ -27,6 +29,7 @@ func SendingKeepers(t testing.TB) (
 	ctx sdk.Context,
 	sendingKeeper *keeper.Keeper,
 	accountKeeper *authkeeper.AccountKeeper,
+	bankKeeper *bankkeeper.BaseKeeper,
```
teddyding, line 32: Instead of returning everything, can we use similar style as clob unit tests where a context object is returned? 
See [definition](https://github.com/dydxprotocol/v4-chain/blob/60d09718d807189f9ebc8522ac9cb15e66c6d8e9/protocol/testutil/keeper/clob.go#L55) and [usage](https://github.com/dydxprotocol/v4-chain/blob/60d09718d807189f9ebc8522ac9cb15e66c6d8e9/protocol/x/clob/keeper/get_price_premium_test.go#L159-L187)

protocol/x/sending/keeper/transfer.go
```diff
@@ -215,6 +213,14 @@ func (k Keeper) SendFromModuleToAccount(
 	ctx sdk.Context,
 	msg *types.MsgSendFromModuleToAccount,
 ) (err error) {
-	// TODO(CORE-559): Implement this method.
-	return status.Errorf(codes.Unimplemented, "SendFromModuleToAccount not implemented")
+	if err = msg.ValidateBasic(); err != nil {
+		return err
+	}
+
+	return k.bankKeeper.SendCoins(
```
teddyding, line 220: Any particular reason to not use [SendCoinsFromModuleToAccount](https://github.com/dydxprotocol/cosmos-sdk/blob/4fadfe5a4606b6dc76644d377ed34420f3b80801/x/bank/keeper/keeper.go#L316)?

protocol/x/sending/keeper/msg_server_create_transfer_test.go
```diff
@@ -256,3 +256,69 @@ func TestWithdrawFromSubaccount(t *testing.T) {
...
+				SenderModuleName: "community_treasury",
+				Recipient:        constants.CarlAccAddress.String(),
+				Coin:             sdk.NewCoin("dv4tnt", sdk.NewInt(1)),
+			},
+			keeperResp:  fmt.Errorf("keeper error"),
+			expectedErr: "keeper error",
+		},
+	}
+
+	for name, tc := range tests {
```
teddyding, line 302: Let's prefer to not have mocked unit tests unless it's really hard to set up. Setting balances with a real keeper and checking balance afterwards should be relatively simple. 

Consider using the `testapp` framework which enables setting up account balances through genesis state ([example](https://github.com/dydxprotocol/v4-chain/blob/60d09718d807189f9ebc8522ac9cb15e66c6d8e9/protocol/x/rewards/keeper/keeper_test.go#L600-L616))

protocol/x/sending/keeper/transfer_test.go
teddyding: Can you add a test that sends funds from a module account to itself? Assume this causes an error, let's check

tqin7: it doesn't cause an error. It subtracts funds from the module account and then adds funds to it. Adding that test doesn't seem quite necessary to me and it also requires changing the way unit tests are currently written. Let me know your thoughts

protocol/x/sending/keeper/msg_server_create_transfer_test.go
```diff
@@ -63,14 +63,14 @@ func setUpTestCase(
 ) {
 	// Initialize Mocks and Context.
 	mockKeeper = &mocks.SendingKeeper{}
-	ctx, _, _, _, _, _, _, _, _ := keepertest.SendingKeepers(t)
-	ctx = ctx.WithBlockHeight(25)
+	ks := keepertest.SendingKeepers(t)
+	ks.Ctx = ks.Ctx.WithBlockHeight(25)
```
teddyding, line 67: Inline:

```
tc.setupMocks(ks.Ctx.WithBlockHeight(25), mockKeeper)
```
Note `WithBlockHeight` returns a copy and doesn't modify the original context. `ks.Ctx = ks.Ctx.WithBlockHeight(25)` is actively modifying `ks.Ctx`

teddyding (approved):

protocol/x/sending/keeper/msg_server_create_transfer_test.go
tqin7: Yeah I think in this case we want to return the modified context at the end of this function. However, I can change it to `ctx := ks.Ctx.WithBlockHeight(25)` and return `ctx` at the end.

PR #272, opened 2023-09-14
clean up clob hydration functions and unit tests

jayy04:


protocol/app/app.go
```diff
@@ -759,7 +759,6 @@ func New(
 		app.AccountKeeper,
 		app.BankKeeper,
 		app.SubaccountsKeeper,
-		memClob,
```
jayy04, line 762: note that I didn't choose to internalize this (it's still initialized in app.go) because there are a lot of tests setup that needs to be updated. so don't think it's worth it right now

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -141,20 +142,6 @@ func (k Keeper) validateClobPair(ctx sdk.Context, clobPair *types.ClobPair) erro
 func (k Keeper) createOrderbook(ctx sdk.Context, clobPair types.ClobPair) {
 	// Create the corresponding orderbook in the memclob.
 	k.MemClob.CreateOrderbook(ctx, clobPair)
-
-	// If this `ClobPair` is for a perpetual, add the `clobPairId` to the list of CLOB pair IDs
```
lucas-dydx, line 145: Should we update the comment to note this isn't done within this method anymore?

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -198,6 +188,35 @@ func (k Keeper) InitMemClobOrderbooks(ctx sdk.Context) {
...
+// SetClobPairIdForPerpetual sets the mapping between clob pair and perpetual.
+func (k Keeper) SetClobPairIdForPerpetual(ctx sdk.Context, clobPair types.ClobPair) {
+	// If this `ClobPair` is for a perpetual, add the `clobPairId` to the list of CLOB pair IDs
+	// that facilitate trading of this perpetual.
+	if perpetualClobMetadata := clobPair.GetPerpetualClobMetadata(); perpetualClobMetadata != nil {
+		perpetualId := perpetualClobMetadata.PerpetualId
+		clobPairIds, exists := k.PerpetualIdToClobPairId[perpetualId]
+		if !exists {
+			clobPairIds = make([]types.ClobPairId, 0)
+		}
```
lucas-dydx, line 212: Should we panic if it already exists, since that indicates there is more than one CLOB pair for a perp?

protocol/x/clob/keeper/clob_pair_test.go
```diff
@@ -493,6 +493,42 @@ func TestInitMemClobOrderbooks(t *testing.T) {
 
 	// Initialize the `ClobPairs` from Keeper state.
 	ks.ClobKeeper.InitMemClobOrderbooks(ks.Ctx)
+}
+
+func TestHydratClobPairAndPerpetualMapping(t *testing.T) {
```
lucas-dydx, line 498:
```suggestion
func TestHydrateClobPairAndPerpetualMapping(t *testing.T) {
```

protocol/app/app.go
lucas-dydx: Wdym internalize? Do you think it should be initialized in the CLOB keeper constructor?

If so I don't think we should init it within the CLOB keeper constructor, since the current method of usage allows for better dependency injection.

lucas-dydx (approved):

protocol/app/app.go
jayy04: yeah by internalize I meant instantiate this within the clob keeper constructor. 

agree this approach allows for better dependency injection for unit tests. but for the application itself, I don't think there is a reason to expose memclob to other modules.

anyways, leaving this as is for now

protocol/x/clob/keeper/clob_pair.go
jayy04: we panic when we call the getter function  `GetClobPairForPerpetual` so I think we are fine

protocol/x/clob/abci.go
```diff
@@ -110,7 +110,6 @@ func EndBlocker(
 func PrepareCheckState(
 	ctx sdk.Context,
 	keeper keeper.Keeper,
```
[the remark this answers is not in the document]
jayy04: updated to use ref

PR #275, opened 2023-09-15
add BridgeEvents script

BrendanChou:
Adds a script that can be used to test bridge events and also help construct a `genesis.json` file if needed

### DIFF
--- a/protocol/scripts/bridge_events.go
+++ b/protocol/scripts/bridge_events.go
@@ -1,0 +1,144 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"math/big"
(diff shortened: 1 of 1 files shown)


protocol/scripts/bridge_events.go
```diff
@@ -0,0 +1,144 @@
...
+		if !exists {
+			v = big.NewInt(0)
+		}
+
+		// Store the new balance in the balances map.
+		balances[event.Address] = v.Add(v, event.Coin.Amount.BigInt())
+	}
+
+	// ------------ OUTPUT ------------
+
```
tqin7, line 112: nit: should we print out something like "use below for genesis configuration"?

tqin7 (approved):

PR #280, opened 2023-09-15
[DEC-2082] update log level + add comments + add sanity checks in tests

ttl33:


protocol/app/prepare/other_msgs.go
```diff
@@ -69,7 +69,7 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 
 		// If tx contains disallowed msg, skip it.
 		if containsDisllowMsg {
-			ctx.Logger().Info(
+			ctx.Logger().Error(
```
ttl33, line 72: nit: escalating this to error since this shouldn't happen

protocol/app/msgs/all_msgs_test.go
```diff
@@ -33,6 +33,7 @@ func TestAllTypeMessages(t *testing.T) {
 
 	// Assert.
 	require.Equal(t, allTypes, lib.GetSortedKeys[sort.StringSlice](msgs.AllTypeMessages))
+	require.Len(t, msgs.AllTypeMessages, 210)
```
ttl33, line 36: not the biggest fan of using direct numbers here, i can define some constants if this bothers anyone

tqin7 (approved):

PR #282, opened 2023-09-15
[CORE-535] - Remove duplicate daemon flag for grpc.address config.

clemire:
As a P1 result from the prices outage, eliminate duplicate flag configuring gRPC address for daemons. Additionally, panic if the app starts with gRCP disabled, as for now it is required to operate. (The application would have panicked anyway if an expected daemon did not come up, but this way the error source is more clear.)

Tests:
- existing unit / integration
- make localnet-startd looks good
- built this image locally and modified container to have grpc disabled - causes panic.
- built this image locally and modified container to have grpc address of 10101 - protocol runs and netstat on the container shows port 10101 in use, 9090 not in use.

Sanity checks:
- vanilla build from main with grcp disabled in config causes daemon panics
- vanilla build from main with grpc address changed, but daemon flag unchanged, causes daemon panics.

### DIFF
--- a/protocol/app/flags/flags.go
+++ b/protocol/app/flags/flags.go
@@ -1,6 +1,7 @@
 package flags
 
 import (
+	"fmt"
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/spf13/cobra"
@@ -14,6 +15,7 @@ type Flags struct {
 
 	// Existing flags
 	GrpcAddress string
+	GrpcEnable  bool
 }
 
 // List of CLI flags.
@@ -24,6 +26,7 @@ const (
 
 	// Cosmos flags below. These config values can be set as flags or in config.toml.
 	GrpcAddress = "grpc.address"
+	GrpcEnable  = "grpc.enable"
 )
 
 // Default values.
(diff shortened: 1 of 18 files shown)


protocol/app/flags/flags.go
```diff
@@ -49,6 +59,15 @@ func AddFlagsToCmd(cmd *cobra.Command) {
 	)
 }
 
+// Validate checks that the flags are valid.
+func (f *Flags) Validate() error {
+	// Validtors must have cosmos grpc services enabled.
+	if !f.NonValidatingFullNode && !f.GrpcEnable {
+		return fmt.Errorf("grpc.enable must be set to true - validating requires gRPC server")
```
teddyding, line 66: Nit: explicitly say sth like:
`grpc.enable must be set to true when 'non-validating-full-node' flag is provided. `

protocol/app/flags/flags.go
```diff
@@ -74,5 +97,13 @@ func GetFlagValuesFromOptions(
 		result.DdTraceAgentPort = v
 	}
 
+	if v, ok := appOpts.Get(GrpcAddress).(string); ok {
+		result.GrpcAddress = v
+	}
+
+	if v, ok := appOpts.Get(GrpcEnable).(bool); ok {
```
teddyding, line 104: Will this logic be triggered if `grpc.enable` is set to `false` in the `config.toml`?

protocol/app/flags/flags.go
```diff
@@ -59,6 +78,10 @@ func GetFlagValuesFromOptions(
 		NonValidatingFullNode: DefaultNonValidatingFullNode,
 		DdAgentHost:           DefaultDdAgentHost,
 		DdTraceAgentPort:      DefaultDdTraceAgentPort,
+
+		// These are the default values from the Cosmos flags.
+		GrpcAddress: config.DefaultGRPCAddress,
```
teddyding, line 83: For my understanding, what happens when
- The validator does not pass in `grpc.addr` and `grpc.enable` in command flags
- They update these values in their `config.toml` to something that's not the default value

In this case would `flags.GrpcAddr` and `flags.GrpcEnble` different from the actual config value loaded to the app?

teddyding (approved):

PR #285, opened 2023-09-18
[CORE-565] define community treasury module account

tqin7:


protocol/x/vest/types/keys.go
```diff
@@ -15,6 +15,9 @@ const (
 
 	// VestEntryKeyPrefix is the prefix used when storing a VestEntry in the state.
 	VestEntryKeyPrefix = "vest_entry"
+
+	// CommunityTreasury defines the root string for community treasury module account.
+	CommunityTreasury = "community_treasury"
```
teddyding, line 20: Can you add a vester account for `community_vester` too?

tqin7: Sure! Is it mainly to be consistent with the `vester` / `treasury` pattern?

teddyding: Yeah, there's always a vester/treasury pair for these accounts AFAIK. Same for V3 as well.

protocol/x/vest/types/keys.go
```diff
@@ -15,6 +15,12 @@ const (
...
```
ttl33, line 20: Should we suffix these with `AccountName` to be consistent with other module account names?

protocol/app/module_accounts_test.go
```diff
@@ -29,6 +30,8 @@ func TestModuleAccountsToAddresses(t *testing.T) {
 		clobmoduletypes.InsuranceFundName:      "dydx1c7ptc87hkd54e3r7zjy92q29xkq7t79w64slrq",
 		rewardsmoduletypes.TreasuryAccountName: "dydx16wrau2x4tsg033xfrrdpae6kxfn9kyuerr5jjp",
 		rewardsmoduletypes.VesterAccountName:   "dydx1ltyc6y4skclzafvpznpt2qjwmfwgsndp458rmp",
+		vestmoduletypes.CommunityTreasury:      "dydx15ztc7xy42tn2ukkc0qjthkucw9ac63pgp70urn",
```
ttl33, line 33: To confirm, will this address change if the root `community_treasury` string changes? It's possible someone makes a mistake and changes the root string. We should be able to catch such issues

tqin7: address changes if root string changes!

protocol/x/vest/types/keys.go
tqin7: Sounds good!

PR #303, opened 2023-09-19
[CLOB-865] - disallow fok maker orders

jakob-dydx:


protocol/x/clob/types/match_with_orders.go
```diff
@@ -76,5 +76,10 @@ func (match MatchWithOrders) Validate() error {
 	if makerOrder.MustGetOrder().TimeInForce == Order_TIME_IN_FORCE_IOC {
 		return errors.New("IOC order cannot be matched as a maker order")
 	}
+
+	// Make sure the maker order is not a FOK order.
+	if makerOrder.MustGetOrder().TimeInForce == Order_TIME_IN_FORCE_FILL_OR_KILL {
+		return errors.New("FOK order cannot be matched as a maker order")
```
jiajames, line 82: very nit, can be a little more descriptive: "Order with Order_TIME_IN_FORCE_FILL_OR_KILL cannot be matched as a maker order"

jiajames (approved):

protocol/x/clob/types/match_with_orders.go
jiajames: Consider also combining this statement with the above IOC one. I think logically its better to think of maker orders not being IOC or FOK as a single idea rather than separate ones (if that makes sense)

protocol/x/clob/types/match_with_orders.go
```diff
@@ -76,5 +76,10 @@ func (match MatchWithOrders) Validate() error {
 	if makerOrder.MustGetOrder().TimeInForce == Order_TIME_IN_FORCE_IOC {
 		return errors.New("IOC order cannot be matched as a maker order")
 	}
+
+	// Make sure the maker order is not a FOK order.
+	if makerOrder.MustGetOrder().TimeInForce == Order_TIME_IN_FORCE_FILL_OR_KILL {
```
lucas-dydx, line 81: Can you add an process operations test that matches a FOK order as a maker and verify whether it would have failed after this change?

protocol/x/clob/types/match_with_orders.go
lucas-dydx: Agreed with James, also include order ID in error and use [`RequiresImmediateExecution`](https://github.com/dydxprotocol/v4-chain/blob/2168a49e61d8d2d327ea818b63cea832755c9c91/protocol/x/clob/types/order.go#L157).

lucas-dydx, line 82: Nit: define errors in https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/types/errors.go.

jakob-dydx: Yeah that's fair, just thought it might be more descriptive to have FOK/IOC in the actual error message itself. But I can change this, no problem

PR #309, opened 2023-09-19
[DEC-2084] - `lib/error` audit refactors

clemire:
- coverage was already 100%
- added more comments
- removed homecooked mock logger and replaced it with a generated mock class, then refactored test case to use expectations to validate logger calls.
- added an additinoal test case for logging plain errors.

### DIFF
--- a/protocol/lib/error/log_contextualizer.go
+++ b/protocol/lib/error/log_contextualizer.go
@@ -2,6 +2,10 @@ package error
 
 import "github.com/cometbft/cometbft/libs/log"
 
+// LogContextualizer describes an object that can add context - that is, descriptive key-value pairs, to a logger.
+// This interface is implemented by ErrorWithLogContext, which wraps some errors that are returned from the protocol,
+// and can be used to add context to log statements from the process proposal handler that helps with debugging
+// erroneous unexpected and/or non-deterministic behaviors.
 type LogContextualizer interface {
 	AddLoggingContext(logger log.Logger) log.Logger
 	Unwrap() error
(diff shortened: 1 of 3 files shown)


protocol/lib/error/log_contextualizer.go
```diff
@@ -2,6 +2,10 @@ package error
 
 import "github.com/cometbft/cometbft/libs/log"
 
+// LogContextualizer describes an object that can add context - that is, descriptive key-value pairs, to a logger.
+// This interface is implemented by ErrorWithLogContext, which wraps some errors that are returned from the protocol,
```
ttl33, line 6:
```suggestion
// This interface is implemented by ErrorWithLogContext, which wraps some errors that are returned from the protocol. For example, this can be used to add context to log statements from the ProcessProposalHandler ...
```

protocol/lib/error/logging.go
```diff
@@ -27,9 +28,10 @@ func LogErrorWithOptionalContext(
 	logger.Error(msg, "error", err)
 }
 
-// WrapErrorWithPricesSourceModuleContext wraps an error with a LogContextualizer that the spercified source module.
-// This is useful for logging the error within the process proposal handler (or any other location that uses
-// LogErrorWithOptionalContext) with metadata that can be used to identify the source of the error.
+// WrapErrorWithPricesSourceModuleContext wraps an error with a LogContextualizer that contains a key-value pair for
+// the specified source module. This is useful for logging the error within the process proposal handler (or from any
```
ttl33, line 32:
```suggestion
// the specified source module. For example, this is useful for logging the error within the process proposal handler (or from any
```

ttl33 (approved):

PR #319, opened 2023-09-20
implement governance message MsgUpdateParams in perpetuals module

tqin7:


protocol/x/perpetuals/keeper/perpetual.go
```diff
@@ -1372,10 +1370,26 @@ func (k Keeper) GetParams(ctx sdk.Context) types.Params {
...
+		return err
+	}
+	if err := k.SetPremiumVoteClampFactorPpm(cacheCtx, params.PremiumVoteClampFactorPpm); err != nil {
+		return err
+	}
+	if err := k.SetMinNumVotesPerSample(cacheCtx, params.MinNumVotesPerSample); err != nil {
+		return err
+	}
+	// Write cached store to main store.
+	writeCache()
```
teddyding, line 1391: Why this is this needed here?

protocol/x/perpetuals/keeper/perpetual.go
```diff
@@ -1372,10 +1370,26 @@ func (k Keeper) GetParams(ctx sdk.Context) types.Params {
...
-	// TODO(CORE-560): implement this method
-	return status.Errorf(codes.Unimplemented, "SetParams not implemented")
+// `SetParams` atomically sets all perpetuals module parameters in store.
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) error {
+	// Validate params.
+	if err := params.Validate(); err != nil {
+		return err
+	}
+	// Try to set params in cached store.
+	cacheCtx, writeCache := ctx.CacheContext()
```
teddyding, line 1380: Why is this and below `writeCache()` needed here? Message handlers should not be responsible for using a branched context

tqin7: I was thinking that `Params` update should succeed atomically for all three parameters. For example, without the branched context, `FundingRateClampFactorPpm` can be updated and then `PremiumVoteClampFactorPpm` update can fail. which I feel like is not the best behavior for `UpdateParams` message

In the future when we migrate `Params` to use just one key, a branched ctx is no longer needed.

teddyding: It's not the responsibility of a message handler to maintain atomicity - in this case this would be handled in `x/gov` endblocker where it uses a cached context, and only writes to cache if the message handler succeeds

tqin7: Ah that makes sense. Will update the PR!

teddyding (approved):

PR #323, opened 2023-09-20
[DEC-2081] Audit app/ante/sigverify.go

lcwik:
Note that I dropped tests that just ran over the SDKs gas cost and set pubkey decorators as these we don't need to maintain. This allowed me to remove some additional methods.

I also confirmed that the sigverification tests that the SDK has hadn't changed and we cover the new functionality with our own tests that we added.


protocol/app/ante/sigverify.go
```diff
@@ -34,27 +33,6 @@ func NewSigVerificationDecorator(
 	}
 }
 
-// OnlyLegacyAminoSigners checks SignatureData to see if all
-// signers are using SIGN_MODE_LEGACY_AMINO_JSON. If this is the case
-// then the corresponding SignatureV2 struct will not have account sequence
-// explicitly set, and we should skip the explicit verification of sig.Sequence
-// in the SigVerificationDecorator's AnteHandler function.
-func OnlyLegacyAminoSigners(sigData signing.SignatureData) bool {
```
ttl33, line 42: To double confirm, we are deleting this because we can directly use CosmosSDK's method?

protocol/app/ante/sigverify_test.go
```diff
@@ -25,160 +19,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestSetPubKey(t *testing.T) {
```
ttl33, line 28: Noob question: why is it okay to drop these tests?

protocol/app/ante/sigverify_test.go
```diff
@@ -25,160 +19,6 @@ import (
...
-	// Require that all accounts have pubkey set after Decorator runs
-	for i, addr := range addrs {
-		pk, err := suite.AccountKeeper.GetPubKey(ctx, addr)
-		require.NoError(t, err, "Error on retrieving pubkey from account")
-		require.True(t, pubs[i].Equals(pk),
-			"Wrong Pubkey retrieved from AccountKeeper, idx=%d\nexpected=%s\n     got=%s", i, pubs[i], pk)
-	}
-}
-
-func TestConsumeSignatureVerificationGas(t *testing.T) {
```
ttl33, line 71: Same question as above, why can we drop this test?

protocol/app/ante/sigverify.go
lcwik: Yes

protocol/app/ante/sigverify_test.go
lcwik: We are verifying functionality of stuff imported from `sdkante` which is from the Cosmos SDK and not verifying stuff that we modified since this test file was a duplicate of the Cosmos SDK test file for this space. The only thing that was modified was the imports.

Note that we don't have a forked version of the `SetPubKey` ante handler.

protocol/app/ante/sigverify_test.go
lcwik: Ditto as a above but in this case we don't have a fork of the `SigVerificationGasConsumer` ante handler.

ttl33 (approved):

PR #327, opened 2023-09-20
[DEC-2081] Audit app/middleware/middleware.go

lcwik:
Changes:
* swap to use test cases
* ensure that the middle ware logger is restored after the test finishes

### DIFF
--- a/protocol/app/middleware/middleware_test.go
+++ b/protocol/app/middleware/middleware_test.go
@@ -1,68 +1,73 @@
 package middleware_test
 
 import (
+	"bytes"
 	"fmt"
+	"github.com/cometbft/cometbft/libs/log"
 	"testing"
 
 	"github.com/dydxprotocol/v4-chain/protocol/app/middleware"
-	"github.com/dydxprotocol/v4-chain/protocol/mocks"
-	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
 func TestRunTxPanicLoggingMiddleware(t *testing.T) {
-	logger := &mocks.Logger{}
-	middleware.Logger = logger
+	tests := map[string]struct {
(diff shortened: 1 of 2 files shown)


protocol/app/middleware/middleware_test.go
```diff
@@ -1,68 +1,73 @@
...
+			}()
 
-	func() {
-		defer func() {
-			if r := recover(); r != nil {
-				handler := middleware.NewRunTxPanicLoggingMiddleware()
-				err := handler(r)
-				require.Nil(t, err)
+			if tc.expectedLogs == nil {
+				require.Equal(t, "", buf.String())
```
ttl33, line 66: nit: does `require.Empty()` work here?

protocol/app/middleware/middleware_test.go
```diff
@@ -1,68 +1,73 @@
...
-		"Error",
-		"runTx panic'ed with a problem",
-		"stack trace",
-		mock.Anything,
-	).Return(nil).Once()
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			// Restore the old logger after the test runs
+			oldLogger := middleware.Logger
+			defer func() { middleware.Logger = oldLogger }()
```
ttl33, line 49: for my understanding, why do we need to restore the logger? is it to reset the logger for each test run?

ttl33 (approved):

protocol/app/middleware/middleware_test.go
lcwik: middleware.Logger is a global variable, clarified in comment

protocol/app/middleware/middleware_test.go
lcwik: It does, updated.

PR #330, opened 2023-09-20
[DEC-2118] - Remove panics from DelayMessage

clemire:
[Slack thread](https://dydx-team.slack.com/archives/C05FTK5D8QG/p1695049275814479)

### DIFF
--- a/protocol/x/delaymsg/keeper/msg_server.go
+++ b/protocol/x/delaymsg/keeper/msg_server.go
@@ -2,7 +2,9 @@ package keeper
 
 import (
 	"context"
+	errorsmod "cosmossdk.io/errors"
 	"fmt"
+	"github.com/dydxprotocol/v4-chain/protocol/daemons/pricefeed/client/constants"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/x/delaymsg/types"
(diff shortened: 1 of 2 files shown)


protocol/x/delaymsg/keeper/msg_server.go
```diff
@@ -23,21 +25,41 @@ func (k msgServer) DelayMessage(
...
-		))
+		)
+		return nil, errorsmod.Wrapf(
+			types.ErrInvalidInput,
+			"%v is not recognized as a valid authority for sending messages",
+			msg.GetAuthority(),
+		)
 	}
 
 	sdkMsg, err := msg.GetMessage()
```
clemire, line 40: During manual testing to see if I could provoke an error here for testing I discovered nil messages cause panics, filed audit task: https://linear.app/dydx/issue/CORE-593/duplicate-validatebasic-into-msgserver-delaymessage-so-that-getmessage

teddyding (approved):

protocol/x/delaymsg/keeper/msg_server.go
teddyding: Can you explain a bit further why it panics? Does it panic [here](https://github.com/dydxprotocol/v4-chain/blob/8b749db88ca08b3514150d5b24ae67651d22073d/protocol/x/delaymsg/types/msg_delay_message.go#L35)?

clemire: yes, I think so. This issue should be caught by ValidateBasic.

PR #333, opened 2023-09-21
Comment out under-collateralized removal validation

jayy04:


protocol/x/clob/keeper/process_operations.go
```diff
@@ -394,6 +367,35 @@ func (k Keeper) PersistOrderRemovalToState(
 				orderToRemove.GetBaseQuantums(),
 			)
 		}
+	// TODO - uncomment when reduce only orders are enabled. Order Removals of this type will fail ValidateBasic.
```
lucas-dydx, line 370: Suggestion: add the ticket?

```suggestion
	// TODO (CLOB-877) - uncomment when reduce only orders are enabled. Order Removals of this type will fail ValidateBasic.
```

lucas-dydx (approved):

PR #337, opened 2023-09-21
[DEC-2083] Rename encoding config function

ttl33:


lcwik (approved):
Overall LGTM

Any reason why we don't just merge the `params/encoding.go` logic here?, not seeing how having them separate helps.

ttl33:
> Overall LGTM
> 
> Any reason why we don't just merge the `params/encoding.go` logic here?, not seeing how having them separate helps.

No specific reason. Was mostly a copy-paste from the source. I merged it into the `app/encoding.go` file!

lcwik (approved):

PR #342, opened 2023-09-21
Re-enable evidence module

teddyding:
* [Context](https://dydx-team.slack.com/archives/C03FY21CSF6/p1695325960116449)
* Compare with [PR that removed it](https://github.com/dydxprotocol/v4/pull/938/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R890)

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -66,6 +66,9 @@ import (
 	distr "github.com/cosmos/cosmos-sdk/x/distribution"
 	distrkeeper "github.com/cosmos/cosmos-sdk/x/distribution/keeper"
 	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
+	"github.com/cosmos/cosmos-sdk/x/evidence"
+	evidencekeeper "github.com/cosmos/cosmos-sdk/x/evidence/keeper"
+	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
 	"github.com/cosmos/cosmos-sdk/x/feegrant"
 	feegrantkeeper "github.com/cosmos/cosmos-sdk/x/feegrant/keeper"
 	feegrantmodule "github.com/cosmos/cosmos-sdk/x/feegrant/module"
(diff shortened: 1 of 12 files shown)


protocol/app/app.go
```diff
@@ -234,6 +237,7 @@ type App struct {
 	ParamsKeeper     paramskeeper.Keeper
 	// IBC Keeper must be a pointer in the app, so we can SetRouter on it correctly
 	IBCKeeper             *ibckeeper.Keeper
+	EvidenceKeeper        evidencekeeper.Keeper
```
ttl33, line 240: I think you also need to add `x/evidence` to module basic [here](https://github.com/dydxprotocol/v4-chain/blob/6d30844950ba73e1d1003e11f320e1ee3529ca70/protocol/app/basic_manager/basic_manager.go#L49)

protocol/app/app.go
```diff
@@ -66,6 +66,9 @@ import (
 	distr "github.com/cosmos/cosmos-sdk/x/distribution"
 	distrkeeper "github.com/cosmos/cosmos-sdk/x/distribution/keeper"
 	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
+	"github.com/cosmos/cosmos-sdk/x/evidence"
+	evidencekeeper "github.com/cosmos/cosmos-sdk/x/evidence/keeper"
+	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
```
ttl33, line 71: Given that the `x/evidence` defines a few msg txs ([code](https://github.com/cosmos/cosmos-sdk/blob/79a64d07d0bf2964b3c38f11241c90a84b2f1262/proto/cosmos/evidence/v1beta1/tx.proto)), I suspect you will need to update a few files and tests in `app/msgs` [here](https://github.com/dydxprotocol/v4-chain/tree/6d30844950ba73e1d1003e11f320e1ee3529ca70/protocol/app/msgs).

Can you run the unit tests there and see if you need to update anything there?

jayy04 (approved):

protocol/app/app.go
teddyding: Fixed tests

ttl33 (approved):

PR #343, opened 2023-09-21
[IND-372] Add caching / rate-limiting logic to screen endpoint.

vincentwschau:
Add logic to `/screen` endpoint to either retrieve the compliance data from the provider for new addresses / addresses with out-dated data or gets the data from the database.

Added rate-limiting logic directly into the controller as the custom logic depends on finding the compliance data in the database first.

Added `upsert` function for compliance data table.

### DIFF
--- a/indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
@@ -1,46 +1,78 @@
-import { stats } from '@dydxprotocol-indexer/base';
+import { logger, stats } from '@dydxprotocol-indexer/base';
 import { ComplianceClientResponse } from '@dydxprotocol-indexer/compliance';
+import { ComplianceDataFromDatabase, ComplianceTable } from '@dydxprotocol-indexer/postgres';
 import express from 'express';
 import { checkSchema, matchedData } from 'express-validator';
+import { DateTime } from 'luxon';
 import {
   Controller, Get, Query, Route,
 } from 'tsoa';
 
-import { getReqRateLimiter } from '../../../caches/rate-limiters';
+import {
+  getReqRateLimiter,
+  screenProviderLimiter,
+  screenProviderGlobalLimiter,
+} from '../../../caches/rate-limiters';
 import config from '../../../config';
 import { placeHolderProvider } from '../../../helpers/compliance/compliance-clients';
 import { complianceCheck } from '../../../lib/compliance-check';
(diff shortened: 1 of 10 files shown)


indexer/services/comlink/src/controllers/api/v4/height-controller.ts
```diff
@@ -44,8 +44,8 @@ router.get(
       return res.send(response);
     } catch (error) {
       return handleControllerError(
-        'FillsController GET /',
-        'Fills error',
+        'HeightController GET /',
+        'Height error',
         error,
```
vincentwschau, line 49: Random typo I noticed and bundled into this PR.

indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
```diff
@@ -1,46 +1,78 @@
...
+import { create4xxResponse, handleControllerError } from '../../../lib/helpers';
+import { getIpAddr, rateLimiterMiddleware } from '../../../lib/rate-limit';
 import { handleValidationErrors } from '../../../request-helpers/error-handler';
 import ExportResponseCodeStats from '../../../request-helpers/export-response-code-stats';
 import { ComplianceRequest, ComplianceResponse } from '../../../types';
 
 const router: express.Router = express.Router();
 const controllerName: string = 'compliance-controller';
+const uncachedQueryPoints: number = 1;
+const globalRateLimitKey: string = 'screenQueryProviderGlobal';
```
Christopher-Li, line 29: nit: global const should be all caps?

indexer/services/comlink/src/controllers/api/v4/compliance-controller.ts
```diff
@@ -81,4 +124,23 @@ router.get(
   },
 );
 
+async function checkRateLimit(ipAddress: string) {
+  try {
+    await Promise.all([
+      screenProviderLimiter.consume(ipAddress, uncachedQueryPoints),
+      screenProviderGlobalLimiter.consume(globalRateLimitKey, uncachedQueryPoints),
```
Christopher-Li, line 131: iiuc,  `screenProviderLimiter` is ensuring a single IP address doesn't make this request too often and GlobalLimiter is ensuring that this endpoint overall doesn't get used too often right? Can we document this in the code? EIther here or in  `rate-limiter.ts`

vincentwschau: Added comments to `rate-limiter.ts`.

dydxwill (approved):

PR #344, opened 2023-09-21
[IND-370] Run task to update compliance data.

vincentwschau:
See [spec](https://www.notion.so/dydx/OFAC-Checks-and-Geo-blocking-d4f749e8f4314fa1b258cd52033fb30d?pvs=4#2fe4fcf8a39642fca90b21b25d3e6e48).

### DIFF
--- a/indexer/services/roundtable/src/config.ts
+++ b/indexer/services/roundtable/src/config.ts
@@ -64,6 +64,10 @@ export const configSchema = {
     // 1 hour
     default: 3_600_000,
   }),
+  LOOPS_INTERVAL_MS_UPDATE_COMPLIANCE_DATA: parseInteger({
+    // 5 minutes
+    default: 300_000,
+  }),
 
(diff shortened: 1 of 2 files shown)


indexer/services/roundtable/src/config.ts
```diff
@@ -64,6 +64,10 @@ export const configSchema = {
     // 1 hour
     default: 3_600_000,
   }),
+  LOOPS_INTERVAL_MS_UPDATE_COMPLIANCE_DATA: parseInteger({
+    // 5 minutes
+    default: 300_000,
```
Christopher-Li, line 69: nit: maybe not for this PR, but at some point we should just create a bunch for constants like `5 * FIVE_MINUTES_IN_MS` which would be cleaner

Christopher-Li (approved):

dydxwill (approved):

PR #345, opened 2023-09-22
v0.3.0 upgrade handler

jayy04:


protocol/app/upgrades/v0.3.0/upgrade.go
```diff
@@ -0,0 +1,28 @@
...
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	"github.com/dydxprotocol/v4-chain/protocol/app/upgrades"
+)
+
+var (
+	Upgrade = upgrades.Upgrade{
+		UpgradeName: UpgradeName,
+		StoreUpgrades: store.StoreUpgrades{
+			Added: []string{
+				evidencetypes.ModuleName,
```
roy-dydx, line 17: Technically should be `StoreKey` right? but it's the same I think

roy-dydx (approved):

proto/dydxprotocol/indexer/events/events.proto
```diff
@@ -107,7 +107,18 @@ message MarketCreateEventV1 {
...
+  MarketBaseEventV1 base = 1;
+
+  // Static value. The exponent of the price.
+  // For example if Exponent == -5 then a `exponent_price` of 1,000,000,000
+  // represents $10,000. Therefore 10 ^ Exponent represents the smallest
+  // price step (in dollars) that can be recorded.
+  //
+  // Updates to this value are allowed, but are unsafe - they can cause
+  // the network to have innacurate prices for up to ~1 minute.
+  sint32 exponent = 2;
```
teddyding, line 120: What's the reasoning behind adding `exponent` as a field in `MarketModifyEventV1` as opposed to `MarketBaseEventV1`?

Are we using the proto definition permanently, or just for this upgrade?

vincentwschau: Can we note that this is just for upgrade testing as well?

clemire: I was trying to minimize code change footprints - this way, I only needed to update emitted MarketModifyEvents in the protocol. I agree that putting it in BaseEvent is better, but my understanding was that this would be not permanently included in the repository.

protocol/app/upgrades/v0.3.0/upgrade.go
```diff
@@ -0,0 +1,77 @@
...
+
+		pepePerpetual.Params.AtomicResolution = 1
+		perpetualsKeeper.UnsafeSetPerpetual(ctx, pepePerpetual)
+
+		// Update market price exponent for PEPE.
+		pepePrice, err := pricesKeeper.GetMarketPrice(ctx, PEPE_ID)
+		if err != nil {
+			panic(err)
+		}
+		pepePrice.Exponent = -16
```
clemire, line 61: @jayy04 , if we're changing the exponent from -13 to -16, couldn't we multiply the old price by 1000 and just set that?

PR #346, opened 2023-09-22
Height gate missing index price err log

teddyding:
Do not log error/increment stats if missing index price is observed during the first 20 blocks (price daemon initialization period)
[Context](https://dydx-team.slack.com/archives/C05AWN29RS4/p1695396499623849?thread_ts=1695391605.690219&cid=C05AWN29RS4)
Safe to backport - no application behavior change


clemire (approved):
I'm fine with this as a solution because it fixes our error in the majority of cases, but FYI this will not fix the issue of missing index prices due to markets recently added via governance. I think there's no need to work on that today if you're pressed for time, but if you want to commit here, can you file a follow-up task for this?

PR #349, opened 2023-09-22
Speed up order fill handler by creating an index on [subaccountId, perpetualId, createdAtHeight] in perpetual positions table

dydxwill:
Speed up order fill handler by creating an index on subaccountId, perpetualId, createdAtHeight in perpetual positions table

### DIFF
--- a/indexer/packages/postgres/src/db/migrations/migration_files/20230922141210_create_perpetual_positions_subaccount_perpetual_created_at_index.ts
+++ b/indexer/packages/postgres/src/db/migrations/migration_files/20230922141210_create_perpetual_positions_subaccount_perpetual_created_at_index.ts
@@ -1,0 +1,18 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/quotes
+  await knex.raw(`
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS "perpetual_positions_subaccount_perpetual_created_at_index" ON "perpetual_positions" ("subaccountId", "perpetualId", "createdAtHeight");
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.raw(`
+    DROP INDEX CONCURRENTLY IF EXISTS "perpetual_positions_subaccount_perpetual_created_at_index";
+  `);
+}
+
+export const config = {
+  transaction: false,
+};


indexer/packages/postgres/src/db/migrations/migration_files/20230922141210_create_perpetual_positions_subaccount_perpetual_created_at_index.ts
```diff
@@ -0,0 +1,18 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/quotes
+  await knex.raw(`
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS "perpetual_positions_subaccount_perpetual_created_at_index" ON "perpetual_positions" ("subaccountId", "perpetualId", "createdAtHeight");
```
vincentwschau, line 6: nit: Do you think it'd be possible to make this a partial index on the status of the position? E.g. we wouldn't touch `CLOSED` positions so there's no need to keep them in the index? Just want to ensure our indices don't grow too large.

Christopher-Li, line 6: do we have any idea how much this will improve latency? Just like maybe even anecdotally, if you apply this index to staging and perform a query to find the most recent perpetual position for a subaccount and perpetual id, how does it affect latency?

dydxwill: in both order fill/liquidation handler, we grab the latest perpetual position like so: 
    SELECT * INTO perpetual_position_record FROM perpetual_positions WHERE "subaccountId" = subaccount_uuid
                                                                       AND "perpetualId" = perpetual_market_record."id"
                                                                       ORDER BY "createdAtHeight" DESC;

dydxwill: not sure, we didn't have performance insights enabled in RDS and we only see latency increase after the chain has been running for a while. Staging has been only running for a day, so I doubt we'll see any meaningful latency differences there. Fore reference, the query we are running in the handler is:

SELECT * INTO perpetual_position_record FROM perpetual_positions WHERE "subaccountId" = subaccount_uuid
AND "perpetualId" = perpetual_market_record."id"
ORDER BY "createdAtHeight" DESC;

vincentwschau: Would that work or would it break the logic?

vincentwschau: What does a `explain analyze` for such a query show today?

vincentwschau: We already have an index on `subaccount id, createdAtHeight` so this new index would reduce the # of rows scanned by a multiple of how many perpetual ids exist. That could be good, but we should also remove the `subaccountId, createdAtHeight` index if we're adding this additional index.

dydxwill: I don't think we are guaranteed that the last perpetual position is necessarily `OPEN` - so that would break the existing logic.

dydxwill: explain analyze select * from perpetual_positions where "subaccountId"='8772e9ed-516c-5402-9118-7f032e305a5e' and "perpetualId"=29 order by "createdAtHeight" desc;

before index creation:

```
 Sort  (cost=78.06..78.08 rows=9 width=158) (actual time=0.508..0.510 rows=25 loops=1)
   Sort Key: "createdAtHeight" DESC
   Sort Method: quicksort  Memory: 31kB
   ->  Bitmap Heap Scan on perpetual_positions  (cost=68.78..77.92 rows=9 width=158) (actual time=0.454..0.497 rows=25 loops=1)
         Recheck Cond: (("subaccountId" = '8772e9ed-516c-5402-9118-7f032e305a5e'::uuid) AND ("perpetualId" = 29))
         Heap Blocks: exact=23
         ->  BitmapAnd  (cost=68.78..68.78 rows=9 width=0) (actual time=0.450..0.450 rows=0 loops=1)
               ->  Bitmap Index Scan on perpetual_positions_subaccountid_createdatheight_index  (cost=0.00..5.72 rows=307 width=0) (actual time=0.037..0.038 rows=346 loops=1)
                     Index Cond: ("subaccountId" = '8772e9ed-516c-5402-9118-7f032e305a5e'::uuid)
               ->  Bitmap Index Scan on perpetual_positions_perpetualid_status_index  (cost=0.00..62.81 rows=4185 width=0) (actual time=0.397..0.397 rows=4503 loops=1)
                     Index Cond: ("perpetualId" = 29)
 Planning Time: 0.350 ms
 Execution Time: 0.542 ms
```

after index creation:

```
 Index Scan Backward using perpetual_positions_subaccount_perpetual_created_at_index on perpetual_positions  (cost=0.42..11.62 rows=10 width=158) (actual time=0.024..0.048 rows=25 loops=1)
   Index Cond: (("subaccountId" = '8772e9ed-516c-5402-9118-7f032e305a5e'::uuid) AND ("perpetualId" = 29))
 Planning Time: 0.439 ms
 Execution Time: 0.081 ms
(4 rows)```

dydxwill: for the index removal, do we not ever query perpetual_positions table on these 2 columns alone? in any case, I'd prefer to do that in a separate PR.

dydxwill: we do query on `subaccountId`,`createdAtHeight` in `perpetual-positions-controller`.

vincentwschau: If we change the ordering of the index to `subaccountId, createdAtHeight, perpetualId` then it can be used for this query as well and have a single index rather than 2.

Christopher-Li (approved):

indexer/packages/postgres/src/db/migrations/migration_files/20230922141210_create_perpetual_positions_subaccount_perpetual_created_at_index.ts
dydxwill: done

vincentwschau (approved):

PR #354, opened 2023-09-22
Only iterate over a fixed number of subaccounts per loop iteration

jayy04:


jakob-dydx (approved):

protocol/daemons/liquidation/client/client.go
```diff
@@ -16,6 +17,8 @@ import (
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+var nextKeyToFetch []byte
```
lucas-dydx, line 20: Suggestion: add a comment noting what this is used for (pagination)

protocol/daemons/liquidation/client/client.go
```diff
@@ -229,7 +204,7 @@ func SendLiquidatableSubaccountIds(
 	return nil
 }
 
-func getSubaccountsFromKey(
+func GetSubaccountsFromKey(
```
lucas-dydx, line 207: Suggestion: add a GoDoc comment

protocol/daemons/liquidation/client/client.go
lucas-dydx, line 20: Small suggestion: I think defining this as a variable [here](https://github.com/dydxprotocol/v4-chain/blob/a62e329fe1dc6befcc53caf9803abf4125160697/protocol/daemons/liquidation/client/client.go#L59) and then passing it in as a function argument to `RunLiquidationDaemonTaskLoop` would avoid needing a global variable and be cleaner

lucas-dydx (approved):

PR #357, opened 2023-09-22
[DEC-2079] - Comments, test coverage for daemons/* (minus bridge, liquidations)

clemire:
1 of multiple PRs

This PR covers `daemons/configs`, `daemons/constants`, `daemons/flags`, and `daemons/pricefeed/client/{constants,handler,price_encoder,price_fetcher}`.

This PR:
- adds a few comments to shared daemon flags.
- increases test coverage for `price_encoder`, a critical piece of functionality for the price daemon. the `sub_task_runner.go` file, which also has lowish coverage, is actually adequately tested where the coverage tool highlights missing coverage via integration tests.
- increases test coverage for common daemon server code.


protocol/daemons/pricefeed/client/price_encoder/price_encoder_test.go
```diff
@@ -272,3 +284,135 @@ func TestConvertPriceUpdate_Mixed(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdatePrice_Failure(t *testing.T) {
```
clemire, line 288: Improve price_encoder coverage.

protocol/daemons/constants/pricefeed.go
```diff
@@ -1,9 +1,9 @@
 package constants
```
clemire, line 1: rename file because all constants are pricefeed-specific.

protocol/daemons/flags/flags.go
```diff
@@ -22,26 +22,41 @@ const (
 	FlagLiquidationDaemonSubaccountPageLimit = "liquidation-daemon-subaccount-page-limit"
 )
 
+// Shared flags contains configuration flags shared by all daemons.
```
clemire, line 25: Document flags.

protocol/daemons/pricefeed/client/price_encoder/price_encoder_test.go
tqin7: nit: maybe a comment here like: failure within grace period is an Info log, after grace period should be Error log

tqin7 (approved):

protocol/daemons/pricefeed/client/price_encoder/price_encoder_test.go
clemire: Updated the inline comment below!

PR #359, opened 2023-09-23
liquidation matches stats for quote size / quote fill amounts and Negative TNC matches

jakob-dydx:
- adds metric for tracking liquidation orders for subaccounts with negative TNC in DeliverTx
- adds metric for tracking total quote quantums for liquidation orders and also the total filled amount in quote quantums


protocol/x/clob/keeper/liquidations.go
```diff
@@ -452,6 +452,28 @@ func (k Keeper) GetFillablePrice(
 		return nil, err
 	}
 
+	// stat liquidation order for negative TNC when in DeliverTx
+	if tncBig.Sign() < 0 && !ctx.IsCheckTx() && !ctx.IsReCheckTx() {
```
lucas-dydx, line 456: Why don't we want to also stat this in `PrepareCheckState`?

Wouldn't it help determine whether some liq daemons are seeing negative TNC subaccounts in a block?

protocol/x/clob/keeper/process_operations.go
```diff
@@ -566,6 +581,19 @@ func (k Keeper) PersistMatchLiquidationToState(
 				),
 			)
 		}
+
+		filledQuoteQuantums := new(big.Int)
+		filledQuoteQuantums.Mul(
```
lucas-dydx, line 586:
```suggestion
		filledQuoteQuantums := new(big.Int).Mul(
```

protocol/x/clob/types/match_perpetual_liquidation.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+import (
+	gometrics "github.com/armon/go-metrics"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+)
+
+func (m MatchPerpetualLiquidation) GetMetricLabels(callback string) []gometrics.Label {
```
lucas-dydx, line 8: small nit: GoDoc comment

protocol/x/clob/types/match_perpetual_liquidation.go
```diff
@@ -0,0 +1,23 @@
...
+	return []gometrics.Label{
+		metrics.GetLabelForStringValue(
+			metrics.Callback,
```
lucas-dydx, line 11: Why add this as a label? Seems like it should go in metric name since it's always `DeliverTx`

protocol/x/clob/keeper/liquidations.go
jakob-dydx: My reasoning was that this function can be called twice per liquidation in PrepareCheckState which would make the metric confusing and difficult to interpret. I don't think we'd be able to meaningfully compare the PrepareCheckState stats with the DeliverTx stats

protocol/x/clob/keeper/liquidations.go
```diff
@@ -452,6 +452,28 @@ func (k Keeper) GetFillablePrice(
...
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.Liquidations, metrics.LiquidationMatchNegativeTNC},
+			1,
+			[]gometrics.Label{
+				metrics.GetLabelForStringValue(
+					metrics.Callback,
+					metrics.DeliverTx,
+				),
+				metrics.GetLabelForStringValue(
```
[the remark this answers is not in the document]
jakob-dydx: subaccount owner cardinality is infinite, but this is only emitted for subaccounts with negative TNC. Don't imagine this metric will be emitted very frequently and when it is it would be helpful to have this information IMO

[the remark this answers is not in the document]
jakob-dydx: LMK if you think I should remove though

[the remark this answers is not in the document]
jakob-dydx: maybe I can leave it in for this but for the liquidation notional quote quantums stat I can leave it out. That metric will be emitted frequently

protocol/x/clob/types/match_perpetual_liquidation.go
jakob-dydx: removing this

protocol/x/clob/keeper/process_operations.go
```diff
@@ -566,6 +581,18 @@ func (k Keeper) PersistMatchLiquidationToState(
 				),
 			)
 		}
+
+		filledQuoteQuantums := new(big.Int).Mul(
+			new(big.Int).Div(matchWithOrders.FillAmount.ToBigInt(), new(big.Int).SetUint64(matchLiquidation.TotalSize)),
```
lucas-dydx, line 586: For my understanding, isn't `matchWithOrders.FillAmount <= matchLiquidation.TotalSize` always true?

Due to integer division rounding, does that mean this number will be either 0 or 1?

protocol/x/clob/keeper/liquidations.go
lucas-dydx: What happens if there are negative TNC subaccounts but they can't be liquidated, and insurance fund balance is too high to deleverage them? Wouldn't we want to know how many negative TNC subaccounts existed if that was the case?

Can we just return the subaccount TNC from `GetFillablePrice` as well? Would that allow us to stat this in `PrepareCheckState`?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -452,6 +452,24 @@ func (k Keeper) GetFillablePrice(
...
+	// stat liquidation order for negative TNC when in DeliverTx
+	if tncBig.Sign() < 0 && !ctx.IsCheckTx() && !ctx.IsReCheckTx() {
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.Liquidations, metrics.LiquidationMatchNegativeTNC, metrics.DeliverTx},
+			1,
+			[]gometrics.Label{
+				metrics.GetLabelForStringValue(
+					metrics.SubaccountOwner,
+					subaccountId.Owner,
+				),
```
lucas-dydx, line 464: IMO if you want to leave this metric label as-is, would recommend one of the following approaches:
- Add a comment noting that metric is infinite cardinality but it was left here because it's useful for observability.
- Emit an info log instead.

I'm leaning towards 2nd approach but no strong opinion.

protocol/x/clob/keeper/process_operations.go
jakob-dydx: Yes thank you, I should be using Quo not Div

jakob-dydx: good callout, updating to use big.Rat and Quo to get the proper result here:
```
fillAmountToTotalSizeRat := new(big.Rat).Quo(
			new(big.Rat).SetUint64(matchWithOrders.FillAmount.ToUint64()),
			new(big.Rat).SetUint64(matchLiquidation.TotalSize),
		)
		filledQuoteQuantums := lib.BigRatRound(
			new(big.Rat).Mul(
				fillAmountToTotalSizeRat,
				new(big.Rat).SetInt(notionalQuoteQuantums),
			),
			true,
		)
		telemetry.IncrCounterWithLabels(
			[]string{types.ModuleName, metrics.LiquidationOrderNotionalQuoteQuantums, metrics.Filled, metrics.DeliverTx},
			float32(filledQuoteQuantums.Uint64()),
			matchLiquidation.GetMetricLabels(),
		)
```

protocol/x/clob/keeper/liquidations.go
jakob-dydx: will update to info log

protocol/x/clob/keeper/process_operations.go
```diff
@@ -566,6 +581,25 @@ func (k Keeper) PersistMatchLiquidationToState(
...
+
+		// Stat fill amount in quote quantums as ratio of notional quote quantums
+		fillAmountToTotalSizeRat := new(big.Rat).Quo(
+			new(big.Rat).SetUint64(matchWithOrders.FillAmount.ToUint64()),
+			new(big.Rat).SetUint64(matchLiquidation.TotalSize),
+		)
+		filledQuoteQuantums := lib.BigRatRound(
+			new(big.Rat).Mul(
+				fillAmountToTotalSizeRat,
+				new(big.Rat).SetInt(notionalQuoteQuantums),
```
lucas-dydx, line 593: What happens if `notionalQuoteQuantums` is negative?

Note that the return value of [`GetNetNotional`](https://github.com/dydxprotocol/v4-chain/blob/6d7544d23fd9fc20657089d70c616f8882f675ea/protocol/x/perpetuals/keeper/perpetual.go#L744) is signed, so it can't be cast to a `uint64` right?

protocol/x/clob/keeper/process_operations.go
```diff
@@ -532,6 +532,21 @@ func (k Keeper) PersistMatchLiquidationToState(
...
+		matchLiquidation.PerpetualId,
+		new(big.Int).SetUint64(matchLiquidation.TotalSize),
+	)
+	if err != nil {
+		return err
+	}
+
+	telemetry.IncrCounterWithLabels(
+		[]string{types.ModuleName, metrics.LiquidationOrderNotionalQuoteQuantums, metrics.DeliverTx},
+		float32(notionalQuoteQuantums.Uint64()),
```
lucas-dydx, line 546: For my understanding wouldn't this overflow for longs?

protocol/x/clob/keeper/process_operations.go
jakob-dydx: Will update to take the absolute value of notionalQuoteQuantums

protocol/x/clob/keeper/process_operations.go
jakob-dydx: Changing this to take absolute value of notional quote quantums

protocol/x/clob/keeper/process_operations.go
```diff
@@ -532,6 +532,22 @@ func (k Keeper) PersistMatchLiquidationToState(
...
 
+	notionalQuoteQuantums, err := k.perpetualsKeeper.GetNetNotional(
+		ctx,
+		matchLiquidation.PerpetualId,
+		new(big.Int).SetUint64(matchLiquidation.TotalSize),
+	)
+	if err != nil {
+		return err
+	}
+	notionalQuoteQuantums.Abs(notionalQuoteQuantums)
```
lucas-dydx, line 543: Small suggestion, imo defining a new variable is more clear:

```suggestion
	absNotionalQuoteQuantums := new(big.Int).Abs(notionalQuoteQuantums)
```

protocol/x/clob/keeper/liquidations.go
```diff
@@ -452,6 +452,35 @@ func (k Keeper) GetFillablePrice(
 		return nil, err
 	}
 
+	// stat liquidation order for negative TNC
+	// CLOB(906) Prevent duplicated stat emissions for liquidation orders in PrepareCheckState.
```
lucas-dydx, line 456: Small nit (makes it easier to lookup all TODOs and / or CLOB-906 in the codebase):

```suggestion
	// TODO(CLOB-906): Prevent duplicated stat emissions for liquidation orders in PrepareCheckState.
```

protocol/x/clob/keeper/liquidations.go
```diff
@@ -452,6 +452,35 @@ func (k Keeper) GetFillablePrice(
...
+					metrics.Callback,
+					callback,
+				),
+			},
+		)
+
+		ctx.Logger().Info(
+			"GetFillablePrice: Subaccount has negative TNC. SubaccountId: %+v, TNC: %+v",
+			subaccountId,
+			tncBig,
```
lucas-dydx, line 480: Should you cast this to a string?

lucas-dydx (approved):

PR #362, opened 2023-09-25
Add backport rule for mergify bot

roy-dydx:
This rule will allow users to backport to the latest release branch.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,15 @@
+defaults:
+  actions:
+    backport:
+      assignees:
+        - "{{ author }}"
+
+pull_request_rules:
+  - name: backport to release/protocol/v0.3.x branch
+    conditions:
+      - base=main
+      - label=backport/v0.3.x
+    actions:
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -0,0 +1,15 @@
+defaults:
+  actions:
+    backport:
+      assignees:
+        - "{{ author }}"
+
+pull_request_rules:
```
teddyding, line 7: Should we add `"#approved-reviews-by>=1"` in `queue_rules` as suggested in in mergify [doc](https://docs.mergify.com/merge-queue/setup)?

It's also used by [Osmosis](https://github.com/osmosis-labs/osmosis/blob/6f033c5aad6edf4e07f15e616532986eec55b6da/.github/mergify.yml#L7)

roy-dydx: We can add a merge queue later, but for now this is just to set up the backport PRs

teddyding (approved):

PR #367, opened 2023-09-25
[CLOB-902] - disallow fok order fills beyond one block

jakob-dydx:
This PR prevents IOC orders from being filled beyond the block they are originally filled in

The proposed solution in https://linear.app/dydx/issue/CLOB-902/return-error-when-attempting-to-match-partially-filled-ioc-orders suggests we modify `ProcessSingleMatch` but this is not a viable solution because ProcessSingleMatch is called once for each fill. If an IOC order had 2 maker fills, the first one would pass the check but the second would fail because there is a fill amount for the IOC order in state already. Because of this, I created a `statefulValidateMatchTaker` function for use in DeliverTx which we can extend with more validation in the future if needed.

CheckTx validation is done in `validateNewOrder` as suggested in the ticket


protocol/x/clob/e2e/order_matches_test.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// BlockAdvancement holds orders and matches to be placed in a block. Using this struct and building
+// the ops queue with the getOperationsQueue helper function allows us to build the operations queue
+// without going through CheckTx and, therefore, not affect the local memclob state. This also allows us to propose
+// an invalid set of operations that an honest validator would not generate.
+type BlockAdvancement struct {
+	ordersAndMatches []interface{} // should hold Order and OperationRaw. OperationRaw are assumed to be matches.
+	expectedError    string
+}
```
lucas-dydx, line 27: Can we add this struct definition in a testutil file? It seems like this could be used across multiple types of tests.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -455,8 +455,16 @@ func (k Keeper) PersistMatchOrdersToState(
 		)
 	}
 
-	makerFills := matchOrders.GetFills()
+	err = k.statefulValidateMatchTaker(ctx, takerOrder.MustGetOrder())
+	if err != nil {
```
lucas-dydx, line 459: Nit: combine into one line

protocol/x/clob/keeper/process_operations.go
```diff
@@ -532,6 +540,15 @@ func (k Keeper) PersistMatchLiquidationToState(
 		return err
 	}
 
+	err = k.statefulValidateMatchTaker(ctx, takerOrder.MustGetOrder())
```
lucas-dydx, line 543: For my understanding why doesn't this panic for liquidation orders?

Doesn't it call [this function](https://github.com/dydxprotocol/v4-chain/blob/76ef1aefc1bc7ab393c20512e0940ea2be018cdc/protocol/x/clob/types/liquidation_order.go#L100-L104)?

protocol/x/clob/keeper/process_operations_stateful_validation.go
```diff
@@ -80,6 +80,25 @@ func (k Keeper) MustFetchOrderFromOrderId(
 	return order
 }
 
+// statefulValidateMatchTaker performs stateful validation on the taker order of a specified match.
+func (k Keeper) statefulValidateMatchTaker(
+	ctx sdk.Context,
+	takerOrder types.Order,
+) error {
+	if takerOrder.GetTimeInForce() == types.Order_TIME_IN_FORCE_IOC {
```
lucas-dydx, line 88: Nit: for completeness also block FOK orders as well (use `RequiresImmediateExecution`)

protocol/x/clob/keeper/process_operations_stateful_validation.go
```diff
@@ -80,6 +80,25 @@ func (k Keeper) MustFetchOrderFromOrderId(
 	return order
 }
 
+// statefulValidateMatchTaker performs stateful validation on the taker order of a specified match.
+func (k Keeper) statefulValidateMatchTaker(
```
lucas-dydx, line 84: Can we rename this to note it verifies that immediate execution taker orders aren't already partially filled?

protocol/x/clob/memclob/memclob.go
```diff
@@ -1307,6 +1307,24 @@ func (m *MemClobPriceTimePriority) validateNewOrder(
 		)
 	}
 
+	// Immediate-or-cancel orders may only be filled once. The remaining size becomes unfillable. This prevents
+	// the case where an IOC order is partially filled multiple times over the course of multiple blocks.
+	if order.TimeInForce == types.Order_TIME_IN_FORCE_IOC && remainingAmount < order.GetBaseQuantums() {
```
lucas-dydx, line 1312: Suggestion: Also block FOK orders here

protocol/x/clob/memclob/memclob.go
```diff
@@ -1307,6 +1307,24 @@ func (m *MemClobPriceTimePriority) validateNewOrder(
...
+		// Prevent IOC orders from replacing partially filled orders.
+		if restingOrderExists {
+			return errorsmod.Wrap(
+				types.ErrInvalidReplacement,
+				"Cannot replace partially filled order with IOC order",
```
lucas-dydx, line 1317: Suggestion: add more detail to the error (filled amount of current order probably)

protocol/x/clob/keeper/process_operations.go
jakob-dydx: It does panic and was causing test failures. We don't need to statefully validate the taker for liquidations (because it isn't a real order). Will remove this validation.

protocol/x/clob/keeper/process_operations_stateful_validation.go
jakob-dydx: Sure, was leaving the function name generic for future extensibility but this is all the logic it holds for now so maybe I don't even need this funciton

lucas-dydx: Yeah up to you. Figured we can change function names later once we add more validation

protocol/x/clob/e2e/order_matches_test.go
jakob-dydx: Will do

protocol/x/clob/e2e/order_matches_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/cometbft/cometbft/types"
+	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
+	clobtestutils "github.com/dydxprotocol/v4-chain/protocol/testutil/clob"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+	feetiertypes "github.com/dydxprotocol/v4-chain/protocol/x/feetiers/types"
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+)
+
+func TestOrderMatches(t *testing.T) {
```
lucas-dydx, line 15: Suggestion:
```suggestion
func TestDeliverTxMatchValidation(t *testing.T) {
```

protocol/x/clob/e2e/order_matches_test.go
```diff
@@ -0,0 +1,90 @@
...
+	tests := map[string]struct {
+		subaccounts       []satypes.Subaccount
+		blockAdvancements []testapp.BlockAdvancementWithError
+	}{
+		"Error: partially filled IOC taker order cannot be matched twice": {
```
lucas-dydx, line 20: Suggestion: can we add the following two tests?

- Order is partially-filled, then replaced with an FOK order and the FOK order is rejected.
- IOC order is filled as a taker across multiple matches.

protocol/x/clob/memclob/memclob_place_order_test.go
```diff
@@ -2048,6 +2048,131 @@ func TestPlaceOrder_MatchOrders_PreexistingMatches(t *testing.T) {
 			},
 			expectedErr: types.ErrOrderFullyFilled,
 		},
+		"Error: Taker is IOC replacement for partially filled IOC": {
```
lucas-dydx, line 2051: Add some of these tests for FOK orders as well?

lucas-dydx (approved):

jayy04 (approved):

PR #368, opened 2023-09-26
[CORE-607] - Apply all valid params on MarketParamUpdate

clemire:
Previously, the entire list of params was discarded if a single param was invalid. This update changes the daemon to pass through all Param updates individually based on whether or not they pass validation.


protocol/daemons/pricefeed/client/client_integration_test.go
```diff
@@ -446,9 +453,10 @@ func (s *PriceDaemonIntegrationTestSuite) TestUpdateMarkets_AddMarketWithUSDTCon
 	)
 }
 
-// TestUpdateMarkets_InvalidUpdates tests that the pricefeed daemon ignores invalid updates.
-func (s *PriceDaemonIntegrationTestSuite) TestUpdateMarkets_InvalidUpdate() {
-	// Start the daemon with a single market. Then, update the endpoint to return invalid params.
+// TestUpdateMarkets_PartialUpdates tests that the pricefeed daemon applies valid market params and discards invalid
+// params whenever an update is partially valid.
+func (s *PriceDaemonIntegrationTestSuite) TestUpdateMarkets_PartialUpdate() {
```
clemire, line 458: Since the entire update is no longer valid/invalid, I updated the test to check for the application of valid updates when some param updates are invalid.

protocol/daemons/pricefeed/client/sub_task_runner.go
```diff
@@ -340,16 +340,22 @@ func RunMarketParamUpdaterTaskLoop(
 	}
 
 	// Update shared, in-memory config with the latest market params. Report update success/failure via logging/metrics.
-	err = configs.UpdateMarkets(getAllMarketsResponse.MarketParams)
-	if err == nil {
+	marketParamErrors, err := configs.UpdateMarkets(getAllMarketsResponse.MarketParams)
+	if err == nil && len(marketParamErrors) == 0 {
```
teddyding, line 344: Feel like it's misleading count "one market fails to update" as general "market update fails". Consider:
- gov introduces invalid oracle market "FOO-USD". It's never removed
- gov introduces another valid oracle market "OSMO-USD". It's actually successfully applied. 

Perhaps makes sense to add a `marketParam.Id` label to this counter?

protocol/daemons/pricefeed/client/types/price_feed_mutable_market_configs.go
```diff
@@ -264,8 +277,9 @@ func (pfmmc *PricefeedMutableMarketConfigsImpl) ValidateAndTransformParams(marke
 			if exchangeConfig.AdjustByMarket != "" {
 				adjustByMarketId, ok := marketNameToId[exchangeConfig.AdjustByMarket]
 				if !ok {
-					return nil, nil, fmt.Errorf(
-						"invalid exchange config json for exchange '%v' on market param %v: adjustByMarket '%v' not found",
+					return nil, nil, nil, fmt.Errorf(
```
teddyding, line 280: Does this mean if gov passes a market with unsupported/mis-spelled exchange, it would still block other valid market updates?

clemire: The errors in this loop should be unreachable because of the validation that happens before this loop is entered - see the comment on line 258, and also "unexpected internal error". In this case I am ok to return wholesale because a.) this is specifically already checked for, and should not happen, and b.) the `MarketParam` update could be partially but not fully applied if we continue, and that puts the daemon into a state that's difficult to predict. I prefer for each `MarketParam` update to succeed or fail atomically.

protocol/daemons/pricefeed/client/sub_task_runner.go
clemire: Good point! I updated the metrics code to emit 1 count for each market individually as success or error.

protocol/daemons/pricefeed/client/types/price_feed_mutable_market_configs.go
teddyding: Ah I see, I missed the validation on line 248.

teddyding (approved):

PR #371, opened 2023-09-26
Chunk liquidation daemon collat check request

jayy04:


protocol/daemons/liquidation/client/client.go
```diff
@@ -192,9 +156,76 @@ func GetAllSubaccounts(
...
+		metrics.Count,
+	)
+
+	// Query the gRPC server in chunks of size `liqFlags.RequestChunkSize`.
+	liquidatableSubaccountIds = make([]satypes.SubaccountId, 0)
+	for start := 0; start < len(subaccountsToCheck); start += int(liqFlags.RequestChunkSize) {
+		end := start + int(liqFlags.RequestChunkSize)
+		if end > len(subaccountsToCheck) {
+			end = len(subaccountsToCheck)
+		}
```
BrendanChou, line 209: nit: use `end := lib.Min(len(subaccountsToCheck), int(liqFlags.RequestChunkSize))`
https://github.com/dydxprotocol/v4-chain/blob/main/protocol/lib/math.go#L59

BrendanChou (approved):

BrendanChou (approved):

protocol/daemons/liquidation/client/client.go
```diff
@@ -192,9 +156,73 @@ func GetAllSubaccounts(
...
+		float32(len(subaccounts)),
+		metrics.GetAllSubaccounts,
+		metrics.Count,
+	)
+
 	return subaccounts, nil
 }
 
+// GetLiquidatableSubaccountIds verifies collateralization statuses of subaccounts with
+// at least one open position and returns a list of unique and potentially liquidatable subaccount ids.
```
lucas-dydx, line 171: Where does this function dedupe subaccount IDs?

protocol/daemons/flags/flags.go
```diff
@@ -127,6 +130,11 @@ func AddDaemonFlagsToCmd(
 		df.Liquidation.SubaccountPageLimit,
 		"Limit on the number of subaccounts to fetch per query in the Liquidation Daemon task loop.",
 	)
+	cmd.Flags().Uint64(
+		FlagLiquidationDaemonRequestChunkSize,
```
lucas-dydx, line 134: Is there any logic that validates this value?

Specifically if it's easy, would be nice to validate this is non-zero.

protocol/daemons/liquidation/client/client.go
```diff
@@ -236,6 +264,13 @@ func SendLiquidatableSubaccountIds(
 		metrics.Latency,
 	)
 
+	telemetry.ModuleSetGauge(
```
lucas-dydx, line 267: Should we stat the number of non-liquidatable subaccounts somewhere?

protocol/daemons/liquidation/client/client.go
jayy04: why do we need to dedupe? there won't be any duplicates, right?

protocol/daemons/flags/flags.go
jayy04: I don't think we validate any of the flags, so going to leave it as is for now. Created this [ticket](https://linear.app/dydx/issue/CLOB-904/validate-appdaemonclob-flags) to keep track

protocol/daemons/liquidation/client/client.go
jayy04: we can get that with `allSubaccounts - liquidatableSubaccounts`

PR #383, opened 2023-09-26
[DEC-2081] Audit upgrades.go

lcwik:
Add a test that covers setting up the upgrade handlers.

### DIFF
--- a/protocol/app/upgrades_test.go
+++ b/protocol/app/upgrades_test.go
@@ -14,41 +14,19 @@ import (
 )
 
 func TestSetupUpgradeHandlers(t *testing.T) {
-	upgradeName := "test-upgrade"
-	createHandlerWasCalled := 0
-	handlerWasCalled := 0
-	upgradeHandlerFn := func(manager *module.Manager, configurator module.Configurator) types.UpgradeHandler {
-		createHandlerWasCalled++
-		return func(ctx sdk.Context, plan types.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
-			handlerWasCalled++
-			return nil, errors.New("upgrade failed")
-		}
-	}
-
 	tests := map[string]struct {
-		upgrades      []upgrades.Upgrade
+		upgradeNames  []string
 		expectedPanic string
 	}{
-		"Successfully setup handler": {
-			upgrades: []upgrades.Upgrade{
-				{
-					UpgradeName:          upgradeName,
-					CreateUpgradeHandler: upgradeHandlerFn,
-				},
-			},
+		"Successfully setup upgrade handler": {
+			upgradeNames: []string{"test1"},
 		},
-		"Panic due to duplicate handlers": {
-			upgrades: []upgrades.Upgrade{
-				{
-					UpgradeName:          upgradeName,
-					CreateUpgradeHandler: upgradeHandlerFn,
-				},
-				{
-					UpgradeName:          upgradeName,
-					CreateUpgradeHandler: upgradeHandlerFn,
-				},
-			},
-			expectedPanic: "Cannot register duplicate upgrade handler 'test-upgrade'",
+		"Successfully setup multiple upgrade handlers": {
+			upgradeNames: []string{"test1", "test2"},
+		},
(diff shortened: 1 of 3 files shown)


lcwik:
@ttl33 We would need to add integration tests to validate that these work as expected but https://linear.app/dydx/issue/CORE-424 was recently cancelled. Maybe we should re-open it and add it to a future project?

Adding integration style tests for upgrades seems beyond the scope of the audit.

protocol/app/upgrades_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	"github.com/dydxprotocol/v4-chain/protocol/app"
+	"github.com/dydxprotocol/v4-chain/protocol/app/upgrades"
+	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
+	"github.com/stretchr/testify/require"
+)
+
```
ttl33, line 15: Meta:
1. should we add tests for `setupUpgradeStoreLoaders`?
2. should we add tests for `scheduleForkUpgrade`? i.e. no duplicate heights, fork upgrade gets triggered on the height

protocol/app/upgrades_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestSetupUpgradeHandlers(t *testing.T) {
+	upgradeName := "test-upgrade"
+	createHandlerWasCalled := 0
```
ttl33, line 18: IIUC, these vars are shared across different test cases. I think it's better to isolate these vars per test case so the outcome of this test does not depend on the sequence of the test run (which is non-deterministic when iterating maps)

protocol/app/upgrades_test.go
```diff
@@ -0,0 +1,80 @@
...
+			defer func() {
+				app.Upgrades = originalUpgrades
+			}()
+
+			app.Upgrades = tc.upgrades
+			if tc.expectedPanic == "" {
+				app := testapp.DefaultTestApp(nil)
+				require.True(t, app.UpgradeKeeper.HasHandler(upgradeName))
+				require.Equal(t, 1, createHandlerWasCalled)
+				require.Equal(t, 0, handlerWasCalled)
```
ttl33, line 68: I think we should also verify that the handler is called at the upgrade height

protocol/app/upgrades_test.go
```diff
@@ -0,0 +1,80 @@
...
+			handlerWasCalled++
+			return nil, errors.New("upgrade failed")
+		}
+	}
+
+	tests := map[string]struct {
+		upgrades      []upgrades.Upgrade
+		expectedPanic string
+	}{
+		"Successfully setup handler": {
```
ttl33, line 32: Suggestions for other test cases:
* what about multiple handlers that are successful?
* what about handlers that target the same height? (maybe this is testing upgrade keeper's behavior so it's not worth the test here)

lcwik: Added multiple handlers that are successful. I think testing upgrade keepers behavior makes sense as an integration test that does perform a full upgrade.

protocol/app/upgrades_test.go
lcwik: That would require scheduling a valid plan and this makes a lot more sense as an integration test then a unit test.

protocol/app/upgrades_test.go
lcwik: Both `setupUpgradeStoreLoaders` and `scheduleForkUpgrade` make more sense as integration tests since we will get little value out of the unit test.

ttl33 (approved):

protocol/app/upgrades_test.go
ttl33: ack, opened up the mentioned ticket and added comments

protocol/app/upgrades_test.go
ttl33: reopened the mentioned ticket + added comments!

PR #393, opened 2023-09-27
explicitly propagate events from cached contexts to original context in delaymsg endblocker

tqin7:
**Context:**
Couldn't find bank transfer events when a bridge is completed.
`x/gov` explicitly propagates events from cached contexts (see [here](https://github.com/dydxprotocol/cosmos-sdk/blob/dydx-fork-v0.47.4/x/gov/abci.go#L102-L103))

**Testing:**
Ran localnet
1. Before this change, can't find the bank transfer events in `end_block_events` in `/block_results` response
2. After this change, can find the bridge bank transfer events in `end_block_events` in `/block_results` response
    example:
   ```
   {"type":"transfer","attributes":[{"key":"recipient","value":"dydx1zg69v7yszg69v7yszg69v7yszg69v7ysuz0wqr","index":true},{"key":"sender","value":"dydx1zlefkpe3g0vvm9a4h0jf9000lmqutlh9jwjnsv","index":true},{"key":"amount","value":"123dv4tnt","index":true}]}
   ```


teddyding:
I had question about events being correctly emitted in this [PR](https://github.com/dydxprotocol/v4-chain/pull/214#discussion_r1319283690) (cc @clemire ). It looks like the [test](https://github.com/dydxprotocol/v4-chain/blob/9a28c3ce18a9e821ba9d86f5b232a3cc8d190f0d/protocol/lib/abci/util_test.go#L61-L64) we added here was not sufficient to catch this?

 Can we add a test that correctly verifies events in delayed messages emitted? (we should verify it fails without the fix in this PR)

protocol/x/delaymsg/keeper/dispatch.go
```diff
@@ -33,13 +40,21 @@ func DispatchMessagesForBlock(k types.DelayMsgKeeper, ctx sdk.Context) {
...
+			// Append events emitted in cached context to `events`.
+			events = append(events, res.GetEvents()...)
+			return nil
 		}); err != nil {
 			k.Logger(ctx).Error("failed to execute delayed message with id %v: %v", id, err)
 		}
 	}
 
+	// Propagate events emitted in cached contexts to current context.
+	ctx.EventManager().EmitEvents(events)
```
teddyding, line 56: We should wrap this event emitting logic in `RunCached` util. Does the axelar example [util](https://github.com/axelarnetwork/axelar-core/blob/e9287fd765f7e501238ab85395ba874d5d89b52c/utils/abci.go#L19) work?

dalmirel: There were a lot of issues due to sdk message handler creating new EventManager.  The emitted events could be picked up from the result like in: https://github.com/cosmos/cosmos-sdk/issues/13389
or just  propagate events like axelar did and @teddyding is suggesting.I believe Axelar  as well had issues with losing events with their initial `RunCached` implementation: https://github.com/axelarnetwork/axelar-core/pull/1439.

tqin7: When I emit events in `RunCached` like [axelar example](https://github.com/axelarnetwork/axelar-core/blob/e9287fd765f7e501238ab85395ba874d5d89b52c/utils/abci.go#L19), they don't show up any more. The reason is that message handlers create a new event manager and thus the events aren't present in `ctx.EventManager()` in [this line](https://github.com/axelarnetwork/axelar-core/blob/e9287fd765f7e501238ab85395ba874d5d89b52c/utils/abci.go#L36) (The cosmos issue that @dalmirel linked also mentioned the same issue)

I think for `x/delaymsg`'s use case, we have to emit events that are part of handler response

tqin7:
> I had question about events being correctly emitted in this [PR](https://github.com/dydxprotocol/v4-chain/pull/214#discussion_r1319283690) (cc @clemire ). It looks like the [test](https://github.com/dydxprotocol/v4-chain/blob/9a28c3ce18a9e821ba9d86f5b232a3cc8d190f0d/protocol/lib/abci/util_test.go#L61-L64) we added here was not sufficient to catch this?
> 
> Can we add a test that correctly verifies events in delayed messages emitted? (we should verify it fails without the fix in this PR)

added a test case in second commit that fails without the fix!

protocol/x/delaymsg/keeper/dispatch.go
```diff
@@ -17,6 +17,13 @@ func DispatchMessagesForBlock(k types.DelayMsgKeeper, ctx sdk.Context) {
 		return
 	}
 
+	// Maintain a list of events emitted by all delayed messages executed in this block.
+	// As each delayed message is executed in a cached context, their emitted events need
```
clemire, line 21: Nit: I think, based on your explanation, that this comment is inaccurate? "As each handler creates collects events in it's own EventManger, the emitted events need to be explicitly added to the current context." or some such.

protocol/x/delaymsg/keeper/dispatch.go
```diff
@@ -33,13 +40,21 @@ func DispatchMessagesForBlock(k types.DelayMsgKeeper, ctx sdk.Context) {
 
 		if err = abci.RunCached(ctx, func(ctx sdk.Context) error {
 			handler := k.Router().Handler(msg)
-			_, err := handler(ctx, msg)
-			return err
+			res, err := handler(ctx, msg)
+			if err != nil {
+				return err
+			}
+			// Append events emitted in cached context to `events`.
```
clemire, line 47:
```suggestion
			// Append events emitted in the handler to `events`.
```

clemire (approved):

tqin7:
As a separate ticket, I think we should also add [this line](https://github.com/axelarnetwork/axelar-core/blob/e9287fd765f7e501238ab85395ba874d5d89b52c/utils/abci.go#L36) to our `RunCached` implementation

PR #396, opened 2023-09-27
[DEC-2070] - x/prices audit: add test cases to util, clarifying comments

clemire:
math methods generally look fine in x/prices, with a few apparent exceptions. I added some comments here to show why those exceptions are actually fine. Also added a little extra test coverage for our `isCrossing*Price` methods. Confirmed on slack with Brendan that medianization rounding behavior [is fine](https://dydx-team.slack.com/archives/C03TDPD3MC4/p1695848350401879) for x/prices, also.

### DIFF
--- a/protocol/x/prices/keeper/util.go
+++ b/protocol/x/prices/keeper/util.go
@@ -10,7 +10,8 @@ import (
 )
 
 // getProposalPrice returns the proposed price update for the next block, which is either the smoothed price or the
-// index price - whichever is closer to the current market price.
+// index price - whichever is closer to the current market price. In cases where the smoothed price and the index price
+// are equidistant from the current market price, the smoothed price is chosen.
 func getProposalPrice(smoothedPrice uint64, indexPrice uint64, marketPrice uint64) uint64 {
 	if lib.AbsDiffUint64(smoothedPrice, marketPrice) > lib.AbsDiffUint64(indexPrice, marketPrice) {
 		return indexPrice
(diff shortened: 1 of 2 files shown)


protocol/x/prices/keeper/util_test.go
```diff
@@ -58,6 +58,12 @@ func TestGetProposalPrice(t *testing.T) {
 			marketPrice:   uint64(1_000_000),
 			expectedPrice: uint64(900_000),
 		},
+		"smoothedPrice: smoothedPrice < marketPrice < indexPrice": {
```
clemire, line 61: Test cases to check that in condition where market price is equidistant from smoothed price and index price, smoothed price is chosen.

protocol/x/prices/keeper/util_test.go
```diff
@@ -178,7 +190,7 @@ func TestIsTowardsIndexPrice(t *testing.T) {
 			indexPrice:     2,
 			expectedResult: true,
 		},
-		"Towards: idx < curr == new": {
+		"Towards: idx < new == curr": {
```
clemire, line 193: test result easier to infer with test case renames: should be towards when new is between index and curr, inclusive.

protocol/x/prices/keeper/util_test.go
```diff
@@ -285,6 +297,12 @@ func TestIsCrossingIndexPrice(t *testing.T) {
 			indexPrice:     3,
 			expectedResult: false,
 		},
+		"Not Crossing: new = curr < index": {
```
clemire, line 300: All crossing tests have 13 conditions to check - add missing cases here and below.

protocol/x/prices/keeper/util.go
```diff
@@ -26,7 +27,8 @@ func isAboveRequiredMinPriceChange(marketParamPrice types.MarketParamPrice, newP
 }
 
 // getMinPriceChangeAmountForMarket returns the amount of price change that is needed to trigger
-// a price update in accordance with the min price change parts-per-million value.
+// a price update in accordance with the min price change parts-per-million value. This method always rounds down,
```
clemire, line 30: Making a note of this behavior.

ttl33 (approved):

PR #414, opened 2023-09-28
standardize key format

BrendanChou:
- remove `types/keys_**.go` files, move them to `types/keys.go`
- remove functions from `types/keys.go`, use `lib/bytes.go` functions directly or add `MustMarshal` functions
- format keys similarly across modules (lowercase, snake_case)
- prefixes have `/` endings, direct keys do not
- fix a lot of tests that use magic strings instead of the key constants


protocol/x/perpetuals/types/keys.go
```diff
@@ -1,9 +1,49 @@
...
+
+	// PremiumVoteClampFactorPpmKey is the key to retrieve premium vote clamp factor
+	// in parts-per-million
+	// |V| <= clamp_factor * (initial margin - maintenance margin)
+	PremiumVoteClampFactorPpmKey = "premium_vote_clamp_factor_ppm"
+
+	// MinNumVotesPerSampleKey is the key to retrieve `min_num_votes_per_sample`,
+	// the minimum number of votes needed to calculate a premium sample.
+	MinNumVotesPerSampleKey = "min_num_votes_per_sample"
+)
```
BrendanChou, line 49: @tqin7 I'll aggregate these here for easier modification

protocol/daemons/pricefeed/client/client_integration_test.go
```diff
@@ -26,9 +30,6 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc"
-	"net"
```
teddyding, line 29: Nit: revert changes in this file since it's irrelevant to PR

teddyding: Nit: revert changes in this file since it's irrelevant to PR

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -186,7 +186,7 @@ func (k Keeper) setClobPair(ctx sdk.Context, clobPair types.ClobPair) {
 	b := k.cdc.MustMarshal(&clobPair)
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.ClobPairKeyPrefix))
 	// Write the `ClobPair` to state.
-	store.Set(types.ClobPairKey(clobPair.GetClobPairId()), b)
+	store.Set(lib.Uint32ToBytes(clobPair.GetClobPairId().ToUint32()), b)
```
teddyding, line 189: Nit: In this file at least, seems like there's merit in keeping the function
```
// ClobPairKey returns the store key to retrieve a ClobPair from the index fields
func ClobPairKey(
	id ClobPairId,
) []byte {
	return lib.Uint32ToBytesForState(id.ToUint32())
}
```

Feels like trade-off between simplicity and readability. No strong preference

protocol/x/clob/types/order_id.go
```diff
@@ -96,6 +97,15 @@ func (o *OrderId) Validate() error {
 	return nil
 }
 
+// MustMarshal returns the marshaled bytes representation of an OrderId, panic'ing on error.
+func (o *OrderId) MustMarshal() []byte {
```
teddyding, line 101: Nit: can we add a unit test for these functions?

protocol/x/delaymsg/types/keys.go
```diff
@@ -1,9 +1,22 @@
...
 	ModuleName = "delaymsg"
 
 	// StoreKey defines the primary module store key
 	StoreKey = ModuleName
 )
+
+// State
+const (
+	// BlockMessageIdsPrefix is the prefix to retrieve all BlockMessageIds for a given block height.
+	BlockMessageIdsPrefix = "block_message_ids/"
```
teddyding, line 15: Nit: some of the prefixes string use plural form (example: here, `EpochStatsKeyPrefix` below). Most use singular form (like `delayed_message/` below). Could we maintain consistency?

protocol/x/perpetuals/types/keys.go
```diff
@@ -1,9 +1,49 @@
...
 	StoreKey = ModuleName
 )
+
+// State
+const (
+	// PerpetualKeyPrefix is the prefix to retrieve all Perpetual
+	PerpetualKeyPrefix = "perpetual/"
+
+	// PerpetualKeyPrefix is the prefix to retrieve all funding premium samples
+	FundingSamplesKeyPrefix = "funding_samples/"
```
teddyding, line 18: This is un-used, can delete.
All votes/samples are stored under below keys

teddyding (approved):

protocol/x/delaymsg/types/keys.go
BrendanChou: The name is based on the value ultimately stored at the location, it is not based on the key. So I believe it is already consistent as it is pluralized if the value is pluralized. Pluralization doesn't have to do with whether it is a collection of things or not

teddyding (approved):

PR #415, opened 2023-09-29
Setup cosmovisor correctly in snapshot full node

roy-dydx:


roy-dydx:
The current setup is working because I have this commit included in the PR that's building the current image.

dydxwill (approved):

PR #418, opened 2023-09-29
[IND-413] deprecate old IndexerTendermintEvent.data string field

dydxwill:
### Summary by CodeRabbit

- Refactor: Removed the usage of base64 encoding for event messages across various modules and replaced it with direct byte handling. This change simplifies the code and improves performance by eliminating unnecessary encoding and decoding operations.
- Refactor: Updated the event handling mechanism to directly use the `DataBytes` field instead of calling `GetBytesFromEventData` function. This change simplifies the code and improves readability.
- Test: Updated test cases to reflect changes in event handling and encoding. This ensures that the tests remain accurate and reliable after the refactor.
- Chore: Removed unused `data` properties and parameters from various functions and interfaces, reducing complexity and improving maintainability.


vincentwschau (approved):
Make sure this only is deployed only after the protocol is upgraded. Important mainly for the test-net.

PR #419, opened 2023-09-29
Add verbose mode to bridge_events script

BrendanChou:


protocol/scripts/bridge_events/bridge_events.go
```diff
@@ -125,13 +131,57 @@ func main() {
...
+		return
+	}
+
+	// Print x/auth accounts information.
+	genesisAccounts := authtypes.GenesisAccounts{}
+	for i, address := range sortedAddresses {
+		var ba authtypes.AccountI = &authtypes.BaseAccount{
+			Address:       address,
+			PubKey:        nil,
+			AccountNumber: uint64(i),
```
ttl33, line 162: To confirm, is this the correct `AccountNumber` value to use? It seems like this ordering based on the account address (since we are iterating addresses sorted by the string)

Or does this account number not matter?

ttl33 (approved):

PR #423, opened 2023-09-29
[DEC-2075] - Remove panic from ProcessVesting, called in BeginBlocker

clemire:
This panic is highly unlikely and represents an unexpected error. Replace it with an emitted error metric + a log statement and continue so we don't cause liveness issue in the event of vesting transfer failure. Validator operators can alert on the metric and investigate based on the log statement.

### DIFF
--- a/protocol/x/vest/keeper/keeper.go
+++ b/protocol/x/vest/keeper/keeper.go
@@ -123,7 +123,7 @@ func (k Keeper) ProcessVesting(ctx sdk.Context) {
 			); err != nil {
 				// This should never happen. However, if it does, we should not panic.
 				// ProcessVesting is called in BeginBlocker, and panicking in BeginBlocker could cause liveness issues.
-				// Instead, we log the error and continue.
+				// Instead, we generate an informative error log, emit an error metric, and continue.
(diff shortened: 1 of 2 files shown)


protocol/x/vest/keeper/keeper.go
```diff
@@ -120,7 +121,27 @@ func (k Keeper) ProcessVesting(ctx sdk.Context) {
 				entry.TreasuryAccount,
 				sdk.NewCoins(sdk.NewCoin(entry.Denom, vestAmount)),
 			); err != nil {
-				panic(err)
+				// This should never happen. However, if it does, we should not panic.
```
teddyding, line 124: Thanks for making this change. I explicitly wanted to panic instead of logging an error during implementation, since this is an invariant violation. 

However after second thought, I'm inclined to agree that logging an error makes more sense here. This is because failure to process a vest event does not disrupt core trading flows. It will not lead to protocol/user losing funds (although I can lead to user not getting their entitled rewards). 

I started a thread in protocol to be aligned on similar topic. Will approve after that thread is resolved.

teddyding (approved):

PR #424, opened 2023-10-01
[DEC-2127] Asset Id refactoring

teddyding:
- Drop `NumAsset` and consecutive asset ID requirement (similar to recent changes in `x/clob`, `x/perpetuals`, `x/prices`
- Reserve `assetId == 0` for USDC

### DIFF
--- a/protocol/x/assets/keeper/asset.go
+++ b/protocol/x/assets/keeper/asset.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"math/big"
+	"sort"
 
 	errorsmod "cosmossdk.io/errors"
 	sdkmath "cosmossdk.io/math"
@@ -16,6 +17,7 @@ import (
 
 func (k Keeper) CreateAsset(
 	ctx sdk.Context,
+	assetId uint32,
 	symbol string,
 	denom string,
 	denomExponent int32,
@@ -23,17 +25,21 @@ func (k Keeper) CreateAsset(
 	marketId uint32,
 	atomicResolution int32,
 ) (types.Asset, error) {
-	// Get the `nextId`
-	nextId := k.GetNumAssets(ctx)
+	if prevAsset, exists := k.GetAsset(ctx, assetId); exists {
+		return types.Asset{}, errorsmod.Wrapf(
+			types.ErrAssetIdAlreadyExists,
+			"previous asset = %v",
+			prevAsset,
+		)
+	}
 
-	_, found := k.internalGetIdByDenom(ctx, denom)
-	if found {
+	if _, exists := k.internalGetIdByDenom(ctx, denom); exists {
 		return types.Asset{}, errorsmod.Wrap(types.ErrAssetDenomAlreadyExists, denom)
 	}
 
 	// Create the asset
 	asset := types.Asset{
-		Id:               nextId,
+		Id:               assetId,
 		Symbol:           symbol,
 		Denom:            denom,
 		DenomExponent:    denomExponent,
@@ -59,9 +65,6 @@ func (k Keeper) CreateAsset(
 	// Store the new asset
 	k.setAsset(ctx, asset)
 
-	// Store the new `numAssets`
-	k.setNumAssets(ctx, nextId+1)
-
 	// Store the denom-to-asset-id mapping
 	k.setDenomToId(ctx, asset.Denom, asset.Id)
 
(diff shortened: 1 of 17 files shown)


protocol/x/assets/keeper/asset.go
```diff
@@ -16,24 +17,34 @@ import (
...
+	}
 
-	_, found := k.internalGetIdByDenom(ctx, denom)
-	if found {
+	if _, exists := k.internalGetIdByDenom(ctx, denom); exists {
 		return types.Asset{}, errorsmod.Wrap(types.ErrAssetDenomAlreadyExists, denom)
 	}
 
+	// Ensure assetId zero is always USDC. This is a protocol-wide invariant.
+	if assetId == types.AssetUsdc.Id && denom != types.AssetUsdc.Denom {
```
ttl33, line 41: should we add the other validation? i.e. if denom == usdc, then id must be 0:

there's prob a shorter way to write this in 1 single if statement (some combo of XOR maybe)
```
if assetid == assetUsdc.Id || denom == assetUsdc.denom {
    if assetid != assetUsdc.Id || denom != assetUsdc.denom {
        throw error
    }
}
```

protocol/x/assets/keeper/asset.go
```diff
@@ -240,34 +229,37 @@ func (k Keeper) GetDenomById(
...
-) []types.Asset {
-	num := k.GetNumAssets(ctx)
-	assets := make([]types.Asset, num)
-
-	for i := uint32(0); i < num; i++ {
-		asset, err := k.GetAsset(ctx, i)
-		if err != nil {
-			panic(err)
-		}
+) (list []types.Asset) {
```
ttl33, line 246: Given that this builds on top of [this PR](https://github.com/dydxprotocol/v4-chain/pull/414), should we label this as state-breaking too (since we are introducing the iterator here)? Or are we assuming that labelling the above PR is sufficient?

protocol/x/assets/keeper/asset_test.go
```diff
@@ -38,6 +39,7 @@ func createNAssets(
 		}
 		asset, err := keeper.CreateAsset(
 			ctx,
+			uint32(i+1),                 // AssetId
```
ttl33, line 42: why +1? is this used to skip USDC one?

protocol/x/assets/keeper/asset_test.go
```diff
@@ -61,6 +63,7 @@ func TestCreateAsset_MarketNotFound(t *testing.T) {
 	// Throws error when creating asset for invalid marketId.
 	_, err := keeper.CreateAsset(
 		ctx,
+		1,
```
ttl33, line 66: nit: here and below, define const "firstValidAssetId" and reuse

protocol/x/assets/types/errors.go
```diff
@@ -12,11 +12,12 @@ var (
 	ErrAssetDenomAlreadyExists      = errorsmod.Register(ModuleName, 4, "Existing asset found with the same denom")
 	ErrAssetIdAlreadyExists         = errorsmod.Register(ModuleName, 5, "Existing asset found with the same asset id")
 	ErrGapFoundInAssetId            = errorsmod.Register(ModuleName, 6, "Found gap in asset Id")
-	ErrAssetZeroNotUsdc             = errorsmod.Register(ModuleName, 7, "First asset is not USDC")
+	ErrAssetZeroNotUsdc             = errorsmod.Register(ModuleName, 7, "Asset zero is not USDC")
```
ttl33, line 15:
```suggestion
	ErrAssetZeroNotUsdc             = errorsmod.Register(ModuleName, 7, "Asset ID zero is not USDC")
```

protocol/x/clob/keeper/deleveraging_test.go
```diff
@@ -185,15 +186,7 @@ func TestIsValidInsuranceFundDelta(t *testing.T) {
...
-			_, err := ks.AssetsKeeper.CreateAsset(
-				ks.Ctx,
-				constants.Usdc.Symbol,
-				constants.Usdc.Denom,
-				constants.Usdc.DenomExponent,
-				constants.Usdc.HasMarket,
-				constants.Usdc.MarketId,
-				constants.Usdc.AtomicResolution,
-			)
+			err := keepertest.CreateUsdcAsset(ks.Ctx, ks.AssetsKeeper)
```
ttl33, line 189: 👍

ttl33 (approved):

protocol/x/assets/keeper/asset.go
teddyding: I think it's worth clarifying what `state-breaking` means.
IMO this PR is `state-breaking` not only due to the iterator, but also due to the fact that we added additional validatoin on `CreateAsset` ("state-incompatible")

protocol/x/assets/keeper/asset.go
teddyding: I just added a separate condition - feels cleaner that way

PR #427, opened 2023-10-02
disable equity tier limit validation in DeliverTx

jayy04:


protocol/x/clob/keeper/orders.go
```diff
@@ -201,9 +201,9 @@ func (k Keeper) PlaceStatefulOrder(
 	}
 
 	// 3. Check that adding the order would not exceed the equity tier for the account.
-	if err := k.ValidateSubaccountEquityTierLimitForNewOrder(ctx, order); err != nil {
-		return err
-	}
+	// if err := k.ValidateSubaccountEquityTierLimitForNewOrder(ctx, order); err != nil {
+	// 	return err
+	// }
```
lucas-dydx, line 206: Doesn't this need to be height-gated on the block that had divergent state? Specifically [block 956441](https://testnet.mintscan.io/dydx-testnet/blocks/956441)?

```suggestion
        if ctx.BlockHeight() < 956441 {
            if err := k.ValidateSubaccountEquityTierLimitForNewOrder(ctx, order); err != nil {
	 	return err
	    }
        }
```

lucas-dydx (approved):

lcwik (commented):
LGTM

PR #428, opened 2023-10-02
[CORE-606] Update PowerReduction to reflect 1e-18 stake denom unit

teddyding:
[Context](https://dydx-team.slack.com/archives/C03FP255RRU/p1696027981031659?thread_ts=1696012576.428419&cid=C03FP255RRU)


protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -395,16 +395,23 @@ func TestSendDelayedCompleteBridgeMessage(t *testing.T) {
 
 	aliceAccountAddress := sdk.MustAccAddressFromBech32(constants.BridgeEvent_Id0_Height0.Address)
 
-	// Sanity check: at block 1, balances are as expected before the message is sent.
+	// Sanity check: at block 1, expect bridge balance is expected before the message is sent.
 	expectAccountBalance(t, ctx, &tApp, BridgeAccountAddress, BridgeGenesisAccountBalance)
-	expectAccountBalance(t, ctx, &tApp, aliceAccountAddress, AliceInitialAccountBalance)
```
teddyding, line 400: Instead of assuming the balance of Alice (which depends on the genesis balance state, directly get its initial balance

protocol/app/app.go
```diff
@@ -284,6 +288,14 @@ type App struct {
 	Server              *daemonserver.Server
 }
 
+// assertAppPreconditions assert invariants required for an application to start.
+func assertAppPreconditions() {
+	// Check that the default power reduction is set correctly.
+	if sdk.DefaultPowerReduction.Uint64() != 1_000_000_000_000_000_000 {
```
BrendanChou, line 294: do not cast to a `uint64`. per the [docs](https://pkg.go.dev/math/big#Int.Uint64), 
> If x cannot be represented in a uint64, the result is undefined

protocol/app/app.go
```diff
@@ -293,6 +305,7 @@ func New(
 	appOpts servertypes.AppOptions,
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *App {
+	assertAppPreconditions()
```
BrendanChou, line 308: nit: add space afterwards, or include the `appFlags` validation in this function

protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -32,12 +32,12 @@ var (
...
-	BridgeGenesisAccountBalance = sdk.NewCoin("dv4tnt", sdkmath.NewInt(1000000000))
-	AliceInitialAccountBalance  = sdk.NewCoin("dv4tnt", sdkmath.NewInt(99500000000))
+	testDenom = "dv4tnt"
+
+	BridgeGenesisAccountBalance = sdk.NewCoin(testDenom, sdkmath.NewInt(1000000000))
 
 	delta                        = constants.BridgeEvent_Id0_Height0.Coin.Amount.Int64()
-	BridgeExpectedAccountBalance = sdk.NewCoin("dv4tnt", sdkmath.NewInt(1000000000-delta))
-	AliceExpectedAccountBalance  = sdk.NewCoin("dv4tnt", sdkmath.NewInt(99500000000+delta))
+	BridgeExpectedAccountBalance = sdk.NewCoin(testDenom, sdkmath.NewInt(1000000000-delta))
```
BrendanChou, line 40: I don't think this and the above line should need to use casing to `int64` (?) why not just do math using `sdkmath.Int` directly?

Also should this line use `BridgeGenesisAccountBalance` or does it happen to use the same number for some unrelated reason

teddyding: Yup, refactored to using `sdkmath.Int` math on `BridgeGenesisAccountBalance` and `delta`

protocol/x/delaymsg/keeper/dispatch_test.go
```diff
@@ -395,16 +395,23 @@ func TestSendDelayedCompleteBridgeMessage(t *testing.T) {
 
 	aliceAccountAddress := sdk.MustAccAddressFromBech32(constants.BridgeEvent_Id0_Height0.Address)
 
-	// Sanity check: at block 1, balances are as expected before the message is sent.
+	// Sanity check: at block 1, expect bridge balance is expected before the message is sent.
```
clemire, line 398: Nit:

```suggestion
	// Sanity check: at block 1, expect bridge balance is genesis value before the message is sent.
```

or some such.

BrendanChou (approved):

PR #436, opened 2023-10-02
[CLOB-910] Update stateful order equity tier count to use state.

lcwik:
### Summary by CodeRabbit

- New Feature: Updated the logic for counting stateful orders in `ValidateSubaccountEquityTierLimitForNewOrder` function to include different types of orders and improve equity tier limit validation.
- Refactor: Renamed several functions related to stateful order counts, including `InitStatefulOrdersInMemClob` to `InitStatefulOrders`, `CountSubaccountOrders` to `CountSubaccountShortTermOrders`, and `CountUntriggeredSubaccountOrders` to `CountUntriggeredSubaccountStatefulOrders`.
- Test: Modified test cases for `PlaceOrder` function to advance block height and ensure successful transactions delivery. Also updated tests to verify the count of stateful orders after initialization.
- Chore: Removed the check for equity tier limits in `validateNewOrder` function and the decrement of `to be committed` stateful order count in `CancelStatefulOrder` function.


protocol/x/clob/keeper/orders.go
```diff
@@ -334,6 +329,12 @@ func (k Keeper) placeOrder(
 		return 0, 0, err
 	}
 
+	// Validate that adding the order wouldn't exceed subaccount equity tier limits.
+	err = k.ValidateSubaccountEquityTierLimitForNewOrder(ctx, order)
```
lucas-dydx, line 333: For my understanding why not add this in `PerformStatefulOrderValidation`? Is it specifically to ensure this isn't called for Short-Term order validation in `DeliverTx`?

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -59,9 +59,12 @@ func (k *Keeper) InitializeEquityTierLimit(
...
 // Note that the method is dependent on whether we are executing on `checkState` or on `deliverState` for
-// stateful orders. During `checkState` we rely on the uncommitted order count to tell us how many stateful
-// orders exist outside of the MemClob. For `deliverState` we use the `ProcessProposerMatchesEvents`
-// to find out how many orders (minus removals) exists outside of the `MemClob`.
+// stateful orders. For `deliverState`, we sum:
+//   - the number of long term orders.
+//   - the number of triggered conditional orders.
+//   - the number of untriggered conditional orders.
+//
+// And for `checkState`, we add to the above sum the number of to be committed stateful orders.
```
lucas-dydx, line 67: Do you think this is slightly more clear?

```suggestion
// And for `checkState`, we add to the above sum the number of to uncommitted stateful orders in the mempool.
```

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -59,9 +59,12 @@ func (k *Keeper) InitializeEquityTierLimit(
...
 func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, order types.Order) error {
```
lucas-dydx, line 68: Small suggestion on simplifying the loop between lines 105 - 112 (IMO slightly more readable because no else statement):
```
	for _, limit := range equityTierLimits {
		if netCollateral.Cmp(limit.UsdTncRequired.BigInt()) < 0 {
			break
		}
                 equityTierLimit = limit
	}
```

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +120,55 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
+				panic(fmt.Errorf(
+					"Expected ValidateSubaccountEquityTierLimitForNewOrder for new order %+v to be >= 0. "+
+						"equityTierCount %d, statefulOrderCount %d, untriggeredSubaccountOrders %d, "+
+						"uncommittedStatefulOrderCount %d.",
+					order,
+					equityTierCount,
+					k.GetStatefulOrderCount(ctx, order.OrderId),
+					k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter),
+					k.GetUncommittedStatefulOrderCount(ctx, order.OrderId),
+				))
```
lucas-dydx, line 144: Small suggestion: formatting.

```suggestion
				panic(
				    fmt.Errorf(
					"Expected ValidateSubaccountEquityTierLimitForNewOrder for new order %+v to be >= 0. "+
						"equityTierCount %d, statefulOrderCount %d, untriggeredSubaccountOrders %d, "+
						"uncommittedStatefulOrderCount %d.",
					order,
					equityTierCount,
					k.GetStatefulOrderCount(ctx, order.OrderId),
					k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter),
					k.GetUncommittedStatefulOrderCount(ctx, order.OrderId),
				    ),
				)
```

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +120,55 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
-			)
-		} else {
-			err = fmt.Errorf(
+				k.GetStatefulOrderCount(ctx, order.OrderId),
+				k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter),
+			))
+		}
+	} else {
+		// For short term orders we just count how many orders exist on the memclob.
+		equityTierCount = int32(k.MemClob.CountSubaccountOrders(ctx, subaccountId, filter))
```
lucas-dydx, line 158: Should we add verification that this is only called in `CheckTx`, and panic if that isn't true? Since this would lead to chain halts if it was called in `DeliverTx`.

lucas-dydx: As a metapoint, I'm a little worried about the complexity of this function, specifically in how some logic should only be executed in `CheckTx` / `DeliverTx`.

Maybe in the longer term we should split this logic out into separate functions for readability?

I was thinking something like the following, what do you think? Open to ideas.
- Separate functions for validating Short-Term order equity tier limits / stateful order equity tier limits.
- Function for validating stateful order equity tier limits is called in `CheckTx` and `DeliverTx` (maybe leaving this function call in `PlaceStatefulOrder` would be fine).
- Function for validating Short-Term order equity tier limits is called in `CheckTx` (can be called in `PlaceShortTermOrder`).

protocol/x/clob/keeper/stores.go
```diff
@@ -75,14 +75,24 @@ func (k Keeper) GetUncommittedStatefulOrderCountTransientStore(ctx sdk.Context)
...
-// GetToBeCommittedStatefulOrderCountTransientStore fetches a state store used for creating,
-// reading, updating, and deleting a stateful order count from transient state. This represents
-// the number of to be committed `order placements - order removals` during `DeliverTx`.
-func (k Keeper) GetToBeCommittedStatefulOrderCountTransientStore(ctx sdk.Context) prefix.Store {
-	lib.AssertDeliverTxMode(ctx)
+// GetStatefulOrderCountStore fetches a state store used for creating,
+// reading, updating, and deleting a stateful order count from stores. This represents
+// the number of long term and triggered conditional orders.
+func (k Keeper) GetStatefulOrderCountStore(ctx sdk.Context) prefix.Store {
+	//lib.AssertDeliverTxMode(ctx)
```
lucas-dydx, line 82: Why store this in state as well? Isn't the memstore sufficient?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -59,11 +59,11 @@ func (k Keeper) SetLongTermOrderPlacement(
 	memStore.Set(orderIdBytes, longTermOrderPlacementBytes)
 
 	if !found {
-		// Increment the `to be committed` stateful order count.
-		k.SetToBeCommittedStatefulOrderCount(
+		// Increment the stateful order count.
+		k.SetStatefulOrderCount(
```
lucas-dydx, line 63: This doesn't get called when the memclob is hydrated on app startup right?

Should we add logic in [this function](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/orders.go#L1073) to set the memstore count properly?

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,58 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
-		panic(err)
 	}
 
 	// Verify that opening this order would not exceed the maximum amount of orders for the equity tier.
-	// Note that once we combine the count of orders on the memclob with how many `uncommitted` or `to be committed`
-	// stateful orders on the memclob we should always have a negative number since we only count order
-	// cancellations/removals for orders that exist.
 	if lib.MustConvertIntegerToUint32(equityTierCount) >= equityTierLimit.Limit {
 		return errorsmod.Wrapf(
 			types.ErrOrderWouldExceedMaxOpenOrdersEquityTierLimit,
```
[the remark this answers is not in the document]
jayy04: 😮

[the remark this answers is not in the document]
lcwik: I made this explicit and got rid of the two extra panics.

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,39 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
+		equityTierCount = k.MemClob.CountSubaccountOrders(ctx, subaccountId, filter)
 	}
 
 	// Verify that opening this order would not exceed the maximum amount of orders for the equity tier.
-	// Note that once we combine the count of orders on the memclob with how many `uncommitted` or `to be committed`
-	// stateful orders on the memclob we should always have a negative number since we only count order
-	// cancellations/removals for orders that exist.
 	if lib.MustConvertIntegerToUint32(equityTierCount) >= equityTierLimit.Limit {
 		return errorsmod.Wrapf(
 			types.ErrOrderWouldExceedMaxOpenOrdersEquityTierLimit,
```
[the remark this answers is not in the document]
lcwik: We don't expect > 2 billion orders for one subaccount.

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,39 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
-				k.MemClob.CountSubaccountOrders(ctx, subaccountId, filter),
-				k.GetUncommittedStatefulOrderCount(ctx, order.OrderId),
-			)
+		// For stateful orders we get the stateful order count which represents how many long term and
+		// triggered conditional orders exist in state. We add to that all untriggered conditional orders.
+		// If this is `CheckTx` then we must also add the number of uncommitted stateful orders that this validator
+		// is aware of (orders that are part of the mempool but have yet to proposed in a block).
+		equityTierCount = k.GetStatefulOrderCount(ctx, order.OrderId)
+		equityTierCount += k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter)
+		if !lib.IsDeliverTxMode(ctx) {
```
jayy04, line 130: We don't need to gate this, right? count is just gonna be zero for deliver tx 

probably improves readability

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,39 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
+						k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter),
+						k.GetUncommittedStatefulOrderCount(ctx, order.OrderId),
+					),
+				)
+			}
+			equityTierCount = uint32(equityTierCountMaybeNegative)
 		}
-		panic(err)
+	} else {
+		lib.AssertCheckTxMode(ctx)
```
jayy04, line 149: nit: let's move this to inside `memClob.CountSubaccountOrders`

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,39 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
-			)
+		// For stateful orders we get the stateful order count which represents how many long term and
+		// triggered conditional orders exist in state. We add to that all untriggered conditional orders.
+		// If this is `CheckTx` then we must also add the number of uncommitted stateful orders that this validator
+		// is aware of (orders that are part of the mempool but have yet to proposed in a block).
+		equityTierCount = k.GetStatefulOrderCount(ctx, order.OrderId)
+		equityTierCount += k.CountUntriggeredSubaccountOrders(ctx, subaccountId, filter)
+		if !lib.IsDeliverTxMode(ctx) {
+			equityTierCountMaybeNegative := k.GetUncommittedStatefulOrderCount(ctx, order.OrderId) + int32(equityTierCount)
+			if equityTierCountMaybeNegative < 0 {
```
jayy04, line 132: for my understanding - how is this possible? does this indicate a bug?

protocol/x/clob/keeper/equity_tier_limit.go
lcwik: I think it makes sense to gate this as we don't want to count this number when in `DeliverTx` and keeps the logic of what is meant to be summed clearer instead of having implicit knowledge or a comment that we expect this to be zero during `DeliverTx`.

protocol/x/clob/keeper/equity_tier_limit.go
lcwik: This can happen if a bug is introduced in the future where we double count cancellations, miss counting place orders.

Here is an example of a similar issue that happened in August: https://linear.app/dydx/issue/CLOB-820/debug-why-equity-tier-count-0

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -117,48 +119,38 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
...
+						k.GetUncommittedStatefulOrderCount(ctx, order.OrderId),
+					),
+				)
+			}
+			equityTierCount = uint32(equityTierCountMaybeNegative)
 		}
-		panic(err)
+	} else {
+		// For short term orders we just count how many orders exist on the memclob.
+		equityTierCount = k.MemClob.CountSubaccountOrders(ctx, subaccountId, filter)
```
lucas-dydx, line 150: Note: this doesn't account for orders that are fully matched in the operations queue, which could later get replayed then dropped (which is kind of confusing).

Is this the intended behavior?

protocol/x/clob/memclob/memclob.go
```diff
@@ -153,6 +153,7 @@ func (m *MemClobPriceTimePriority) CountSubaccountOrders(
 	subaccountId satypes.SubaccountId,
 	filter func(types.OrderId) bool,
 ) (count uint32) {
+	lib.AssertCheckTxMode(ctx)
```
lucas-dydx, line 156: Suggestion: can we update this function to just count Short-Term orders? It seems like that's all it's used for now.

I don't think this function should ever be used outside of that context.

Also update the function comment to note this function should only be called in `CheckTx` / `PrepareCheckState` and is inherently non-deterministic.

protocol/testing/e2e/gov/add_new_market_test.go
```diff
@@ -302,7 +302,7 @@ func TestAddNewMarketProposal(t *testing.T) {
 				require.Equal(t, initPerpetuals, tApp.App.PerpetualsKeeper.GetAllPerpetuals(ctx))
 				require.Equal(t, initClobPairs, tApp.App.ClobKeeper.GetAllClobPairs(ctx))
 			case govtypesv1.ProposalStatus_PROPOSAL_STATUS_PASSED:
-				// Proposal passed and succesfully executed, check states are updated.
```
lucas-dydx, line 305: Let me know if I'm missing something, but were we able to add a regression test for the issue we saw?

Specifically, if an order removal occurs in `CheckTx`, then it no longer decrements the stateful order count because it's not committed to the `deliverState`?

protocol/x/clob/keeper/equity_tier_limit.go
lucas-dydx: Yeah personally I think this is more readable if it's gated, since then you don't need to worry about the implementation of `GetUncommittedStatefulOrderCount` when thinking about this function in the context of `DeliverTx`

In the longer term, a larger readability improvement is likely splitting out this function into two, one to use in `DeliverTx` and another to use in `CheckTx`.

protocol/x/clob/keeper/equity_tier_limit.go
lcwik: `placeOrder` is only invoked from `PlaceShortTermOrder` which is only invokedfrom the ante handler when its not a recheck.

So this code isn't executed on the replay path. This would mean that in some cases a user could have more open orders then what is allowed since replaying them could add too many. The user doesn't really have control over whether their short term order is replayed so I don't think its a risk.

protocol/testing/e2e/gov/add_new_market_test.go
lcwik: Not yet, was prioritizing the fix over the test.

protocol/x/clob/memclob/memclob.go
lcwik: I updated the comment and updated this method and the `CountUntriggeredSubaccountOrders` as well.

protocol/x/clob/keeper/untriggered_conditional_orders.go
```diff
@@ -323,21 +323,19 @@ func (k Keeper) MaybeTriggerConditionalOrders(ctx sdk.Context) (triggeredConditi
...
 }
 
-// CountUntriggeredSubaccountOrders will count all untriggered orders for a given subaccount that match the provided
-// filter.
-func (k Keeper) CountUntriggeredSubaccountOrders(ctx sdk.Context,
+// CountUntriggeredSubaccountStatefulOrders will count all untriggered stateful orders for a given subaccount.
+func (k Keeper) CountUntriggeredSubaccountStatefulOrders(ctx sdk.Context,
 	subaccountId satypes.SubaccountId,
-	filter func(types.OrderId) bool,
 ) uint32 {
```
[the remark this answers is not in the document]
lucas-dydx: Terrible comment. This is a value type and can't be nil.

protocol/x/clob/keeper/orders_test.go
```diff
@@ -2114,6 +2117,11 @@ func TestInitStatefulOrdersInMemClob(t *testing.T) {
 					ks.ClobKeeper.MustTriggerConditionalOrder(ks.Ctx, order.OrderId)
 				}
 
+				// Increment the expected count for non-conditional orders and triggered conditional orders.
```
lucas-dydx, line 2120: Shouldn't the number of open stateful orders include untriggered conditional orders in this test expectation?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -478,3 +489,51 @@ func (k Keeper) getUntriggeredConditionalOrdersIterator(ctx sdk.Context) sdk.Ite
...
 	return sdk.KVStorePrefixIterator(store, []byte{})
 }
+
+// GetStatefulOrderCount gets a count of how many stateful orders are written to state. This does not
+// include any untriggered conditional orders.
+//
+// OrderId can be conditional or long term.
+func (k Keeper) GetStatefulOrderCount(
+	ctx sdk.Context,
+	orderId types.OrderId,
```
lucas-dydx, line 499: Any reason why we don't pass in a subaccount ID here instead of an order ID?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -478,3 +489,51 @@ func (k Keeper) getUntriggeredConditionalOrdersIterator(ctx sdk.Context) sdk.Ite
 	)
 	return sdk.KVStorePrefixIterator(store, []byte{})
 }
+
+// GetStatefulOrderCount gets a count of how many stateful orders are written to state. This does not
```
lucas-dydx, line 493:
```suggestion
// GetStatefulOrderCount gets a count of how many stateful orders are written to state for a subaccount. This does not
```

protocol/x/clob/keeper/untriggered_conditional_orders.go
```diff
@@ -323,21 +323,19 @@ func (k Keeper) MaybeTriggerConditionalOrders(ctx sdk.Context) (triggeredConditi
 	return triggeredConditionalOrderIds
 }
 
-// CountUntriggeredSubaccountOrders will count all untriggered orders for a given subaccount that match the provided
-// filter.
-func (k Keeper) CountUntriggeredSubaccountOrders(ctx sdk.Context,
+// CountUntriggeredSubaccountStatefulOrders will count all untriggered stateful orders for a given subaccount.
```
lucas-dydx, line 326: Suggestion, since conditional orders can only be stateful:
```suggestion
// CountUntriggeredSubaccountConditionalOrders will count all untriggered conditional orders for a given subaccount.
```

protocol/x/clob/keeper/untriggered_conditional_orders.go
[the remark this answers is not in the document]
lucas-dydx: Can you resolve this comment?

protocol/x/clob/keeper/untriggered_conditional_orders.go
```diff
@@ -323,21 +323,19 @@ func (k Keeper) MaybeTriggerConditionalOrders(ctx sdk.Context) (triggeredConditi
...
+func (k Keeper) CountUntriggeredSubaccountStatefulOrders(ctx sdk.Context,
 	subaccountId satypes.SubaccountId,
-	filter func(types.OrderId) bool,
 ) uint32 {
 	count := uint32(0)
 	for _, untriggeredConditionalOrders := range k.UntriggeredConditionalOrders {
 		for _, order := range untriggeredConditionalOrders.OrdersToTriggerWhenOraclePriceGTETriggerPrice {
-			if order.OrderId.SubaccountId == subaccountId && filter(order.OrderId) {
+			if order.OrderId.SubaccountId == subaccountId && order.IsStatefulOrder() {
```
lucas-dydx, line 333: When will `order.IsStatefulOrder()` ever be false? Short-Term orders can't be conditional

lcwik:
Rebase to deal with merge conflict and fixed unit test expectations.

lucas-dydx (approved):

jayy04 (approved):

protocol/x/clob/keeper/orders_test.go
lcwik: No, `GetStatefulOrderCount` represents all triggered and non-conditional stateful orders. Untriggered orders are counted separately.

Minor update to comment.

protocol/x/clob/keeper/stateful_order_state.go
lcwik: Swapped to subaccount. 

It was mainly to ensure that the counts are only fetched when processing a stateful order but that seems like overkill.

protocol/x/clob/keeper/untriggered_conditional_orders.go
lcwik: I removed it.

It was mainly to prevent a regression in case conditional orders are expanded to include short-term orders for some reason.

PR #437, opened 2023-10-02
[CORE-627] - Fix daemon panics by delaying daemon start until gRPC service is available

clemire:
The code change to cause daemon panics when daemons fail to respond in a timely fashion has led to an unintended effect: when genesis time is delayed after a node starts, the node will sleep, and the daemon monitoring service will cause a panic.  The root cause of this is unavailability of the cosmos gRPC service - daemons will hang when trying to establish a connection to this service and will never send calls to the daemon server.

This PR delays daemon / daemon service start until after the gRPC service is initilialized.

I implemented with a minimal change by adding a closure function that is called whenever gRPC routes are registered on the gRPC server. Open to other suggestions. This fixes the issue locally and all tests pass.

### DIFF
--- a/protocol/x/prices/keeper/market.go
+++ b/protocol/x/prices/keeper/market.go
@@ -74,17 +74,22 @@ func (k Keeper) CreateMarket(
 	return marketParam, nil
 }
 
-// IsRecentlyAdded returns true if the market was added recently. Since it takes a few seconds for
-// index prices to populate, we would not consider missing index prices for a recently added market
-// to be an error.
-func (k Keeper) IsRecentlyAdded(marketId uint32) bool {
+// IsRecentlyAvailable returns true if the market was recently made available to the pricefeed daemon. A market is
+// considered recently available either if it was recently created, or if the pricefeed daemon was recently started. If
+// an index price does not exist for a recently available market, the protocol does not consider this an error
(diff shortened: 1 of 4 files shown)


protocol/x/prices/keeper/market.go
```diff
@@ -74,17 +74,22 @@ func (k Keeper) CreateMarket(
 	return marketParam, nil
 }
 
-// IsRecentlyAdded returns true if the market was added recently. Since it takes a few seconds for
-// index prices to populate, we would not consider missing index prices for a recently added market
-// to be an error.
-func (k Keeper) IsRecentlyAdded(marketId uint32) bool {
+// IsRecentlyAvailable returns true if the market was recently made available to the pricefeed daemon. A market is
```
clemire, line 77: This method controls when missing index prices are escalated to errors in the logs. When a node starts with a delayed genesis time, the market is created on node start, but prices are still not available until the daemon starts, so this logic was updated to use block height as a proxy for daemon warmup period.

vincentwschau: nit: Can you include the reasoning about the block height being used as a proxy into the comment?

vincentwschau (approved):

lcwik (commented):
LGTM

PR #445, opened 2023-10-03
Add prod_genesis.sh and `sample_pregenesis.json`

teddyding:
# Overview
Goal of script: the genesis coordination party can simply modify  values under `TODO(GENESIS)` and have a production-ready genesis. 

Unfortunately due to historical reasons, the genesis state is written at 3 layers, each overwriting the previous one. In the interest of minimum refactoring, this PR assumes the following convention for "what to put in each layer"

1. `DefaultParams` (in `x/module/types`): 
Generally, values here are basis shared among different networks (Unfortunately due to historical reasons, clob, perpetual, prices state, liquiditiers were all in `genesis.sh` instead of default params)

2. `genesis.sh` (overwrites default params):
Testnet specific setup and other common setup due to historical reasons (e.g. clobpairs)

3. `prod_pregenesis.sh` (further overwrites genesis.sh)
Default production values. 

# Test 
Tested locally by running generated genesis with `.alice` validator

### Summary by CodeRabbit

- New Feature: Introduced a bash script to generate pregenesis files for production networks, allowing customization of various genesis parameters.
- Refactor: Updated the `DefaultParams` function in `params.go` and `GenesisState` constant in `genesis.go` to reflect changes in currency denomination precision and vesting entries.
- Test: Modified tests in `keeper_test.go`, `genesis_test.go`, and `query_vest_entry_test.go` to accommodate changes in vesting entries and denomination precision.
- Chore: Updated `genesis.sh` script to include new functions for setting token metadata and editing genesis file.

### DIFF
--- a/protocol/app/testdata/default_genesis_state.json
+++ b/protocol/app/testdata/default_genesis_state.json
@@ -351,7 +351,7 @@
       "treasury_account":"rewards_treasury",
       "denom":"dv4tnt",
       "denom_exponent":-18,
-      "market_id":11,
+      "market_id":1,
       "fee_multiplier_ppm":990000
     }
   },
--- a/protocol/app/testdata/default_genesis_state.json
+++ b/protocol/app/testdata/default_genesis_state.json
@@ -350,7 +350,7 @@
     "params": {
       "treasury_account":"rewards_treasury",
       "denom":"dv4tnt",
-      "denom_exponent":-6,
+      "denom_exponent":-18,
       "market_id":1,
       "fee_multiplier_ppm":990000
     }
@@ -406,8 +406,15 @@
     "vest_entries": [
       {
         "denom": "dv4tnt",
-        "end_time": "2023-10-13T00:00:00Z",
-        "start_time": "2023-09-13T00:00:00Z",
+        "end_time": "2025-01-01T00:00:00Z",
+        "start_time": "2023-01-01T00:00:00Z",
+        "treasury_account": "community_treasury",
+        "vester_account": "community_vester"
+      },
+      {
+        "denom": "dv4tnt",
+        "end_time": "2025-01-01T00:00:00Z",
+        "start_time": "2023-01-01T00:00:00Z",
         "treasury_account": "rewards_treasury",
         "vester_account": "rewards_vester"
       }
--- a/protocol/app/testdata/default_genesis_state.json
+++ b/protocol/app/testdata/default_genesis_state.json
@@ -351,7 +351,7 @@
       "treasury_account":"rewards_treasury",
       "denom":"dv4tnt",
       "denom_exponent":-18,
-      "market_id":1,
+      "market_id":11,
       "fee_multiplier_ppm":990000
     }
   },
--- a/protocol/scripts/genesis/prod_pregenesis.sh
+++ b/protocol/scripts/genesis/prod_pregenesis.sh
@@ -36,12 +36,12 @@ NATIVE_TOKEN="adv4tnt"
 NATIVE_TOKEN_WHOLE_COIN="dv4tnt"
 # Human readable name of token.
 COIN_NAME="dYdX Testnet Token"
-# Market ID in the oracle price list for the rwards token.
+# Market ID in the oracle price list for the rewards token.
 REWARDS_TOKEN_MARKET_ID=11
 # The numerical chain ID of the Ethereum chain for bridge daemon to query.
 ETH_CHAIN_ID=1
 # The address of the Ethereum contract for bridge daemon to monitor for logs.
-ETH_BRIDGE_ADDRESS="0xcca9D5f0a3c58b6f02BD0985fC7F9420EA24C1f0"
+ETH_BRIDGE_ADDRESS="0xcca9D5f0a3c58b6f02BD0985fC7F9420EA24C1f0" # default value points to a Sepolia contract
 # The next event id (the last processed id plus one) of the logs from the Ethereum contract.
 BRIDGE_GENESIS_ACKNOWLEDGED_NEXT_ID=0
 # The Ethereum block height of the most recently processed bridge event.
@@ -57,7 +57,7 @@ REWARDS_VEST_START_TIME="2001-01-01T00:00:00Z"
 # End time of the rewards vesting schedule.
 REWARDS_VEST_END_TIME="2050-01-01T00:00:00Z"
 
-################## Start of required values to be updated ##################
+################## End of required values to be updated ##################
 
 cleanup_tmp_dir() {
 	if [ -d "$TMP_EXCHANGE_CONFIG_JSON_DIR" ]; then
@@ -121,11 +121,15 @@ function overwrite_genesis_production() {
     dasel put -t int -f "$GENESIS" '.app_state.rewards.params.fee_multiplier_ppm' -v '0'
 
     # Vest params
-    # For communmity treasury
+    # For community treasury
+	dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].vester_account' -v "community_vester"
+	dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].treasury_account' -v "community_treasury"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].denom' -v "$NATIVE_TOKEN"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].start_time' -v "$COMMUNITY_VEST_START_TIME"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].end_time' -v "$COMMUNITY_VEST_END_TIME"
     # For rewards treasury
+	dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].vester_account' -v "rewards_vester"
+	dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].treasury_account' -v "rewards_treasury"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].denom' -v "$NATIVE_TOKEN"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].start_time' -v "$REWARDS_VEST_START_TIME"
     dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].end_time' -v "$REWARDS_VEST_END_TIME"
--- a/protocol/scripts/genesis/prod_pregenesis.sh
+++ b/protocol/scripts/genesis/prod_pregenesis.sh
@@ -30,16 +30,29 @@ source "./testing/genesis.sh"
 # TODO(GENESIS): Update below values before running this script. Sample values are shown.
 ################## Start of required values to be updated ##################
 CHAIN_ID="dydx-1"
-NATIVE_TOKEN="dv4tnt"
+# Base denomination of the native token. Usually comes with a prefix "u-", "a-" to indicate unit.
+NATIVE_TOKEN="adv4tnt"
+# Denomination of the native token in whole coins.
+NATIVE_TOKEN_WHOLE_COIN="dv4tnt"
+# Market ID in the oracle price list for the rwards token.
 REWARDS_TOKEN_MARKET_ID=11
+# The numerical chain ID of the Ethereum chain for bridge daemon to query.
 ETH_CHAIN_ID=1
+# The address of the Ethereum contract for bridge daemon to monitor for logs.
 ETH_BRIDGE_ADDRESS="0xcca9D5f0a3c58b6f02BD0985fC7F9420EA24C1f0"
+# The next event id (the last processed id plus one) of the logs from the Ethereum contract.
 BRIDGE_GENESIS_ACKNOWLEDGED_NEXT_ID=0
+# The Ethereum block height of the most recently processed bridge event.
 BRIDGE_GENESIS_ACKNOWLEDGED_ETH_BLOCK_HEIGHT=0
+# Genesis time of the chain.
 GENESIS_TIME="2023-12-31T00:00:00Z"
+# Start time of the community vesting schedule.
 COMMUNITY_VEST_START_TIME="2001-01-01T00:00:00Z"
+# End time of the community vesting schedule.
 COMMUNITY_VEST_END_TIME="2050-01-01T00:00:00Z"
+# Start time of the rewards vesting schedule.
 REWARDS_VEST_START_TIME="2001-01-01T00:00:00Z"
+# End time of the rewards vesting schedule.
 REWARDS_VEST_END_TIME="2050-01-01T00:00:00Z"
 
 ################## Start of required values to be updated ##################
@@ -74,6 +87,7 @@ function overwrite_genesis_production() {
 	dasel put -t string -f "$GENESIS" '.app_state.distribution.params.community_tax' -v '0.0' # 0%
 	dasel put -t bool -f "$GENESIS" '.app_state.distribution.params.withdraw_addr_enabled' -v 'true'
 
+	# Bank params
     # Initialize bank balance of bridge module account.
     dasel put -t json -f "$GENESIS" ".app_state.bank.balances" -v "[]"
 	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[]" -v "{}"
@@ -81,6 +95,9 @@ function overwrite_genesis_production() {
 	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[0].coins.[]" -v "{}"
 	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].denom" -v "${NATIVE_TOKEN}"
 	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].amount" -v "${BRIDGE_MODACC_BALANCE}"
+	# Set denom metadata
+	set_denom_metadata "$NATIVE_TOKEN" "$NATIVE_TOKEN_WHOLE_COIN"
+
 	# Governance params
 	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].amount' -v "10000$EIGHTEEN_ZEROS" # 10k whole coins of native token
 	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].denom' -v "$NATIVE_TOKEN"
@@ -130,7 +147,7 @@ function overwrite_genesis_production() {
 	dasel put -t int -f "$GENESIS" '.app_state.bridge.acknowledged_event_info.eth_block_height' -v "$BRIDGE_GENESIS_ACKNOWLEDGED_ETH_BLOCK_HEIGHT"
 
     # Crisis module
-    dasel put -t string -f "$GENESIS" '.app_state.crisis.constant_fee.amount' -v "1$EIGHTEEN_ZEROS"
+    dasel put -t string -f "$GENESIS" '.app_state.crisis.constant_fee.amount' -v "1$EIGHTEEN_ZEROS" # 1 whole coin of native denom
     dasel put -t string -f "$GENESIS" '.app_state.crisis.constant_fee.denom' -v "$NATIVE_TOKEN"
 
     # Genesis time
--- a/protocol/scripts/genesis/prod_pregenesis.sh
+++ b/protocol/scripts/genesis/prod_pregenesis.sh
@@ -34,6 +34,8 @@ CHAIN_ID="dydx-1"
 NATIVE_TOKEN="adv4tnt"
 # Denomination of the native token in whole coins.
 NATIVE_TOKEN_WHOLE_COIN="dv4tnt"
+# Human readable name of token.
+COIN_NAME="dYdX Testnet Token"
 # Market ID in the oracle price list for the rwards token.
 REWARDS_TOKEN_MARKET_ID=11
 # The numerical chain ID of the Ethereum chain for bridge daemon to query.
@@ -96,7 +98,7 @@ function overwrite_genesis_production() {
 	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].denom" -v "${NATIVE_TOKEN}"
 	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].amount" -v "${BRIDGE_MODACC_BALANCE}"
 	# Set denom metadata
-	set_denom_metadata "$NATIVE_TOKEN" "$NATIVE_TOKEN_WHOLE_COIN"
+	set_denom_metadata "$NATIVE_TOKEN" "$NATIVE_TOKEN_WHOLE_COIN" "$COIN_NAME"
 
 	# Governance params
 	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].amount' -v "10000$EIGHTEEN_ZEROS" # 10k whole coins of native token
@@ -131,6 +133,7 @@ function overwrite_genesis_production() {
     # Delayed message params
     # Schedule a delayed message to swap fee tiers to the standard schedule after ~120 days of blocks.
 	dasel put -t int -f "$GENESIS" '.app_state.delaymsg.num_messages' -v '1'
+	dasel put -t json -f "$GENESIS" '.app_state.delaymsg.delayed_messages' -v "[]"
 	dasel put -t json -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[]' -v "{}"
 	dasel put -t int -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[0].id' -v '0'
 
--- a/protocol/scripts/genesis/prod_pregenesis.sh
+++ b/protocol/scripts/genesis/prod_pregenesis.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+set -eo pipefail
+
+# This script creates the pregenesis file for the public testnet.
+#
+# The script must be run from the root of the `v4-chain` repo.
+#
+# example usage:
+# $ ./scripts/genesis/prod_pregenesis.sh ./build/dydxprotocold
+
+# Check for missing required arguments
+if [ -z "$1" ]; then
+  echo "Error: Missing required argument DYDX_BINARY."
+  echo "Usage: $0 <DYDX_BINARY> [-s|--SEED_FAUCET_USDC]"
+  exit 1
+fi
+
+# Capture the required argument
+DYDX_BINARY="$1"
+
+TMP_CHAIN_DIR="/tmp/prod-chain"
+TMP_EXCHANGE_CONFIG_JSON_DIR="/tmp/prod-exchange_config"
+FIFTEEN_ZEROS="000000000000000"
+EIGHTEEN_ZEROS="000$FIFTEEN_ZEROS"
+BRIDGE_MODACC_BALANCE="1000000000$EIGHTEEN_ZEROS" # 1e27
+BRIDGE_MODACC_ADDR="dydx1zlefkpe3g0vvm9a4h0jf9000lmqutlh9jwjnsv"
+
+source "./testing/genesis.sh"
+# TODO(GENESIS): Update below values before running this script. Sample values are shown.
+################## Start of required values to be updated ##################
+CHAIN_ID="dydx-1"
+NATIVE_TOKEN="dv4tnt"
+REWARDS_TOKEN_MARKET_ID=11
+ETH_CHAIN_ID=1
+ETH_BRIDGE_ADDRESS="0xcca9D5f0a3c58b6f02BD0985fC7F9420EA24C1f0"
+BRIDGE_GENESIS_ACKNOWLEDGED_NEXT_ID=0
+BRIDGE_GENESIS_ACKNOWLEDGED_ETH_BLOCK_HEIGHT=0
+GENESIS_TIME="2023-12-31T00:00:00Z"
+COMMUNITY_VEST_START_TIME="2001-01-01T00:00:00Z"
+COMMUNITY_VEST_END_TIME="2050-01-01T00:00:00Z"
+REWARDS_VEST_START_TIME="2001-01-01T00:00:00Z"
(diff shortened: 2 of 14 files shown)


protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,156 @@
...
+	dasel put -t int -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[0].block_height' -v '6912000'
+
+    # Bridge module params.
+	dasel put -t string -f "$GENESIS" '.app_state.bridge.event_params.denom' -v "$NATIVE_TOKEN"
+	dasel put -t int -f "$GENESIS" '.app_state.bridge.event_params.eth_chain_id' -v "$ETH_CHAIN_ID"
+	dasel put -t string -f "$GENESIS" '.app_state.bridge.event_params.eth_address' -v "$ETH_BRIDGE_ADDRESS"
+	dasel put -t int -f "$GENESIS" '.app_state.bridge.acknowledged_event_info.next_id' -v "$BRIDGE_GENESIS_ACKNOWLEDGED_NEXT_ID"
+	dasel put -t int -f "$GENESIS" '.app_state.bridge.acknowledged_event_info.eth_block_height' -v "$BRIDGE_GENESIS_ACKNOWLEDGED_ETH_BLOCK_HEIGHT"
+
+    # Crisis module
```
teddyding, line 132: Added documentation for this [here](https://www.notion.so/dydx/AC-Priv-Genesis-Parameters-d2321636dd494ee49cc95b7825cbbc98?pvs=4#376e65ad774d4fd382fa3783ef404482)

protocol/x/rewards/types/params.go
```diff
@@ -10,9 +10,9 @@ import (
 func DefaultParams() Params {
 	return Params{
 		// Corresponds to module account address: dydx16wrau2x4tsg033xfrrdpae6kxfn9kyuerr5jjp
-		TreasuryAccount:  "rewards_treasury",
+		TreasuryAccount:  TreasuryAccountName,
 		Denom:            "dv4tnt",
-		DenomExponent:    -6,
+		DenomExponent:    -18,
```
teddyding, line 15: Should be -18 across all environments

protocol/x/vest/types/genesis.go
```diff
@@ -11,12 +11,19 @@ func DefaultGenesis() *GenesisState {
 	return &GenesisState{
 		// TODO(CORE-530): in genesis.sh, overwrite start and end times dynamically for testnets.
 		VestEntries: []VestEntry{
+			{
```
teddyding, line 14: Should have two entries across all environments.

protocol/scripts/genesis/sample_pregenesis.json
```diff
@@ -0,0 +1,1723 @@
+{
```
topperdogger123, line 1: Is this file the result of running `prod_pregenesis.sh`? Checking it in isn't that useful if it can go out of sync.

roy-dydx, line 1: Is this file the result of running prod_pregenesis.sh? Checking it in isn't that useful if it can go out of sync.

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+	# Consensus params
+	dasel put -t string -f "$GENESIS" '.consensus_params.block.max_bytes' -v '4194304'
+	dasel put -t string -f "$GENESIS" '.consensus_params.block.max_gas' -v '-1'
+
+    # Rewards params
+    dasel put -t string -f "$GENESIS" '.app_state.rewards.params.denom' -v "$NATIVE_TOKEN"
+    dasel put -t int -f "$GENESIS" '.app_state.rewards.params.fee_multiplier_ppm' -v '0'
+
+    # Vest params
+    # For communmity treasury
```
ttl33, line 124:
```suggestion
    # For community treasury
```

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+    dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].denom' -v "$NATIVE_TOKEN"
+    dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[0].start_time' -v "$COMMUNITY_VEST_START_TIME"
```
ttl33, line 126: where do we specify the vester and treasury account module names?

i.e. `community_vester` `community_treasury`

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+    dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].start_time' -v "$REWARDS_VEST_START_TIME"
+    dasel put -t string -f "$GENESIS" '.app_state.vest.vest_entries.[1].end_time' -v "$REWARDS_VEST_END_TIME"
+
+    # Delayed message params
+    # Schedule a delayed message to swap fee tiers to the standard schedule after ~120 days of blocks.
+	dasel put -t int -f "$GENESIS" '.app_state.delaymsg.num_messages' -v '1'
+	dasel put -t json -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[]' -v "{}"
+	dasel put -t int -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[0].id' -v '0'
+
+	delaymsg=$(cat "$DELAY_MSG_JSON_DIR/perpetual_fee_params_msg.json" | jq -c '.')
```
ttl33, line 140: do we specify the `DELAY_MSG_JSON_DIR` somewhere, I dont think this is called out in the above "TODO(GENESIS)"

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+	dasel put -t json -f "$GENESIS" '.app_state.delaymsg.delayed_messages.[0].msg' -v "$delaymsg"
```
ttl33, line 141: where is the $delaymsg var being initialized?

protocol/scripts/genesis/prod_pregenesis.sh
ttl33, line 140: separately, can we get YuJin to validate the json file's content?

protocol/scripts/genesis/prod_pregenesis.sh
ttl33: Should this be higher since we want to discourage spamming?

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+	if [ -d "$TMP_EXCHANGE_CONFIG_JSON_DIR" ]; then
+		rm -r "$TMP_EXCHANGE_CONFIG_JSON_DIR"
+	fi
+	if [ -d "$TMP_CHAIN_DIR" ]; then
+		rm -r "$TMP_CHAIN_DIR"
+	fi
+}
+
+# Set production default genesis params.
+function overwrite_genesis_production() {
```
ttl33, line 72: I believe this was the original meta point you raised, but I wonder if we can do something better for now.

Can we add overrides for all the modules? For example, `blocktime`. I think it would be generally easier to reason about that this script will override all the module parameters rather than having to refer to the `testing/genesis.sh` or the default param.

Later we can think about refactoring these.

protocol/app/testdata/default_genesis_state.json
```diff
@@ -406,8 +406,15 @@
     "vest_entries": [
       {
         "denom": "dv4tnt",
-        "end_time": "2023-10-13T00:00:00Z",
-        "start_time": "2023-09-13T00:00:00Z",
+        "end_time": "2025-01-01T00:00:00Z",
```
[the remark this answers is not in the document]
teddyding: They are applicable to unit test and e2e. Plus, they serve as the basis for scripts to override. If not overridden, the value will remain in the testnet or prod genesis.

protocol/testing/genesis.sh
```diff
@@ -1498,3 +1493,19 @@ update_genesis_complete_bridge_delay() {
...
+	local WHOLE_COIN_DENOM=$2
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata" -v "[]"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.denom_metadata.[0].description" -v "The native token of the network"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units" -v "[]"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[0].denom" -v "$BASE_DENOM"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[1].denom" -v "$WHOLE_COIN_DENOM"
+	dasel put -t int -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[1].exponent" -v 18
```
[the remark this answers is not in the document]
teddyding: Yes, can confirm. Can use below query for osmosis example

grpcurl -plaintext -d '{"denom":"uosmo"}' grpc.osmosis.zone:9090 cosmos.bank.v1beta1.Query/DenomMetadata

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+
+# Capture the required argument
+DYDX_BINARY="$1"
+
+TMP_CHAIN_DIR="/tmp/prod-chain"
+TMP_EXCHANGE_CONFIG_JSON_DIR="/tmp/prod-exchange_config"
+FIFTEEN_ZEROS="000000000000000"
+EIGHTEEN_ZEROS="000$FIFTEEN_ZEROS"
+BRIDGE_MODACC_BALANCE="1000000000$EIGHTEEN_ZEROS" # 1e27
+BRIDGE_MODACC_ADDR="dydx1zlefkpe3g0vvm9a4h0jf9000lmqutlh9jwjnsv"
```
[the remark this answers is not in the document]
teddyding: This is constant. See module_accounts_test.go for mapping between module accounts and addresses

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].address" -v "${BRIDGE_MODACC_ADDR}"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[0].coins.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].denom" -v "${NATIVE_TOKEN}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].amount" -v "${BRIDGE_MODACC_BALANCE}"
+	# Set denom metadata
+	set_denom_metadata "$NATIVE_TOKEN" "$NATIVE_TOKEN_WHOLE_COIN"
+
+	# Governance params
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].amount' -v "10000$EIGHTEEN_ZEROS" # 10k whole coins of native token
```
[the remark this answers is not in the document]
teddyding: No, all balances need to be in base demon.
The chain doesn't understand what a full coin is. Denom metadata is aimed for human readability

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,176 @@
...
+	dasel put -t int -f "$GENESIS" '.app_state.staking.params.max_entries' -v '7'
+	dasel put -t int -f "$GENESIS" '.app_state.staking.params.historical_entries' -v '10000'
+
+	# Distribution params
+	dasel put -t string -f "$GENESIS" '.app_state.distribution.params.community_tax' -v '0.0' # 0%
+	dasel put -t bool -f "$GENESIS" '.app_state.distribution.params.withdraw_addr_enabled' -v 'true'
+
+	# Bank params
+    # Initialize bank balance of bridge module account.
+    dasel put -t json -f "$GENESIS" ".app_state.bank.balances" -v "[]"
```
roy-dydx, line 94: nit: indentation

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,176 @@
...
+
+	# Governance params
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].amount' -v "10000$EIGHTEEN_ZEROS" # 10k whole coins of native token
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.min_deposit.[0].denom' -v "$NATIVE_TOKEN"
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.max_deposit_period' -v '172800s' # 2 days
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.voting_period' -v '345600s' # 4 days
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.quorum' -v '0.33400' # 33.4%
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.threshold' -v '0.50000' # 50%
+	dasel put -t string -f "$GENESIS" '.app_state.gov.params.veto_threshold' -v '0.33400' # 33.4%
+    dasel put -t bool -f "$GENESIS" '.app_state.gov.params.burn_proposal_deposit_prevote' -v 'false'
```
roy-dydx, line 111: nit: indentation

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+  echo "Usage: $0 <DYDX_BINARY> [-s|--SEED_FAUCET_USDC]"
+  exit 1
+fi
+
+# Capture the required argument
+DYDX_BINARY="$1"
+
+TMP_CHAIN_DIR="/tmp/prod-chain"
+TMP_EXCHANGE_CONFIG_JSON_DIR="/tmp/prod-exchange_config"
+FIFTEEN_ZEROS="000000000000000"
```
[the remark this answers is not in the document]
teddyding: Originally added `FIFTEEN_ZEROS` because I wanted to add a some value for `100$FIFTEEN_ZEROS` which turned out wasn't necessary. IMO it's probably cleaner to just define one constant of `$EIGHTEEN_ZEROS`

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: I believe 1 full coin per tx is inline with our conventions wrt to transaction spamming ([example](https://dydx-team.slack.com/archives/C05PD2S76D7/p1692997533646139?thread_ts=1692996873.257109&cid=C05PD2S76D7)). 

Ultimately, it's pretty costly to flood the network with these transactions if each one costs 1 full coin

protocol/testing/genesis.sh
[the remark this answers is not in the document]
teddyding: `nit: can we use 18 from some shared constant to minimize places that we need to change?` This is the only place we are using `18` as an exponent, so explicitly using `18` is probably better than defining a constant

protocol/testing/genesis.sh
```diff
@@ -1498,3 +1493,19 @@ update_genesis_complete_bridge_delay() {
...
+
+set_denom_metadata() {
+	local BASE_DENOM=$1
+	local WHOLE_COIN_DENOM=$2
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata" -v "[]"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.denom_metadata.[0].description" -v "The native token of the network"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units" -v "[]"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.denom_metadata.[0].denom_units.[0].denom" -v "$BASE_DENOM"
```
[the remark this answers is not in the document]
teddyding: Per definition of [exponent](https://github.com/dydxprotocol/cosmos-sdk/blob/93454d9f2fb6f262e63a55641e44c024bd326860/proto/cosmos/bank/v1beta1/bank.proto#L90) field, base denom doesn't have an exponent

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: Yes, in [genesis.sh](https://github.com/dydxprotocol/v4-chain/blob/1eefd073950dfe937341973d8ef0791360e4945a/protocol/testing/genesis.sh#L57)
It's not be part of `TODO(GENESIS)`, because the it's not required to be changed to launch a potential production network.
`TODO(GENESIS)` consists of items that *must be* edited to launch a production network. *All* other parameters can be configured as seen fit by the coordination party.

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: In [genesis.sh](https://github.com/dydxprotocol/v4-chain/blob/1eefd073950dfe937341973d8ef0791360e4945a/protocol/testing/genesis.sh#L57)

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: Tagged Yujin in #p-v4-fee-schedule

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: IMO we should keep minimal logic in `overwrite_genesis_production`. It should only contain values that *must* live in this method, which includes
- Values under `TODO(GENESIS)` that need to be updated by genesis coordination party
- Values that are different between testnets and a production network

Reason for keeping `overwrite_genesis_production` minimal:
- It's not covered by any test, whereas `DefaultParams` in Golang code are.
- It's impractical to override all modules (at that point we are just manually specifying the whole genesis.json file)
- It's not the responsibility of this script to reason that necessary parameters are overridden. Instead, anyone should only rely on the final genesis file to reason if parameters are set up correctly, since that's the final source of truth.

protocol/scripts/genesis/sample_pregenesis.json
teddyding: Agreed. In the interest of time, I'll manually keep them in sync for this PR. Can work on a script later to validate

protocol/app/testdata/default_genesis_state.json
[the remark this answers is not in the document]
ttl33: Why would values set in `testdata` affect testnet or prod genesis?

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+
+# TODO(GENESIS): Update below values before running this script. Sample values are shown.
+################## Start of required values to be updated ##################
+CHAIN_ID="dydx-1"
+# Base denomination of the native token. Usually comes with a prefix "u-", "a-" to indicate unit.
+NATIVE_TOKEN="adv4tnt"
+# Denomination of the native token in whole coins.
+NATIVE_TOKEN_WHOLE_COIN="dv4tnt"
+# Market ID in the oracle price list for the rwards token.
+REWARDS_TOKEN_MARKET_ID=11
```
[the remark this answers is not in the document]
teddyding: Possible, but feels like an overhaul that has limited benefit. In most unit and e2e tests we assume market `0` to be BTC, so prefer to not shake things up this PR 😄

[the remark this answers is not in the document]
ttl33: Actually, I think the USDT market ID is 1_000_000 ([code](https://github.com/dydxprotocol/v4-chain/blob/1eefd073950dfe937341973d8ef0791360e4945a/protocol/testing/genesis.sh#L930))

The reward token market seems to be 1_000_001 ([code](https://github.com/dydxprotocol/v4-chain/pull/455/files#diff-419a625a9d75c24f92e64167f64933ab42ecf6079f27626a2ace77608466c78bR945))

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: Generally, testnet-specific parameters should be done in this separate [overwrite_genesis_public_genesis](https://github.com/dydxprotocol/v4-chain/blob/1eefd073950dfe937341973d8ef0791360e4945a/protocol/testing/testnet-external/pregenesis.sh#L135) method, so I agree there's concern but the risk is low.

We also have a [ticket](https://linear.app/dydx/issue/CORE-632/workflow-to-ensure-sample-pregenesisjson-in-sync-with-prod-genesissh) to ensure the `sample_pregenesis.json` is in sync with the `prod_pregenesis.sh` script, which will prevent unexpected side effect

protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -0,0 +1,173 @@
...
+	dasel put -t bool -f "$GENESIS" '.app_state.distribution.params.withdraw_addr_enabled' -v 'true'
+
+	# Bank params
+    # Initialize bank balance of bridge module account.
+    dasel put -t json -f "$GENESIS" ".app_state.bank.balances" -v "[]"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].address" -v "${BRIDGE_MODACC_ADDR}"
+	dasel put -t json -f "$GENESIS" ".app_state.bank.balances.[0].coins.[]" -v "{}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].denom" -v "${NATIVE_TOKEN}"
+	dasel put -t string -f "$GENESIS" ".app_state.bank.balances.[0].coins.[0].amount" -v "${BRIDGE_MODACC_BALANCE}"
```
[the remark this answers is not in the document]
teddyding: By design of bridge module it mints all native tokens upfront. At bridge event completion, funds are transferred from bridge module account to user account.

protocol/scripts/genesis/sample_pregenesis.json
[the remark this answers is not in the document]
teddyding: Yes, also see [discussion](https://github.com/dydxprotocol/v4-chain/pull/445/files#r1345190452)

protocol/scripts/genesis/prod_pregenesis.sh
teddyding: It's set in `func DefaultGenesis() *GenesisState` function in `x/vest`. I debated whether to explicitly overwrite the account names here. I was concerned that overriding here could lead to typo, but not overriding here also isn't good for readability. Ultimately decided to override explicitly in this script

protocol/testing/delaymsg_config/perpetual_fee_params_msg.json
```diff
@@ -56,15 +56,15 @@
         "absolute_volume_requirement": "125000000000000",
         "total_volume_share_requirement_ppm": 5000,
         "maker_volume_share_requirement_ppm": 10000,
-        "maker_fee_ppm": -90,
+        "maker_fee_ppm": -70,
```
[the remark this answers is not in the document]
teddyding: Makes sense. Alternatively make sense to change directory of `testing` to something like `networks` since it really is used for setting up production and test networks

Filed ticket [here](https://linear.app/dydx/issue/CORE-633/mov-files-used-by-prod-genesis-out-of-the-testing-directory)

protocol/scripts/genesis/prod_pregenesis.sh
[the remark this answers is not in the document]
teddyding: Those IDs look fine to me

protocol/testutil/constants/genesis.go
```diff
@@ -1473,8 +1473,15 @@ const GenesisState = `{
       "vest_entries": [
         {
           "denom": "dv4tnt",
-          "end_time": "2023-10-13T00:00:00Z",
-          "start_time": "2023-09-13T00:00:00Z",
+          "end_time": "2025-01-01T00:00:00Z",
```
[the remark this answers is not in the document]
teddyding: This test file contains the raw literal string of genesis json. I did added a constant in golang file below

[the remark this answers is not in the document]
ttl33: I think if you use a "date" object, you can print the string literal.

protocol/scripts/genesis/prod_pregenesis.sh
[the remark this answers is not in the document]
ttl33: @tqin7 can you double confirm this? just wanted to be extra sure here since this is pretty important

protocol/testing/delaymsg_config/perpetual_fee_params_msg.json
```diff
@@ -56,15 +56,15 @@
...
         "taker_fee_ppm": 250
       },
       {
         "name": "8",
         "absolute_volume_requirement": "125000000000000",
         "total_volume_share_requirement_ppm": 5000,
         "maker_volume_share_requirement_ppm": 20000,
-        "maker_fee_ppm": -110,
+        "maker_fee_ppm": -90,
```
ttl33, line 67: To confirm, is someone from research confirming these changes?

protocol/testutil/constants/genesis.go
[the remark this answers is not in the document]
ttl33: oops, I now understand what you meant by raw literal string (i.e. this entire section is just lines of strings)

protocol/x/rewards/types/params.go
```diff
@@ -10,9 +10,9 @@ import (
 func DefaultParams() Params {
 	return Params{
 		// Corresponds to module account address: dydx16wrau2x4tsg033xfrrdpae6kxfn9kyuerr5jjp
-		TreasuryAccount:  "rewards_treasury",
+		TreasuryAccount:  TreasuryAccountName,
 		Denom:            "dv4tnt",
```
ttl33, line 14: nit: don't we define `dv4tnt` somewhere as a constant? can we reuse that?

protocol/x/rewards/types/params.go
ttl33, line 15: nit: same here, don't we define what the denom exp is for the native staking token? can we use that const?

protocol/x/rewards/types/params.go
```diff
@@ -10,9 +10,9 @@ import (
...
-		DenomExponent:    -6,
+		DenomExponent:    -18,
 		MarketId:         1,
```
ttl33, line 16: is this market id correct?

protocol/x/vest/types/genesis.go
```diff
@@ -6,17 +6,29 @@ import (
...
+
 // DefaultGenesis returns the default genesis state
 func DefaultGenesis() *GenesisState {
 	return &GenesisState{
 		// TODO(CORE-530): in genesis.sh, overwrite start and end times dynamically for testnets.
 		VestEntries: []VestEntry{
+			{
+				VesterAccount:   CommunityVesterAccountName,
+				TreasuryAccount: CommunityTreasuryAccountName,
+				Denom:           "dv4tnt",
```
ttl33, line 22: same here, don't we define some constnat?

protocol/scripts/genesis/prod_pregenesis.sh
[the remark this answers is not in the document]
ttl33: I am confused. So shouldn't this number be `1000001`?

ttl33 (approved):

protocol/x/rewards/types/params.go
teddyding: We actually don't right now - prefer to keep as is to avoid conflict with this other [PR](https://github.com/dydxprotocol/v4-chain/pull/454)

protocol/x/vest/types/genesis.go
teddyding: See above

protocol/scripts/genesis/prod_pregenesis.sh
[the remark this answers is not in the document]
tqin7: Correct!

protocol/app/testdata/default_genesis_state.json
```diff
@@ -350,8 +350,8 @@
     "params": {
       "treasury_account":"rewards_treasury",
       "denom":"dv4tnt",
-      "denom_exponent":-6,
-      "market_id":1,
+      "denom_exponent":-18,
+      "market_id":11,
```
yang-dydx, line 354: Whats the `market_id` for here? Should it be the same as the `MarketId` [here](https://github.com/dydxprotocol/v4-chain/blob/c491adc0054f61a94612c65a294c85b6c91cf57f/protocol/testutil/daemons/pricefeed/exchange_config/market_id.go#L82)?

yang-dydx: oh nvm saw thread below

protocol/x/rewards/types/params.go
teddyding: Actually going to revert this back to `1` since the default state is used in multiple e2e tests.
This field is meant to be overridden in `testnet` and prod network; `1` is a default sample value so prefer to keep it for now.

protocol/scripts/genesis/sample_pregenesis.json
yujin512, line 1: Can we add the safety tier into perpetuals/liquidity tiers? https://www.notion.so/dydx/Market-Parameters-for-V4-markets-d678dbc594bf483293e829dcb57466f5?pvs=4#1b980ed9a4244d95a97eddeaea03e74c

PR #450, opened 2023-10-03
add integration tests for perpetuals module governance messages

tqin7:
add integration tests for `MsgUpdateParams`, `MsgUpdatePerpetualParams`, `MsgSetLiquidityTier`


protocol/testing/e2e/gov/perpetuals_test.go
```diff
@@ -0,0 +1,305 @@
...
+				ctx,
+				&tApp,
+				[]sdk.Msg{tc.msg},
+				tc.expectSubmitProposalFail,
+				tc.expectedProposalStatus,
+			)
+
+			// If governance proposal is supposed to fail submission, verify that perpetuals
+			// params match the ones before proposal submission.
+			if tc.expectSubmitProposalFail {
```
clemire, line 86: Sorry to revisit this after you merged, but... can you add this check-original-state clause to the bridge it tests? https://github.com/dydxprotocol/v4-chain/pull/425/files

protocol/testing/e2e/gov/perpetuals_test.go
```diff
@@ -0,0 +1,305 @@
...
+				Authority: authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+				Params: perptypes.Params{
+					FundingRateClampFactorPpm: 123_456,
+					PremiumVoteClampFactorPpm: 123_456_789,
+					MinNumVotesPerSample:      15,
+				},
+			},
+			expectedProposalStatus: govtypesv1.ProposalStatus_PROPOSAL_STATUS_PASSED,
+		},
+		"Fail: invalid authority": {
```
teddyding, line 35: In general, could we add some other failure tests other than `invalid authority`?

protocol/testing/e2e/gov/perpetuals_test.go
tqin7: will add it in upcoming PR that adds more bridge test cases!

teddyding (approved):

PR #457, opened 2023-10-04
[CLOB-177] - long term e2e tests

jakob-dydx:
### Summary by CodeRabbit

- New Feature: Introduced long-term order placement and matching in the decentralized exchange. This includes handling of partial fills and multiple orders.
- Test: Enhanced test coverage with new scenarios for long-term order placement, expiration, and matching. This includes tests for off-chain and on-chain messages, subaccount updates, and order fill events.
- Refactor: Improved order matching logic and subaccount state updates to accommodate long-term orders.


protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -233,3 +246,942 @@ func TestImmediateExecutionLongTermOrders(t *testing.T) {
...
+		clobtypes.Order{
+			OrderId:      clobtypes.OrderId{SubaccountId: constants.Bob_Num0, ClientId: 1, ClobPairId: 0},
+			Side:         clobtypes.Order_SIDE_SELL,
+			Quantums:     10_000_000_000,
+			Subticks:     500_000_000,
+			GoodTilOneof: &clobtypes.Order_GoodTilBlock{GoodTilBlock: 20},
+		},
+	)
+	LongTermPlaceOrder_Alice_Num0_Id0_Clob0_Buy100_Price10_GTBT15_PO := *clobtypes.NewMsgPlaceOrder(
+		MustScaleOrder(
```
jonfung-dydx, line 342: Why is this scaled but the other ones aren't?

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -233,3 +246,942 @@ func TestImmediateExecutionLongTermOrders(t *testing.T) {
...
+										),
+									},
+								}))),
+							},
+						},
+					)},
+				},
+			},
+		},
+		"Test post-only order placed on the book": {
```
jonfung-dydx, line 668: I think for this test. we should probably have something resting on the book that _could_ be matched with this PO order in order to verify that this PO order doesn't match and instead rests on the book?

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -233,3 +246,942 @@ func TestImmediateExecutionLongTermOrders(t *testing.T) {
...
+
+		// Expectations to verify at end of test
+		orderShouldRestOnOrderbook bool
+		expectedOrderFillAmount    uint64
+		expectedSubaccounts        []satypes.Subaccount
+	}{
+		"Test placing an order": {
+			order:                      LongTermPlaceOrder_Alice_Num0_Id0_Clob0_Buy1_Price50000_GTBT5.Order,
+			orderShouldRestOnOrderbook: true,
+			expectedOrderFillAmount:    0,
```
jonfung-dydx, line 420: do we need `expectedSubaccounts` for this test?

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -233,3 +246,942 @@ func TestImmediateExecutionLongTermOrders(t *testing.T) {
...
+							Events: []*indexer_manager.IndexerTendermintEvent{},
+						},
+					)},
+				},
+			},
+		},
+		"Test matching an order fully as taker": {
+			order:                      LongTermPlaceOrder_Alice_Num0_Id0_Clob0_Buy1_Price50000_GTBT5.Order,
+			orderShouldRestOnOrderbook: false,
+			expectedOrderFillAmount:    0, // order is fully-filled and removed from state
```
jonfung-dydx, line 480: do we need `expectedSubaccounts` for this test?

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -233,3 +246,942 @@ func TestImmediateExecutionLongTermOrders(t *testing.T) {
...
+			},
+		},
+		"Test matching an order partially as taker then fully as maker": {
+			order:                      LongTermPlaceOrder_Alice_Num0_Id0_Clob0_Buy2_Price50000_GTBT5.Order,
+			orderShouldRestOnOrderbook: false,
+			// order is fully-filled and removed from state, resulting in zero fill amount in state
+			expectedOrderFillAmount: 0,
+			expectedSubaccounts: []satypes.Subaccount{
+				{
+					Id: &constants.Alice_Num0,
```
jonfung-dydx, line 730: should we assert  subaccounts for Bob  as well?

jonfung-dydx (approved):

protocol/x/clob/e2e/long_term_orders_test.go
jakob-dydx: Was just re-using a pre-existing order that existed in constants which needed to be scaled. For clarity I'll just create the order manually instead of calling the scale function

jakob-dydx: Just re-using an existing post-only order which was created using the unscaled order format (where quantums should be multiplied by step_base_quantums and subticks should be multiplied by subticks per tick)

PR #473, opened 2023-10-04
Add PR template

roy-dydx:
Adding the initial PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,13 @@
+### Changelist
+[Describe or list the changes made in this PR]
+
+### Test Plan
+[Describe how this PR was tested (if applicable)]
+
+### Author/Reviewer Checklist
+- Add an entry to `protocol/CHANGELOG.md` and/or `indexer/CHANGELOG.md` (if applicable).
+- If this PR has changes that modify the way state is read or written, require state migration, or change app hash, manually add the `state-breaking` label.
+- If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal`or `ProcessProposal`, manually add the label `proposal-breaking`.
+- If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
+- If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
+- Manually add any of the following labels: `refactor`, `chore`, `bug`.


.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+### Changelist
+[Describe or list the changes made in this PR]
+
+### Test Plan
+[Describe how this PR was tested (if applicable)]
+
+### Author/Reviewer Checklist
+- Add an entry to `protocol/CHANGELOG.md` and/or `indexer/CHANGELOG.md` (if applicable).
```
ttl33, line 8: nit: can we make these into checkboxes so authors can check them off?

separately:
1. I don't see `CHANGELOG.md` - can you add them as part of this PR?
2. do you have guidelines around what to enter in the changelog? it would be good to add some examples so people can follow.

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+- If this PR has changes that modify the way state is read or written, require state migration, or change app hash, manually add the `state-breaking` label.
```
ttl33, line 9: As a follow up item, I think it would be good to explain what qualifies as state-breaking with examples

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+- If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal`or `ProcessProposal`, manually add the label `proposal-breaking`.
```
ttl33, line 10: again, links to examples would be nice

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+- If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
+- If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
+- Manually add any of the following labels: `refactor`, `chore`, `bug`.
```
ttl33, line 13: Do we envision that it would be one of these (plus feature) labels?

I would think so and we should encourage ppl to break PRs into logically separate ones if PRs are combinations of these

.github/pull_request_template.md
roy-dydx: I put a better description. I'm not quite sure what to do for examples for now because there are many ways things can change the app state, and an example might not help people know what to look for. Let's think about this later.

.github/pull_request_template.md
roy-dydx: Same as above.

.github/pull_request_template.md
roy-dydx: Yeah ideally we have nice granularity of PRs, but generally github PRs tend to be bigger than if we used something like Phabricator because it's tedious (and not common practice) to stack PRs in github. So I definitely think it's forgivable if a PR fits multiple categories. Just add more than one tag.

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+### Changelist
+[Describe or list the changes made in this PR]
+
+### Test Plan
+[Describe how this PR was tested (if applicable)]
+
+### Author/Reviewer Checklist
+- [ ] If this change affects functionality (features, bug fixes, breaking changes, etc.), update `protocol/CHANGELOG.md` and/or `indexer/CHANGELOG.md` appropriately.
+- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
+- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
```
ttl33, line 10: Can a change be state-breaking and also a proposal-breaking change? I suppose we'd want to break those into two separate PRs?

indexer/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+## [Unreleased]
+
+### Features
+
+* [#473](https://github.com/dydxprotocol/v4-chain/pull/473/files) Changelog example entry.
```
ttl33, line 7: Can we add an actual example per section? For example, this one https://github.com/dydxprotocol/v4-chain/pull/481 for state breaking.

Also, this PR doesn't feel like a good example. Can you pick a different PR for the features section?

.github/pull_request_template.md
ttl33: I think picking recent examples are fine for now. Again, this won't be comprehensive but gives ppl ideas why certain changes are state-breaking.

indexer/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+
+* [#143](https://github.com/dydxprotocol/v4-chain/pull/143) Add websocket events for perpetual markets.
+
+### Improvements
+* [#469](https://github.com/dydxprotocol/v4-chain/pull/469) Added a reason field to `/screen` endpoint to display a reason for blocking an address.
+  
+### Bug Fixes
+* [#460](https://github.com/dydxprotocol/v4-chain/pull/460/files) Fixed track lag timing to output positive numbers.
+
+### API Breaking Changes
```
roy-dydx, line 15: Couldn't find a recent example of this in commit history.

protocol/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# Changelog
+
+## [Unreleased]
+
+### Features
+
+* [#190](https://github.com/dydxprotocol/v4-chain/pull/190) Added `MsgCreateClobPair` to x/clob to allow creation of new clob pairs.
```
ttl33, line 7: nit: should there be subsections for features? for example a group of `feature:gov-updates` PRs should prob be next to each other.

or is this something that we'd do when summarizing the changlogs?

ttl33 (approved):

protocol/CHANGELOG.md
roy-dydx: I guess there could be if there's a lot of entries that could be grouped together? We can continue without for now and see how many entries end up in a changelog and if there's a lot we can consider breaking it up.

PR #476, opened 2023-10-04
[CLOB-934] Update rate limiting for short term orders and cancellations to be per account and update the limit to be 200 per block.

lcwik:
### Summary by CodeRabbit

**Release Notes:**

- New Feature: Increased the maximum limits for short-term orders and order cancellations per block from 50 to 200, allowing more trading activity within a single block.
- Refactor: Updated terminology from "subaccount" to "account" in various places for clarity and consistency.
- Test: Added new test cases for rate limiting of orders and order cancellations, improving coverage and reliability.
- Chore: Introduced telemetry for rate-limited accounts, enhancing monitoring capabilities.


protocol/x/clob/rate_limit/order_rate_limiter.go
```diff
@@ -124,12 +127,20 @@ func (r *placeOrderRateLimiter) RateLimit(ctx sdk.Context, msg *types.MsgPlaceOr
 			1,
 			msg.Order.GetOrderLabels(),
 		)
-		r.rateLimitedSubaccounts[msg.Order.GetSubaccountId()] = true
+		r.rateLimitedAccounts[msg.Order.OrderId.SubaccountId.Owner] = true
+		r.rateLimitedSubaccounts[msg.Order.OrderId.SubaccountId] = true
```
BrendanChou, line 131: is this necessarily the desired behavior to do both? Shouldn't it be an if/else depending on if the ordertype is stateful or not?

lcwik: The purpose here was just to know at what number of accounts/subaccounts are hitting rate limits and thus it makes sense at a high level here to `AND` them.

Separately we could add different metrics to see which rate limit is being hit and at what level inside the rate limiter implementation itself.

protocol/lib/metrics/constants.go
```diff
@@ -129,6 +129,7 @@ const (
 	OrderFlag                                    = "order_flag"
 	OrderSide                                    = "order_side"
 	PlaceOrder                                   = "place_order"
+	PlaceOrderAccounts                           = "place_order_accounts"
```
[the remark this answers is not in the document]
lcwik: Both are being used.

BrendanChou (approved):

PR #477, opened 2023-10-05
[breaking] Shorten State Keys

BrendanChou:
Changelog
- Shorten Keys and Key Prefixes in various modules (use CamelCase and abbreviations)
  - Did not touch `x/clob` keys which still need to be fixed by @jayy04 
- Add generics protos (Uint32, Uint64, Int32, and Int64) to hold single values of respective types
  - This prevents us from marshaling directly between bytes and integer types which is prone to critical marshaling and unmarshaling bugs if the logic changes
- Use BigEndian to translate `uint32` to state keys which orders them lexicographically. Moved the relevant function from `bytes.go` and renamed it to `Uint32ToKey` so it's clear that it is used for keys
- Change `x/delay-msg` to use height as `uint32` which is how it is in the rest of the code and is needed to support lexographically-sorted keys

Other notes:
- Tests for success-cases of `SetDelayMessage` are still missing, to be added by @clemire 
- `indexer/` files are generated


protocol/x/bridge/keeper/grpc_query_test.go
```diff
@@ -254,7 +254,7 @@ func TestDelayedCompleteBridgeMessages(t *testing.T) {
 
 			// Construct expected responses.
 			delayMsgAuthority := authtypes.NewModuleAddress(delaymsgtypes.ModuleName).String()
-			blockOfExecution := int64(k.GetSafetyParams(ctx).DelayBlocks) + ctx.BlockHeight()
+			blockOfExecution := k.GetSafetyParams(ctx).DelayBlocks + uint32(ctx.BlockHeight())
```
jayy04, line 257: nit: lib.MustConvertIntegerToUint32? 

not a big deal though

BrendanChou: Let me fix this consistently in the codebase in a new PR

PR #484, opened 2023-10-05
[breaking] update key prefixes for clob

jayy04:
1. Use abbreviated key prefixes + CamelCase
2. remove unused keys
3. update prunable height to 0 for stateful order as its never used (should get omitted when marshalled)

### Summary by CodeRabbit

- Refactor: Removed the `GetOrdersFilledDuringLatestBlock` function from `protocol/x/clob/keeper/order_state.go`, indicating a change in how order IDs filled during the latest block are retrieved.
- Refactor: Updated the value of `pruneableBlockHeight` in `protocol/x/clob/keeper/process_single_match.go` to ensure stateful orders are never pruned by `BlockHeight`.
- Style: Improved clarity and consistency in `protocol/x/clob/types/keys.go` by modifying key prefixes used for reading/writing state and some constant names.


protocol/x/clob/keeper/process_single_match.go
```diff
@@ -469,8 +468,8 @@ func (k Keeper) setOrderFillAmountsAndPruning(
 	newTotalFillAmount satypes.BaseQuantums,
 	curPruneableBlockHeight uint32,
 ) *types.OffchainUpdates {
-	// Note that stateful orders are never pruned by `BlockHeight`, so we set the value to `MaxUInt32` here.
-	pruneableBlockHeight := uint32(math.MaxUint32)
+	// Note that stateful orders are never pruned by `BlockHeight`, so we set the value to `0` here.
+	pruneableBlockHeight := uint32(0)
```
jayy04, line 472: Setting to 0 here should save us 4 bytes / stateful order

BrendanChou: Are you sure this is a safe assumption to make? Will anything panic if there exists something in state which "should" have already been pruned?

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
-	LiquidationsConfigKey = "liquidations_config"
+	LiquidationsConfigKey = "LiqCfg"
 
 	// EquityTierLimitConfigKey is the key to retrieve the equity tier limit configuration.
-	EquityTierLimitConfigKey = "equity_tier_limit_config"
+	EquityTierLimitConfigKey = "EqTierLim"
 
 	// BlockRateLimitConfigKey is the key to retrieve the block rate limit configuration.
-	BlockRateLimitConfigKey = "block_rate_limit_config"
+	BlockRateLimitConfigKey = "RateLim"
```
BrendanChou, line 34: Nit: I'd add `Cfg` here for clarity. The non-prefix keys are less important since they only occur once.

Maybe if you keep state for the past `X` blocks then it's replicated on disk `X` times but doesn't seem too bad for these param/config keys

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
-	OrderAmountFilledKeyPrefix = "order_amount_filled/"
+	OrderAmountFilledKeyPrefix = "Filled/"
 
 	// BlockHeightToPotentiallyPrunableOrdersPrefix is the prefix to retrieve a list of potentially prunable orders
 	// by block height.
-	BlockHeightToPotentiallyPrunableOrdersPrefix = "block_height_to_potentially_prunable_orders/"
-
-	// OrdersFilledDuringLatestBlockKey is the key to retrieve the list of orders filled during the latest block.
-	OrdersFilledDuringLatestBlockKey = "orders_filled_during_latest_block"
+	BlockHeightToPotentiallyPrunableOrdersPrefix = "HgtToPrunable/"
```
BrendanChou, line 46: Nit: `HtToPrune`?

You also have `ExpTime` below, is `ExpHeight` or `ExpHt` acceptable?

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
 	// NextStatefulOrderBlockTransactionIndexKey is the transient store key that stores the next
 	// transaction index to use for the next newly-placed stateful order.
-	NextStatefulOrderBlockTransactionIndexKey = "next_stateful_order_block_transaction_index"
+	NextStatefulOrderBlockTransactionIndexKey = "NextTxIdx"
 
 	// UncommittedStatefulOrderPlacementKeyPrefix is the key to retrieve an uncommitted stateful order and information
 	// about when it was placed. uncommitted orders are orders that this validator is aware of that have yet to be
 	// committed to a block and are stored in a transient store.
-	UncommittedStatefulOrderPlacementKeyPrefix = StatefulOrderKeyPrefix + "uncommitted/long_term/"
+	UncommittedStatefulOrderPlacementKeyPrefix = StatefulOrderKeyPrefix + "U/LT/"
```
BrendanChou, line 90: Why is `StatefulOrderKeyPrefix` used here? It seems unrelated. Originally this was just a string and I broke it out into explicitly having this prefix since the prefix did match and I thought it was needed for iterating over the keys, but I don't think it is

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
 	// uncommitted orders are orders that this validator is aware of that have yet to be committed to a block and
 	// are stored in a transient store. This count represents the number of uncommitted stateful
 	// `placements - cancellations`.
-	UncommittedStatefulOrderCountPrefix = "stateful_order_count/uncommitted/long_term/"
+	UncommittedStatefulOrderCountPrefix = "Ct/U/LT/"
 
 	// StatefulOrderCountPrefix is the key to retrieve the stateful order count. The stateful order count
 	// represents the number of long term order placements and triggered conditional orders stored in state.
-	StatefulOrderCountPrefix = "stateful_order_count/long_term/"
+	StatefulOrderCountPrefix = "Ct/LT/"
```
BrendanChou, line 105: Nit: prefer not using `/` inside most keys

Nit: maybe consider `Num` or `Delta` instead of `Ct`

protocol/x/clob/types/keys.go
jayy04: not needed for iteration. 

I think having `SOP` here is fine since the mapped value is of type `LongTermOrderPlacement` which is the same as other keys that are prefixed by `SOP`.

No strong opinion though. let me know if you have a strong preference to make these different

jayy04: actually let me just update and remove SOP for simplicity.

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
-	SubaccountLiquidationInfoKeyPrefix = "subaccount_liquidation_info/"
+	SubaccountLiquidationInfoKeyPrefix = "SaLiqInfo/"
 
 	// LiquidationsConfigKey is the key to retrieve the liquidations config.
-	LiquidationsConfigKey = "liquidations_config"
+	LiquidationsConfigKey = "LiqCfg"
 
 	// EquityTierLimitConfigKey is the key to retrieve the equity tier limit configuration.
-	EquityTierLimitConfigKey = "equity_tier_limit_config"
+	EquityTierLimitConfigKey = "EqTierLimCfg"
```
lucas-dydx, line 31: Suggestion: this is probably enough honestly
```suggestion
	EquityTierLimitConfigKey = "EqTierCfg"
```

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
-	OrderAmountFilledKeyPrefix = "order_amount_filled/"
+	OrderAmountFilledKeyPrefix = "Filled/"
 
 	// BlockHeightToPotentiallyPrunableOrdersPrefix is the prefix to retrieve a list of potentially prunable orders
 	// by block height.
-	BlockHeightToPotentiallyPrunableOrdersPrefix = "block_height_to_potentially_prunable_orders/"
-
-	// OrdersFilledDuringLatestBlockKey is the key to retrieve the list of orders filled during the latest block.
-	OrdersFilledDuringLatestBlockKey = "orders_filled_during_latest_block"
+	BlockHeightToPotentiallyPrunableOrdersPrefix = "ExpHt/"
```
lucas-dydx, line 46: Hmm any other ideas for this key name? Agree it's good to decrease length, but not sure `ExpHt` is descriptive

Maybe `PruneHt` instead?

lucas-dydx (approved):

protocol/x/clob/types/keys.go
```diff
@@ -18,94 +18,91 @@ const (
...
 	// UntriggeredConditionalOrderKeyPrefix is the key to retrieve an untriggered conditional order and
 	// information about when it was placed.
-	UntriggeredConditionalOrderKeyPrefix = StatefulOrderKeyPrefix + "untriggered/conditional/"
+	UntriggeredConditionalOrderKeyPrefix = StatefulOrderKeyPrefix + "Untrig/"
 
 	// StatefulOrdersTimeSlicePrefix is the key to retrieve a unique list of the stateful orders that
 	// expire at a given timestamp, sorted by order ID.
-	StatefulOrdersTimeSlicePrefix = "stateful_orders_time_slice/"
+	StatefulOrdersTimeSlicePrefix = "ExpTime/"
 )
```
BrendanChou, line 79: these are probably the most important to shorten in state since there will be a lot.

Suggest:

`SO/`
`P/` (unchanged)
`T/` (triggered)
`L/` (longterm)
`U/` (untriggered)
`ExpTm/`

BrendanChou (approved):

protocol/x/clob/types/keys.go
jayy04: we don't need this to be descriptive right? this is only used for debugging. 

I think this is fine since we have `ExpTm`. let me know if you have a strong preference

PR #491, opened 2023-10-05
[CLOB-184] - e2e short term order replacement tests

jakob-dydx:
### Changelist
- adds e2e tests for short term order replacements

### Summary by CodeRabbit

- New Feature: Added functionality for placing three new types of orders with varying parameters in the trading system.
- Test: Introduced `TestShortTermOrderReplacements` to validate the behavior of order replacements in different scenarios, including same block replacements, next block replacements, and handling of partial fills.
- Test: Enhanced testing for order replacements to include different order types such as FOK and IOC.
- Chore: Removed a comment explaining the logic for checking replacement orders in the `validateShortTermOrderPlacementOperation` function.


protocol/x/clob/e2e/short_term_orders_test.go
```diff
@@ -607,3 +607,370 @@ func TestPlaceOrder(t *testing.T) {
...
+								Side:         clobtypes.Order_SIDE_SELL,
+								Quantums:     3,
+								Subticks:     10,
+								GoodTilOneof: &clobtypes.Order_GoodTilBlock{GoodTilBlock: 20},
+							},
+							testapp.DefaultGenesis(),
+						)),
+						PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB21,
+					},
+					orderIdsExpectations: map[clobtypes.OrderId]orderIdExpectations{
```
jonfung-dydx, line 738: Can we add some expectations that `PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20` does not exist anywhere?

Also the 2nd message that was placed should have a fill amount too?

protocol/x/clob/e2e/short_term_orders_test.go
```diff
@@ -607,3 +607,370 @@ func TestPlaceOrder(t *testing.T) {
...
+						PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB21.Order.OrderId: {
+							shouldExistOnMemclob: true,
+							expectedOrder:        PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB21.Order,
+							expectedFillAmount:   PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20.Order.Quantums / 2,
+						},
+					},
+				},
+			},
+		},
+		"Success: Replacement order increases size in next block after partial fill": {
```
jonfung-dydx, line 748: Can we test replacement order swaps sides in next block after partial fill?

protocol/x/clob/e2e/short_term_orders_test.go
```diff
@@ -607,3 +607,370 @@ func TestPlaceOrder(t *testing.T) {
...
+						PlaceOrder_Alice_Num0_Id0_Clob0_Buy7_Price10_GTB21.Order.OrderId: {
+							shouldExistOnMemclob: true,
+							expectedOrder:        PlaceOrder_Alice_Num0_Id0_Clob0_Buy7_Price10_GTB21.Order,
+							expectedFillAmount:   PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20.Order.Quantums / 2,
+						},
+					},
+				},
+			},
+		},
+		"Success: Replacement order decreases size in next block after partial fill": {
```
jonfung-dydx, line 786: Should we test the case where replacement order decreases size to below the matched partial fill amount?

jonfung-dydx (approved):

protocol/x/clob/e2e/short_term_orders_test.go
jakob-dydx: As for 1) By asserting that `PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB21.Order` exists on the memclob, we are simultaneously asserting that `PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB20.Order` does not exist. These two orders share an OrderId and all data stored on both the memclob and in state is keyed off of OrderId. 

2). Yes the second order should have a fill amount but honestly I feel like it is fine to not assert that here. It makes the test simpler to read by only asserting information relevant to the order being replaced. We have other tests that validate matching

protocol/x/clob/e2e/short_term_orders_test.go
```diff
@@ -606,3 +606,418 @@ func TestPlaceOrder(t *testing.T) {
...
 	}
 }
+
+func TestShortTermOrderReplacements(t *testing.T) {
+	order := PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20
+	fok_replacement := order
+	fok_replacement.Order.GoodTilOneof = &clobtypes.Order_GoodTilBlock{GoodTilBlock: 21}
+	fok_replacement.Order.TimeInForce = clobtypes.Order_TIME_IN_FORCE_FILL_OR_KILL
+	ioc_replacement := fok_replacement
+	ioc_replacement.Order.TimeInForce = clobtypes.Order_TIME_IN_FORCE_IOC
```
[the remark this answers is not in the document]
jakob-dydx: I don't think this is correct. fok_replacement := order is making a copy of the order because order is not a pointer, it is a value

protocol/x/clob/e2e/short_term_orders_test.go
```diff
@@ -606,3 +606,418 @@ func TestPlaceOrder(t *testing.T) {
...
+						PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20,
+						fok_replacement,
+					},
+					orderIdsExpectations: map[clobtypes.OrderId]orderIdExpectations{
+						PlaceOrder_Alice_Num0_Id0_Clob0_Buy6_Price10_GTB20.Order.OrderId: {
+							shouldExistOnMemclob: false,
+						},
+					},
+				},
+			},
```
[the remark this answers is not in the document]
jakob-dydx: good catch

PR #497, opened 2023-10-05
[CLOB-924] Metrics Daemon - Log out app version periodically

jonfung-dydx:
Add a metrics daemon to log out app version periodically

### Summary by CodeRabbit

- New Feature: Introduced a Metrics Daemon for enhanced observability. This daemon periodically reports application metrics, including the app version and git commit, improving monitoring capabilities.
- Refactor: Reorganized import statements across several files for better code readability and maintainability.
- Chore: Updated the application logger to include the app version and git commit in the log message, providing more context for debugging and analysis.
- New Feature: Added a gRPC service `MetricsService` in `metrics.proto` for future RPC methods related to metrics.
- Refactor: Removed the reporting of app version and git commit from the `BeginBlocker` function, as this is now handled by the new Metrics Daemon.


protocol/app/app.go
```diff
@@ -1242,6 +1243,15 @@ func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.R
 		)
 	}
 
+	// Log out app version every hour.
+	if ctx.BlockTime().Unix()%3600 == 0 {
```
jiajames, line 1247: If we're stuck on a block, will this `BlockTime` still update? 

Also, sorry I mispoke but we should provide telemetry on this rather than a log (or maybe both? but put them in the same spot)

jiajames: oops didn't realize they were already in the same spot

protocol/app/app.go
```diff
@@ -1167,6 +1168,24 @@ func New(
 	app.ScopedIBCKeeper = scopedIBCKeeper
 	app.ScopedTransferKeeper = scopedTransferKeeper
 
+	// Report out app version and git commit. This will be run when validators restart.
+	version := version.NewInfo()
```
lucas-dydx, line 1172: What do you think about putting this logic into a separate daemon thread that loops every 30 minutes or so?

For example, we could do the following:
- Create a goroutine to log out app version.
- Recover from all panics in goroutine.
- Log out app version immediately upon goroutine start + every 30 minutes afterwards.

lucas-dydx (approved):

protocol/daemons/metrics/client/client.go
```diff
@@ -0,0 +1,72 @@
...
+	ticker := time.NewTicker(time.Duration(METRICS_DAEMON_LOOP_DELAY_MS) * time.Millisecond)
+	// TODO prometheus flag enable
+	defer ticker.Stop()
+	for ; true; <-ticker.C {
+		RunMetricsDaemonTaskLoop(
+			ctx,
+			logger,
+		)
+	}
+}
```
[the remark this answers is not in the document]
jonfung-dydx: Hello rabbit, the Start function is only called in a goroutine.

protocol/app/app.go
```diff
@@ -645,6 +646,20 @@ func New(
...
+			app.Server.ExpectMetricsDaemon(
+				daemonservertypes.MaximumAcceptableUpdateDelay(metricsclient.METRICS_DAEMON_LOOP_DELAY_MS),
+			)
+			metricsclient.Start(
+				// The client will use `context.Background` so that it can have a different context from
+				// the main application.
+				context.Background(),
+				logger,
+			)
+		}()
```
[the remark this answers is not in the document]
jonfung-dydx: I do not see a `if app.Config.Daemons.StartDaemons` block.

protocol/daemons/metrics/client/client.go
```diff
@@ -0,0 +1,72 @@
...
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Error(
+				"panic when reporting metrics from metrics daemon",
+				"panic",
+				r,
+				"stack",
+				string(debug.Stack()),
+			)
+		}
```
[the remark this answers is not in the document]
jonfung-dydx: I do not want to panic in a function used purely for observability.

protocol/app/app.go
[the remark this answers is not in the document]
jonfung-dydx: `startDaemons` is a function not a boolean flag

protocol/daemons/metrics/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+		)
+	}
+}
+
+// RunMetricsDaemonTaskLoop contains the logic to emit metrics about the application running.
+func RunMetricsDaemonTaskLoop(
+	ctx context.Context,
+	logger log.Logger,
+) {
+	defer func() {
```
lucas-dydx, line 42: Why not move this defer statement up to the where the go routine is started?

protocol/app/app.go
```diff
@@ -1167,6 +1185,24 @@ func New(
 	app.ScopedIBCKeeper = scopedIBCKeeper
 	app.ScopedTransferKeeper = scopedTransferKeeper
 
+	// Report out app version and git commit. This will be run when validators restart.
+	version := version.NewInfo()
```
lucas-dydx, line 1189: Won't this be a no-op since the prometheus server isn't enabled? Why leave it?

PR #500, opened 2023-10-05
update MsgCancelOrder DeliverTx failure logging

jakob-dydx:
- catches case where `MsgCancelOrder` fails DeliverTx because of a previous order removal. Logs this as info instead of as an error

### Summary by CodeRabbit

**Release Notes:**

- Bug Fix: Improved error handling in the `CancelOrder` function to handle cases where an order was already removed from state. This change prevents unexpected behavior and enhances system stability.
- Test: Added new test cases for the `CancelOrder` function, ensuring that it behaves as expected when an order is not found or has been previously removed.
- Chore: Introduced new constants and error messages to improve code readability and maintainability. These changes will make future debugging and development more efficient.


protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -22,6 +26,20 @@ func (k msgServer) CancelOrder(
...
 	defer func() {
 		if err != nil {
+			// Gracefully handle the case where the order was already removed from state.
+			if errors.Is(err, types.ErrStatefulOrderDoesNotExist) {
+				processProposerMatchesEvents := k.Keeper.GetProcessProposerMatchesEvents(ctx)
+				removedOrderIds := lib.SliceToSet(processProposerMatchesEvents.RemovedStatefulOrderIds)
+				if _, found := removedOrderIds[msg.GetOrderId()]; found {
+					k.Keeper.Logger(ctx).Info(
+						errorsmod.Wrap(
+							err,
```
lucas-dydx, line 36: Can we return a more informative error up from the handler (ideally one with an error code defined in https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/types/errors.go)? Specifically telling the user the order existed but was fully-filled / removed in the last block.

Note that any potential deployments of this software would likely use this error in a block explorer. Ideally users can view the block explorer and the error is informative enough that they understand why their order was not actually canceled and potentially was partially filled.

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -22,6 +26,20 @@ func (k msgServer) CancelOrder(
...
+							"MsgCancelOrder failed DeliverTx because the order was already removed from state",
```
lucas-dydx, line 37: Can you include the order ID in the log?

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -22,6 +26,20 @@ func (k msgServer) CancelOrder(
 
 	defer func() {
 		if err != nil {
+			// Gracefully handle the case where the order was already removed from state.
```
lucas-dydx, line 29: Suggestion: update this comment to explain why this happens

Also add a TODO here with a ticket to fix this for most cases (can use this ticket: https://linear.app/dydx/issue/CLOB-778/write-order-removals-to-state-store-uncommitted-transient-store). Ideally we also create a low priority M6 ticket to fix this for fully-filled orders (likely requires `PrepareProposal` changes).

Additionally, can we stat how often this happens?

protocol/x/clob/keeper/msg_server_cancel_orders_test.go
```diff
@@ -55,6 +57,46 @@ func TestCancelOrder_PanicIfValidationSucceedsButOrderNotFound(t *testing.T) {
...
+		ctx,
+		types.ProcessProposerMatchesEvents{BlockHeight: 2, RemovedStatefulOrderIds: []types.OrderId{orderToCancel.OrderId}},
+	)
+
+	_, err := msgServer.CancelOrder(ctx, &orderToCancel)
+	require.ErrorIs(t, err, types.ErrStatefulOrderDoesNotExist)
+	mockLogger.AssertExpectations(t)
+}
+
 func TestCancelOrder_Error(t *testing.T) {
```
lucas-dydx, line 100: Suggestion: add a test case where `ExpectedError` is something different than `types.ErrStatefulOrderDoesNotExist`.

lucas-dydx (approved):

jayy04 (approved):

protocol/x/clob/keeper/msg_server_cancel_orders.go
jakob-dydx: the order id is included in the original error returned from PerformOrderCancellationStatefulValidation

PR #518, opened 2023-10-06
Quick fix in `prod_genesis.sh`

teddyding:


protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -19,15 +19,13 @@ fi
 # Capture the required argument
 DYDX_BINARY="$1"
 
+source "./testing/genesis.sh"
+
 TMP_CHAIN_DIR="/tmp/prod-chain"
 TMP_EXCHANGE_CONFIG_JSON_DIR="/tmp/prod-exchange_config"
-NINE_ZEROS="000000000"
```
teddyding, line 24: This was previously not declared in but used in `genesis.sh`

PR #519, opened 2023-10-06
Quick fix in `prod_genesis.sh` 

teddyding:


protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -19,15 +19,15 @@ fi
 # Capture the required argument
 DYDX_BINARY="$1"
 
+source "./testing/genesis.sh"
```
teddyding, line 22: Make sure not to override any envvars by sourcing `"./testing/genesis.sh"`

PR #526, opened 2023-10-06
[CLOB-943] validation and observability for stateful cancellations / placements

jakob-dydx:
### Summary by CodeRabbit

- Refactor: Enhanced error logging by renaming `LogErrorWithBlockHeight` to `LogDeliverTxError` and adding additional parameters for more context.
- New Feature: Added telemetry metrics tracking in the `CancelStatefulOrder` and `PlaceStatefulOrder` functions.
- Refactor: Improved block height validation logic in the `PerformOrderCancellationStatefulValidation` function based on transaction mode.
- Test: Updated test cases to reflect changes in error handling and logging.
- Chore: Removed unnecessary import statements across multiple files.

### DIFF
--- a/protocol/lib/error/logging.go
+++ b/protocol/lib/error/logging.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 
 	"github.com/cometbft/cometbft/libs/log"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	errorsmod "cosmossdk.io/errors"
 )
@@ -40,14 +43,16 @@ func WrapErrorWithSourceModuleContext(err error, module string) error {
 		WithLogKeyValue(SourceModuleKey, fmt.Sprintf("x/%v", module))
 }
 
-// LogErrorWithBlockHeight logs an error, appending the block height and ABCI callback to the error message.
-func LogErrorWithBlockHeight(logger log.Logger, err error, blockHeight int64, callback string) {
+// LogDeliverTxError logs an error, appending the block height and ABCI callback to the error message.
+func LogDeliverTxError(logger log.Logger, err error, blockHeight int64, handler string, msg sdk.Msg) {
 	if err != nil {
 		err = errorsmod.Wrapf(
 			err,
-			"Block height: %d, Callback: %s",
+			"Block height: %d, Handler: %s, Callback: %s, Msg: %+v",
 			blockHeight,
-			callback,
+			handler,
+			metrics.DeliverTx,
+			msg,
 		)
 		logger.Error(err.Error())
 	} else {
--- a/protocol/lib/error/logging.go
+++ b/protocol/lib/error/logging.go
@@ -8,8 +8,6 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	errorsmod "cosmossdk.io/errors"
 )
 
 const (
@@ -46,15 +44,13 @@ func WrapErrorWithSourceModuleContext(err error, module string) error {
 // LogDeliverTxError logs an error, appending the block height and ABCI callback to the error message.
 func LogDeliverTxError(logger log.Logger, err error, blockHeight int64, handler string, msg sdk.Msg) {
 	if err != nil {
-		err = errorsmod.Wrapf(
-			err,
-			"Block height: %d, Handler: %s, Callback: %s, Msg: %+v",
-			blockHeight,
-			handler,
-			metrics.DeliverTx,
-			msg,
+		logger.Error(
+			err.Error(),
+			metrics.BlockHeight, blockHeight,
+			metrics.Handler, handler,
+			metrics.Callback, metrics.DeliverTx,
+			metrics.Msg, msg,
 		)
-		logger.Error(err.Error())
 	} else {
 		logger.Error("LogErrorWithBlockHeight called with nil error")
 	}
--- a/protocol/x/clob/keeper/msg_server_cancel_orders.go
+++ b/protocol/x/clob/keeper/msg_server_cancel_orders.go
@@ -44,14 +44,14 @@ func (k msgServer) CancelOrder(
 						1,
 						msg.OrderId.GetOrderIdLabels(),
 					)
-					k.Keeper.Logger(ctx).Info(
-						errorsmod.Wrapf(
-							types.ErrStatefulOrderCancellationFailedForAlreadyRemovedOrder,
-							"Error: %s",
-							err.Error(),
-						).Error(),
+					err = errorsmod.Wrapf(
+						types.ErrStatefulOrderCancellationFailedForAlreadyRemovedOrder,
+						"Error: %s",
+						err.Error(),
 					)
-					return
+					k.Keeper.Logger(ctx).Info(
+						err.Error(),
+					)
 				}
 			}
 			errorlib.LogDeliverTxError(k.Keeper.Logger(ctx), err, ctx.BlockHeight(), "CancelOrder", msg)
--- a/protocol/x/clob/keeper/msg_server_cancel_orders.go
+++ b/protocol/x/clob/keeper/msg_server_cancel_orders.go
@@ -54,7 +54,7 @@ func (k msgServer) CancelOrder(
 					return
 				}
 			}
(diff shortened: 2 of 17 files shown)


protocol/lib/error/logging.go
```diff
@@ -40,14 +43,16 @@ func WrapErrorWithSourceModuleContext(err error, module string) error {
...
 			err,
-			"Block height: %d, Callback: %s",
+			"Block height: %d, Handler: %s, Callback: %s, Msg: %+v",
 			blockHeight,
-			callback,
+			handler,
+			metrics.DeliverTx,
+			msg,
 		)
 		logger.Error(err.Error())
```
jayy04, line 57: nit: let's not format this  - instead we should do 

logger.Error(
    "error string", 
    "key", value,
    "key1", value1,
    ....
)

this makes it easier to filter/search on datadog

protocol/x/clob/keeper/orders.go
```diff
@@ -130,18 +131,36 @@ func (k Keeper) PlaceShortTermOrder(
...
+) (err error) {
+	defer func() {
+		if err != nil {
+			callback := metrics.DeliverTx
+			if ctx.IsCheckTx() {
+				callback = metrics.CheckTx
+			}
+			if ctx.IsReCheckTx() {
+				callback = metrics.ReCheckTx
+			}
```
jayy04, line 143: nit: I assume we are going to do this a lot in the future . let's move this to a helper function

protocol/x/clob/keeper/orders.go
```diff
@@ -516,9 +554,30 @@ func (k Keeper) PerformOrderCancellationStatefulValidation(
 ) error {
 	orderIdToCancel := msgCancelOrder.GetOrderId()
 	if orderIdToCancel.IsStatefulOrder() {
-		cancelGoodTilBlockTime := msgCancelOrder.GetGoodTilBlockTime()
 		previousBlockTime := k.blockTimeKeeper.GetPreviousBlockInfo(ctx).Timestamp
 
+		prevBlockHeight := k.blockTimeKeeper.GetPreviousBlockInfo(ctx).Height
+		if lib.IsDeliverTxMode(ctx) && prevBlockHeight != uint32(ctx.BlockHeight()-1) {
```
jayy04, line 560: nit: I think we should also build a `GetBlockHeight(ctx)` function which returns height based on deliver tx/ check tx mode

jayy04 (approved):

protocol/x/clob/keeper/orders.go
lucas-dydx: Yeah maybe a `GetCallback` function

Also should note that we assume if it's not `CheckTx` and `RecheckTx` that it's `DeliverTx`, which isn't always true

protocol/x/clob/keeper/orders.go
```diff
@@ -516,9 +554,31 @@ func (k Keeper) PerformOrderCancellationStatefulValidation(
 ) error {
 	orderIdToCancel := msgCancelOrder.GetOrderId()
 	if orderIdToCancel.IsStatefulOrder() {
-		cancelGoodTilBlockTime := msgCancelOrder.GetGoodTilBlockTime()
 		previousBlockTime := k.blockTimeKeeper.GetPreviousBlockInfo(ctx).Timestamp
 
+		prevBlockHeight := k.blockTimeKeeper.GetPreviousBlockInfo(ctx).Height
```
lucas-dydx, line 559: Nit: save `k.blockTimeKeeper.GetPreviousBlockInfo(ctx)` in a var

protocol/x/clob/keeper/orders.go
```diff
@@ -516,9 +554,31 @@ func (k Keeper) PerformOrderCancellationStatefulValidation(
...
+		currBlockHeight := uint32(ctx.BlockHeight())
+		if lib.IsDeliverTxMode(ctx) && prevBlockHeight != currBlockHeight-1 {
+			k.Logger(ctx).Error(
+				fmt.Sprintf(
+					"PerformOrderCancellationStatefulValidation: prev block height is not one below current block height"+
+						"in DeliverTx. Prev: %d, Current: %d", prevBlockHeight, ctx.BlockHeight(),
```
lucas-dydx, line 565: Nit: use `currBlockHeight` if possible
```suggestion
						"in DeliverTx. Prev: %d, Current: %d", prevBlockHeight, currBlockHeight,
```

protocol/x/clob/keeper/orders.go
```diff
@@ -516,9 +554,31 @@ func (k Keeper) PerformOrderCancellationStatefulValidation(
...
+				),
+			)
+		}
+
+		// CheckTx or ReCheckTx
+		if !lib.IsDeliverTxMode(ctx) && currBlockHeight > 1 && prevBlockHeight != currBlockHeight {
+			k.Logger(ctx).Error(
+				fmt.Sprintf(
+					"PerformOrderCancellationStatefulValidation: prev block height is not equal to current block height"+
+						"in CheckTx/ReCheckTx. Prev: %d, Current: %d", prevBlockHeight, ctx.BlockHeight(),
```
lucas-dydx, line 575: Can you add the actual callback (`CheckTx` or `RecheckTx`) to the log?

protocol/x/clob/keeper/orders.go
lucas-dydx: Should you add the `msgCancelOrder` in the error log?

protocol/x/clob/keeper/orders.go
```diff
@@ -130,18 +131,36 @@ func (k Keeper) PlaceShortTermOrder(
...
+			if ctx.IsReCheckTx() {
+				callback = metrics.ReCheckTx
+			}
+			telemetry.IncrCounterWithLabels(
+				[]string{types.ModuleName, metrics.CancelStatefulOrder, metrics.Error, metrics.Count},
+				1,
+				[]gometrics.Label{
+					metrics.GetLabelForStringValue(metrics.Callback, callback),
+				},
+			)
```
lucas-dydx, line 150: Where do you emit a success metric? Ideally we have enough metrics to get a % success rate for stateful order placements + cancellations

protocol/x/clob/keeper/msg_server_proposed_operations.go
```diff
@@ -17,7 +16,7 @@ func (k msgServer) ProposedOperations(
 
 	defer func() {
 		if err != nil {
-			errorlib.LogErrorWithBlockHeight(k.Keeper.Logger(ctx), err, ctx.BlockHeight(), metrics.DeliverTx)
```
lucas-dydx, line 20: Possible we could also add a basic helper function to add success / failure metrics for each message handler?

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -54,7 +54,7 @@ func (k msgServer) CancelOrder(
 					return
```
lucas-dydx, line 54: Can you reassign the returned error in this case? Some software deployments might have associated block explorers for viewing TXs, and in this case the user's cancellation technically wasn't received in time rather than order not found (order was already fully-filled / undercollateralized).

So something like this:

err = types.ErrStatefulOrderCancellationFailedForAlreadyRemovedOrder

lucas-dydx (approved):

protocol/x/clob/keeper/orders.go
jakob-dydx: will add this in

protocol/x/clob/keeper/msg_server_proposed_operations.go
jakob-dydx: yeah that's a good idea. I think it would be a nice pattern to use `defer` and emit failure if `err != nil` and success otherwise. Would be pretty easy, won't do it in this PR though don't want to overload

jakob-dydx: https://github.com/dydxprotocol/v4-chain/pull/529/files

protocol/x/clob/keeper/orders.go
```diff
@@ -179,7 +199,20 @@ func (k Keeper) CancelStatefulOrder(
...
+		}
+	}()
+
 	// 1. Ensure the order is not a Short-Term order.
 	order := msg.Order
 	order.OrderId.MustBeStatefulOrder()
```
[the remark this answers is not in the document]
jakob-dydx: will follow up in separate PR, after https://github.com/dydxprotocol/v4-chain/pull/529/files is merged

PR #530, opened 2023-10-07
[DEC-2075] [fix-halt] Avoid overflow in vest and rewards stats

teddyding:
After we switch to 18 decimals for native coin, balance-realted stats are prone to overflow since it was previously casted to `int64`

Also audited remaining call to `bigIntAmount.Int64()`

Also made vester balance more realistic in test environments in https://github.com/dydxprotocol/v4-chain/pull/539

### Summary by CodeRabbit

- New Feature: Added a new function `Int64MulPow10` in `big_math.go` for calculating the result of `val * 10^exponent` using `*big.Int`.
- Refactor: Updated telemetry metrics in `ProcessRewardsForBlock` function in `keeper.go` of `rewards` and `vest` modules to use `metrics.GetMetricValueFromBigInt` for accurate metric value conversion.
- Test: Added new test function `TestInt64MulPow10` in `big_math_test.go` to validate the `Int64MulPow10` function. Updated test cases in `keeper_test.go` of `rewards` and `vest` modules with more realistic values.
- Chore: Cleaned up import statements in `util_test.go` and `keeper.go` of `rewards` and `vest` modules.


protocol/lib/big_math.go
```diff
@@ -264,3 +264,16 @@ func warmCache() map[uint64]*big.Int {
 
 	return bigExponentValues
 }
+
+// Int64MulPow10 returns the result of `val * 10^exponent`, in *big.Int.
+func Int64MulPow10(
```
BrendanChou, line 269: if this is only used in testing, I would prefer not introducing more very-specific functions into the `lib` folder

this generally creates more surface area for bugs which is a bad thing

if only for testing, is it okay to put it in `testutils`?

teddyding: Done

BrendanChou (approved):

PR #532, opened 2023-10-07
Use new binance endpoint

BrendanChou:
### Changelist
Use `data-api.binance.vision` instead of `data.binance.com`
https://binance-docs.github.io/apidocs/spot/en/#change-log:~:text=API%20Market%20data%20from%20data.binance.com%20can%20now%20be%20accessed%20from%20data%2Dapi.binance.vision.

Even though this technically can affect `PrepareProposal` and `ProcessProposal`, I would consider it non-breaking since prices from exchanges are already medianized and very similar to each other

### Test Plan
Tested locally and in internal `funder` service

### Author/Reviewer Checklist
- [x] If this change affects functionality (features, bug fixes, breaking changes, etc.), update `protocol/CHANGELOG.md` and/or `indexer/CHANGELOG.md` appropriately.

### DIFF
--- a/protocol/CHANGELOG.md
+++ b/protocol/CHANGELOG.md
@@ -9,7 +9,9 @@
 ### Improvements
 
 * [#479](https://github.com/dydxprotocol/v4-chain/pull/479) Ensure that rate limiting of short term order placements/cancellations is guarded against replay attacks.
-  
+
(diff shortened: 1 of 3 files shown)


ttl33 (approved):
1. is this endpoint more reliable and how do we confirm that?
2. does this need to part of the v1.0?

PR #544, opened 2023-10-09
[DEC-2143] use cast to parse command line flags

jayy04:
### Changelist
Use cast to parse command line flags.

### Test Plan

### Author/Reviewer Checklist

### Summary by CodeRabbit

- Refactor: Enhanced type safety in the `flags` package across multiple modules (`protocol/app`, `protocol/daemons`, `protocol/indexer`, `protocol/x/clob`). The `GetFlagValuesFromOptions` and similar functions now use the `cast` package for type conversion, ensuring that option values are correctly converted to their expected types. This change improves error handling and reduces the risk of runtime type assertion errors.

### DIFF
--- a/protocol/app/flags/flags.go
+++ b/protocol/app/flags/flags.go
@@ -2,8 +2,10 @@ package flags
 
 import (
 	"fmt"
+
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	"github.com/spf13/cast"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


protocol/app/flags/flags.go
```diff
@@ -85,24 +87,34 @@ func GetFlagValuesFromOptions(
 	}
 
 	// Populate the flags if they exist.
-	if v, ok := appOpts.Get(NonValidatingFullNodeFlag).(bool); ok {
-		result.NonValidatingFullNode = v
+	if option := appOpts.Get(NonValidatingFullNodeFlag); option != nil {
+		if v, err := cast.ToBoolE(option); err == nil {
+			result.NonValidatingFullNode = v
+		}
```
jayy04, line 93: we can maybe just do `result.NonValidatingFullNode = cast.ToBool(option)` for simplicity (i.e. ignore error) but let's worry about that later

jakob-dydx (approved):

protocol/app/flags/flags.go
ttl33: I wonder if it would be more useful to throw an error or panic if the flag value can't be converted.

If I was an operator and I accidentally typed `tru` insted of `true`, I would want to software to fail and tell me my input was incorrect rather than using the default `false` value.

wdyt?

jayy04: yup Brendan brought this up too - we decided to punt this to M6 (since our current behavior can't handle this either)

ticket is [here](https://linear.app/dydx/issue/DEC-2146)

ttl33 (approved):

vincentwschau (approved):

PR #553, opened 2023-10-10
[DEC-2081] Audit app/prepare/full_node_prepare_proposal.go

lcwik:
### Changelist
* Swap tests to be more faithful with how a full node starts.
* Add the ability to capture logs for test apps.
* Validate full nodes metric and log is produced.

### Test Plan
These are test only changes.

### Author/Reviewer Checklist

### Summary by CodeRabbit

- Test: Enhanced `TestFullNodePrepareProposalHandler` in `full_node_prepare_proposal_test.go` with metrics and logging integration, and improved test setup with a test application builder.
- Refactor: Introduced a new constant `LoggerInstanceForTest` in `app.go` for retrieving the logger instance from the `appOptions` map, allowing for more flexible logger customization in tests.
- Test: Updated `TestIncrCountMetricWithLabels` and `TestModuleMeasureSinceWithLabels` in `util_test.go` to ensure `gometrics.Shutdown` is called at the end of each test case, improving test reliability.

### DIFF
--- a/protocol/app/prepare/full_node_prepare_proposal_test.go
+++ b/protocol/app/prepare/full_node_prepare_proposal_test.go
@@ -1,220 +1,62 @@
 package prepare_test
 
 import (
+	"bytes"
 	"testing"
+	"time"
 
+	gometrics "github.com/armon/go-metrics"
 	abci "github.com/cometbft/cometbft/abci/types"
+	"github.com/cometbft/cometbft/libs/log"
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
-	"github.com/dydxprotocol/v4-chain/protocol/app/prepare"
-	"github.com/dydxprotocol/v4-chain/protocol/mocks"
-	keepertest "github.com/dydxprotocol/v4-chain/protocol/testutil/keeper"
-	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
-	perpetualtypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
-	pricestypes "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
-	"github.com/stretchr/testify/mock"
+	"github.com/dydxprotocol/v4-chain/protocol/app/flags"
+	testApp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
 	"github.com/stretchr/testify/require"
 )
 
 // TestFullNodePrepareProposalHandler test that the full-node PrepareProposal handler always returns
 // an empty result.
 func TestFullNodePrepareProposalHandler(t *testing.T) {
-	tests := map[string]struct {
-		txs      [][]byte
-		maxBytes int64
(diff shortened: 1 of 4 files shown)


protocol/app/prepare/full_node_prepare_proposal_test.go
```diff
@@ -1,220 +1,67 @@
...
+			require.Empty(t, proposal.Txs)
+			return true
+		},
+	})
+
+	for _, metrics := range sink.Data() {
+		metrics.RLock()
+		defer metrics.RUnlock()
+
+		if metric, ok := metrics.Counters["testService.prepare_proposal.handler.error.count;detail=prepare_proposal_txs"]; ok {
```
ttl33, line 48: noob question: when do we expect `ok` to be false? is this iterating all the metrics that's been generated?

protocol/app/prepare/full_node_prepare_proposal_test.go
```diff
@@ -1,220 +1,67 @@
...
+	_, err := gometrics.NewGlobal(conf, sink)
+	require.NoError(t, err)
 
-		fundingResp    *perpetualtypes.MsgAddPremiumVotes
-		fundingEncoder sdktypes.TxEncoder
-
-		clobResp    *clobtypes.MsgProposedOperations
-		clobEncoder sdktypes.TxEncoder
-	}{
-		"Error: newPrepareProposalTransactions fails": {
```
ttl33, line 34: to confirm, are we removing these because these cases are irrelevant? (i.e. we don't care about the content of `RequestPrepareProposal` because we simply care that this method to be NOT called)

ttl33 (approved):

protocol/app/prepare/full_node_prepare_proposal_test.go
```diff
@@ -1,220 +1,67 @@
...
 func TestFullNodePrepareProposalHandler(t *testing.T) {
-	tests := map[string]struct {
-		txs      [][]byte
-		maxBytes int64
+	t.Cleanup(gometrics.Shutdown)
 
-		pricesResp    *pricestypes.MsgUpdateMarketPrices
-		pricesEncoder sdktypes.TxEncoder
+	conf := gometrics.DefaultConfig("testService")
+	sink := gometrics.NewInmemSink(time.Hour, time.Hour)
```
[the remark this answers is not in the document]
lcwik: I don't think adding them as constants will make sense since they aren't something we care whether they are consistent within the application or across tests.

I don't expect these tests to run longer than an hour so retaining the metrics for an hour makes sense to me.

protocol/app/prepare/full_node_prepare_proposal_test.go
lcwik: Yes, it is iterating all the metrics that have been produced and there are a few through this.

protocol/app/prepare/full_node_prepare_proposal_test.go
lcwik: That is right.

PR #554, opened 2023-10-10
[DEC-2081] Audit app/prepare/other_msgs.go and app/prepare/transactions.go

lcwik:
Note that the testing of what is and isn't allowed should be covered by lib/ante/disallow_msg.go audit.

### Changelist
Minor test clean-up and swap to use uint value type for values that are never meant to be negative and hid visibility of `UpdateUsedBytes`.

### Test Plan
N/A

### Author/Reviewer Checklist

### Summary by CodeRabbit

- Refactor: Updated data types from `int64` to `uint64` in `other_msgs.go`, `prepare_proposal.go`, and `transactions.go` for better handling of unsigned integer values.
- Test: Modified test setup in `other_msgs_test.go` and `transactions_test.go` to align with the refactored code.
- Chore: Renamed `UpdateUsedBytes` to `updateUsedBytes` in `transactions.go` for better code readability and consistency.

### DIFF
--- a/protocol/app/prepare/transactions.go
+++ b/protocol/app/prepare/transactions.go
@@ -20,8 +20,8 @@ type PrepareProposalTxs struct {
 	// Bytes.
 	// In general, there's no need to check for int64 overflow given that it would require
 	// exabytes of memory to hit the max int64 value in bytes.
-	MaxBytes  int64
-	UsedBytes int64
+	MaxBytes  uint64
+	UsedBytes uint64
 }
 
 // NewPrepareProposalTxs returns a new `PrepareProposalTxs` given the request.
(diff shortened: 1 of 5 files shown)


protocol/app/prepare/transactions.go
```diff
@@ -33,15 +33,14 @@ func NewPrepareProposalTxs(
 	}
 
 	return PrepareProposalTxs{
-		MaxBytes:  req.MaxTxBytes,
-		UsedBytes: 0,
+		MaxBytes: uint64(req.MaxTxBytes),
```
ttl33, line 36: The main reason I ended up using int64 instead of uint64 is because the `PrepareProposalRequest.MaxTxBytes` ([code](https://github.com/cosmos/cosmos-sdk/blob/bff3104781cca3c409fecfa3efebd80a40eb48c3/proto/tendermint/abci/types.proto#L128)) uses `int64` so decided to keep the same data type.

I agree that negative numbers for these do not make sense.

I wonder for the sake of consistency, it's better to keep it as int64 instead of uint64.

@BrendanChou wdyt?

lcwik: We validate that MaxTxBytes can't be negative though.

BrendanChou: If the `PreparePropoalRequest` is from the ABCI call and is positive, then we proceed with our own internal logic to make sure that we will not return more than the number of bytes requested.

We are not really storing this information anywhere or comparing it to other `int64` types so I think it's fine to convert the type.

Since it is a positive `int64` it will fit into a `uint64` so I see no issue there either.

BrendanChou (approved):

ttl33 (approved):

BrendanChou (approved):

PR #559, opened 2023-10-10
Revert [CLOB-942] do not replay stateful orders

jakob-dydx:
Follow up to https://github.com/dydxprotocol/v4-chain/commit/8a209f9e3f8fe2297f079ec9e715d098b11fa086. Logic for the fix was incorrect but we will now log an error instead. Explained in this linear ticket: https://linear.app/dydx/issue/CLOB-942/fix-bug-where-stateful-order-replays-can-conflict-with-order-removals. Original PR: https://github.com/dydxprotocol/v4-chain/pull/504

### Summary by CodeRabbit

"Refactor: 
- Removed unused constants 'ReplayOperations' and 'SkipStatefulReplayPlaceOrder' from the metrics library.
- Simplified order placement logic in the memclob module by eliminating the pre-check for order removal in the operations queue. This change streamlines the order placement process and may result in minor performance improvements."


lucas-dydx (approved):
Want to include in the PR description why the previous change isn't necessary, and link the previous PR?

protocol/x/clob/memclob/memclob.go
```diff
@@ -935,19 +935,6 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
...
-				// place the order. Any generated matches would fail in DeliverTx as the order would be missing from
-				// state (due to the order removal being processed).
-				telemetry.IncrCounter(
-					1,
-					types.ModuleName,
-					metrics.ReplayOperations,
-					metrics.SkipStatefulReplayPlaceOrder,
-				)
-				continue
-			}
```
lucas-dydx, line 949: Actually can we change this to an error log instead? I don't think this should ever happen and if it does it's an error.

jakob-dydx: sure

PR #562, opened 2023-10-11
[CLOB-953] port over changes from v0.3.x release branch to main

jayy04:
### Changelist
- Port over changes from v0.3.x release branch to main.

### Test Plan

### Author/Reviewer Checklist

### Summary by CodeRabbit

- New Feature: Added a new flag `MaxDeleveragingSubaccountsToIterate` to the `ClobFlags` struct, allowing users to specify the maximum number of subaccounts to iterate for deleveraging.
- Refactor: Updated the `OffsetSubaccountPerpetualPosition` function to iterate over subaccounts, check their positions, and perform offsetting operations more efficiently.
- New Feature: Introduced a pseudorandom number generator in the `liquidations.go` file to process a maximum of `MaxLiquidationOrdersPerBlock` subaccounts from a pseudorandom location.
- New Feature: Added a new function `GetPseudoRand` to the `keeper` package, providing a random number generator seeded with a pseudorandom seed based on the previous block timestamp.
- Test: Updated test cases in `flags_test.go` to accommodate the new `MaxDeleveragingSubaccountsToIterate` flag.

### DIFF
--- a/protocol/x/clob/flags/flags.go
+++ b/protocol/x/clob/flags/flags.go
@@ -32,7 +32,7 @@ const (
 
 // Default values.
 const (
-	DefaultMaxLiquidationOrdersPerBlock    = 100
+	DefaultMaxLiquidationOrdersPerBlock    = 20
 	DefaultMaxDeleveragingAttemptsPerBlock = 5
 
 	DefaultMevTelemetryEnabled    = false
--- a/protocol/x/clob/flags/flags.go
+++ b/protocol/x/clob/flags/flags.go
@@ -138,8 +138,10 @@ func GetClobFlagValuesFromOptions(
 		}
 	}
 
-	if v, ok := appOpts.Get(MaxDeleveragingSubaccountsToIterate).(uint32); ok {
-		result.MaxDeleveragingSubaccountsToIterate = v
+	if option := appOpts.Get(MaxDeleveragingSubaccountsToIterate); option != nil {
+		if v, err := cast.ToUint32E(option); err == nil {
+			result.MaxDeleveragingSubaccountsToIterate = v
+		}
 	}
 
 	return result
--- a/protocol/x/clob/flags/flags.go
+++ b/protocol/x/clob/flags/flags.go
@@ -10,7 +10,7 @@ import (
 
 // A struct containing the values of all flags.
 type ClobFlags struct {
-	MaxLiquidationOrdersPerBlock        uint32
+	MaxLiquidationAttemptsPerBlock      uint32
 	MaxDeleveragingAttemptsPerBlock     uint32
 	MaxDeleveragingSubaccountsToIterate uint32
 
@@ -22,7 +22,7 @@ type ClobFlags struct {
 // List of CLI flags.
 const (
 	// Liquidations and deleveraging.
-	MaxLiquidationOrdersPerBlock        = "max-liquidation-orders-per-block"
+	MaxLiquidationAttemptsPerBlock      = "max-liquidation-attempts-per-block"
 	MaxDeleveragingAttemptsPerBlock     = "max-deleveraging-attempts-per-block"
 	MaxDeleveragingSubaccountsToIterate = "max-deleveraging-subaccounts-to-iterate"
 
@@ -34,8 +34,8 @@ const (
 
 // Default values.
 const (
-	DefaultMaxLiquidationOrdersPerBlock        = 20
-	DefaultMaxDeleveragingAttemptsPerBlock     = 5
+	DefaultMaxLiquidationAttemptsPerBlock      = 50
+	DefaultMaxDeleveragingAttemptsPerBlock     = 10
 	DefaultMaxDeleveragingSubaccountsToIterate = 500
 
 	DefaultMevTelemetryEnabled    = false
@@ -48,11 +48,11 @@ const (
 // E.g. `dydxprotocold start --non-validating-full-node true`.
 func AddClobFlagsToCmd(cmd *cobra.Command) {
 	cmd.Flags().Uint32(
-		MaxLiquidationOrdersPerBlock,
-		DefaultMaxLiquidationOrdersPerBlock,
+		MaxLiquidationAttemptsPerBlock,
+		DefaultMaxLiquidationAttemptsPerBlock,
 		fmt.Sprintf(
 			"Sets the maximum number of liquidation orders to process per block. Default = %d",
-			DefaultMaxLiquidationOrdersPerBlock,
+			DefaultMaxLiquidationAttemptsPerBlock,
 		),
 	)
 	cmd.Flags().Uint32(
@@ -90,7 +90,7 @@ func AddClobFlagsToCmd(cmd *cobra.Command) {
 
 func GetDefaultClobFlags() ClobFlags {
 	return ClobFlags{
-		MaxLiquidationOrdersPerBlock:        DefaultMaxLiquidationOrdersPerBlock,
+		MaxLiquidationAttemptsPerBlock:      DefaultMaxLiquidationAttemptsPerBlock,
 		MaxDeleveragingAttemptsPerBlock:     DefaultMaxDeleveragingAttemptsPerBlock,
 		MaxDeleveragingSubaccountsToIterate: DefaultMaxDeleveragingSubaccountsToIterate,
 		MevTelemetryEnabled:                 DefaultMevTelemetryEnabled,
@@ -126,9 +126,9 @@ func GetClobFlagValuesFromOptions(
 		}
 	}
 
-	if option := appOpts.Get(MaxLiquidationOrdersPerBlock); option != nil {
+	if option := appOpts.Get(MaxLiquidationAttemptsPerBlock); option != nil {
 		if v, err := cast.ToUint32E(option); err == nil {
-			result.MaxLiquidationOrdersPerBlock = v
+			result.MaxLiquidationAttemptsPerBlock = v
 		}
 	}
 
--- a/protocol/x/clob/flags/flags.go
+++ b/protocol/x/clob/flags/flags.go
@@ -10,8 +10,9 @@ import (
 
 // A struct containing the values of all flags.
 type ClobFlags struct {
-	MaxLiquidationOrdersPerBlock    uint32
-	MaxDeleveragingAttemptsPerBlock uint32
+	MaxLiquidationOrdersPerBlock        uint32
+	MaxDeleveragingAttemptsPerBlock     uint32
+	MaxDeleveragingSubaccountsToIterate uint32
 
 	MevTelemetryEnabled    bool
 	MevTelemetryHost       string
@@ -21,8 +22,9 @@ type ClobFlags struct {
 // List of CLI flags.
 const (
 	// Liquidations and deleveraging.
-	MaxLiquidationOrdersPerBlock    = "max-liquidation-orders-per-block"
-	MaxDeleveragingAttemptsPerBlock = "max-deleveraging-attempts-per-block"
+	MaxLiquidationOrdersPerBlock        = "max-liquidation-orders-per-block"
+	MaxDeleveragingAttemptsPerBlock     = "max-deleveraging-attempts-per-block"
+	MaxDeleveragingSubaccountsToIterate = "max-deleveraging-subaccounts-to-iterate"
 
 	// Mev.
 	MevTelemetryEnabled    = "mev-telemetry-enabled"
@@ -32,8 +34,9 @@ const (
 
 // Default values.
 const (
-	DefaultMaxLiquidationOrdersPerBlock    = 20
-	DefaultMaxDeleveragingAttemptsPerBlock = 5
+	DefaultMaxLiquidationOrdersPerBlock        = 20
+	DefaultMaxDeleveragingAttemptsPerBlock     = 5
+	DefaultMaxDeleveragingSubaccountsToIterate = 500
 
 	DefaultMevTelemetryEnabled    = false
 	DefaultMevTelemetryHost       = ""
@@ -60,6 +63,14 @@ func AddClobFlagsToCmd(cmd *cobra.Command) {
 			DefaultMaxDeleveragingAttemptsPerBlock,
 		),
 	)
+	cmd.Flags().Uint32(
+		MaxDeleveragingSubaccountsToIterate,
+		DefaultMaxDeleveragingSubaccountsToIterate,
+		fmt.Sprintf(
+			"Sets the maximum number of subaccounts iterated for each deleveraging event. Default = %d",
+			DefaultMaxDeleveragingSubaccountsToIterate,
+		),
+	)
 	cmd.Flags().Bool(
 		MevTelemetryEnabled,
 		DefaultMevTelemetryEnabled,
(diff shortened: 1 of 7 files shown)


protocol/x/clob/flags/flags.go
```diff
@@ -33,8 +34,9 @@ const (
 
 // Default values.
 const (
-	DefaultMaxLiquidationOrdersPerBlock    = math.MaxUint32
-	DefaultMaxDeleveragingAttemptsPerBlock = 35
+	DefaultMaxLiquidationOrdersPerBlock        = 20
+	DefaultMaxDeleveragingAttemptsPerBlock     = 5
+	DefaultMaxDeleveragingSubaccountsToIterate = 500
```
jayy04, line 39: @BrendanChou are these numbers reasonable? 

I think after the metrics optimization, we might be able to increase these numbers (maybe 3x or 5x, given the performance improvement). let me know what you think

protocol/x/clob/keeper/keeper.go
```diff
@@ -108,10 +109,11 @@ func NewKeeper(
 			Host:       clobFlags.MevTelemetryHost,
 			Identifier: clobFlags.MevTelemetryIdentifier,
 		},
-		MaxLiquidationOrdersPerBlock:    clobFlags.MaxLiquidationOrdersPerBlock,
-		MaxDeleveragingAttemptsPerBlock: clobFlags.MaxDeleveragingAttemptsPerBlock,
-		placeOrderRateLimiter:           placeOrderRateLimiter,
-		cancelOrderRateLimiter:          cancelOrderRateLimiter,
+		MaxLiquidationOrdersPerBlock:        clobFlags.MaxLiquidationOrdersPerBlock,
```
BrendanChou, line 112: nit: `MaxLiquidationAttemptsPerBlock` ? Or `PerPrepareCheckState`?

protocol/x/clob/keeper/keeper.go
```diff
@@ -44,8 +44,9 @@ type (
 
 		memStoreInitialized *atomic.Bool
 
-		MaxLiquidationOrdersPerBlock    uint32
-		MaxDeleveragingAttemptsPerBlock uint32
+		MaxLiquidationOrdersPerBlock        uint32
+		MaxDeleveragingAttemptsPerBlock     uint32
+		MaxDeleveragingSubaccountsToIterate uint32
```
BrendanChou, line 49: Nit: Store these as a single `Flags ClobFlags` struct so that they don't have to manually be updated if we add more flags

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -215,96 +220,36 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
+				numSubaccountsWithNonOverlappingBankruptcyPrices++
 			}
 			return deltaQuantumsRemaining.Sign() == 0
 		},
-		rand,
+		k.GetPseudoRand(ctx),
 	)
 
-	telemetry.SetGauge(float32(numSubaccountsIterated), metrics.NumSubaccountsIterated, metrics.Count)
+	telemetry.IncrCounter(
```
BrendanChou, line 241: want to change these from `IncrCounter`?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -315,15 +260,12 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
 			1,
 			types.ModuleName, metrics.CheckTx, metrics.Deleveraging, metrics.NotEnoughPositionToFullyOffset, metrics.Count,
```
BrendanChou, line 261: consider modifying these metrics to tell a clearer story. There is a difference between not finding anything to deleverage against vs finding partial positions to deleverage against vs finding a full fill (deltaQuantumsRemaining = 0)

protocol/x/clob/keeper/liquidations.go
```diff
@@ -29,9 +29,21 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+
 	// Get the liquidation order for each subaccount.
+	// Process at-most `MaxLiquidationOrdersPerBlock` subaccounts, starting from a pseudorandom location
+	// in the slice.
 	liquidationOrdersForSorting := make([]types.LiquidationOrder, 0)
-	for _, subaccountId := range subaccountIds {
+	numSubaccounts := len(subaccountIds)
+	numLiqOrders := lib.Min(numSubaccounts, int(k.MaxLiquidationOrdersPerBlock))
+	indexOffset := 0
+	if numSubaccounts > 0 {
```
lucas-dydx, line 41: For my understanding why not just early return if `numSubaccounts == 0`?

IMO that might make sense to early return because currently the stat `numFilledLiquidations` will be 0 if there are no liquidatable subaccounts (which makes sense but IMO that stat should be used for verifying percent of filled liquidation orders).

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -162,21 +162,26 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
 	)
 
 	numSubaccountsIterated := 0
+	numSubaccountsWithNonOverlappingBankruptcyPrices := 0
+	numSubaccountsWithNoOpenPositionOnOppositeSide := 0
```
lucas-dydx, line 166: Suggestion, up to you: for clarity why not use the `uint` type?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -215,96 +220,36 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
-					"offsettingBankruptcyPriceQuoteQuantums", offsettingBankruptcyPrice,
-					"offsettingTnc", offsettingTnc,
-				)
-				telemetry.IncrCounter(
-					1,
-					types.ModuleName, metrics.Deleveraging, metrics.NonOverlappingBankruptcyPrices, metrics.Count,
+					"liquidatedSubaccount", log.NewLazySprintf("%+v", liquidatedSubaccount),
+					"offsettingSubaccount", log.NewLazySprintf("%+v", offsettingSubaccount),
 				)
+				numSubaccountsWithNonOverlappingBankruptcyPrices++
```
lucas-dydx, line 234: Unrelated to this PR, but should we error log in `ProcessDeleveraging` if the update for the deleveraged subaccount (not the offsetting subaccount) fails?

Based on [this logic](https://github.com/dydxprotocol/v4-chain/blob/a68b735cbe6031fd3dae3dfc9870dbe0ff7ce8e3/protocol/x/clob/keeper/deleveraging.go#L433-L442) it doesn't look like we differentiate between collat checks failing for deleveraged vs offsetting subaccount, and this code assumes collat check failed for offsetting subaccount.

protocol/x/clob/flags/flags.go
```diff
@@ -127,5 +138,11 @@ func GetClobFlagValuesFromOptions(
 		}
 	}
 
+	if option := appOpts.Get(MaxDeleveragingSubaccountsToIterate); option != nil {
+		if v, err := cast.ToUint32E(option); err == nil {
+			result.MaxDeleveragingSubaccountsToIterate = v
```
lucas-dydx, line 143: Is it worth creating a low-priority M6 ticket to validate this? Slightly worried that when people are setting this they could accidentally set it to zero and make it impossible to deleverage subaccounts.

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -162,21 +162,26 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
-	s := rand.NewSource(k.blockTimeKeeper.GetPreviousBlockInfo(ctx).Timestamp.Unix())
-	rand := rand.New(s)
-
 	k.subaccountsKeeper.ForEachSubaccountRandomStart(
 		ctx,
 		func(offsettingSubaccount satypes.Subaccount) (finished bool) {
+			// Iterate at most `MaxDeleveragingSubaccountsToIterate` subaccounts.
+			if numSubaccountsIterated >= int(k.MaxDeleveragingSubaccountsToIterate) {
+				return true
+			}
```
lucas-dydx, line 176: Question for my understanding: why not put this at the end of the loop?

Won't this logic technically read `MaxDeleveragingSubaccountsToIterate + 1` subaccounts from state because this finishing condition is at the beginning of the loop (a subaccount is read from state then not used)?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -99,9 +111,13 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 			if numAttemptedDeleveraging < k.MaxDeleveragingAttemptsPerBlock {
 				// The liquidation order was unfilled. Try to deleverage the subaccount.
 				subaccountId := liquidationOrder.GetSubaccountId()
-				perpetualId := liquidationOrder.MustGetLiquidatedPerpetualId()
 				deltaQuantums := liquidationOrder.GetDeltaQuantums()
 
+				// Get a pseudorandom perpetualId from the list of open positions.
+				subaccount := k.subaccountsKeeper.GetSubaccount(ctx, subaccountId)
+				positions := subaccount.GetPerpetualPositions()
+				perpetualId := positions[pseudoRand.Intn(len(positions))].PerpetualId
```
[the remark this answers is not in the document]
lucas-dydx: I think a panic here is justified since [this liquidation order validation](https://github.com/dydxprotocol/v4-chain/blob/a68b735cbe6031fd3dae3dfc9870dbe0ff7ce8e3/protocol/x/clob/keeper/liquidations.go#L152-L155) must have passed for this logic to be executed.

Understand this is an implicit dependency though. Maybe worth adding a comment that this code assumes that the subaccount has at least one open position? Would prefer to reduce the number of errors returned upstream since it makes the code unreadable + errors in the liquidation flow are usually worth halting the chain for.

protocol/x/clob/keeper/random.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"math/rand"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GetPseudoRand returns a random number generator seeded with a pseudorandom seed.
+// The seed is based on the previous block timestamp.
+func (k *Keeper) GetPseudoRand(ctx sdk.Context) *rand.Rand {
```
lucas-dydx, line 11: Suggestion, up to you: should this function be unit tested? Maybe a simple unit test for determinism (just verify that same block info produces same values, and updating it produces diff values)?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -29,9 +29,21 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+	numSubaccounts := len(subaccountIds)
+	numLiqOrders := lib.Min(numSubaccounts, int(k.MaxLiquidationOrdersPerBlock))
+	indexOffset := 0
+	if numSubaccounts > 0 {
+		indexOffset = pseudoRand.Intn(numSubaccounts)
+	}
+	for i := 0; i < numLiqOrders; i++ {
+		index := (i + indexOffset) % numSubaccounts
+		subaccountId := subaccountIds[index]
 		// If attempting to liquidate a subaccount returns an error, panic.
```
lucas-dydx, line 47: Suggestion: we should update this comment since this logic doesn't liquidate subaccounts anymore

protocol/x/clob/flags/flags.go
lucas-dydx: I'd personally prefer the more aggressive numbers Brendan suggested unless the performance tradeoff is meaningful (liquidating / deleveraging bad debt quickly is important for the exchange).

lucas-dydx, line 39: For my understanding, what methodology should people use for configuring these values in the future?

Fine with creating a medium priority M6 ticket for research to figure this out later.

lucas-dydx (approved):

protocol/x/clob/keeper/liquidations.go
```diff
@@ -29,9 +29,21 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+	// in the slice.
 	liquidationOrdersForSorting := make([]types.LiquidationOrder, 0)
-	for _, subaccountId := range subaccountIds {
+	numSubaccounts := len(subaccountIds)
+	numLiqOrders := lib.Min(numSubaccounts, int(k.MaxLiquidationOrdersPerBlock))
+	indexOffset := 0
+	if numSubaccounts > 0 {
+		indexOffset = pseudoRand.Intn(numSubaccounts)
+	}
+	for i := 0; i < numLiqOrders; i++ {
```
lucas-dydx, line 44: Is it worth measuring the latency of generating the liquidation orders? It seems like this would be helpful for determining how to tune `MaxLiquidationOrdersPerBlock` going forwards.

Or are we trying to limit metric usage due to how expensive it is?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -60,7 +72,7 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 	// Attempt to place each liquidation order and perform deleveraging if necessary.
 	numFilledLiquidations := uint32(0)
 	numAttemptedDeleveraging := uint32(0)
-	for i := 0; numFilledLiquidations < k.MaxLiquidationOrdersPerBlock && i < len(liquidationOrdersForSorting); i++ {
+	for i := 0; i < len(liquidationOrdersForSorting); i++ {
```
lucas-dydx, line 75: Is it worth measuring the latency of how long it takes to liquidate subaccounts as well (will help determine how to configure the max liq orders per block)?

protocol/x/clob/keeper/liquidations.go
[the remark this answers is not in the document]
jayy04: Will refactor code and do this in a separate PR

protocol/x/clob/keeper/liquidations.go
```diff
@@ -99,9 +111,13 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 			if numAttemptedDeleveraging < k.MaxDeleveragingAttemptsPerBlock {
 				// The liquidation order was unfilled. Try to deleverage the subaccount.
 				subaccountId := liquidationOrder.GetSubaccountId()
-				perpetualId := liquidationOrder.MustGetLiquidatedPerpetualId()
 				deltaQuantums := liquidationOrder.GetDeltaQuantums()
 
+				// Get a pseudorandom perpetualId from the list of open positions.
```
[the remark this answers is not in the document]
jayy04: I think this code will not live past this week so will punt on this for now.

protocol/x/clob/keeper/liquidations.go
```diff
@@ -29,9 +29,21 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
 
+	pseudoRand := k.GetPseudoRand(ctx)
+
 	// Get the liquidation order for each subaccount.
+	// Process at-most `MaxLiquidationOrdersPerBlock` subaccounts, starting from a pseudorandom location
+	// in the slice.
 	liquidationOrdersForSorting := make([]types.LiquidationOrder, 0)
-	for _, subaccountId := range subaccountIds {
+	numSubaccounts := len(subaccountIds)
+	numLiqOrders := lib.Min(numSubaccounts, int(k.MaxLiquidationOrdersPerBlock))
```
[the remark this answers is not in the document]
jayy04: we have a ticket in M6 to validate flags in general. currently there is also nothing that's preventing node operators to turn off liquidation daemon.

protocol/x/clob/keeper/deleveraging.go
jayy04: I think it's pretty clear that it's a counter (non negative) by their names. they get converted to float32 anyways, so doesn't really matter

updated since it's an easy change

protocol/x/clob/flags/flags.go
jayy04: https://linear.app/dydx/issue/DEC-2146

protocol/x/clob/keeper/deleveraging.go
jayy04: putting at the end of loop will technically not be correct when Max = 0. reading an extra subaccount is not really a big deal in my opinion.

protocol/x/clob/keeper/random.go
jayy04: math/rand should have pretty comprehensive unit tests for determinism. personally don't feel like it's necessary.

protocol/x/clob/keeper/liquidations.go
jayy04: going to do metrics in a separate PR

protocol/x/clob/keeper/liquidations.go
jayy04: going to do metrics in a separate PR

protocol/x/clob/keeper/deleveraging.go
jayy04: let me add an error log in a future PR

protocol/x/clob/keeper/liquidations.go
BrendanChou: Yeah just think about this more in general when you do metrics. I also noticed we always time `k.SortLiquidationOrders` even when there are no orders to sort

BrendanChou (approved):

PR #580, opened 2023-10-11
[CORE-654] Audit lib/eth

BrendanChou:
### Changelist
- Cleanup lib/eth code and add tests

### Test Plan
- Added more unit tests

### DIFF
--- a/protocol/lib/eth/util.go
+++ b/protocol/lib/eth/util.go
@@ -1,9 +1,11 @@
 package eth
 
 import (
-	sdkmath "cosmossdk.io/math"
 	"math/big"
 	"strings"
+	"sync"
+
+	sdkmath "cosmossdk.io/math"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/app/config"
@@ -14,33 +16,41 @@ import (
 	ethcoretypes "github.com/ethereum/go-ethereum/core/types"
 )
 
-var bridgeEventAbi *ethabi.ABI
-
 const (
-	MIN_ADDRESS_BYTES = 20
-	MAX_ADDRESS_BYTES = 32
+	MinAddrLen = 20
+	MaxAddrLen = 32
 )
 
+// ABI Singleton and mutex.
+var bridgeEventAbi *ethabi.ABI
+var once sync.Once
+
 // getBridgeEventAbi returns the ABI (application binary interface) for the Bridge contract.
-func getBridgeEventAbi() *ethabi.ABI {
-	// Initialize the singleton if it does not exist.
-	if bridgeEventAbi == nil {
+func GetBridgeEventAbi() *ethabi.ABI {
+	// Initialize the singleton in a thread-safe way.
+	once.Do(func() {
 		bridgeAbi, err := ethabi.JSON(strings.NewReader(constants.BridgeEventABI))
 		if err != nil {
 			panic(err)
 		}
 		bridgeEventAbi = &bridgeAbi
+	})
+
+	// This could happen if the initialization fails in another thread.
+	if bridgeEventAbi == nil {
+		panic("getBridgeEventAbi: bridgeEventAbi not initialized")
 	}
+
 	return bridgeEventAbi
 }
 
-// padOrTruncateAddress right-pads an address with zeros if it's shorter than `MIN_ADDRESS_BYTES` or
-// takes the first `MAX_ADDRESS_BYTES` if it's longer than that.
-func padOrTruncateAddress(address []byte) []byte {
-	if len(address) > MAX_ADDRESS_BYTES {
-		return address[:MAX_ADDRESS_BYTES]
-	} else if len(address) < MIN_ADDRESS_BYTES {
-		return append(address, make([]byte, MIN_ADDRESS_BYTES-len(address))...)
+// PadOrTruncateAddress right-pads an address with zeros if it's shorter than `minAddrLen` or
+// takes the first `maxAddrLen` if it's longer than that.
+func PadOrTruncateAddress(address []byte) []byte {
+	if len(address) > MaxAddrLen {
+		return address[:MaxAddrLen]
+	} else if len(address) < MinAddrLen {
+		return append(address, make([]byte, MinAddrLen-len(address))...)
 	}
 	return address
 }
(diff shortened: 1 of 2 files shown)


protocol/lib/eth/util_test.go
```diff
@@ -9,10 +10,104 @@ import (
...
 	bridgetypes "github.com/dydxprotocol/v4-chain/protocol/x/bridge/types"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcoretypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/stretchr/testify/require"
 )
 
+func TestGetBridgeEventAbi(t *testing.T) {
+	results := make([]*abi.ABI, 0)
+	mu := &sync.Mutex{}
+	wg := new(sync.WaitGroup)
```
lcwik, line 21: nit: consider using a non-pointer version of the waitgroup so it is allocated on the stack.
```suggestion
	wg := sync.WaitGroup{}
```
or `var wg sync.WaitGroup{}`

protocol/lib/eth/util_test.go
```diff
@@ -9,10 +10,104 @@ import (
...
 	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
 	bridgetypes "github.com/dydxprotocol/v4-chain/protocol/x/bridge/types"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcoretypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/stretchr/testify/require"
 )
 
+func TestGetBridgeEventAbi(t *testing.T) {
+	results := make([]*abi.ABI, 0)
+	mu := &sync.Mutex{}
```
lcwik, line 20: nit: consider using the value type.
```suggestion
	mu := sync.Mutex{}
```

the function your using captures `mu` in its closure as a reference to `mu`

protocol/lib/eth/util.go
```diff
@@ -14,33 +16,41 @@ import (
...
+	once.Do(func() {
 		bridgeAbi, err := ethabi.JSON(strings.NewReader(constants.BridgeEventABI))
 		if err != nil {
 			panic(err)
 		}
 		bridgeEventAbi = &bridgeAbi
+	})
+
+	// This could happen if the initialization fails in another thread.
+	if bridgeEventAbi == nil {
```
lcwik, line 40: Consider using [OnceValue](https://pkg.go.dev/sync#OnceValue) as it is simpler and guaranteed to panic when invoked instead of needing to check. This also allows you to combine the once with the value being stored and prevents erroneous usage of using `bridgeEventAbi` without the once.

BrendanChou: Unfortunately I don't think it exists in go 1.19, I think it's new as of 1.21

tqin7 (approved):

protocol/lib/eth/util.go
lcwik: We'll need to upgrade to 1.21 for Cosmos 0.50 so might as well and do it now: https://github.com/dydxprotocol/v4-chain/pull/586

lcwik: The upgrade to go 1.21 is now merged.

protocol/lib/eth/util.go
```diff
@@ -21,27 +21,16 @@ const (
...
-var bridgeEventAbi *ethabi.ABI
-var once sync.Once
-
-// getBridgeEventAbi returns the ABI (application binary interface) for the Bridge contract.
-func GetBridgeEventAbi() *ethabi.ABI {
-	// Initialize the singleton in a thread-safe way.
-	once.Do(func() {
-		bridgeAbi, err := ethabi.JSON(strings.NewReader(constants.BridgeEventABI))
+// GetBridgeEventAbi returns the ABI (application binary interface) for the Bridge contract.
+func GetBridgeEventAbi() ethabi.ABI {
```
lcwik, line 25: I would use a pointer type since the struct seems like it could be pretty large so not pay the copy tax on each invocation of this method.

The caveat is that it isn't truly safe as a pointer type since `UnmarhsalJSON` mutates it so returning a copy would make each instance truly thread safe if that was important otherwise it is only thread safe as long as nobody else calls `UnmarshalJSON` on it.

protocol/lib/eth/util.go
```diff
@@ -21,27 +21,16 @@ const (
...
+	result := sync.OnceValue[ethabi.ABI](func() ethabi.ABI {
```
lcwik, line 26: You want to define `result` outside of this method as a global variable otherwise `sync.OnceValue()` will be created each time you invoke this method.

protocol/lib/eth/util_test.go
```diff
@@ -9,10 +10,104 @@ import (
...
+	wg.Wait()
+
+	// Call the function one more time.
+	// Ensure that all the pointers are equal.
+	expected := libeth.GetBridgeEventAbi()
+	require.NotNil(t, expected)
+	for _, r := range results {
+		require.Same(t, expected, r)
+	}
+}
```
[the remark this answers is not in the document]
BrendanChou: Can you explain what code changes would be needed?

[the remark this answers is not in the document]
BrendanChou: Can you explain how the contents would be different if the pointers are the same?

protocol/lib/eth/util.go
```diff
@@ -14,33 +16,36 @@ import (
...
 
-// padOrTruncateAddress right-pads an address with zeros if it's shorter than `MIN_ADDRESS_BYTES` or
-// takes the first `MAX_ADDRESS_BYTES` if it's longer than that.
-func padOrTruncateAddress(address []byte) []byte {
-	if len(address) > MAX_ADDRESS_BYTES {
-		return address[:MAX_ADDRESS_BYTES]
-	} else if len(address) < MIN_ADDRESS_BYTES {
-		return append(address, make([]byte, MIN_ADDRESS_BYTES-len(address))...)
+// PadOrTruncateAddress right-pads an address with zeros if it's shorter than `minAddrLen` or
+// takes the first `maxAddrLen` if it's longer than that.
```
lcwik, line 43:
```suggestion
// PadOrTruncateAddress right-pads an address with zeros if it's shorter than `MinAddrLen` or
// takes the first `MaxAddrLen` if it's longer than that.
```

lcwik (approved):

PR #581, opened 2023-10-11
[DEC-2097] - lib/math audit

clemire:
### Changelist
Comments and method renames to clarify cases when methods may cause panics, rounding behavior.
- Confirmed that panics which bubble up to x/perpetuals are expected/desired.
- Rounding behavior in modules was already audited, but added documentation.

### Test Plan
No behavior changes.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/lib/big_math.go
+++ b/protocol/lib/big_math.go
@@ -45,7 +45,8 @@ func RatPow10(exponent int32) *big.Rat {
 	return result
 }
 
-// BigIntMulPpm takes a `big.Int` and returns the result of `input * ppm / 1_000_000`.
+// BigIntMulPpm takes a `big.Int` and returns the result of `input * ppm / 1_000_000`. This method rounds towards
+// zero.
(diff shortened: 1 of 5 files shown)


protocol/lib/big_math.go
```diff
@@ -45,7 +45,8 @@ func RatPow10(exponent int32) *big.Rat {
 	return result
 }
 
-// BigIntMulPpm takes a `big.Int` and returns the result of `input * ppm / 1_000_000`.
+// BigIntMulPpm takes a `big.Int` and returns the result of `input * ppm / 1_000_000`. This method rounds towards
+// zero.
```
teddyding, line 49: I think this rounds towards neg infinity, since input can be negative. Could you update comment/add unit test accordingly?

teddyding (approved):

protocol/lib/big_math.go
clemire: You're right! Updating comment, adding test cases.

PR #596, opened 2023-10-12
[CLOB-956] improve liquidations and deleveraging telemetry

jayy04:
### Changelist
- Improve telemetry for liquidations and deleveraging.

### Test Plan

### Author/Reviewer Checklist


protocol/x/clob/abci.go
```diff
@@ -180,20 +186,6 @@ func PrepareCheckState(
...
-		metrics.Liquidations,
-		metrics.LiquidatableSubaccountIds,
-		metrics.Count,
-	)
-
-	telemetry.ModuleSetGauge(
-		types.ModuleName,
-		metrics.GetMetricValueFromBigInt(keeper.GetInsuranceFundBalance(ctx)),
-		metrics.InsuranceFundBalance,
-	)
```
jayy04, line 195: I think this is more appropriate in `EndBlocker`

protocol/x/clob/abci.go
```diff
@@ -180,20 +186,6 @@ func PrepareCheckState(
 		panic(err)
 	}
 
-	telemetry.ModuleSetGauge(
-		types.ModuleName,
-		float32(len(subaccountIds)),
-		metrics.Liquidations,
-		metrics.LiquidatableSubaccountIds,
-		metrics.Count,
-	)
```
jayy04, line 189: moved to `LiquidateSubaccountsAgainstOrderbook`

protocol/x/clob/keeper/process_operations.go
```diff
@@ -590,24 +590,6 @@ func (k Keeper) PersistMatchLiquidationToState(
...
-				fillAmountToTotalSizeRat,
-				new(big.Rat).SetInt(absNotionalQuoteQuantums),
-			),
-			true,
-		)
-		telemetry.IncrCounterWithLabels(
-			[]string{types.ModuleName, metrics.LiquidationOrderNotionalQuoteQuantums, metrics.Filled, metrics.DeliverTx},
-			float32(filledQuoteQuantums.Uint64()),
-			matchLiquidation.GetMetricLabels(),
-		)
```
jayy04, line 609: this is being stat'ed in `process_single_match` which is shared between check tx and deliver tx

protocol/x/clob/abci.go
```diff
@@ -104,6 +104,12 @@ func EndBlocker(
 
 	// Prune any rate limiting information that is no longer relevant.
 	keeper.PruneRateLimits(ctx)
+
+	telemetry.ModuleSetGauge(
+		types.ModuleName,
+		metrics.GetMetricValueFromBigInt(keeper.GetInsuranceFundBalance(ctx)),
+		metrics.InsuranceFundBalance,
+	)
```
BrendanChou, line 112: nit: could put in a different function or have a clear comment that this is for any metrics that you want to emit every block

protocol/x/clob/abci.go
BrendanChou: You already also have this metric already on the daemon/server side right? It's kind of showing up here a 3rd time (which may be fine but may also be unnecessary)

protocol/lib/metrics/constants.go
```diff
@@ -106,8 +106,11 @@ const (
 	ConvertToUpdates                             = "convert_to_updates"
 	CreateClobPair                               = "create_clob_pair"
 	Expired                                      = "expired"
+	FullyFilled                                  = "fully_filled"
 	GetFillQuoteQuantums                         = "get_fill_quote_quantums"
 	Hydrate                                      = "hydrate"
+	IsBuy                                        = "is_buy"
+	IsLong                                       = "is_long"
```
BrendanChou, line 113: Is `isBuy: true/false` going to be better than `side:bid/ask`?
Basically, consider a string tag rather than a boolean tag (which I think gets converted into a string anyway?)

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -243,6 +246,25 @@ func (k Keeper) ProcessSingleMatch(
...
+			[]string{metrics.Liquidations, metrics.PlacePerpetualLiquidation, metrics.Filled, metrics.QuoteQuantums},
+			metrics.GetMetricValueFromBigInt(notionalLiquidatedQuoteQuantums),
+			labels,
+		)
+		// Stat insurance fund delta.
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.Liquidations, metrics.InsuranceFundDelta},
+			metrics.GetMetricValueFromBigInt(new(big.Int).Abs(takerInsuranceFundDelta)),
+			append(labels, metrics.GetLabelForBoolValue(metrics.Positive, takerInsuranceFundDelta.Sign() == 1)),
+		)
```
BrendanChou, line 267: when you use a counter, is it possible to get a nice graph where you can see that $100k got liquidated in this hour and $200k got liquidated in the next hour? Or are you stuck with having only cumulative-sum graphs?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -49,14 +50,40 @@ func (k Keeper) MaybeDeleverageSubaccount(
...
+					metrics.FullyFilled,
+					quantumsDeleveraged.CmpAbs(deltaQuantums) == 0,
+				),
+			},
+			labels...,
+		),
 	)
 
-	return k.MemClob.DeleverageSubaccount(ctx, subaccountId, perpetualId, deltaQuantums)
+	if quoteQuantums, err := k.perpetualsKeeper.GetNetNotional(
```
lucas-dydx, line 74: Should we log out `err` if it's non-nil?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -383,6 +403,24 @@ func (k Keeper) ProcessDeleveraging(
 		return updateErr
 	}
 
+	// Stat quantums deleveraged in quote quantums.
+	if delevergedQuoteQuantums, err := k.perpetualsKeeper.GetNetCollateral(
```
lucas-dydx, line 407:
```suggestion
	if deleveragedQuoteQuantums, err := k.perpetualsKeeper.GetNetCollateral(
```

protocol/x/clob/keeper/liquidations.go
```diff
@@ -92,23 +98,14 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 		optimisticallyFilledQuantums, _, err := k.PlacePerpetualLiquidation(ctx, *liquidationOrder)
 		if err != nil {
 			k.Logger(ctx).Error(
-				fmt.Sprintf(
-					"Failed to liquidate subaccount. Liquidation Order: (%+v). Err: %v",
-					liquidationOrder,
-					err,
-				),
+				"Failed to liquidate subaccount",
+				"liquidationOrder", liquidationOrder,
```
lucas-dydx, line 102: Will this print out as a pointer? Should it be `*liquidationOrder`?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -222,34 +224,49 @@ func (k Keeper) PlacePerpetualLiquidation(
...
-		metrics.Liquidations,
-		metrics.PlacePerpetualLiquidation,
-		metrics.Count,
-	)
+	labels := []gometrics.Label{
+		metrics.GetLabelForIntValue(metrics.PerpetualId, int(perpetualId)),
+		metrics.GetLabelForBoolValue(metrics.IsBuy, liquidationOrder.IsBuy()),
+	}
 
+	// Stat the number of liquidation orders placed.
```
lucas-dydx, line 239: Should we stat this outside of `PlacePerpetualLiquidation`, maybe [here](https://github.com/dydxprotocol/v4-chain/blob/d9fc60643fabfe028afd21e26a0820b3b042d070/protocol/x/clob/keeper/liquidations.go#L80) in `LiquidateSubaccountsAgainstOrderbook`?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -164,6 +154,12 @@ func (k Keeper) MaybeGetLiquidationOrder(
 		return nil, types.ErrSubaccountNotLiquidatable
 	}
 
+	defer telemetry.ModuleMeasureSince(
+		types.ModuleName,
+		time.Now(),
+		metrics.MaybeGetLiquidationOrder,
+	)
```
BrendanChou, line 161: this naming be misleading because you are leaving out the `IsLiquidatable` call above which presumably is a relatively "expensive" call. You are reading from state which is a lot of marshaling and unmarshaling, then looking through every position and every perpetual and every price that's relevant. Then doing calculations using `big.Int` and `big.Rat` types

protocol/x/clob/keeper/liquidations.go
```diff
@@ -222,34 +224,49 @@ func (k Keeper) PlacePerpetualLiquidation(
...
 	)
 
-	telemetry.IncrCounterWithLabels(
-		[]string{metrics.Liquidations, metrics.PlacePerpetualLiquidation, metrics.Filled, metrics.BaseQuantums},
-		metrics.GetMetricValueFromBigInt(orderSizeOptimisticallyFilledFromMatchingQuantums.ToBigInt()),
-		[]gometrics.Label{
-			metrics.GetLabelForIntValue(metrics.PerpetualId, int(liquidationOrder.MustGetLiquidatedPerpetualId())),
-		},
-	)
+	// Stat the volume of liquidation orders placed.
```
lucas-dydx, line 258: Suggestion: might be more useful to aggregrate this value per market and then stat it in `LiquidateSubaccountsAgainstOrderbook`. Up to you.

protocol/x/clob/keeper/liquidations.go
```diff
@@ -222,34 +224,49 @@ func (k Keeper) PlacePerpetualLiquidation(
...
 	)
 
-	telemetry.IncrCounter(
-		1,
-		metrics.Liquidations,
-		metrics.PlacePerpetualLiquidation,
-		metrics.Count,
-	)
+	labels := []gometrics.Label{
+		metrics.GetLabelForIntValue(metrics.PerpetualId, int(perpetualId)),
```
BrendanChou, line 235: nit, can be other PR: `GetLabelForIntValue` is so long. Maybe `metrics.LabelInt(...)` `metrics.LabelBool(...)`

protocol/x/clob/keeper/process_operations.go
```diff
@@ -713,6 +687,21 @@ func (k Keeper) PersistMatchDeleveragingToState(
 				err,
 			)
 		}
+
+		if quoteQuantums, err := k.perpetualsKeeper.GetNetNotional(
+			ctx,
+			perpetualId,
+			new(big.Int).SetUint64(fill.FillAmount),
```
lucas-dydx, line 694: Suggestion: use [`ToBigInt`](https://github.com/dydxprotocol/v4-chain/blob/d9fc60643fabfe028afd21e26a0820b3b042d070/protocol/x/subaccounts/types/subaccount.go#L20-L23).

```suggestion
			fill.FillAmount.ToBigInt(),
```

lucas-dydx (approved):

protocol/x/clob/keeper/liquidations.go
```diff
@@ -222,34 +224,49 @@ func (k Keeper) PlacePerpetualLiquidation(
...
+		append(
+			[]gometrics.Label{
+				metrics.GetLabelForBoolValue(
+					metrics.Unfilled,
+					orderSizeOptimisticallyFilledFromMatchingQuantums == 0,
+				),
+				metrics.GetLabelForBoolValue(
+					metrics.FullyFilled,
+					orderSizeOptimisticallyFilledFromMatchingQuantums == liquidationOrder.GetBaseQuantums(),
+				),
```
BrendanChou, line 252: Consider making this a single label like "filled: unfilled/partially/fully" since it's impossible for both these labels to be true or both to be false. Also will make dashboards easier to split out into these three cases. You could also consider a metric of what percentage filled it is and emit it as some sort of distribution or something

protocol/x/clob/keeper/liquidations.go
BrendanChou, line 258: to confirm, there is a metric for liquidations FILLED somewhere else right?

protocol/x/clob/abci.go
```diff
@@ -104,6 +104,12 @@ func EndBlocker(
 
 	// Prune any rate limiting information that is no longer relevant.
 	keeper.PruneRateLimits(ctx)
+
+	telemetry.ModuleSetGauge(
```
lucas-dydx, line 108: Question: do you think this `ModuleSetGauge` function is helpful? IIRC most of our metrics are pre-pended with the module name, and this function [adds the module name as a label](https://github.com/dydxprotocol/cosmos-sdk/blob/93454d9f2fb6f262e63a55641e44c024bd326860/telemetry/wrapper.go#L36) which IMO makes it harder to find module metrics since we mix styles

I prefer [`SetGaugeWithLabels`](https://github.com/dydxprotocol/cosmos-sdk/blob/93454d9f2fb6f262e63a55641e44c024bd326860/telemetry/wrapper.go#L64) for this reason

protocol/x/clob/keeper/liquidations.go
```diff
@@ -222,34 +224,49 @@ func (k Keeper) PlacePerpetualLiquidation(
...
+	if totalQuoteQuantums, err := k.perpetualsKeeper.GetNetNotional(
+		ctx,
+		perpetualId,
+		liquidationOrder.GetBaseQuantums().ToBigInt(),
+	); err == nil {
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.Liquidations, metrics.PlacePerpetualLiquidation, metrics.QuoteQuantums},
+			metrics.GetMetricValueFromBigInt(totalQuoteQuantums),
+			labels,
+		)
```
BrendanChou, line 268: strongly consider some sort of distribution for these metrics

Then you can make analysis like "90% of liquidation orders are less than $10k in size". I think currently this only allows us to say "there were 30 liquidation events with a total amount of $100k" but not know if most of that is from a small number of orders, or if they are relatively similar size

protocol/x/clob/abci.go
lucas-dydx: Is it acceptable that this will only stat the insurance fund balance in the `deliverState`, and never stat the insurance fund balance of the `checkState`?

Would we ever want to know what each validator thinks the insurance fund balance would be after processing their liquidations?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -92,23 +98,14 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
-					err,
-				),
+				"Failed to liquidate subaccount",
+				"liquidationOrder", liquidationOrder,
+				"error", err,
 			)
 			return err
 		}
 
-		// Keep a count of partially and fully filled liquidations for this block.
```
lucas-dydx, line 104: Do you think it would be helpful to stat out the total amount of quantums successfully deleveraged, per market, in each block? Asking because I think statting out each individual value could be hard to aggregrate in datadog since the value would likely fluctuate a lot between each deleveraging event

protocol/x/clob/keeper/deleveraging.go
jayy04: ideally we would panic since this is not really supposed to happen. but don't want to panic for telemetry. 

if this returns an error, I am pretty sure we will have logs somewhere, so leaning towards not logging this here

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -243,6 +246,25 @@ func (k Keeper) ProcessSingleMatch(
...
+
+		labels := []gometrics.Label{
+			metrics.GetLabelForIntValue(metrics.PerpetualId, int(perpetualId)),
+			metrics.GetLabelForBoolValue(metrics.IsBuy, matchWithOrders.TakerOrder.IsBuy()),
+			metrics.GetLabelForBoolValue(metrics.CheckTx, ctx.IsCheckTx()),
+		}
+
+		// Stat quote quantums liquidated.
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.Liquidations, metrics.PlacePerpetualLiquidation, metrics.Filled, metrics.QuoteQuantums},
```
lucas-dydx, line 258: Should you include CLOB module name as a prefix in this stat?

lucas-dydx (approved):

protocol/x/clob/keeper/liquidations.go
jayy04: I think it's more appropriate here since this is the entry point for each liquidation?

protocol/x/clob/keeper/liquidations.go
jayy04: this is also per market (there is a label). 

I think they are effectively the same, and don't think perf is a concern here.

protocol/x/clob/keeper/process_operations.go
jayy04: This field is a uint64, not serializable int so doesn't have a ToBigInt function 😭

protocol/x/clob/abci.go
jayy04: Currently it's on the client side (# of ids sent by daemon). there is nothing on the server side IIRC.

This effectively is the count for the server (application to be more accurate).

protocol/x/clob/keeper/liquidations.go
lucas-dydx: Won't this make it difficult to tell how much liquidation volume there was per block?

protocol/x/clob/keeper/process_single_match.go
jayy04: I think you will need to zoom in to this hour to see the cumulative sum. then zoom in to next hour to see the cumulative sum for the next hour.

protocol/x/clob/keeper/liquidations.go
lucas-dydx: Yeah basically was hoping to also have a stat for liquidation volume per block (along with this stat)

For example, if you wanted to check how many liquidation orders were placed after a large price update

protocol/x/clob/keeper/liquidations.go
jayy04: hmm how would you make this per block on datadog without adding block height as labels?

protocol/x/clob/keeper/process_single_match.go
jayy04: Lol I was actually going to drop `metrics.Liquidations`. 

I think having too many strings here can have perf implications because of string concatenation. so let's minimize the number of strings if possible. let me know what you think.

protocol/x/clob/abci.go
jayy04: We have insurance fund delta info so we can technically sort of calculate from that.

protocol/x/clob/abci.go
jayy04: no preference here. updated.

protocol/x/clob/keeper/liquidations.go
jayy04: Agreed. Will save that for a future PR.

protocol/x/clob/keeper/liquidations.go
jayy04: This metric has `Filled/PartiallyFilled/Unfilled` as label, meant to be a simple count.

I added a separate metric in `process_single_match` for the actual filled quote quantums.

protocol/x/clob/keeper/liquidations.go
jayy04: Initially wanted to avoid multiple if statements, but agree it might be a little bit easier to read with a single label.

updated.

protocol/x/clob/keeper/process_single_match.go
lucas-dydx: Ideally we have the module name as a prefix. Maybe we can add the CLOB prefix to `metrics.PlacePerpetualLiquidation`? Up to you though.

protocol/x/clob/keeper/liquidations.go
lucas-dydx: If you sum up the total amount of quantums deleveraged and emit it as one stat, wouldn't you know that's for one block?

Agree you wouldn't know the exact block height though

protocol/x/clob/keeper/liquidations.go
jayy04: Will look into how `AddSample` from `go-metrics` works

protocol/x/clob/keeper/liquidations.go
jayy04: hmm maybe I am missing something - IMO we don't really care about per block info on human readable graphs, right? we usually care about per minute or per 5 minutes, etc. Our alerts are also set up this way. 

If you want exact block info, they should be emitted as logs (similar to MEV logs).

jayy04: Also I don't think you can zoom in to like 1 second precision on datadog. so even if we aggregate / emit as a single metric, we probably won't be able to look at this block in isolation

protocol/x/clob/keeper/liquidations.go
jayy04: I think some quick math will get us something pretty close. e.g. aggregation over 1 minute period divided by block time.

protocol/lib/metrics/constants.go
jayy04: Got rid of `IsBuy` in favor of buy/sell.

Left `IsLong` as is since it's only used for deleveraging and is more appropriate.

PR #608, opened 2023-10-13
Prevent stateful orders from being replayed twice

jakob-dydx:
explanation in slack: https://dydx-team.slack.com/archives/C03SLFHC3L7/p1697151824501269?thread_ts=1697058864.556479&cid=C03SLFHC3L7

Basically, a Preexisting Stateful Order and Order Removal operation both being in the ops queue can result in duplicate placements in ReplayOperations. 

Order Removals should never precede Preexisting stateful orders in the queue. If a preexisting stateful order placement precedes an order removal, skip the order removal.

### DIFF
--- a/protocol/x/clob/memclob/memclob.go
+++ b/protocol/x/clob/memclob/memclob.go
@@ -861,6 +861,7 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 		}
 	}()
 
+	placedStatefulOrderIds := make(map[types.OrderId]struct{})
 	// Iterate over all provided operations.
 	for _, operation := range localOperations {
 		switch operation.Operation.(type) {
@@ -935,13 +936,14 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 				continue
 			}
 
-			if m.operationsToPropose.IsOrderRemovalInOperationsQueue(statefulOrderPlacement.Order.OrderId) {
-				// If an order removal for this order is found in the ops queue, we have encountered an error. A
-				// stateful order should not be on the book if it hasn't been replayed yet which means it should not
-				// be possible for an order removal to precede a stateful order placement.
+			// Log an error if the order was already placed. We should not see duplicate Preexisting Stateful Orders
+			// and Order Removals (which, when replayed, can result in placements) should never precede a Preexisting
+			// Stateful Order operation.
+			if _, found := placedStatefulOrderIds[*orderId]; found {
 				m.clobKeeper.Logger(ctx).Error(
-					"ReplayOperations: Order removal found in operations queue for a stateful order which was not replayed yet.",
-					metrics.OrderId, statefulOrderPlacement.Order.OrderId,
+					"ReplayOperations: PreexistingStatefulOrder operation for order which was already placed",
+					metrics.OrderId, *orderId,
+					metrics.BlockHeight, ctx.BlockHeight(),
 				)
 				continue
 			}
@@ -952,6 +954,7 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 				ctx,
 				statefulOrderPlacement.Order,
 			)
+			placedStatefulOrderIds[*orderId] = struct{}{}
 			existingOffchainUpdates = m.GenerateOffchainUpdatesForReplayPlaceOrder(
 				ctx,
 				err,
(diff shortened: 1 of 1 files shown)


protocol/x/clob/memclob/memclob.go
```diff
@@ -964,10 +967,17 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 		// Matches are a no-op.
 		case *types.InternalOperation_Match:
 		case *types.InternalOperation_OrderRemoval:
+			orderId := operation.GetOrderRemoval().OrderId
+
+			// Prevent double placement caused by PreexistingStatefulOrder and Order Removal
+			// both existing in local operations.
+			if _, found := placedStatefulOrderIds[orderId]; found {
+				continue
+			}
```
jayy04, line 976: For my understanding - it's valid to have [preexisting placement, removal] (we don't error log here), but not valid to have [removal, preexisting placement], right?

jayy04 (approved):

protocol/x/clob/memclob/memclob.go
```diff
@@ -935,13 +936,14 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
...
+			// Log an error if the order was already placed. We should not see duplicate Preexisting Stateful Orders
+			// and Order Removals (which, when replayed, can result in placements) should never precede a Preexisting
+			// Stateful Order operation.
+			if _, found := placedStatefulOrderIds[*orderId]; found {
 				m.clobKeeper.Logger(ctx).Error(
-					"ReplayOperations: Order removal found in operations queue for a stateful order which was not replayed yet.",
-					metrics.OrderId, statefulOrderPlacement.Order.OrderId,
+					"ReplayOperations: PreexistingStatefulOrder operation for order which was already placed",
+					metrics.OrderId, *orderId,
+					metrics.BlockHeight, ctx.BlockHeight(),
```
lucas-dydx, line 946: This is technically `CheckTx` so it should be the block height of the next block right?

```suggestion
					metrics.BlockHeight, ctx.BlockHeight() + 1,
```

protocol/x/clob/memclob/memclob.go
```diff
@@ -964,10 +967,17 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 		// Matches are a no-op.
 		case *types.InternalOperation_Match:
 		case *types.InternalOperation_OrderRemoval:
+			orderId := operation.GetOrderRemoval().OrderId
+
+			// Prevent double placement caused by PreexistingStatefulOrder and Order Removal
+			// both existing in local operations.
+			if _, found := placedStatefulOrderIds[orderId]; found {
+				continue
```
lucas-dydx, line 975: Wdyt of adding a stat here to see how often this happens?

protocol/x/clob/memclob/memclob.go
```diff
@@ -964,10 +967,17 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 		// Matches are a no-op.
 		case *types.InternalOperation_Match:
 		case *types.InternalOperation_OrderRemoval:
+			orderId := operation.GetOrderRemoval().OrderId
```
lucas-dydx, line 970: Question: do you think it's worth keeping track of which order IDs have already been removed, and then error logging if you see multiple order removals? I believe we currently swallow that error if it occurs.

lucas-dydx (approved):

protocol/x/clob/memclob/memclob.go
```diff
@@ -935,13 +936,14 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
...
 			}
 
-			if m.operationsToPropose.IsOrderRemovalInOperationsQueue(statefulOrderPlacement.Order.OrderId) {
-				// If an order removal for this order is found in the ops queue, we have encountered an error. A
-				// stateful order should not be on the book if it hasn't been replayed yet which means it should not
-				// be possible for an order removal to precede a stateful order placement.
+			// Log an error if the order was already placed. We should not see duplicate Preexisting Stateful Orders
+			// and Order Removals (which, when replayed, can result in placements) should never precede a Preexisting
+			// Stateful Order operation.
+			if _, found := placedStatefulOrderIds[*orderId]; found {
```
jonfung-dydx, line 942: If order removals should never precede a preexisting stateful order operation, should we add this check to ValidateBasic for the `MsgProposedOperation`? We would catch this before we enter DeliverTx.

jakob-dydx: Preexisting stateful order operations are not included in the operations queue that is proposed in MsgProposedOperations. Preexisting stateful orders are kept only as internal operations and are only used during the ReplayOperations step. So I don't think we could add validation for this

jakob-dydx: I was actually wrong about this. Preexisting stateful orders are not proposed, however we should be able to validate that a stateful order is not present as a taker in a match after an Order Removal in the queue. This will always cause the set of operations to fail execution in DeliverTx. Will cut an issue for this, thanks for suggesting @jonfung-dydx

PR #611, opened 2023-10-13
[DEC-2081] Simplify app/process/transactions.go to not use idx map

lcwik:
This migrates panics from during block execution to application start-up.

### Changelist
Simplify app/process/transactions.go to not use idx map.
This migrates panics from during block execution to application start-up.

### Test Plan
Existing tests cover the update. Was also able to remove the idx map specific tests.

### Author/Reviewer Checklist


protocol/app/process/transactions.go
```diff
@@ -4,20 +4,37 @@ import (
...
+	}
+	if minTxsCount != len(txIndicesAndOffsets) {
+		panic("minTxsCount does not match expected count of Txs.")
+	}
+	if lib.ContainsDuplicates(txIndicesAndOffsets) {
+		panic("Duplicate indices/offsets defined for Txs.")
+	}
+	if slices.Min[[]int](txIndicesAndOffsets) != lastOtherTxLenOffset {
+		panic("lastTxLenOffset is not the lowest offset")
+	}
```
BrendanChou, line 36: Add a check that `slices.Max` equals `firstOtherTxOffset`?

protocol/app/process/transactions.go
```diff
@@ -4,20 +4,37 @@ import (
...
-	AcknowledgeBridgesTxType txtype = 2
-	AddPremiumVotesTxType    txtype = 3
-	UpdateMarketPricesTxType txtype = 4
+	minTxsCount                   = 4
+	proposedOperationsTxIndex     = 0
+	updateMarketPricesTxLenOffset = -1
+	addPremiumVotesTxLenOffset    = -2
+	acknowledgeBridgesTxLenOffset = -3
+	lastOtherTxLenOffset          = acknowledgeBridgesTxLenOffset
+	firstOtherTxOffset            = 1
```
BrendanChou, line 18: for naming consistency, this should be `TxIndex` instead of `TxOffset`?

protocol/app/process/transactions.go
```diff
@@ -4,20 +4,37 @@ import (
...
+		updateMarketPricesTxLenOffset,
+		addPremiumVotesTxLenOffset,
+		acknowledgeBridgesTxLenOffset,
+	}
+	if minTxsCount != len(txIndicesAndOffsets) {
+		panic("minTxsCount does not match expected count of Txs.")
+	}
+	if lib.ContainsDuplicates(txIndicesAndOffsets) {
+		panic("Duplicate indices/offsets defined for Txs.")
+	}
```
BrendanChou, line 33: Add an additional duplicates check after adding `minTxsCount` to each of the `LenOffsets`? This ensures that the `LenOffset` variables don't collide with the normal `TxIndex` variables (?)

protocol/app/process/transactions.go
```diff
@@ -4,20 +4,37 @@ import (
...
+	addPremiumVotesTxLenOffset    = -2
+	acknowledgeBridgesTxLenOffset = -3
+	lastOtherTxLenOffset          = acknowledgeBridgesTxLenOffset
+	firstOtherTxOffset            = 1
 )
 
-const (
-	MinTxsCount = 4
-)
+func init() {
```
BrendanChou, line 21: Was going to ask "is this function ever called?", but just learned about this special function in golang. Cool!

protocol/app/process/transactions.go
```diff
@@ -42,77 +59,58 @@ func DecodeProcessProposalTxs(
...
 	// Other txs.
-	allOtherTxs := make([]*OtherMsgsTx, numTxs-len(txTypeToIdx))
-	idx := 0
-	for i, txBytes := range req.Txs {
-		if _, exists := idxToTxType[i]; exists { // skip, because tx is not part of "others".
-			continue
-		}
-
+	allOtherTxs := make([]*OtherMsgsTx, numTxs-minTxsCount)
+	for i, txBytes := range req.Txs[firstOtherTxOffset : numTxs+lastOtherTxLenOffset] {
```
BrendanChou, line 107: do we want some sort of sanity-check that `numTxs-minTxsCount == (numTxs+lastOtherTxLenOffset) - firstOtherTxOffset` ?

Is it preferable to continue in this case?

protocol/app/process/transactions.go
lcwik: I used `TxIndex` for cases where it is a raw index while `TxLenOffset` to indicate its an offset of the `len` of the `Tx` list. I can swap to just use `TxOffset`. WDYT?

protocol/app/process/transactions.go
lcwik: added `minTxsCount != firstOtherTxOffset-lastOtherTxLenOffset` to init for our sanity check.

protocol/app/process/transactions.go
BrendanChou: I agree with the naming convention, I am talking in particular about `firstOtherTxOffset` which I think is really an Index and not an Offset of some sort.

For example, imagine adding another mandatory transaction that must come second. Then we would change the `firstOtherTxOffset` variable to be `2`.

protocol/app/process/transactions.go
```diff
@@ -34,6 +34,24 @@ func init() {
...
+	}
+	if minTxsCount != len(txIndicesForMinTxsCount) {
+		panic("minTxsCount does not match expected count of Txs.")
+	}
+	if lib.ContainsDuplicates(txIndicesForMinTxsCount) {
+		panic("Overlapping indices and offsets defined for Txs.")
+	}
+	if minTxsCount != firstOtherTxOffset-lastOtherTxLenOffset {
+		panic("Unexpected gap between firstOtherTxOffset and lastOtherTxLenOffset which is greater than minTxsCount")
+	}
```
BrendanChou, line 54: Wait won't this panic with our current values?
Also I think use comparison rather than equality. That is:
```
if minTxsCount < firstOtherTxOffset-lastOtherTxLenOffset {
```

BrendanChou (approved):
✅ besides my comment above
https://github.com/dydxprotocol/v4-chain/pull/611#discussion_r1358517604

protocol/app/process/transactions.go
```diff
@@ -4,20 +4,55 @@ import (
...
-const (
-	MinTxsCount = 4
-)
+func init() {
+	txIndicesAndOffsets := []int{
+		proposedOperationsTxIndex,
+		updateMarketPricesTxLenOffset,
+		addPremiumVotesTxLenOffset,
+		acknowledgeBridgesTxLenOffset,
+	}
```
ttl33, line 27: nit: prefer the ordering to match the order of txs that we expect to see

```suggestion
		acknowledgeBridgesTxLenOffset,
		addPremiumVotesTxLenOffset,
		updateMarketPricesTxLenOffset,
	}
```

protocol/app/process/transactions.go
```diff
@@ -4,20 +4,55 @@ import (
...
+		panic("lastTxLenOffset is not the lowest offset")
+	}
+	if slices.Max[[]int](txIndicesAndOffsets)+1 != firstOtherTxOffset {
+		panic("firstOtherTxOffset is <= the maximum offset")
+	}
+	txIndicesForMinTxsCount := []int{
+		proposedOperationsTxIndex,
+		updateMarketPricesTxLenOffset + minTxsCount,
+		addPremiumVotesTxLenOffset + minTxsCount,
+		acknowledgeBridgesTxLenOffset + minTxsCount,
```
ttl33, line 44: nit: same as above

```suggestion
		acknowledgeBridgesTxLenOffset + minTxsCount,
		addPremiumVotesTxLenOffset + minTxsCount,
		updateMarketPricesTxLenOffset + minTxsCount,
```

ttl33 (approved):

protocol/app/process/transactions.go
lcwik: SGTM

BrendanChou (approved):

PR #613, opened 2023-10-13
[CLOB-959] loosen deliver tx liquidation validation and allow arbitra…

jayy04:
…ry perpetual

### Changelist

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/liquidations.go
+++ b/protocol/x/clob/keeper/liquidations.go
@@ -161,20 +161,45 @@ func (k Keeper) MaybeGetLiquidationOrder(
 	)
 
 	// The subaccount is liquidatable. Get the perpetual position and position size to liquidate.
-	perpetualId, baseQuantumsBig, err := k.GetPerpetualPositionToLiquidate(ctx, subaccountId)
+	perpetualId, err := k.GetPerpetualPositionToLiquidate(ctx, subaccountId)
+	if err != nil {
+		return nil, err
+	}
+
+	return k.GetLiquidationOrderForPerpetual(
+		ctx,
+		subaccountId,
+		perpetualId,
+	)
+}
+
+// GetLiquidationOrderForPerpetual returns a liquidation order for a subaccount
+// given a perpetual ID and position size delta.
+func (k Keeper) GetLiquidationOrderForPerpetual(
+	ctx sdk.Context,
+	subaccountId satypes.SubaccountId,
+	perpetualId uint32,
+) (
+	liquidationOrder *types.LiquidationOrder,
+	err error,
+) {
+	deltaQuantums, err := k.GetLiquidatablePositionSizeDelta(
+		ctx,
+		subaccountId,
+		perpetualId,
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	// Get the fillable price of the liquidation order in subticks.
-	deltaQuantumsBig := baseQuantumsBig.Neg(baseQuantumsBig)
-	fillablePriceRat, err := k.GetFillablePrice(ctx, subaccountId, perpetualId, deltaQuantumsBig)
+	fillablePriceRat, err := k.GetFillablePrice(ctx, subaccountId, perpetualId, deltaQuantums)
 	if err != nil {
 		return nil, err
 	}
 
 	// Calculate the fillable price.
-	isLiquidatingLong := deltaQuantumsBig.Sign() == -1
+	isLiquidatingLong := deltaQuantums.Sign() == -1
 	clobPair := k.mustGetClobPairForPerpetualId(ctx, perpetualId)
 	fillablePriceSubticks := k.ConvertFillablePriceToSubticks(
 		ctx,
@@ -184,7 +209,7 @@ func (k Keeper) MaybeGetLiquidationOrder(
 	)
 
 	// Create the liquidation order.
-	absBaseQuantums := deltaQuantumsBig.Abs(deltaQuantumsBig)
+	absBaseQuantums := deltaQuantums.Abs(deltaQuantums)
 	liquidationOrder = types.NewLiquidationOrder(
 		subaccountId,
 		clobPair,
@@ -664,15 +689,13 @@ func (k Keeper) GetLiquidationInsuranceFundDelta(
 
 // GetPerpetualPositionToLiquidate determines which position and position size to liquidate on the
 // passed-in subaccount (after accounting for the `update`). It will return the perpetual id that
-// will be used for liquidating the perpetual position and the number of quantums to liquidate
-// from the perpetual position (positive if long, negative if short).
+// will be used for liquidating the perpetual position
 // This function returns an error if the subaccount has no perpetual positions to liquidate.
 func (k Keeper) GetPerpetualPositionToLiquidate(
 	ctx sdk.Context,
 	subaccountId satypes.SubaccountId,
 ) (
 	perpetualId uint32,
-	quantums *big.Int,
 	err error,
 ) {
 	// Fetch the subaccount from state.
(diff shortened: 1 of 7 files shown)


protocol/x/clob/keeper/liquidations.go
```diff
@@ -106,28 +106,28 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+		_, err := k.MaybeDeleverageSubaccount(ctx, subaccountId, perpetualId, deltaQuantums)
+		if err != nil {
+			k.Logger(ctx).Error(
+				"Failed to deleverage subaccount.",
+				"subaccount", subaccountId,
+				"perpetualId", perpetualId,
+				"deltaQuantums", deltaQuantums,
+				"error", err,
+			)
+			return err
```
jayy04, line 130: moved deleveraging to a separate loop for readability

protocol/x/clob/keeper/liquidations_test.go
```diff
@@ -3829,7 +3828,9 @@ func TestGetPerpetualPositionToLiquidate(t *testing.T) {
 			},
 
 			expectedClobPair: constants.ClobPair_Btc,
-			expectedQuantums: constants.PerpetualPosition_OneTenthBTCLong.GetBigQuantums(),
+			expectedQuantums: new(big.Int).Neg(
+				constants.PerpetualPosition_OneTenthBTCLong.GetBigQuantums(),
+			),
```
jayy04, line 3833: `GetPerpetualPositionToLiquidate` used to return position size in base quantums (not delta). 

The new `GetLiquidatablePositionSizeDelta` function returns delta, so flipping all the signs here.

protocol/x/clob/keeper/liquidations_test.go
```diff
@@ -4229,27 +4232,7 @@ func TestGetPerpetualPositionToLiquidate(t *testing.T) {
...
-			},
-			liquidationConfig: constants.LiquidationsConfig_No_Limit,
-
-			clobPairs: []types.ClobPair{
-				constants.ClobPair_Btc,
-				constants.ClobPair_Eth,
-			},
-
-			expectedClobPair: constants.ClobPair_Eth,
-			expectedQuantums: constants.PerpetualPosition_OneTenthEthLong.GetBigQuantums(),
```
jayy04, line 4252: not relevant anymore since selection is random

protocol/x/clob/keeper/process_operations.go
```diff
@@ -532,7 +532,20 @@ func (k Keeper) PersistMatchLiquidationToState(
...
+	}
+	if !isLiquidatable {
+		return types.ErrSubaccountNotLiquidatable
+	}
+
+	takerOrder, err := k.GetLiquidationOrderForPerpetual(
+		ctx,
+		matchLiquidation.Liquidated,
+		matchLiquidation.PerpetualId,
+	)
```
jayy04, line 548: now bypasses `GetPerpetualPositionToLiquidate`

protocol/x/clob/keeper/liquidations.go
```diff
@@ -106,28 +106,28 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+
+	// For each unfilled liquidation, attempt to deleverage the subaccount.
+	for i := 0; i < int(k.Flags.MaxDeleveragingAttemptsPerBlock) && i < len(unfilledLiquidations); i++ {
+		liquidationOrder := unfilledLiquidations[i]
+
+		subaccountId := liquidationOrder.GetSubaccountId()
+		perpetualId := liquidationOrder.MustGetLiquidatedPerpetualId()
+		deltaQuantums := liquidationOrder.GetDeltaQuantums()
+
+		_, err := k.MaybeDeleverageSubaccount(ctx, subaccountId, perpetualId, deltaQuantums)
```
BrendanChou, line 121: btw one thing in 0.3.X I did was use the entire position size as `deltaQuantums` to deleverage as fast as possible. This is preferable to using the liquidation order because 1. we already have a limit on the number of accounts iterated for deleveraging and 2. there is no orderbook impact to really worry about

protocol/x/clob/keeper/liquidations.go
```diff
@@ -681,7 +704,11 @@ func (k Keeper) GetPerpetualPositionToLiquidate(
 
 	var perpetualPosition *satypes.PerpetualPosition
 
-	for _, position := range subaccount.PerpetualPositions {
+	numPositions := len(subaccount.PerpetualPositions)
+	indexOffset := k.GetPseudoRand(ctx).Intn(numPositions)
+
+	for i := 0; i < len(subaccount.PerpetualPositions); i++ {
```
BrendanChou, line 710: `len(subaccount.PerpetualPositions)` -> `numPositions`

protocol/x/clob/keeper/liquidations.go
```diff
@@ -161,20 +161,45 @@ func (k Keeper) MaybeGetLiquidationOrder(
...
+		ctx,
+		subaccountId,
+		perpetualId,
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	// Get the fillable price of the liquidation order in subticks.
-	deltaQuantumsBig := baseQuantumsBig.Neg(baseQuantumsBig)
```
BrendanChou, line 170: not immediately clear to me why this was removed

protocol/x/clob/keeper/liquidations.go
```diff
@@ -694,33 +721,56 @@ func (k Keeper) GetPerpetualPositionToLiquidate(
...
 	if perpetualPosition == nil {
 		return 0,
-			nil,
 			errorsmod.Wrapf(
 				types.ErrNoPerpetualPositionsToLiquidate,
 				"Subaccount ID: %v",
 				subaccount.Id,
 			)
 	}
+	return perpetualPosition.PerpetualId, nil
```
BrendanChou, line 730: instead of returning here, just return where there is `break` above. Then you can also remove the `if perpetualPosition == nil` conditional

protocol/x/clob/keeper/process_operations.go
```diff
@@ -532,7 +532,20 @@ func (k Keeper) PersistMatchLiquidationToState(
...
 ) error {
-	takerOrder, err := k.MaybeGetLiquidationOrder(ctx, matchLiquidation.Liquidated)
+	// If the subaccount is not liquidatable, do nothing.
+	isLiquidatable, err := k.IsLiquidatable(ctx, matchLiquidation.Liquidated)
+	if err != nil {
+		return err
+	}
+	if !isLiquidatable {
+		return types.ErrSubaccountNotLiquidatable
+	}
```
BrendanChou, line 542: nit: this could be a function like `k.EnsureIsLiquidatable` that return an `error` since it's used twice. Up to you

protocol/x/clob/keeper/liquidations.go
jayy04: ah this was moved to `GetLiquidatablePositionSizeDelta`.

GetPerpetualPositionToLiquidate used to return position size in base quantums. and delta was calculated outside of the function.

protocol/x/clob/keeper/liquidations.go
jayy04: makes sense. Updated!

BrendanChou (approved):

protocol/x/clob/keeper/liquidations.go
```diff
@@ -664,63 +700,82 @@ func (k Keeper) GetLiquidationInsuranceFundDelta(
...
+	ctx sdk.Context,
+	subaccountId satypes.SubaccountId,
+	perpetualId uint32,
+) (
+	deltaQuantums *big.Int,
+	err error,
+) {
+	subaccount := k.subaccountsKeeper.GetSubaccount(ctx, subaccountId)
+	perpetualPosition, exists := subaccount.GetPerpetualPositionForId(perpetualId)
+	if !exists {
```
lucas-dydx, line 750: IMO this should be a panic, since I think [this logic](https://github.com/dydxprotocol/v4-chain/blob/1a8bafb3b294df1188db5c879435d0486167fa62/protocol/x/clob/abci.go#L188) in `PrepareCheckState` will panic on this case. Up to you

lucas-dydx (approved):

protocol/x/clob/keeper/liquidations.go
jayy04: then we can just panic upstream, right?

PR #618, opened 2023-10-13
[chore] Add `PrepareCheckState` metrics

lucas-dydx:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/x/clob/abci.go
```diff
@@ -156,11 +164,26 @@ func PrepareCheckState(
 	)
 
 	// 3. Place all stateful order placements included in the last block on the memclob.
+	// Note telemetry is measured outside of the function call because `PlaceStatefulOrdersFromLastBlock`
+	// is called within `PlaceConditionalOrdersTriggeredInLastBlock`.
+	startPlaceLongTermOrders := time.Now()
```
lucas-dydx, line 169: Avoided adding this within `PlaceStatefulOrdersFromLastBlock` since I thought the double counting would be confusing.

protocol/x/clob/abci.go
```diff
@@ -120,6 +121,13 @@ func PrepareCheckState(
 	keeper *keeper.Keeper,
 	liquidatableSubaccountIds *liquidationtypes.LiquidatableSubaccountIds,
 ) {
+	defer telemetry.MeasureSince(
```
lucas-dydx, line 124: Added latency metrics on every step of this function except two function calls:
- `GetProcessProposerMatchesEvents`
- `SetMemclobGauges`

Let me know if anyone wants me to add latency metrics to those.

protocol/x/clob/keeper/liquidations.go
```diff
@@ -109,8 +123,14 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
 		}
 	}
+	telemetry.MeasureSince(
+		startPlaceLiquidationOrders,
+		types.ModuleName,
+		metrics.ClobLiquidateSubaccounts_PlaceLiquidations,
+	)
 
 	// For each unfilled liquidation, attempt to deleverage the subaccount.
+	startDeleverageSubaccounts := time.Now()
```
lucas-dydx, line 133: LMK if anyone can think of a cleaner way to do this. Could potentially abstract these loops into functions so a `defer` statement could be used to measure latency.

protocol/x/clob/memclob/memclob.go
```diff
@@ -848,6 +848,20 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
...
+		types.ModuleName,
+		metrics.MemClobReplayOperations,
+		metrics.Latency,
+	)
+
+	telemetry.SetGauge(
+		float32(len(localOperations)),
+		types.ModuleName,
+		metrics.MemClobReplayOperations,
+		metrics.OperationsQueueLength,
```
lucas-dydx, line 862: Note: this isn't as useful as the number of each operation type since match operations are a no-op when replaying.

LMK if anyone prefers this stat is more granular and counts each operation type.

protocol/x/clob/keeper/liquidations.go
```diff
@@ -48,6 +54,8 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 	if numSubaccounts > 0 {
 		indexOffset = pseudoRand.Intn(numSubaccounts)
 	}
+
+	startGetLiquidationOrders := time.Now()
```
lucas-dydx, line 58: Avoided using gauges for number of subaccounts liquidated / deleveraged since those are already added in the individual liquidation / deleveraging functions.

protocol/x/clob/memclob/memclob.go
```diff
@@ -1073,6 +1087,13 @@ func (m *MemClobPriceTimePriority) RemoveAndClearOperationsQueue(
 	ctx sdk.Context,
 	localValidatorOperationsQueue []types.InternalOperation,
 ) {
+	defer telemetry.MeasureSince(
```
lucas-dydx, line 1090: LMK if anyone prefers I use `ModuleMeasureSince` instead of `MeasureSince`. IMO think it's clearer to not use module name as a label.

Also can update stat names like `RemoveAndClearOperationsQueue` to include the module name as a prefix to avoid unnecessary string concatenation, but these stats aren't in the hot path so I _think_ this is fine.

protocol/x/clob/keeper/liquidations.go
```diff
@@ -67,6 +76,12 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 
 		liquidationOrders = append(liquidationOrders, *liquidationOrder)
 	}
+	telemetry.MeasureSince(
+		startGetLiquidationOrders,
```
lucas-dydx, line 80: LMK if you all think this metric name is a misnomer, since I know this doesn't technically "get" the liquidation orders since they're recreated when placing them.

protocol/x/clob/abci.go
```diff
@@ -156,11 +164,26 @@ func PrepareCheckState(
...
+		types.ModuleName,
+		metrics.PlaceLongTermOrdersFromLastBlock,
+		metrics.Latency,
+	)
+	telemetry.SetGauge(
+		float32(len(processProposerMatchesEvents.PlacedLongTermOrderIds)),
+		types.ModuleName,
+		metrics.PlaceLongTermOrdersFromLastBlock,
+		metrics.Count,
+	)
```
jayy04, line 186: nit: I personally would put these inside `PlaceStatefulOrdersFromLastBlock`

protocol/x/clob/keeper/liquidations.go
jayy04: should we just modify the sorting latency below to include this? since this is only used for sorting

protocol/x/clob/keeper/liquidations.go
jayy04: yeah I think breaking this down to individual functions would be cleaner since this function is kind of long. 

no need to block on this though, we can do that separately

protocol/lib/metrics/constants.go
```diff
@@ -100,6 +100,11 @@ const (
 	CancelShortTermOrder                         = "cancel_short_term_order"
 	CancelStatefulOrder                          = "cancel_stateful_order"
 	ClobPairId                                   = "clob_pair_id"
+	ClobPrepareCheckState                        = "clob_prepare_check_state"
+	ClobLiquidateSubaccountsAgainstOrderbook     = "clob_liquidate_subaccounts_against_orderbook"
+	ClobLiquidateSubaccounts_GetLiquidations     = "clob_liquidate_subaccounts_against_orderbook_get_liquidations"
+	ClobLiquidateSubaccounts_PlaceLiquidations   = "clob_liquidate_subaccounts_against_orderbook_place_liquidations"
+	ClobLiquidateSubaccounts_Deleverage          = "clob_liquidate_subaccounts_against_orderbook_deleverage"
```
jayy04, line 107: I think we can drop the clob prefix. it's kind of verbose now

jayy04 (approved):

protocol/x/clob/abci.go
lucas-dydx: Personally I avoided it since I wanted to differentiate between long-term order placements and stateful order placements. Let me know if you disagree though and I can change it.

protocol/x/clob/keeper/liquidations.go
lucas-dydx: Agreed, created an M6 ticket [here](https://linear.app/dydx/issue/CLOB-963/break-up-liquidatesubaccountsagainstorderbook-steps-into-functions) to consolidate this logic. LMK if you think it should be M5 and I can move it up + potentially take it on immediately

protocol/x/clob/keeper/liquidations.go
lucas-dydx: Agreed, created an M6 ticket [here](https://linear.app/dydx/issue/CLOB-963/break-up-liquidatesubaccountsagainstorderbook-steps-into-functions). LMK if you think this should be M5

protocol/lib/metrics/constants.go
lucas-dydx: If it's fine I'm going to leave `ClobPrepareCheckState` +  `ClobLiquidateSubaccountsAgainstOrderbook ` as variable names since removing the `Clob` prefix causes them to conflict with function names.

I'll remove it from the other ones though + remove it from all stat names since agreed it's redundant

protocol/x/clob/keeper/liquidations.go
```diff
@@ -128,6 +151,12 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 			return err
 		}
 	}
+	telemetry.MeasureSince(
+		startDeleverageSubaccounts,
+		types.ModuleName,
+		metrics.LiquidateSubaccounts_Deleverage,
+		metrics.Latency,
+	)
```
BrendanChou, line 159: nit: this should be part of the following paragraph, no the one above

BrendanChou (approved):

PR #603, opened 2023-10-12
[CORE-658] - Telemetry updates, remove commented out test file

clemire:
### Changelist
Add telemetry for new msg service methods. Remove unused test file.

### Test Plan
[Describe how this PR was tested (if applicable)]

### Author/Reviewer Checklist
- [ ] If this change affects functionality (features, bug fixes, breaking changes, etc.), update `protocol/CHANGELOG.md` and/or `indexer/CHANGELOG.md` appropriately.
- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [ ] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [ ] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
- [ ] Manually add any of the following labels: `refactor`, `chore`, `bug`.

PR #625, opened 2023-10-13
[CORE-656] add test cases, metrics, and comments to sending module

tqin7:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/x/sending/keeper/msg_server_create_transfer.go
```diff
@@ -97,7 +100,7 @@ func (k msgServer) WithdrawFromSubaccount(
 	return &types.MsgWithdrawFromSubaccountResponse{}, nil
 }
 
-// SendFromModuleToAccount sends a coin from a module to an account.
+// SendFromModuleToAccount sends coins from a module to an account.
```
clemire, line 103: To confirm: (I'm not super familiar with x/sender) - this is to an account and not a subaccount?

protocol/x/sending/keeper/msg_server_create_transfer.go
```diff
@@ -113,8 +116,22 @@ func (k msgServer) SendFromModuleToAccount(
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	if err := k.Keeper.SendFromModuleToAccount(ctx, msg); err != nil {
+		telemetry.IncrCounterWithLabels(
```
clemire, line 119: This may not be a big save for total lines of code, but FYI you can use named parameters with a defer pattern to have a single telemetry call in code, see [example](https://github.com/dydxprotocol/v4-chain/pull/603/files)

protocol/x/sending/keeper/msg_server_create_transfer.go
clemire: (Is this also an x/bank recipient?)

clemire (approved):

protocol/x/sending/keeper/msg_server_create_transfer.go
tqin7: yep this is an x/bank recipient

protocol/x/sending/keeper/msg_server_create_transfer.go
tqin7: Ah I see! Will keep that in mind for the future.

PR #627, opened 2023-10-13
Remove pricefeed config toml entirely.

clemire:
### Changelist
- Remove pricefeed config toml from protocol entirely:
  - toml generation functionality, including post-init hook to create file on protocol init
  - all code dependencies
- Update daemon to configure itself from the static exchange startup config in code.
- Rename `ExchangeStartupConfig` to `ExchangeQueryConfig`, which IMO is more specific and self-documenting.

As part of CORE-469, I plan to follow up with a PR that allows users to pass in an optional pricefeed daemon flag populated with json config that disables and/or changes query configuration of particular exchanges.

### Test Plan
test-ci and container-test pass locally.

### Author/Reviewer Checklist


protocol/cmd/dydxprotocold/cmd/init.go
clemire, line 1: This file added a hook to the init command that generated pricefeed config toml.

protocol/daemons/configs/default_pricefeed_exchange_config.go
clemire, line 1: This directory contained logic for generating and reading pricefeed config tomls.

protocol/daemons/pricefeed/client/client.go
```diff
@@ -117,7 +117,7 @@ func (c *Client) start(ctx context.Context,
 	appFlags appflags.Flags,
 	logger log.Logger,
 	grpcClient daemontypes.GrpcClient,
-	exchangeIdToStartupConfig map[types.ExchangeId]*types.ExchangeStartupConfig,
+	exchangeIdToStartupConfig map[types.ExchangeId]*types.ExchangeQueryConfig,
```
clemire, line 120: type rename: `ExchangeStartupConfig` -> `ExchangeQueryConfig`

protocol/testing/containertest/config/pricefeed_exchange_config.toml
clemire, line 1: Remove pricefeed config from containertests.

protocol/testing/genesis.sh
```diff
@@ -1476,7 +1476,7 @@ function update_genesis_use_test_exchange() {
 	# All remaining markets can just use the LINK ticker so the daemon will start. All markets must have at least 1
 	# exchange. With only one exchange configured, there should not be enough prices to meet the minimum exchange
 	# count, and these markets will not have index prices.
-	for market_idx in {3..33}
+	for market_idx in {3..34}
```
clemire, line 1479: dydx token was not accounted for here when I added it a few weeks ago, this caused problems because it mentions non-existent markets as adjustment markets. Was not a problem before because we had only enabled the test exchange for container tests.

RN, container tests are running a daemon with all exchanges. I can update this once I add a flag that allows users to disable exchanges.

teddyding: Could run `make update-sample-pregenesis` to check if prod_sample_genesis is changed as a result of this?

protocol/daemons/pricefeed/client/client.go
teddyding: Probably should rename variable `exchangeIdToStartupConfig` accordingly?

protocol/app/app.go
```diff
@@ -603,7 +602,7 @@ func New(
 
 		// Non-validating full-nodes have no need to run the price daemon.
 		if !appFlags.NonValidatingFullNode && daemonFlags.Price.Enabled {
-			exchangeStartupConfig := configs.ReadExchangeStartupConfigFile(homePath)
+			exchangeStartupConfig := constants.StaticExchangeQueryConfig
```
teddyding, line 605: For my understanding:
- This PR is no-op for the actual value of `exchangeStartupConfig` right? Meaning it's still initialized as the same value as before, just not read from a config file
- Is there currently any sanity check for the `exchangeStartupConfig` being populated correctly?

ttl33 (approved):
approved, pls address Teddy's comments and get his approval before merging!

BrendanChou (approved):

protocol/app/app.go
clemire: Yes and yes. See [this file](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/daemons/pricefeed/client/constants/static_exchange_startup_config_test.go).

protocol/testing/genesis.sh
clemire: No updates!

protocol/daemons/pricefeed/client/client.go
clemire: Ok, renamed everywhere!

teddyding (approved):

PR #638, opened 2023-10-17
update base position notional in genesis.sh

yujin512:
update the genesis.sh file to reflect recommendation on the base position notional values for midcap and long tail liquidity tiers  https://dydx-team.slack.com/archives/C04BNGM82HM/p1697493986734849?thread_ts=1697487833.033209&cid=C04BNGM82HM

PR #640, opened 2023-10-17
update base position notional in genesis.sh

yujin512:
update the genesis.sh file to reflect recommendation on the base position notional values for midcap and long tail liquidity tiers https://dydx-team.slack.com/archives/C04BNGM82HM/p1697493986734849?thread_ts=1697487833.033209&cid=C04BNGM82HM

same changes as in this pr but verified commit https://github.com/dydxprotocol/v4-chain/pull/638

### DIFF
--- a/protocol/scripts/genesis/sample_pregenesis.json
+++ b/protocol/scripts/genesis/sample_pregenesis.json
@@ -1,72 +1,44 @@
 {
-  "genesis_time": "2023-12-31T00:00:00Z",
-  "chain_id": "dydx-sample-1",
-  "initial_height": "1",
-  "consensus_params": {
-    "block": {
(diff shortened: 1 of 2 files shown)


IvoCrnkovic:
teddy indicates that in some cases you have to run make update-sample-pregenesis and commit the resulting changes after you update genesis.sh in [this thread](https://dydx-team.slack.com/archives/C04BNGM82HM/p1697563625631579?thread_ts=1697559168.985669&cid=C04BNGM82HM)

@teddyding for our learning can you clarify if thats required here? IE even if it would do nothing, am I right that YJ should do that as its best practice?

tqin7 (approved):

PR #636, opened 2023-10-17
Update default slashing params

teddyding:
### Changelist
Update default slashing params. Latest changes are discussed [here](https://www.notion.so/dydx/AC-Priv-Genesis-Parameters-d2321636dd494ee49cc95b7825cbbc98?pvs=4#1fae11680769402d89167d9e4c43a426)

### Test Plan
Ran `make update-sample-pregenesis` to update `sample_pregenesis.json`

### Author/Reviewer Checklist
- [x] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [x] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [x] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [x] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
- [x] Manually add any of the following labels: `refactor`, `chore`, `bug`.

PR #634, opened 2023-10-16
[CLOB-947] Split TEST-USD liquidity tier with differing base_position_notional

jonfung-dydx:

PR #632, opened 2023-10-16
fix edge case where position is closed before deleveraging

jayy04:
### Changelist
- technically it's possible that the position is closed before deleveraging is attempted. so this PR is meant to address that extreme edge case.

### Test Plan
[Describe how this PR was tested (if applicable)]

### Author/Reviewer Checklist
- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [ ] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [ ] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
- [ ] Manually add any of the following labels: `refactor`, `chore`, `bug`.

PR #628, opened 2023-10-16
Update `pregenesis.sh` script for testnet4

teddyding:
### Changelist
Update pregenesis.sh script for testnet 4
Generate pregenesis file [here](https://github.com/dydxprotocol/v4-testnets/pull/90)

### Test Plan
N/A

PR #642, opened 2023-10-17
Add initial simple script to generate release notes

roy-dydx:
### Changelist
Add a script that generates a markdown list of changes between two commits.

### Test Plan
Tried the script on some examples:

`GITHUB_TOKEN=<redacted> python3 main.py --old release/protocol/v1.x --new main --path protocol` gives:
- [DEC-2082] add + clarify comments ([#495](https://github.com/dydxprotocol/v4-chain/pull/495))
- [CLOB-947] Split TEST-USD liquidity tier with differing base_position_notional ([#634](https://github.com/dydxprotocol/v4-chain/pull/634))
- Update default slashing params ([#636](https://github.com/dydxprotocol/v4-chain/pull/636))
- Conditionally info log in PlaceOrder DeliverTx if collat check fails ([#635](https://github.com/dydxprotocol/v4-chain/pull/635))
- [CLOB-177] - long term e2e tests ([#457](https://github.com/dydxprotocol/v4-chain/pull/457))
- fix edge case where position is closed before deleveraging ([#632](https://github.com/dydxprotocol/v4-chain/pull/632))
- [refactor] Remove duplicate CLOB `PrepareCheckState` latency metric ([#631](https://github.com/dydxprotocol/v4-chain/pull/631))
- Test exchange - use public dummy http server ([#630](https://github.com/dydxprotocol/v4-chain/pull/630))
- Update `pregenesis.sh` script for testnet4 ([#628](https://github.com/dydxprotocol/v4-chain/pull/628))
- [CLOB-947] New liquidity tier for TEST-USD ([#626](https://github.com/dydxprotocol/v4-chain/pull/626))
- [DEC-2081] Add tests for app/process/proposed_operations.go ([#609](https://github.com/dydxprotocol/v4-chain/pull/609))

`GITHUB_TOKEN=<redacted> python3 main.py --old a814748dfb39dc70302becbca95c19606e3bab8e --new release/indexer/v0.4.x --path indexer` gives (this tests cherry-picks that don't have associated PR):
- Revert "[IND-412]: Correctly set order status for all order types (#577)" ([0dbc3ee](https://github.com/dydxprotocol/v4-chain/commit/0dbc3eeeedc47a60c8ff309c80d8531e087da12f))
- Revert "Fix FOK orders so status is always FILLED (#594)" ([d720ed8](https://github.com/dydxprotocol/v4-chain/commit/d720ed85f227197a0569beb37c0a78964f21ef07))

### Author/Reviewer Checklist


docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
+# Usage:
+# GITHUB_TOKEN=<token> python main.py --old <old> --new <new> --path <path>
```
ttl33, line 2:
```suggestion
# GITHUB_TOKEN=<token> python main.py --old <old commit> --new <new commit> --path <path>
```

docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if len(pulls) > 1:
+        # I don't know when a commit would have more than one PR. If this happens, report a bug.
+        raise NotImplementedError("Commit unexpectedly had more than one associated PR. Please file a bug.")
+    elif len(pulls) == 1:
+        # If there is an associated PR, use PR title and link to PR.
+        pull = pulls[0]
+        return "- %s ([#%s](%s))" % (pull["title"], pull["number"], pull["html_url"])
+    else:
+        # If there is no associated PR, use first line of commit message and link to commit.
```
ttl33, line 34: when would we hit this case?

docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
...
+    ret = []
+    for commit in commits:
+        ret.append(commit_to_entry(session, commit))
+    return "\n".join(ret)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--old", help="Oldest commit, exclusive")
+    parser.add_argument("--new", help="Newest commit, inclusive")
```
ttl33, line 64: is there a way to sanity check that new > old commit?

docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
+# Usage:
```
ttl33, line 1: I wonder if there are other existing tools that basically do this job instead of us maintaining our own script. Have we looked at 3rd party solutions?

It could be useful to use 3rd party solutions if they add new extensions / reduce our maintenance burden.

docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
...
+    r = session.get(LIST_COMMIT_PULLS_ENDPOINT % commit_json["sha"])
+    pulls = r.json()
+
+    if len(pulls) > 1:
+        # I don't know when a commit would have more than one PR. If this happens, report a bug.
+        raise NotImplementedError("Commit unexpectedly had more than one associated PR. Please file a bug.")
+    elif len(pulls) == 1:
+        # If there is an associated PR, use PR title and link to PR.
+        pull = pulls[0]
+        return "- %s ([#%s](%s))" % (pull["title"], pull["number"], pull["html_url"])
```
ttl33, line 32: is it useful to also pull the labels?

ttl33, line 32: Now that we will rely on the `title` more for release notes, is there a `title` format that would make the release note preparer's job easy?

For example, prefixing `(feat/xyz)` or `(bug)` might make it easier to reason about the PR from just the title.

We can also experiment w different title format as we go

docs/release_note_generator/main.py
```diff
@@ -0,0 +1,71 @@
...
+# Only commits that change files in <path> will be included. A github token is required to avoid rate limits.
+
+import argparse
+import os
+import requests
+
+GITHUB_TOKEN_ENV_VAR = "GITHUB_TOKEN"
+COMMITS_ENDPOINT = "https://api.github.com/repos/dydxprotocol/v4-chain/commits"
+GET_COMMIT_ENDPOINT = "https://api.github.com/repos/dydxprotocol/v4-chain/commits/%s"
+LIST_COMMIT_PULLS_ENDPOINT = "https://api.github.com/repos/dydxprotocol/v4-chain/commits/%s/pulls"
```
ttl33, line 14: Now that `v4-chain` includes indexer and js/py client code, do we want to keep track of non-protocol changes too? I am assuming that this will pull all changes including indexer + client changes

docs/release_note_generator/main.py
roy-dydx: In the future I was thinking to use labels to organize these into sections. For now I don't think it would be very useful, so it'll have to be done manually.

docs/release_note_generator/main.py
roy-dydx: Added comment.

docs/release_note_generator/main.py
roy-dydx: No real need. It would just give you empty results. I'm planning on adding this to release workflow anyways so people don't have to run it themselves.

docs/release_note_generator/main.py
roy-dydx: Being consistent about labels will help us organize it into sections. And also document things like state-breaking.

I think formatting the same way will be a little helpful, but it's not too difficult to go through the draft and reword everything. I care more that titles contain enough context so that most of the time, it won't be necessary to click into the PR to find out what it's doing.

docs/release_note_generator/main.py
ttl33: I am still wondering when we'd hit the case where there's a no associated PR? 

I think we should prevent contributors from pushing commits directly without the review process. I might be missing some context here tho. Can you elaborate?

docs/release_note_generator/main.py
ttl33: Can we create a ticket and add a TODO here to do that? What you outlined seems useful!

docs/release_note_generator/main.py
ttl33: Got it. Can you add more comments around the path? For example, add an example to use `/protocol` to only grab protocol related commits

ttl33 (approved):

docs/release_note_generator/main.py
ttl33: Non-blocking: can you suggest better title format in the near future to all the contributors?

teddyding (approved):

docs/release_note_generator/main.py
roy-dydx: I originally put this to catch the cherry-picks we've been doing without a PR. But if we're write protecting the branches in the future we won't hit this case.

PR #643, opened 2023-10-17
[CLOB-948] - non-linear margin requirement liquidations test

jakob-dydx:


protocol/x/clob/e2e/liquidation_deleveraging_test.go
```diff
@@ -67,6 +78,134 @@ func TestPlacePerpetualLiquidation_Deleveraging(t *testing.T) {
...
+						},
+					},
+					PerpetualPositions: []*satypes.PerpetualPosition{
+						{
+							PerpetualId:  0,
+							Quantums:     dtypes.NewInt(100_000_000), // 1 BTC
+							FundingIndex: dtypes.NewInt(0),
+						},
+					},
+				},
```
jayy04, line 205: is this going to be used for other tests in the future? should we move it to testutil/constants?

jayy04 (approved):

protocol/x/clob/e2e/liquidation_deleveraging_test.go
jakob-dydx: idk it's pretty specific to this test, don't really see it as a generally re-usable subaccount because it assumes non-linear margin requirements for its Asset and Perpetual positions to be relevant and frequently used values in tests

PR #648, opened 2023-10-17
Fix `AdvanceToBlock` blocktime behavior

teddyding:
### Changelist
Discussed offline with @lcwik to fix the blocktime behavior of `AdvanceToBlock` function. Previously comment indicates that `options.BlockTime` is only for the requested (latest) block, but the implementation actually applies `options.BlockTime` to the first advanced block.

### Test Plan
Fixed a few e2e tests affected by this behavior change. All other e2e tests still passing

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/bridge/app_test.go
+++ b/protocol/x/bridge/app_test.go
@@ -142,6 +142,7 @@ func TestBridge_Success(t *testing.T) {
 						genesisState.SafetyParams = tc.safetyParams
 					},
 				)
+				genesis.GenesisTime = tc.blockTime
 				return genesis
 			}).WithTesting(t).Build()
 			ctx := tApp.InitChain()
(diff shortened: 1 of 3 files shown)


protocol/x/bridge/app_test.go
```diff
@@ -142,6 +142,7 @@ func TestBridge_Success(t *testing.T) {
 						genesisState.SafetyParams = tc.safetyParams
 					},
 				)
+				genesis.GenesisTime = tc.blockTime
```
teddyding, line 145: @tqin7 The `blockTime` in the test case is ambiguous. I believe setting genesis time with this time best suits the intention of this test (so the BlockTime is already set to `tc.blockTime` by the time bridge module looks at the bridge events

protocol/x/stats/keeper/keeper_test.go
```diff
@@ -197,12 +197,13 @@ func TestExpireOldStats(t *testing.T) {
 	})
 	windowDuration := tApp.App.StatsKeeper.GetWindowDuration(ctx)
 	// 5 epochs are out of the window
-	ctx = tApp.AdvanceToBlock(100, testapp.AdvanceToBlockOptions{
+	tApp.AdvanceToBlock(3, testapp.AdvanceToBlockOptions{
```
teddyding, line 200: @roy-dydx I didn't fully under stand this test case, so I just edit'ed the test set up to retain the previous behavior (setting specified `BlockTime` at block 3, then advance to 100)

lcwik (approved):

protocol/x/stats/keeper/keeper_test.go
roy-dydx: Does the test fail without this change?

It's supposed to simulate advancing to block `2` at `time.Unix(int64(1), 0).UTC()` and then advancing to block 100 at this specified time. I feel like it should work without the additional advance to 3?

protocol/x/bridge/app_test.go
tqin7: Yeah that works!

tqin7 (approved):

PR #652, opened 2023-10-18
[CORE-632] Add workflow to check sample_pregenesis is uptodate; sort pregenesis file so ordering is deterministic

teddyding:
### Changelist
- Add workflow to check sample_pregenesis is uptodate
- Sort pregenesis file so ordering is deterministic

### Test Plan
Check workflow on this PR. 

### Author/Reviewer Checklist


protocol/scripts/genesis/prod_pregenesis.sh
```diff
@@ -170,6 +170,11 @@ create_pregenesis_file() {
 	overwrite_genesis_production
 }
 
+sort_genesis_file(){
+	jq -S . $VAL_CONFIG_DIR/genesis.json > $VAL_CONFIG_DIR/sorted_genesis.json
```
teddyding, line 174: Sorted the output json so ordering is deterministic and can be diff'ed

.github/workflows/protocol-pregenesis.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    paths:
+      - 'protocol/scripts/genesis/**'
+      - 'protocol/testing/**'
+  push:
+    branches:
+      - main
+      - 'release/protocol/v0.[0-9]+.x'  # e.g. release/protocol/v0.1.x
```
ttl33, line 11: for my understanding, are we going to use the letter `x`? I thought it was just a placeholder

teddyding: The branch name explicitly has letter `x`, like `release/protocol/v1.x`

protocol/scripts/genesis/Dockerfile
```diff
@@ -0,0 +1,56 @@
+# NB: This is a digest for a multi-arch manifest list, you will want to get this by running
```
roy-dydx, line 1: What's the difference between this dockerfile and `protocol/Dockerfile`? Can we avoid the duplication and extra maintenance by reusing it like we do in the staging dockerfile, localnet dockerfile, etc?

If the only difference is installing jq and dasel, it's OK to add it to the original docker image IMO.

teddyding: Updated to use the `dydxprotoocl-base` image as base

PR #666, opened 2023-10-19
[CLO-987] Subaccount equity tier count was double counting conditional orders.

lcwik:
### Changelist
Remove counting of untriggered conditional orders since they are already in the subaccount order state count.

### Test Plan
The change to the error message contains the actual count which was 2 in the case of the conditional order and not 1 showing the double counting.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/e2e/equity_tier_limit_test.go
+++ b/protocol/x/clob/e2e/equity_tier_limit_test.go
@@ -1,6 +1,7 @@
 package clob_test
 
 import (
+	"fmt"
 	"github.com/cometbft/cometbft/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/dtypes"
@@ -16,19 +17,21 @@ import (
 
 func TestPlaceOrder_EquityTierLimit(t *testing.T) {
 	tests := map[string]struct {
-		firstOrder                   clobtypes.Order
-		secondOrder                  clobtypes.Order
+		allowedOrders                []clobtypes.Order
+		limitedOrder                 clobtypes.Order
 		equityTierLimitConfiguration clobtypes.EquityTierLimitConfiguration
 		cancellation                 *clobtypes.MsgCancelOrder
 		advanceBlock                 bool
 		expectError                  bool
 	}{
 		"Short-term order would exceed max open short-term orders in same block": {
-			firstOrder: MustScaleOrder(
-				constants.Order_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB20,
-				testapp.DefaultGenesis(),
-			),
-			secondOrder: MustScaleOrder(
+			allowedOrders: []clobtypes.Order{
+				MustScaleOrder(
+					constants.Order_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB20,
+					testapp.DefaultGenesis(),
+				),
+			},
+			limitedOrder: MustScaleOrder(
 				constants.Order_Alice_Num0_Id0_Clob1_Buy5_Price10_GTB15,
 				testapp.DefaultGenesis(),
 			),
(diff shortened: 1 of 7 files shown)


protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -149,8 +145,9 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
 	if lib.MustConvertIntegerToUint32(equityTierCount) >= equityTierLimit.Limit {
 		return errorsmod.Wrapf(
 			types.ErrOrderWouldExceedMaxOpenOrdersEquityTierLimit,
-			"Opening order would exceed equity tier limit of %d. Order id: %+v",
+			"Opening order would exceed equity tier limit of %d. Order count: %d, order id: %+v",
```
lucas-dydx, line 148: Wdyt of adding the TNC in this error?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -518,8 +518,7 @@ func (k Keeper) getUntriggeredConditionalOrdersIterator(ctx sdk.Context) sdk.Ite
 	return sdk.KVStorePrefixIterator(store, []byte{})
 }
 
-// GetStatefulOrderCount gets a count of how many stateful orders are written to state for a subaccount. This does not
-// include any untriggered conditional orders.
+// GetStatefulOrderCount gets a count of how many stateful orders are written to state for a subaccount.
```
lucas-dydx, line 521: Can we update [this comment](https://github.com/dydxprotocol/v4-chain/blob/e48f473f1771b6c9e8d6f4862cc9a7131ea5b701/protocol/x/clob/keeper/stateful_order_state.go#L538) too?

protocol/x/clob/e2e/equity_tier_limit_test.go
```diff
@@ -726,7 +726,7 @@ func TestPlaceOrder_EquityTierLimit_OrderExpiry(t *testing.T) {
 				resp := tApp.CheckTx(tx)
 				if tc.expectError {
 					require.Conditionf(t, resp.IsErr, "Expected CheckTx to error. Response: %+v", resp)
-					require.Contains(t, resp.Log, "Opening order would exceed equity tier limit of 1.")
+					require.Contains(t, resp.Log, "Opening order would exceed equity tier limit of 1. Order count: 1,")
```
lucas-dydx, line 729: Do we have any E2E tests that verify equity tier limits that are greater than 1?

jayy04 (approved):

protocol/x/clob/keeper/equity_tier_limit.go
jayy04: Getting TNC is actually quite a bit of work in terms of CPU time. if we only expect a small amount of orders to fail equity tier limit, then I am fine with adding. but if we expect a large number of subaccounts failing, imo we should just leave it out

jayy04: probably better to add the entire subaccount instead of TNC since that's way cheaper (we can calculate TNC offline if there are issues)

lucas-dydx: We already calculate net collateral [here](https://github.com/dydxprotocol/v4-chain/blob/79b327ce906ffd74219927be62b6079eaa2bdfd5/protocol/x/clob/keeper/equity_tier_limit.go#L88) in this function though right? So no extra computation is required.

lucas-dydx:
Should we update these comments as well?
- https://github.com/dydxprotocol/v4-chain/blob/79b327ce906ffd74219927be62b6079eaa2bdfd5/protocol/x/clob/types/keys.go#L105-L107
- https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/stores.go#L80

I would suggest updating the state keys as well but I believe that change is backwards-incompatible 😕

protocol/x/clob/e2e/equity_tier_limit_test.go
lcwik: Added cases with an equity tier count of 2.

lcwik:
> Should we update these comments as well?
> 
> * https://github.com/dydxprotocol/v4-chain/blob/79b327ce906ffd74219927be62b6079eaa2bdfd5/protocol/x/clob/types/keys.go#L105-L107
> * https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/stores.go#L80
> 
> I would suggest updating the state keys as well but I believe that change is backwards-incompatible 😕

Since this is only in memstore changing they key doesn't change what is written in state just where it is referenced in the memstore so it would be safe.

protocol/x/clob/types/keys.go
```diff
@@ -101,10 +105,6 @@ const (
 	// are stored in a transient store. This count represents the number of uncommitted stateful
 	// `placements - cancellations`.
 	UncommittedStatefulOrderCountPrefix = "NumUncmtLT:"
```
lucas-dydx, line 107: This is all up to you, fine with leaving as-is if you prefer.

Should we update `UncommittedStatefulOrderCountPrefix` to be SO instead of LT?

Same with `UncommittedStatefulOrderCancellationKeyPrefix` and `UncommittedStatefulOrderPlacementKeyPrefix`.

lucas-dydx (approved):

protocol/x/clob/keeper/equity_tier_limit.go
```diff
@@ -149,8 +145,10 @@ func (k Keeper) ValidateSubaccountEquityTierLimitForNewOrder(ctx sdk.Context, or
 	if lib.MustConvertIntegerToUint32(equityTierCount) >= equityTierLimit.Limit {
 		return errorsmod.Wrapf(
 			types.ErrOrderWouldExceedMaxOpenOrdersEquityTierLimit,
-			"Opening order would exceed equity tier limit of %d. Order id: %+v",
+			"Opening order would exceed equity tier limit of %d. Order count: %d, total net collateral: %+v, order id: %+v",
 			equityTierLimit.Limit,
+			equityTierCount,
+			netCollateral,
```
lucas-dydx, line 151: Do you think we should include the block height in the error as well? Up to you.

lcwik: That seems like information that should be added to all CheckTx errors outside of just this one and would make sense for it to be added elsewhere.

PR #678, opened 2023-10-20
[refactor] Remove errors from subaccounts keeper functions that don't return errors

lucas-dydx:
### Changelist

Remove errors from subaccounts keeper functions that don't return errors.

This is part of a larger subaccounts keeper refactor to simplify the code by removing returned errors when invariants are broken, and instead panicing earlier.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/subaccounts/keeper/subaccount_helper.go
+++ b/protocol/x/subaccounts/keeper/subaccount_helper.go
@@ -106,7 +106,6 @@ func UpdatePerpetualPositions(
 	perpIdToFundingIndex map[uint32]dtypes.SerializableInt,
 ) (
 	success bool,
-	err error,
 ) {
 	// Apply the updates.
 	for i, u := range settledUpdates {
@@ -164,7 +163,8 @@ func UpdatePerpetualPositions(
 
 		settledUpdates[i].SettledSubaccount.PerpetualPositions = perpetualPositions
 	}
-	return true, nil
+
+	return true
 }
 
 // For each settledUpdate in settledUpdates, updates its SettledSubaccount.AssetPositions
@@ -173,7 +173,6 @@ func UpdateAssetPositions(
 	settledUpdates []settledUpdate,
 ) (
 	success bool,
-	err error,
 ) {
 	// Apply the updates.
 	for i, u := range settledUpdates {
@@ -226,5 +225,6 @@ func UpdateAssetPositions(
 
 		settledUpdates[i].SettledSubaccount.AssetPositions = assetPositions
 	}
-	return true, nil
+
+	return true
 }
--- a/protocol/x/subaccounts/keeper/subaccount_helper.go
+++ b/protocol/x/subaccounts/keeper/subaccount_helper.go
@@ -104,8 +104,6 @@ func getUpdatedPerpetualPositions(
 func UpdatePerpetualPositions(
 	settledUpdates []settledUpdate,
 	perpIdToFundingIndex map[uint32]dtypes.SerializableInt,
-) (
-	success bool,
 ) {
 	// Apply the updates.
 	for i, u := range settledUpdates {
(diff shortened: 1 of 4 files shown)


protocol/x/subaccounts/types/subaccount.go
```diff
@@ -73,26 +73,25 @@ func (m *Subaccount) GetUsdcPosition() *big.Int {
...
-			}
-		} else {
-			if usdcAssetPosition == nil {
-				usdcAssetPosition = &AssetPosition{
-					AssetId: assettypes.AssetUsdc.Id,
-				}
-				m.AssetPositions = append([]*AssetPosition{usdcAssetPosition}, m.AssetPositions...)
+func (m *Subaccount) SetUsdcAssetPosition(newUsdcPosition *big.Int) {
+	if m == nil {
+		return
```
lucas-dydx, line 78: Should we panic / error log if this ever happens?

It doesn't seem like this function should silently fail if called with a nil pointer.

protocol/x/subaccounts/types/subaccount.go
```diff
@@ -73,26 +73,25 @@ func (m *Subaccount) GetUsdcPosition() *big.Int {
...
+	}
+
+	usdcAssetPosition := m.getUsdcAssetPosition()
+	if newUsdcPosition == nil || newUsdcPosition.Sign() == 0 {
```
lucas-dydx, line 82: When do we ever expect this function to be provided a `nil` pointer for `newUsdcPosition`? Should we panic / error log in that case as well?

protocol/x/subaccounts/keeper/subaccount_helper.go
```diff
@@ -106,7 +106,6 @@ func UpdatePerpetualPositions(
 	perpIdToFundingIndex map[uint32]dtypes.SerializableInt,
 ) (
 	success bool,
-	err error,
```
lucas-dydx, line 109: This function doesn't return an error anymore, so removed it.

protocol/x/subaccounts/types/subaccount.go
```diff
@@ -73,26 +73,25 @@ func (m *Subaccount) GetUsdcPosition() *big.Int {
...
-			if usdcAssetPosition != nil {
-				m.AssetPositions = m.AssetPositions[1:]
-			}
-		} else {
-			if usdcAssetPosition == nil {
-				usdcAssetPosition = &AssetPosition{
-					AssetId: assettypes.AssetUsdc.Id,
-				}
-				m.AssetPositions = append([]*AssetPosition{usdcAssetPosition}, m.AssetPositions...)
+func (m *Subaccount) SetUsdcAssetPosition(newUsdcPosition *big.Int) {
```
lucas-dydx, line 76: This function also doesn't return an error anymore, so removed it + refactored it to early return.

Let me know if anyone would prefer I remove the refactor logic to early return. This function has good test coverage though so I felt like the change is safe.

protocol/x/subaccounts/keeper/subaccount_helper.go
```diff
@@ -106,7 +106,6 @@ func UpdatePerpetualPositions(
 	perpIdToFundingIndex map[uint32]dtypes.SerializableInt,
 ) (
 	success bool,
```
lucas-dydx, line 108: Hmm actually on re-reading this function, this always returns `success = true`. I'm going to remove this variable as well.

lucas-dydx: Done!

jayy04 (approved):

protocol/x/subaccounts/types/subaccount.go
jayy04: I think this was meant to be a convenience method for tests only. I don't think there is a reason to use this instead of `UpdateAssetPositions`. wdyt?

protocol/x/subaccounts/types/subaccount.go
jayy04: see my other comment. wdyt about moving this to testutil package and make this used in tests only?

teddyding (approved):

protocol/x/subaccounts/types/subaccount.go
lucas-dydx: It's currently used in subaccounts keeper [here](https://github.com/dydxprotocol/v4-chain/blob/dcd6ecb083145ccc4021aef8bbbcb2a3897f7111/protocol/x/subaccounts/keeper/subaccount.go#L436)

Created a ticket [here](https://linear.app/dydx/issue/CLOB-993/remove-setusdcassetposition-function) to remove it later.

protocol/x/subaccounts/types/subaccount.go
lucas-dydx: Responded in other comment!

PR #679, opened 2023-10-20
[CORE-704] Initial E2E testing framework for funding

teddyding:
### Changelist
Added an funding e2e test that can be built upon in the future. Current behavior tested:
- Set up orders on the book
- Advance to start of full `funding-tick` period
- Measure a full `60 min` `funding-tick` period and check expected funding samples and resulting `fundingIndex`. 

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testing/e2e/funding/funding_e2e_test.go
+++ b/protocol/testing/e2e/funding/funding_e2e_test.go
@@ -174,8 +174,8 @@ func TestFunding(t *testing.T) {
 			)
 
 			ctx = tApp.AdvanceToBlock(BlockHeightAtFirstFundingTick, testapp.AdvanceToBlockOptions{
-				BlockTime:                 FirstFundingTick,
-				GradualBlockTimeIncrement: true,
+				BlockTime:                    FirstFundingTick,
+				LinearBlockTimeInterpolation: true,
 			})
 
 			premiumSamples := tApp.App.PerpetualsKeeper.GetPremiumSamples(ctx)
@@ -201,8 +201,8 @@ func TestFunding(t *testing.T) {
 			ctx = tApp.AdvanceToBlock(
 				testapp.EstimatedHeightForBlockTime(GenesisTime, LastFundingSampleOfSecondFundingTickEpoch, BlockTimeDuration),
 				testapp.AdvanceToBlockOptions{
-					BlockTime:                 LastFundingSampleOfSecondFundingTickEpoch,
-					GradualBlockTimeIncrement: true,
+					BlockTime:                    LastFundingSampleOfSecondFundingTickEpoch,
+					LinearBlockTimeInterpolation: true,
 				})
 
 			premiumSamples = tApp.App.PerpetualsKeeper.GetPremiumSamples(ctx)
--- a/protocol/testing/e2e/funding/funding_e2e_test.go
+++ b/protocol/testing/e2e/funding/funding_e2e_test.go
@@ -170,7 +170,9 @@ func TestFunding(t *testing.T) {
 				ctx,
 				tApp.App,
 				pricestest.MustHumanPriceToMarketPrice(tc.initialIndexPrice[0], -5),
-				time.Now(),
+				// Only index price past a certain threshold is used for premium calculation.
+				// Use additional buffer here to ensure `test-race` passes.
+				time.Now().Add(1*time.Hour),
 			)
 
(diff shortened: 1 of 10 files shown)


protocol/testutil/app/app.go
```diff
@@ -83,6 +83,11 @@ type AdvanceToBlockOptions struct {
 	// The time associated with the block. If left at the default value then block time will be left unchanged.
 	BlockTime time.Time
 
+	// Whether to increment the block time gradually and evenly among the advanced blocks.
+	// TODO(DEC-2156): Instead of an option, pass in a `BlockTimeFunc` to map each block to a
+	// time. This gives user more flexibility.
```
lcwik, line 88:
```suggestion
	// TODO(DEC-2156): Instead of an option, pass in a `BlockTimeFunc` to map each block to a
	// time giving user greater flexibility.
```

protocol/testutil/app/app.go
```diff
@@ -407,8 +412,12 @@ func (tApp *TestApp) AdvanceToBlock(
 	for tApp.App.LastBlockHeight() < int64(block) {
 		tApp.panicIfChainIsHalted()
 		tApp.header.Height = tApp.App.LastBlockHeight() + 1
-		// By default, only update block time at the requested block.
-		if tApp.header.Height == int64(block) {
+		if options.GradualBlockTimeIncrement {
```
lcwik, line 415: We want the last block to end up with the exact time the user choose so I would suggest making the `tApp.header.Height == int64(block)` check first so there is no rounding issue for the last block.

protocol/testutil/app/app.go
```diff
@@ -83,6 +83,11 @@ type AdvanceToBlockOptions struct {
 	// The time associated with the block. If left at the default value then block time will be left unchanged.
 	BlockTime time.Time
 
+	// Whether to increment the block time gradually and evenly among the advanced blocks.
```
lcwik, line 86:
```suggestion
	// Whether to increment the block time using linear extrapolation among the blocks.
```

protocol/testutil/app/app.go
```diff
@@ -83,6 +83,11 @@ type AdvanceToBlockOptions struct {
...
+	// TODO(DEC-2156): Instead of an option, pass in a `BlockTimeFunc` to map each block to a
+	// time. This gives user more flexibility.
+	GradualBlockTimeIncrement bool
```
lcwik, line 89: nit: name change
```suggestion
	LinearBlockTimeInterpolation bool
```

lcwik (changes requested):

lcwik (approved):

protocol/testing/e2e/funding/funding_e2e_test.go
```diff
@@ -0,0 +1,247 @@
...
+			tApp := testapp.NewTestAppBuilder(t).WithGenesisDocFn(func() (genesis types.GenesisDoc) {
+				genesis = testapp.DefaultGenesis()
+				genesis.GenesisTime = GenesisTime
+				return genesis
+			}).Build()
+			ctx := tApp.InitChain()
+
+			// Place orders on the book.
+			for _, testHumanOrder := range tc.testHumanOrders {
+				order := clobtest.MustMakeOrderFromHumanInput(
```
lcwik, line 153: It would be nice if we could just use [MustScaleOrder](https://github.com/dydxprotocol/v4-chain/blob/a612627dc7bea419a23bbe2307636818a36b0285/protocol/x/clob/e2e/app_test.go#L681) but understand that it may not have the granularity that you need for these tests.

protocol/testing/e2e/funding/funding_e2e_test.go
```diff
@@ -0,0 +1,247 @@
...
+		testHumanOrders   []TestHumanOrder
+		initialIndexPrice map[uint32]string
+		// index price to be used in premium calculation
+		indexPriceForPremium map[uint32]string
+		// oracle price for funding index calculation
+		oracelPriceForFundingIndex map[uint32]string
+		expectedFundingPremium     int32
+		expectedFundingIndex       int64
+	}{
+		"Test funding": {
```
tqin7, line 102: should we in the future add cases where funding is 0?

tqin7 (approved):

protocol/testing/e2e/funding/funding_e2e_test.go
teddyding: Yes, planning to do that in a follow-up PR. Added a TODO.

PR #691, opened 2023-10-24
[CORE-709] do not error when completing a bridge with non-positive amount

tqin7:
### Changelist
When completing a bridge with non-positive amount, do not error. Simply ignore bank transfer and consider the bridge completed.

Approaches considered for bridges with 0 amount
1. Still complete it but skip bank transfer (what this PR implements)
* pros
  * all bridges follow the same logic path in bridge module, leading to more consistent behaviors and metrics (for example, we emit metric on completed bridge ID in `CompleteBridge`)
  * state-compatible (delaymsg module state is the same as before)
* cons:
  * does not save storage (but I think this is fine as any bridge, as long as acknowledged by the chain, "deserves" to have a `MsgCompleteBridge` scheduled for it (as gas has been paid anyways on Ethereum side)
2. When acknowledging the bridge, do not schedule a delayed `MsgCompleteBridge`
* pros
  * saves storage
* cons:
  * not state-compatible (delaymsg module state is not the same as before)

### Test Plan

### Author/Reviewer Checklist


clemire (approved):

protocol/x/bridge/keeper/complete_bridge.go
```diff
@@ -35,14 +35,18 @@ func (k Keeper) CompleteBridge(
...
-		ctx,
-		types.ModuleName,
-		bridgeAccAddress,
-		sdk.Coins{bridge.Coin},
-	); err != nil {
-		return err
+	// If coin amount is positive, send coin from bridge module account to
+	// specified account.
+	if bridge.Coin.Amount.IsPositive() {
+		err = k.bankKeeper.SendCoinsFromModuleToAccount(
```
teddyding, line 41: nit: `if err = .....; err != nil {`

protocol/x/bridge/keeper/complete_bridge.go
```diff
@@ -35,14 +35,18 @@ func (k Keeper) CompleteBridge(
...
-	if err = k.bankKeeper.SendCoinsFromModuleToAccount(
-		ctx,
-		types.ModuleName,
-		bridgeAccAddress,
-		sdk.Coins{bridge.Coin},
-	); err != nil {
-		return err
+	// If coin amount is positive, send coin from bridge module account to
+	// specified account.
+	if bridge.Coin.Amount.IsPositive() {
```
teddyding, line 40: I see we have taken the approach to still schedule delayed messages for `0` transfers. 
Can you document somewhere the rationale behind tradeoffs of different approaches, and why we are allowing zero-transfer delay messages to be added to state?

tqin7: Documented rationale in PR description!

teddyding (approved):

PR #692, opened 2023-10-24
[IND-387] add script to parse block and print json

dydxwill:
### Changelist
add script to parse block and return json

### Test Plan
ran on devbox

### Author/Reviewer Checklist


indexer/services/scripts/src/helpers/block-helpers.ts
```diff
@@ -0,0 +1,124 @@
...
+  StatefulOrderEventV1,
+  SubaccountUpdateEventV1,
+  TransferEventV1,
+  UpdateClobPairEventV1,
+  UpdatePerpetualEventV1,
+} from '@dydxprotocol-indexer/v4-protos';
+
+import { AnnotatedIndexerTendermintEvent } from './types';
+
+export enum DydxIndexerSubtypes {
```
Christopher-Li, line 19: nit: move to types

indexer/services/scripts/src/print-block.ts
```diff
@@ -0,0 +1,169 @@
...
+    message: `Connecting to kafka brokers: ${config.KAFKA_BROKER_URLS}`,
+  });
+  await startKafka(height);
+  logger.info({
+    at: 'index#start',
+    message: `Successfully connected to kafka brokers: ${config.KAFKA_BROKER_URLS}`,
+  });
+}
+
+const args = yargs.options({
```
Christopher-Li, line 160: nit: only if it's easy to do, can we also add support to just parse a block at a particular offset

vincentwschau (approved):

indexer/services/scripts/src/helpers/block-helpers.ts
will530: done

indexer/services/scripts/src/print-block.ts
will530: I don't think this is needed. I believe all debug logs in Ender that print out offset should also print out block height, so we can just use the block height.

indexer/services/scripts/src/helpers/block-helpers.ts
dydxwill: done

Christopher-Li (approved):

PR #694, opened 2023-10-24
[CORE-711] Always log market id when pair information is missing.

clemire:
### Changelist
When getting market id labels, always return the printed market id if the pair string is not available.

### Test Plan
Updated existing tests.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/daemons/pricefeed/metrics/market_pairs.go
+++ b/protocol/daemons/pricefeed/metrics/market_pairs.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"fmt"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


protocol/daemons/pricefeed/metrics/market_pairs.go
```diff
@@ -39,7 +40,7 @@ func GetMarketPairForTelemetry(marketId types.MarketId) string {
 
 	marketPair, exists := marketToPair[marketId]
 	if !exists {
-		return INVALID
+		return fmt.Sprintf("id:%v", marketId)
```
ttl33, line 43:
```suggestion
		return fmt.Sprintf("invalid id:%v", marketId)
```

ttl33 (approved):

PR #695, opened 2023-10-24
Initial `PrepareCheckState` audit

lucas-dydx:
### Changelist
Small `PrepareCheckState` fixes from audit (nothing major that needs to be backported to the release).

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/liquidations.go
+++ b/protocol/x/clob/keeper/liquidations.go
@@ -271,6 +271,9 @@ func (k Keeper) PlacePerpetualLiquidation(
 			ctx,
 			liquidationOrder,
 		)
+	if err != nil {
+		return err
+	}
 
 	// TODO(DEC-1323): Potentially allow liquidating the same perpetual + subaccount
 	// multiple times in a block.
--- a/protocol/x/clob/keeper/liquidations.go
+++ b/protocol/x/clob/keeper/liquidations.go
@@ -272,7 +272,7 @@ func (k Keeper) PlacePerpetualLiquidation(
 			liquidationOrder,
 		)
 	if err != nil {
-		return err
+		return 0, 0, err
 	}
 
 	// TODO(DEC-1323): Potentially allow liquidating the same perpetual + subaccount
--- a/protocol/x/clob/keeper/orders.go
+++ b/protocol/x/clob/keeper/orders.go
@@ -454,6 +454,17 @@ func (k Keeper) PlaceStatefulOrdersFromLastBlock(
 
 		orderPlacement, exists := k.GetLongTermOrderPlacement(ctx, orderId)
 		if !exists {
+			// Error log if this is a conditional orders and it doesn't exist in state, since it
+			// can't have been canceled.
+			if orderId.IsConditionalOrder() {
+				k.Logger(ctx).Error(
+					fmt.Sprintf(
+						"PlaceStatefulOrdersFromLastBlock: Triggered conditional Order ID %+v doesn't exist in state",
+						orderId,
+					),
+				)
+			}
+
 			// Order does not exist in state and therefore should not be placed. This likely
 			// indicates that the order was cancelled.
 			continue
--- a/protocol/x/clob/keeper/orders.go
+++ b/protocol/x/clob/keeper/orders.go
@@ -454,7 +454,7 @@ func (k Keeper) PlaceStatefulOrdersFromLastBlock(
 
 		orderPlacement, exists := k.GetLongTermOrderPlacement(ctx, orderId)
 		if !exists {
-			// Error log if this is a conditional orders and it doesn't exist in state, since it
+			// Error log if this is a conditional orders and it doesn't exist in triggered state, since it
 			// can't have been canceled.
 			if orderId.IsConditionalOrder() {
 				k.Logger(ctx).Error(
--- a/protocol/x/clob/memclob/memclob.go
+++ b/protocol/x/clob/memclob/memclob.go
@@ -1004,7 +1004,7 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 			}
 
 			// Log an error if there are two Order Removals for the same OrderId
-			if _, found := placedPreexistingStatefulOrderIds[orderId]; found {
+			if _, found := placedOrderRemovalOrderIds[orderId]; found {
 				m.clobKeeper.Logger(ctx).Error(
 					"ReplayOperations: OrderRemoval operation for order which was already removed",
 					metrics.OrderId, orderId,
@@ -1132,12 +1132,12 @@ func (m *MemClobPriceTimePriority) RemoveAndClearOperationsQueue(
 			otpOrderHash := operation.GetShortTermOrderPlacement().Order.GetOrderHash()
 
 			// If the order exists in the book, remove it.
-			// Else if the order is a Short-Term order, since it's no longer on the book or operations
-			// queue we should remove the order hash from `ShortTermOrderTxBytes`.
+			// Else, since the Short-Term order is no longer on the book or operations queue we
+			// should remove the order hash from `ShortTermOrderTxBytes`.
 			existingOrder, found := m.openOrders.getOrder(ctx, otpOrderId)
 			if found && existingOrder.GetOrderHash() == otpOrderHash {
 				m.mustRemoveOrder(ctx, otpOrderId)
-			} else if otpOrderId.IsShortTermOrder() {
+			} else {
 				order := operation.GetShortTermOrderPlacement().Order
 				m.operationsToPropose.RemoveShortTermOrderTxBytes(order)
 			}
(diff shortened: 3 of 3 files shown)


protocol/x/clob/keeper/liquidations.go
```diff
@@ -271,6 +271,9 @@ func (k Keeper) PlacePerpetualLiquidation(
 			ctx,
 			liquidationOrder,
 		)
+	if err != nil {
```
lucas-dydx, line 274: IIRC we were previously ignoring this error and re-assigning it below

protocol/x/clob/keeper/orders.go
```diff
@@ -454,6 +454,17 @@ func (k Keeper) PlaceStatefulOrdersFromLastBlock(
 
 		orderPlacement, exists := k.GetLongTermOrderPlacement(ctx, orderId)
 		if !exists {
+			// Error log if this is a conditional orders and it doesn't exist in state, since it
+			// can't have been canceled.
+			if orderId.IsConditionalOrder() {
```
lucas-dydx, line 459: Conditional orders should always exist in state, so error logging here if they don't exist

protocol/x/clob/memclob/memclob.go
```diff
@@ -1004,7 +1004,7 @@ func (m *MemClobPriceTimePriority) ReplayOperations(
 			}
 
 			// Log an error if there are two Order Removals for the same OrderId
-			if _, found := placedPreexistingStatefulOrderIds[orderId]; found {
+			if _, found := placedOrderRemovalOrderIds[orderId]; found {
```
lucas-dydx, line 1007: This check was previously the exact same as the above check, which was an oversight

protocol/x/clob/memclob/memclob.go
```diff
@@ -1132,12 +1132,12 @@ func (m *MemClobPriceTimePriority) RemoveAndClearOperationsQueue(
...
 			// If the order exists in the book, remove it.
-			// Else if the order is a Short-Term order, since it's no longer on the book or operations
-			// queue we should remove the order hash from `ShortTermOrderTxBytes`.
+			// Else, since the Short-Term order is no longer on the book or operations queue we
+			// should remove the order hash from `ShortTermOrderTxBytes`.
 			existingOrder, found := m.openOrders.getOrder(ctx, otpOrderId)
 			if found && existingOrder.GetOrderHash() == otpOrderHash {
 				m.mustRemoveOrder(ctx, otpOrderId)
-			} else if otpOrderId.IsShortTermOrder() {
+			} else {
```
lucas-dydx, line 1140: This shouldn't be necessary since all of these operations should be short-term orders

protocol/x/clob/memclob/memclob.go
```diff
@@ -2003,8 +2003,14 @@ func (m *MemClobPriceTimePriority) RemoveOrderIfFilled(
 	// Get current fill amount for this order.
 	exists, orderStateFillAmount, _ := m.clobKeeper.GetOrderFillAmount(ctx, orderId)
 
-	// If there is no fill amount for this order, return early.
+	// If there is no fill amount for this order, error log and return early. Note this is an
+	// error since this function should only be called for orders that were partially or fully filled.
 	if !exists {
+		m.clobKeeper.Logger(ctx).Error(
```
lucas-dydx, line 2009: These orders were filled in the last block so they should always have a non-zero fill amount in state

protocol/x/clob/keeper/orders.go
jonfung-dydx: Also might be good to assert `IsConditionalOrderTriggered` is True. Since the conditional order has to be triggered at this point in time. It is possible to read out from untriggered state in`GetLongTermOrderPlacement`.

protocol/x/clob/keeper/orders.go
```diff
@@ -454,6 +454,17 @@ func (k Keeper) PlaceStatefulOrdersFromLastBlock(
 
 		orderPlacement, exists := k.GetLongTermOrderPlacement(ctx, orderId)
 		if !exists {
+			// Error log if this is a conditional orders and it doesn't exist in state, since it
```
jonfung-dydx, line 457:
```suggestion
			// Error log if this is a conditional orders and it doesn't exist in triggered state, since it
```

protocol/x/clob/memclob/memclob.go
jonfung-dydx: Do all the callsites for this function only contain orders filled in last block?

jonfung-dydx: nvm, mixed up fn names. I see that this has only one callsite populating the `filledOrderIds` from PPME.

jonfung-dydx: In the future do we want to panic in this situation instead of early returning? Having an error log that results in a no-op seems a bit awkward to me.

jonfung-dydx (approved):

protocol/x/clob/keeper/orders.go
jakob-dydx: ^ should definitely assert that the conditional order is triggered

protocol/x/clob/memclob/memclob.go
jakob-dydx: Won't this error log always trigger for fully filled stateful orders, stateful orders which were filled and then cancelled, stateful orders which were filled and then removed, and stateful orders that were filled and then expired? Also, if we wrongfully trigger this error log we will also not remove the order from the book as a result because we return early.

The `PPME.OrderIdsFilledInLastBlock` does not filter out cancellations, removals, or expirations.

lucas-dydx: >In the future do we want to panic in this situation instead of early returning? Having an error log that results in a no-op seems a bit awkward to me.

Panicing would halt the chain so I avoided it. I think there are cases where a bug with fill amounts could make sense to halt the chain (Short-Term orders not storing ill amounts before expiration could lead to replay attacks), but generally I don't think it makes sense to halt. Let me know if you disagree.

jonfung-dydx: > Also, if we wrongfully trigger this error log we will also not remove the order from the book as a result because we return early.

Isn't this just an error log, so it doesn't affect application behavior? It's fine to print out whatever we want as long as we don't panic

> Agreed regarding error cases for stateful orders that were fully-filled, partially-filled then canceled, expired or removed.

Makes sense to me! Agreed that if an order was cancelled or removed this would trigger.

>Panicing would halt the chain so I avoided it.
I think if we have stronger invariants that we are 100% sure must be valid we should halt. In this specific case I think that we shouldn't halt because of the edge cases Jakob said above.

Honestly for this specific function I wouldn't be against calling it `RemoveOrderIfFilled` (which implies it only removes orders if filled, and if not filled, does nothing) and doing all the invariant checking in the PCS callsite.

protocol/x/clob/memclob/memclob.go
```diff
@@ -2003,8 +2003,17 @@ func (m *MemClobPriceTimePriority) RemoveOrderIfFilled(
 	// Get current fill amount for this order.
 	exists, orderStateFillAmount, _ := m.clobKeeper.GetOrderFillAmount(ctx, orderId)
 
-	// If there is no fill amount for this order, return early.
+	// If there is no fill amount for this order, return early. Note this is only valid if the
+	// order is a stateful order that was fully-filled or partially-filled and expired / canceled /
+	// removed in the last block.
```
jonfung-dydx, line 2008:
```suggestion
	// If there is no fill amount for this order, return early. Note this is only valid if the
	// order is a stateful order that was fully-filled or partially-filled and expired / canceled /
	// removed in the last block. For short-term orders, their fill amounts are stored past expiration,
	// so this return is invalid.
```

jakob-dydx (approved):

protocol/x/clob/memclob/memclob.go
lucas-dydx: >I think if we have stronger invariants that we are 100% sure must be valid we should halt. In this specific case I think that we shouldn't halt because of the edge cases Jakob said above.

Personally I think it depends on the problem. If the issue doesn't cause traders to lose money (rate-limiting that is too aggressive and leads to dropped orders in `DeliverTx` for example) then I don't think that's worth halting the chain for.

However if it causes traders to lose significant amounts of money, the chain should be halted.

>Honestly for this specific function I wouldn't be against calling it RemoveOrderIfFilled (which implies it only removes orders if filled, and if not filled, does nothing) and doing all the invariant checking in the PCS callsite.

LMK if I'm missing your point, but isn't the function already called `RemoveOrderIfFilled`?

Overall agree about the invariant checking though, I also think that should be done in the PCS callsite especially since that function has the `ProcessProposerMatchesEvents` variables defined. Ideally functions like this return a boolean indicating whether the order was removed or not, which can allow `PurgeInvalidMemclobState` to perform invariant checks

protocol/x/clob/memclob/memclob.go
lucas-dydx: Updated, lmk what you think!

protocol/x/clob/memclob/memclob.go
jonfung-dydx: > If the issue doesn't cause traders to lose money

Yeah, this school of thought sounds fine to me.

> Isn't the function already called RemoveOrderIfFilled?

Yep, it's already called `RemoveOrderIfFilled`, was commenting on the fact that the fn name lends itself to having invariant checks done in PCS

PR #703, opened 2023-10-25
[CLOB-930, CORE-29] Extend the e2e test framework to start the cosmos app using cmd.

lcwik:
### Changelist
 Extend the e2e test framework to start the cosmos app using cmd.

This works towards enabling and passing in flags for flag based testing. This also works towards enabling daemons on e2e app start-up once the daemons can be cleaned up. This relies on https://github.com/dydxprotocol/cosmos-sdk/pull/29

### Test Plan
Existing tests updated and pass.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testutil/app/app.go
+++ b/protocol/testutil/app/app.go
@@ -761,6 +761,8 @@ func launchValidator(
 		"false",
 		"--liquidation-daemon-enabled",
 		"false",
+		"--bridge-daemon-eth-rpc-endpoint",
+		"https://eth-sepolia.g.alchemy.com/v2/demo",
 	})
 
 	ctx := svrcmd.CreateExecuteContext(parentCtx)
--- a/protocol/testutil/app/app.go
+++ b/protocol/testutil/app/app.go
@@ -2,12 +2,21 @@ package app
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	tmcfg "github.com/cometbft/cometbft/config"
+	tmcli "github.com/cometbft/cometbft/libs/cli"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/server"
+	svrcmd "github.com/cosmos/cosmos-sdk/server/cmd"
+	srvtypes "github.com/cosmos/cosmos-sdk/server/types"
+	"github.com/dydxprotocol/v4-chain/protocol/cmd/dydxprotocold/cmd"
 	"math"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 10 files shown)


protocol/testutil/app/app.go
```diff
@@ -603,6 +660,135 @@ func (tApp *TestApp) PrepareProposal() abcitypes.ResponsePrepareProposal {
...
+			for key, value := range appOptions {
+				serverCtxPtr.Viper.Set(key, value)
+			}
+
+			// Set the test logger instance based upon AppOptions.
+			if logger, ok := appOptions[testlog.LoggerInstanceForTest]; ok {
+				serverCtxPtr.Logger = logger.(log.Logger)
+			}
+		},
+		// Override the addresses to use domain sockets to avoid port conflicts.
```
clemire, line 717: niiiiice

protocol/testutil/app/app.go
```diff
@@ -603,6 +660,135 @@ func (tApp *TestApp) PrepareProposal() abcitypes.ResponsePrepareProposal {
...
+	case a = <-appCaptor:
+	case err = <-done:
+		// Send the error to done channel so that `Cleanup` function will not block.
+		cancelFn()
+		done <- err
+		return "", nil, nil, err
+	}
+
+	shutdownFn = func() error {
+		cancelFn()
```
clemire, line 785: Why are we maybe calling `cancelFn` twice?

protocol/testutil/app/app.go
```diff
@@ -603,6 +660,135 @@ func (tApp *TestApp) PrepareProposal() abcitypes.ResponsePrepareProposal {
...
+// shuts down.
+func launchValidator(
+	genesis types.GenesisDoc,
+	appOptions map[string]interface{},
+) (homeDir string, a *app.App, shutdownFn func() error, err error) {
+	// Create the validators home directory as a temporary directory and fill it with:
+	//  - config/priv_validator_key.json
+	//  - config/node_key.json
+	//  - config/genesis.json
+	validatorHomeDir := filepath.Join(os.TempDir(), fmt.Sprint(time.Now().UnixNano()))
```
clemire, line 674: perhaps silly question, but: is it possible we'll ever get conflicts here? Would something like a uuid4 work better?

lcwik: Its unlikely since I'm using nanoseconds.

protocol/testutil/app/app.go
lcwik: Its a different one since the names are limited to the enclosing block.

clemire: ICIC, go case statements do not fall through.

clemire (approved):

protocol/testutil/app/app.go
lcwik: that is right

PR #704, opened 2023-10-25
[CLOB-930] Run three other instances of the application during e2e tests that check for non-determinism

lcwik:
### Changelist
Add 3 additional validator invariant checkers (PR) for almost all existing e2e tests:
* parallelApp is responsible for seeing all CheckTx requests, block proposals, blocks, and RecheckTx requests. This allows it to detect state differences due to inconsistent in-memory structures (for example iteration order in maps).
* noCheckTxApp is responsible for seeing all block proposals and blocks. This allows it to simulate a validator that never received any of the CheckTx requests and that it will still accept blocks and arrive at the same state hash.
* crashingApp is responsible for restarting before processing a block and sees all CheckTx requests, block proposals, and blocks. This allows it to check that in memory state can be restored successfully on application restart and that it will accept a block after a crash and arrive at the same state hash.

### Test Plan
Existing tests pass beyond the few places where I had to disable non-determinism checks in certain tests that mutate keeper state directly in a way where non-determinism would be falsely detected or they rely on in-memory state that can't survive app restart.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"math"
 	"math/big"
@@ -212,6 +213,8 @@ type App struct {
 	appCodec          codec.Codec
 	txConfig          client.TxConfig
 	interfaceRegistry types.InterfaceRegistry
+	db                dbm.DB
+	snapshotDB        dbm.DB
 
 	// keys to access the substores
 	keys    map[string]*storetypes.KVStoreKey
@@ -298,6 +301,7 @@ func assertAppPreconditions() {
 func New(
 	logger log.Logger,
 	db dbm.DB,
+	snapshotDB dbm.DB,
 	traceStore io.Writer,
 	loadLatest bool,
 	appOpts servertypes.AppOptions,
@@ -369,6 +373,8 @@ func New(
 		keys:              keys,
 		tkeys:             tkeys,
 		memKeys:           memKeys,
+		db:                db,
+		snapshotDB:        snapshotDB,
 	}
 
 	app.ParamsKeeper = initParamsKeeper(appCodec, cdc, keys[paramstypes.StoreKey], tkeys[paramstypes.TStoreKey])
@@ -1424,7 +1430,11 @@ func (app *App) Close() error {
 	if app.Server != nil {
 		app.Server.Stop()
 	}
-	return nil
+	return errors.Join(
+		// TODO(CORE-538): Remove this if possible during upgrade to Cosmos 0.50.
+		app.db.Close(),
+		app.snapshotDB.Close(),
+	)
 }
 
 // RegisterSwaggerAPI registers swagger route with API Server
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/debug"
+	"sync"
 
 	autocliv1 "cosmossdk.io/api/cosmos/autocli/v1"
 	reflectionv1 "cosmossdk.io/api/cosmos/reflection/v1"
@@ -215,6 +216,7 @@ type App struct {
 	interfaceRegistry types.InterfaceRegistry
 	db                dbm.DB
 	snapshotDB        dbm.DB
+	closeOnce         func() error
 
 	// keys to access the substores
 	keys    map[string]*storetypes.KVStoreKey
@@ -376,6 +378,21 @@ func New(
 		db:                db,
 		snapshotDB:        snapshotDB,
 	}
+	app.closeOnce = sync.OnceValue[error](
+		func() error {
+			if app.PriceFeedClient != nil {
+				app.PriceFeedClient.Stop()
+			}
+			if app.Server != nil {
+				app.Server.Stop()
+			}
+			return errors.Join(
+				// TODO(CORE-538): Remove this if possible during upgrade to Cosmos 0.50.
+				app.db.Close(),
+				app.snapshotDB.Close(),
+			)
+		},
+	)
 
 	app.ParamsKeeper = initParamsKeeper(appCodec, cdc, keys[paramstypes.StoreKey], tkeys[paramstypes.TStoreKey])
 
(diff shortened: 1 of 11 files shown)


protocol/app/app.go
```diff
@@ -212,6 +213,8 @@ type App struct {
 	appCodec          codec.Codec
 	txConfig          client.TxConfig
 	interfaceRegistry types.InterfaceRegistry
+	db                dbm.DB
+	snapshotDB        dbm.DB
```
teddyding, line 217: For my understanding, where/how is this `snapshotDB` used?

protocol/testing/e2e/funding/funding_e2e_test.go
```diff
@@ -141,11 +141,15 @@ func TestFunding(t *testing.T) {
...
 		t.Run(name, func(t *testing.T) {
-			tApp := testapp.NewTestAppBuilder(t).WithGenesisDocFn(func() (genesis types.GenesisDoc) {
-				genesis = testapp.DefaultGenesis()
-				genesis.GenesisTime = GenesisTime
-				return genesis
-			}).Build()
+			tApp := testapp.NewTestAppBuilder(t).
+				// UpdateIndexPrice only contacts the tApp.App.Server causing non-determinism in the
+				// other App instances in TestApp used for non-determinism checking.
+				WithNonDeterminismChecksEnabled(false).
```
teddyding, line 147: For e2e tests that require setting up in-memory state, instead of disabling non-determinism checks altogether (zero coverage), would it make sense to apply the required in-memory setup to the 3 new instances (at least some coverage)? 

For instance, allow test cases to pass in `func SetupApp(app *app.App)` so changes like `pricefeed_testutil.UpdateIndexPrice(` can be applied to not just `tApp.App` but the 3 other instances. 

In the case of the funding e2e test, the better way is probably to overwrite `Prepare/ProcessProposal` to accept the new oracle price, but unsure if other tests could benefit from above.

protocol/testutil/app/app.go
```diff
@@ -273,18 +276,44 @@ type ExecuteCheckTxs func(ctx sdk.Context, app *app.App) (stop bool)
...
-//   - DefaultTestAppCreatorFn with no custom flags
+//   - no custom flags
 //   - an ExecuteCheckTxs function that will stop on after the first block
+//   - non-determinism checks enabled
+//
+// Note that the TestApp instance will have 3 non-determinism state checking apps:
+//   - `parallelApp` is responsible for seeing all CheckTx requests, block proposals, blocks, and RecheckTx requests.
+//     This allows it to detect state differences due to inconsistent in-memory structures (for example iteration order
+//     in maps).
+//   - `noCheckTxApp` is responsible for seeing all block proposals and blocks. This allows it to simulate a validator
```
teddyding, line 289: In similar spirit, would it make sense to have a `noisyCheckTxApp` that just takes arbitrary `checkTx` that aren't sent to the main `tApp.App`? This is getting closer to fuzz testing but could be valuable.

protocol/testing/e2e/funding/funding_e2e_test.go
lcwik: For things that mutate keepers, we should rely on changing the tests to use genesis/CheckTx instead.

For queries, we could expose new capabilities like how TestApp.CheckTx is invoked, we could have something like TestApp.ServerMsg that invokes it on all the instances but I'm pretty undecided as to what to do here as this change moves the needle pretty far.

teddyding:
https://github.com/dydxprotocol/v4-chain/actions/runs/6671376530/job/18134582211?pr=704 simapp failing and I can reproduce locally. Seems related to DB closing?

protocol/testutil/app/app.go
lcwik: Yeah, we could totally do that. I don't think we are limited to 3 invariant checkers. Feel free to file a ticket with additional invariants that you can think of.

protocol/app/app.go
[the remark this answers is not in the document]
lcwik: app.Close() is the location where we should be shutting down resources on application shutdown.

The DB backs the snapshot store and Cosmos SDK doesn't do this for us so we need to shut it down.

protocol/testutil/app/app.go
```diff
@@ -382,49 +455,181 @@ func (tApp *TestApp) initChainIfNeeded() {
 	originalServerStartTime := srvtypes.ServerStartTime.Load()
 	srvtypes.ServerStartTime.Store(int64(time.Millisecond * 10))
 
-	validatorHomeDir, app, shutdownFn, err := launchValidator(tApp.genesis, tApp.builder.appOptions)
-	if err != nil {
-		tApp.builder.t.Fatal(err)
-	}
-	tApp.App = app
+	// Launch the main instance of the application
+	{
```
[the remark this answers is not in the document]
lcwik: The issue that I didn't find a good way to handle the `restart fn` since I rely on updating the references that are captured outside which are shared between the `restart fn` and `cleanup fn`.

I filed https://linear.app/dydx/issue/CORE-721

lcwik:
> https://github.com/dydxprotocol/v4-chain/actions/runs/6671376530/job/18134582211?pr=704 simapp failing and I can reproduce locally. Seems related to DB closing?

Turns out that `NewSimApp` is calling App.Close() which is shutting the db down and the call of sim app is also adding a `db.Close()` which is causing a double close. I removed the second close and guarded the Close() call so it can only have effect once. I also updated the test to use clean-up instead of defer to ensure that the Close() call can happen to shutdown the DB before we try to delete everything.

protocol/testing/e2e/funding/funding_e2e_test.go
teddyding: > For things that mutate keepers, we should rely on changing the tests to use genesis/CheckTx instead.

For tests where oracle prices need to be be updated more than once during the lifespan, `genesis` isn't sufficient. I guess in this case we should use `DeliverTxsOverride` to force a `MsgUpdatePrice` transaction?

protocol/testutil/app/app.go
teddyding: Filed https://linear.app/dydx/issue/CLOB-998/create-a-noisychecktx-app-instance-in-e2e-test-suite

teddyding (approved):

protocol/testing/e2e/funding/funding_e2e_test.go
lcwik: We could ensure that they source the same pricing information.

PR #708, opened 2023-10-26
Duplicate validation checking for ProcessProposerMatchesEvents

jonfung-dydx:
Panic if PPME is ever set with duplicate order ids.

### DIFF
--- a/protocol/x/clob/keeper/process_proposer_matches_events.go
+++ b/protocol/x/clob/keeper/process_proposer_matches_events.go
@@ -1,8 +1,6 @@
 package keeper
 
 import (
-	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
@@ -36,10 +34,6 @@ func (k Keeper) MustSetProcessProposerMatchesEvents(
 	processProposerMatchesEvents types.ProcessProposerMatchesEvents,
 ) {
 	lib.AssertDeliverTxMode(ctx)
-	if ctx.BlockHeight() != int64(processProposerMatchesEvents.BlockHeight) {
-		panic(fmt.Errorf("block height %d for ProcessProposerMatchesEvents does not equal current block height %d",
-			processProposerMatchesEvents.BlockHeight, ctx.BlockHeight()))
-	}
 
 	if err := processProposerMatchesEvents.ValidateProcessProposerMatchesEvents(); err != nil {
 		panic(err)
--- a/protocol/x/clob/keeper/process_proposer_matches_events.go
+++ b/protocol/x/clob/keeper/process_proposer_matches_events.go
@@ -28,6 +28,7 @@ func (k Keeper) GetProcessProposerMatchesEvents(ctx sdk.Context) types.ProcessPr
 // This function panics if:
 //   - the current block height does not match the block height of the ProcessProposerMatchesEvents
 //   - called outside of deliver TX mode
+//   - Any of the ProcessProposerMatchesEvents fields have duplicates.
 //
 // TODO(DEC-1281): add parameter validation.
 func (k Keeper) MustSetProcessProposerMatchesEvents(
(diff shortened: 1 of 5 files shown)


protocol/x/clob/memclob/memclob.go
```diff
@@ -442,17 +442,15 @@ func (m *MemClobPriceTimePriority) PlaceOrder(
...
-			if _, found := m.openOrders.getOrder(ctx, orderId); found {
-				if message, success := off_chain_updates.CreateOrderRemoveMessageWithReason(
-					m.clobKeeper.Logger(ctx),
-					orderId,
-					indexershared.OrderRemovalReason_ORDER_REMOVAL_REASON_REPLACED,
-					off_chain_updates.OrderRemoveV1_ORDER_REMOVAL_STATUS_BEST_EFFORT_CANCELED,
-				); success {
-					offchainUpdates.AddRemoveMessage(orderId, message)
-				}
+		orderId := order.OrderId
```
jonfung-dydx, line 445: I believe that since order is a `types.Order`, it can't be a liquidation order. Removing this if statement.

protocol/x/clob/types/process_proposer_matches_events.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
+)
+
+// ValidateProcessProposerMatchesEvents performs basic stateless validation on ProcessProposerMatchesEvents.
+// It returns an error if:
+// - Any of the fields have duplicate OrderIds.
+func (ppme *ProcessProposerMatchesEvents) ValidateProcessProposerMatchesEvents() error {
```
jonfung-dydx, line 12: If anyone has additional validation, lmk. I think there can be duplicates between the fields since we do all the filtering out (i.e filtering out placed, then cancelled stateful orders in the same block) when we read out PPME. 

In the future I think it would be nice to have specific update functions for each field so we don't have to validate all fields when we're only updating one subfield.

It would also be nice to do all the filtering out upon update instead of upon readout of PPME.

protocol/x/clob/types/process_proposer_matches_events.go
```diff
@@ -0,0 +1,56 @@
...
+		)
+	}
+	if lib.ContainsDuplicates(ppme.PlacedConditionalOrderIds) {
+		return fmt.Errorf(
+			"ProcessProposerMatchesEvents contains duplicate PlacedConditionalOrderIds: %+v",
+			ppme.PlacedConditionalOrderIds,
+		)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jonfung-dydx: How can you just assume the `lib.ContainsDuplicates` function is likely to have a O(n^2) complexity? Could you look at the `lib.ContainsDuplicates` function and let me know the runtime?

protocol/x/clob/keeper/process_proposer_matches_events.go
```diff
@@ -40,6 +41,10 @@ func (k Keeper) MustSetProcessProposerMatchesEvents(
 			processProposerMatchesEvents.BlockHeight, ctx.BlockHeight()))
 	}
 
+	if err := processProposerMatchesEvents.ValidateProcessProposerMatchesEvents(); err != nil {
```
lucas-dydx, line 44: Should we move the block height validation on lines 39-42 within this function as well?

protocol/x/clob/types/process_proposer_matches_events.go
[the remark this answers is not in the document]
jonfung-dydx: What files do you have read access to when reviewing this PR?

protocol/x/clob/types/process_proposer_matches_events.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
+)
+
+// ValidateProcessProposerMatchesEvents performs basic stateless validation on ProcessProposerMatchesEvents.
+// It returns an error if:
+// - Any of the fields have duplicate OrderIds.
```
lucas-dydx, line 11: Is it worth noting that this is currently invalid since stateful order replacements are not enabled?

protocol/x/clob/types/process_proposer_matches_events.go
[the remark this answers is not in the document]
jonfung-dydx:
```
// ContainsDuplicates returns true if the slice contains duplicates, false if not.
func ContainsDuplicates[V comparable](values []V) bool {
	for _, val := range values {
		for _, val2 := range values {
			for _, val3 := range values {
				for _, val4 := range values {
					for _, val5 := range values {
						for _, val6 := range values {
							if val == val2 && val2 == val3 && val3 == val4 && val4 == val5 && val5 == val6 {
								return true
							}
						}
					}
				}
			}
		}
	}
	return false
}
```
Here is the implementation of `lib.ContainsDuplicates`. Could you please review this code again with this in mind?

lucas-dydx (approved):

PR #715, opened 2023-10-27
[CORE-666] - Robust health check

clemire:
### Changelist
Add HealthCheck interface along with an implementation, and migrate the pricefeed daemon to conform to the interface.

The pricefeed daemon actually does expect to update on a regular cadence, since the update loop does not do any work / computation besides iterating over a cache - so we continue to use this logic to determine daemon health.

Note: I removed the panic that was in the pricefeed daemon on invalid updates. We'll consolidate this logic and rely on this panic to be triggered by the update monitor instead. This also gives users the future option to disable panics for the pricefeed daemon, if they'd like, once we add flag support for delaying or disabling panics on daemon failure.

Upcoming PRs:
- migrate liquidations with appropriate health-check logic
- migrate bridge with appropriate health-check logic
- update the UpdateMonitor code to use the HealthCheck interface instead of relying on daemon response frequency to detect errors.

### Test Plan
Existing tests
- added a new HealthCheck unit tests that validates expected daemon health state after a successful or failed price update.

### Author/Reviewer Checklist


protocol/daemons/pricefeed/client/client.go
```diff
@@ -49,6 +50,33 @@ type Client struct {
...
+// Ensure Client implements the DaemonClient interface.
+var _ daemontypes.DaemonClient = (*Client)(nil)
+
+// HealthCheck returns an error if the daemon is unhealthy.
+// The daemon is considered unhealthy if it is not producing non-empty price updates. We expect a short period of time
+// to pass on daemon startup where the daemon is unhealthy as the price cache warms up.
+// This method is synchronized by isHealthy, which is an atomic.Bool.
+func (c *Client) HealthCheck(_ context.Context) error {
+	if !c.isHealthy.Load() {
+		return errors.New(
```
lcwik, line 69: Consider using a `mutex` around an `error` so that you can provide details around the error.

This would allow you to differentiate between an error due to initialization failure or due to price update issues.

protocol/daemons/types/daemon_client.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+type DaemonClient interface {
```
lcwik, line 3: nit: I would only add the `DaemonClient` once it is necessary.

Also, `dydx.App` is likely to know the concrete types of the `daemons` so it won't need an interface type to call stop on so it is unlikely we would need the `DaemonClient` interface any time soon.

protocol/daemons/pricefeed/client/client_test.go
```diff
@@ -721,6 +727,89 @@ func TestPriceUpdater_Mixed(t *testing.T) {
...
+		time.Sleep(100 * time.Millisecond)
+		ticker.Stop()
+		stop <- true
+	}()
+
+	return ticker, stop
+}
+
+func TestHealthCheck(t *testing.T) {
+	// Setup.
```
lcwik, line 754: nit: convert this into test cases like in `TestMarketUpdater_Mixed`

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import "context"
+
+// HealthCheckable is a common interface for services that can be health checked.
+type HealthCheckable interface {
+	// HealthCheck returns an error if a service is unhealthy. This method takes
+	// a context to allow the caller to use a context-based timeout with the call.
+	// If the service is healthy, this method returns nil.
+	HealthCheck(ctx context.Context) (err error)
```
ttl33, line 10: What should be the behavior if the daemon is in STOPPED state? Or is this not a concern because `HealthCheck` method won't be able to get called at all?

protocol/daemons/pricefeed/client/sub_task_runner.go
```diff
@@ -87,8 +91,17 @@ func (s *SubTaskRunnerImpl) StartPriceUpdater(
 		select {
 		case <-ticker.C:
 			err := RunPriceUpdaterTaskLoop(ctx, exchangeToMarketPrices, priceFeedServiceClient, logger)
-			if err != nil {
-				panic(err)
+
+			if err == nil {
+				// Record update success for the daemon health check.
+				c.RecordUpdateSuccess(&libtime.TimeProviderImpl{})
```
lcwik, line 97: nit: Your creating and destroying the `TimeProviderImpl` on every call. Adding it as a member variable to `SubTaskRunnerImpl` would allow setting it for testing purposes in the future or if you take my suggestion below about taking a time then you could pass in `time.Now()`.

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+	HealthCheck(provider libtime.TimeProvider) (err error)
+}
+
+// HealthCheckableImpl implements the HealthCheckable interface by tracking the timestamps of the last successful and
+// failed updates. If:
+// - no update has occurred
+// - the most recent update failed, or
+// - the daemon has not seen a successful update in at least 5 minutes,
+// the service is considered unhealthy.
+type HealthCheckableImpl struct {
```
lcwik, line 26: don't expose this, instead have the interface expose what methods you want like
`HealthCheck` and `Report*` and have `healthCheckImpl` implement them. This way the internal implementation doesn't matter.

This also allows you to merge instantiation (via `NewHealthCheckableImpl`) and initialization (via `InitializeHealthStatus`) into a single method that instantiates and initializes.

If you don't want to expose the `Report*` on the `HealthCheckable` then you could have two interfaces, one which is the `HealthCheckable` and `HealthReporter` that splits them up but have `HealthCheckableImpl` implement both of them.

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+func (h *HealthCheckableImpl) HealthCheck(timeProvider libtime.TimeProvider) error {
+	h.Lock()
+	defer h.Unlock()
+
+	if !h.initialized {
+		panic("HealthCheckableImpl has not been initialized")
+	}
+
+	if h.lastSuccessfulUpdate.IsZero() {
+		return fmt.Errorf("no successful update has occurred")
```
lcwik, line 92: This error message should state that there has been no successful update and also provide the current error if there has been one.

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if h.lastFailedUpdate.After(h.lastSuccessfulUpdate) {
+		return fmt.Errorf(
+			"last update failed at %v with error: %w",
+			h.lastFailedUpdate,
```
lcwik, line 98: Should we include when the last success was?

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+			h.lastUpdateError,
+		)
+	}
+
+	// If the last successful update was more than 5 minutes ago, report the specific error.
+	if timeProvider.Now().Sub(h.lastSuccessfulUpdate) > MaximumAcceptableUpdateDelay {
+		return fmt.Errorf(
+			"last successful update occurred at %v, which is more than %v ago",
```
lcwik, line 106: should we include the last error and when it occurred to if it had?

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+	h.lastUpdateError = err
+}
+
+// HealthCheck returns an error if a service is unhealthy.
+// The service is unhealthy if any of the following are true:
+// - no successful update has occurred
+// - the most recent update failed, or
+// - the daemon has not seen a successful update in at least 5 minutes,
+// This method is synchronized.
+func (h *HealthCheckableImpl) HealthCheck(timeProvider libtime.TimeProvider) error {
```
lcwik, line 83: You might as well just take the time itself in all these methods since the caller is either a test or something that should take/have a `libtime.TimeProvider`.

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+	// lastUpdateError is the error describing the failure reason for the last failed update.
+	lastUpdateError error
+
+	// initialized is true if the HealthCheckableImpl has been initialized. If false, the service will panic
+	// when the HealthCheck method is called. This error should occur during or soon after app initialization.
+	initialized bool
+}
+
+// NewHealthCheckableImpl creates a new HealthCheckableImpl instance.
+func NewHealthCheckableImpl(daemon string, timeProvider libtime.TimeProvider) *HealthCheckableImpl {
```
lcwik, line 43: You could rename this to `NewTimeBoundedHealthCheckable`.

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+func (h *HealthCheckableImpl) InitializeHealthStatus(serviceName string, timeProvider libtime.TimeProvider) {
+	h.RecordUpdateFailure(timeProvider, fmt.Errorf("%v is initializing", serviceName))
+
+	h.Lock()
+	defer h.Unlock()
+	h.initialized = true
+}
+
+// RecordUpdateSuccess records a successful update. This method is synchronized.
+func (h *HealthCheckableImpl) RecordUpdateSuccess(timeProvider libtime.TimeProvider) {
```
lcwik, line 61: nit: RecordUpdateSuccess -> ReportSuccess
RecordUpdateFailure -> ReportFailure

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+	// HealthCheck returns an error if a service is unhealthy. If the service is healthy, this method returns nil.
+	HealthCheck(provider libtime.TimeProvider) (err error)
+}
+
+// HealthCheckableImpl implements the HealthCheckable interface by tracking the timestamps of the last successful and
+// failed updates. If:
+// - no update has occurred
+// - the most recent update failed, or
+// - the daemon has not seen a successful update in at least 5 minutes,
+// the service is considered unhealthy.
```
ttl33, line 25:
```suggestion
// failed updates.
// If any of the following occurs, then the service should be considered unhealthy:
// - no update has occurred
// - the most recent update failed, or
// - the daemon has not seen a successful update within `MaxAcceptableUpdateDelay`,
```

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	libtime "github.com/dydxprotocol/v4-chain/protocol/lib/time"
+	"sync"
+	"time"
+)
+
+const (
+	MaximumAcceptableUpdateDelay = 5 * time.Minute
```
ttl33, line 11:
```suggestion
	MaxAcceptableUpdateDelay = 5 * time.Minute
```

protocol/daemons/types/health_checkable.go
```diff
@@ -0,0 +1,113 @@
...
+// - the daemon has not seen a successful update in at least 5 minutes,
+// the service is considered unhealthy.
+type HealthCheckableImpl struct {
+	sync.Mutex
+
+	// lastSuccessfulUpdate is the timestamp of the last successful update.
+	lastSuccessfulUpdate time.Time
+	// lastFailedUpdate is the timestamp of the last failed update. After the HealthCheckableImpl is initialized,
+	// this should never be a zero value.
+	lastFailedUpdate time.Time
```
ttl33, line 33: nit: I assume that `lastFailedUpdate` and `lastUpdateError` will be updated together. It could be useful to define a new struct to couple them

```
type updateTimeAndError struct {
  ...
}
```

protocol/daemons/pricefeed/client/client.go
```diff
@@ -49,8 +49,14 @@ type Client struct {
 
 	// Ensure stop only executes one time.
 	stopDaemon sync.Once
+
+	// include HealthCheckableImpl to track the health of the daemon.
+	daemontypes.HealthCheckableImpl
```
ttl33, line 54: nit: move this to the top of `Client` so it's clear that this struct implements a client / has an anonymous field

protocol/daemons/types/health_checkable.go
clemire: In the stopped state, the last successful update timestamp will never be refreshed and the daemon will become unhealthy after a waiting period of 5 min.

protocol/daemons/pricefeed/client/sub_task_runner.go
```diff
@@ -87,8 +91,17 @@ func (s *SubTaskRunnerImpl) StartPriceUpdater(
 		select {
 		case <-ticker.C:
 			err := RunPriceUpdaterTaskLoop(ctx, exchangeToMarketPrices, priceFeedServiceClient, logger)
-			if err != nil {
-				panic(err)
+
+			if err == nil {
```
ttl33, line 95: Should something similar happen for the following subtasks too?
* `StartPriceEncoder`
* `StartPriceFetcher`
* `StartMarketParamUpdater`

Why or why not?

clemire: - StartPriceEncoder
- StartPriceFetcher

No, because I think measuring the contents of the market update already captures more or less the same information, and is much simpler, than instrumenting these subtasks as well.

- StartMarketParamUpdater
I think we made the decision that if a market update fails, we will continue to operate the protocol with the current market configuration on the daemon so that trading can continue for existing markets. (The reasonable IMO assumption here is that market param updates are failing across the network.) Marking unhealthy and panicking on a failure here could bring down the network instead of allowing degraded performance by continuing to support trading for existing markets. The better approach would probably be to continue the network and allow another governance vote to fix any bad parameters in the exchange_config_json.

One possibility to consider is that some daemons could fail and others succeed due to disagreeing app versions, but I think we should catch this by upgrading via governance to enforce versions if there is a breaking change in daemon configuration.

ttl33: Okay, makes sense.

For `StartMarketParamUpdater` do we have clear msgs to indicate this failed? Do we have dashboard + alerting on this?

protocol/daemons/types/health_checkable_test.go
```diff
@@ -0,0 +1,163 @@
...
+	tests := map[string]struct {
+		updates []struct {
+			timestamp time.Time
+			// leave error nil for a successful update
+			err error
+		}
+		healthCheckTime      time.Time
+		expectedHealthStatus error
+	}{
+		"unhealthy: no updates": {
```
ttl33, line 44:
```suggestion
		"unhealthy: no updates, returns initializing error": {
```

protocol/daemons/types/health_checkable_test.go
```diff
@@ -0,0 +1,163 @@
...
+				timestamp time.Time
+				err       error
+			}{
+				{Time1, TestError}, // failed update
+				{Time2, nil},       // successful update
+			},
+			healthCheckTime:      Time3,
+			expectedHealthStatus: nil, // expect healthy
+		},
+		"unhealthy: last successful update was more than 5 minutes ago": {
```
ttl33, line 103:
```suggestion
		"unhealthy: last successful update was more than max delay": {
```

ttl33 (approved):

clemire:
... for failed market param updates, yes we have monitoring, error logs and dashboard. monitor [here](https://app.datadoghq.com/monitors/124945670)

CodeRabbitUser:
CodeRabbit experienced a temporary issue resulting in some of the walkthrough comments getting garbled. This should be resolved with subsequent commits to the PR.

PR #719, opened 2023-10-27
Make iaviewer usage more human friendly

roy-dydx:
### Changelist
- Migrate cli to use cobra
- Add `data-all` command to list iavl tree for all modules
- Unmarshal proto values and output human readable strings if they are in the registry

### Test Plan
- Manually tried all commands on a db from localnet

### Author/Reviewer Checklist

### DIFF
--- a/protocol/scripts/iaviewer/iaviewer.go
+++ b/protocol/scripts/iaviewer/iaviewer.go
@@ -6,14 +6,13 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"os"
 	"path/filepath"
-	"strconv"
+	"regexp"
 	"strings"
 
 	dbm "github.com/cometbft/cometbft-db"
-
 	"github.com/cosmos/iavl"
+	"github.com/spf13/cobra"
 )
 
 // TODO: make this configurable?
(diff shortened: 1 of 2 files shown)


protocol/scripts/iaviewer/unmarshaller.go
```diff
@@ -0,0 +1,28 @@
...
+
+func protoUnmarshaller[M codec.ProtoMarshaler](b []byte) string {
+	cdc := app.GetEncodingConfig().Codec
+	var m M
+	// TODO: avoid reflection?
+	msgType := reflect.TypeOf(m).Elem()
+	m = reflect.New(msgType).Interface().(M)
+	cdc.MustUnmarshal(b, m)
+	return m.String()
+}
```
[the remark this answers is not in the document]
roy-dydx: I want this to panic for now.

protocol/scripts/iaviewer/unmarshaller.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func protoUnmarshaller[M codec.ProtoMarshaler](b []byte) string {
+	cdc := app.GetEncodingConfig().Codec
+	var m M
+	// TODO: avoid reflection?
+	msgType := reflect.TypeOf(m).Elem()
+	m = reflect.New(msgType).Interface().(M)
+	cdc.MustUnmarshal(b, m)
+	return m.String()
```
jayy04, line 18: Q: should we return string here and use %s? or should we return `m` and use %+v?

not sure if %+v works

protocol/scripts/iaviewer/iaviewer.go
```diff
@@ -8,59 +8,177 @@ import (
...
-		fmt.Printf("Size: %X\n", tree.Size())
-	case "shape":
-		PrintShape(tree)
-	case "versions":
-		PrintVersions(tree)
+func AddVersionFlag(cmd *cobra.Command) {
+	cmd.Flags().Uint64("version", 0, "version of the tree (default: latest version)")
+}
+
+func DataAllCmd() *cobra.Command {
```
jayy04, line 65: for my understanding - how did this work before `prefix` option was added? kind of surprised that this is not already supported

jayy04 (approved):

protocol/scripts/iaviewer/iaviewer.go
roy-dydx: It doesn't work. If you don't pass prefix in, it will assume that all the leveldb keys are unprefixed and try to interpret it as an iavl tree. But for dbs created by cosmos, all the keys are prefixed and not ready to be interpreted as iavl tree unless the prefix is removed. So you'd just find no values unless you pass in a prefix of a tree.

protocol/scripts/iaviewer/unmarshaller.go
roy-dydx: The functions would have generic return value which results in different return values for each proto. We need all the return values to be the same so we can put them in a map like we're doing now.

protocol/scripts/iaviewer/iaviewer.go
jayy04: 🤯

protocol/scripts/iaviewer/unmarshaller.go
jayy04: you can use `any`? 

not a big deal though

roy-dydx: Then to use `String()` I'd need to type assert which would require another generic function. (or maybe reflection works?)

I don't know how %+v works, but I don't think it would work on `interface{}`.

PR #736, opened 2023-11-02
[IND-460] Emit deleveraging events from protocol

dydxwill:
### Changelist
Emit deleveraging on-chain events from protocol
Add skeleton for Ender handling/validation

### Test Plan
Unit tests

### Author/Reviewer Checklist


protocol/x/clob/keeper/liquidations.go
```diff
@@ -152,6 +155,27 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+						subaccountId,
+						fill.OffsettingSubaccountId,
+						clobPairId,
+						satypes.BaseQuantums(fill.FillAmount),
+						uint64(subticks),
+						isBuy,
+					),
+				),
+			)
+		}
```
vincentwschau, line 178: Are there any tests we need to update to check that these new events are correctly emitted?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -152,6 +155,27 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 			)
 			return err
 		}
+		clobPairId := k.mustGetClobPairForPerpetualId(ctx, perpetualId).Id
+		subticks := liquidationOrder.GetOrderSubticks()
+		isBuy := liquidationOrder.IsBuy()
+		// Send on-chain deleveraging event for each fill.
+		for _, fill := range fills {
+			k.GetIndexerEventManager().AddTxnEvent(
```
vincentwschau, line 163: @jayy04 Is there any preference on whether to emit the event here as compared to where the CometBFT events are emitted ([here](https://github.com/dydxprotocol/v4-chain/blob/f892a43fccdad4bcdf4e1b79fc122680bbc888ab/protocol/x/clob/keeper/deleveraging.go#L436))?

Christopher-Li (approved):
LGTM but wait on @vincentwschau and @jayy04 's review

proto/dydxprotocol/indexer/events/events.proto
```diff
@@ -165,6 +165,28 @@ message OrderFillEventV1 {
...
+// the amount filled.
+message DeleveragingEventV1 {
+  // ID of the subaccount that was liquidated.
+  dydxprotocol.indexer.protocol.v1.IndexerSubaccountId liquidated = 1
+      [ (gogoproto.nullable) = false ];
+  // ID of the subaccount that was used to offset the position.
+  dydxprotocol.indexer.protocol.v1.IndexerSubaccountId offsetting = 2
+      [ (gogoproto.nullable) = false ];
+  // The ID of the clob pair that was liquidated.
+  uint32 clob_pair_id = 3;
```
jayy04, line 179: nit: let's maybe use `perpetual_id` instead of clob_pair_id here.

CLOB actually assumes a one to many relationship between perp id and clob pair id ([link](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/keeper.go#L34)). We currently only have 1 to 1 mapping, but using perpetual id is probably more resilient to changes in the future

protocol/x/clob/keeper/liquidations.go
jayy04: wait just to make sure we are on the same page - this is happening in `PrepareCheckState` / CheckState, so deleveraging fills are not on-chain yet. I don't think we should be sending an on-chain event?

deleveraging fills, similar to other order matches, get persisted to consensus state in DeliverTx [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/keeper/process_operations.go#L619)

jayy04: and for my understanding - CometBFT events are offchain events?

vincentwschau: CometBFT events are not stored in consensus so yes they would be off-chain.
The indexer events are added to a `TransientStore` and only sent in the EndBlocker call back.

proto/dydxprotocol/indexer/events/events.proto
dydxwill: for liquidations we send both. I'm going to include both.

protocol/x/clob/keeper/process_operations_test.go
```diff
@@ -1912,7 +1912,28 @@ func setupNewMockEventManager(
 				),
 			).Once().Return()
 		}
+		if isClobMatchPerpetualDeleveraging(operation) {
+			mockIndexerEventManager.On("AddTxnEvent",
+				mock.Anything,
+				indexerevents.SubtypeDeleveraging,
+				indexerevents.DeleveragingEventVersion,
+				mock.Anything,
```
vincentwschau, line 1920: nit: Is it possible to derive the deleveraging event here from the operation?

protocol/x/clob/keeper/liquidations_test.go
```diff
@@ -2018,6 +2018,9 @@ func TestPlacePerpetualLiquidation_Deleveraging(t *testing.T) {
 			}
 
 			if tc.expectedFilledSize == 0 {
+				mockIndexerEventManager.On("AddTxnEvent",
+					mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything,
```
vincentwschau, line 2022: nit: Can the actual deleveraging event be derived here?

jayy04 (approved):

protocol/x/clob/keeper/process_operations_test.go
dydxwill: Bankruptcy price isn't exposed by the API. I could add it to the test params, but that would require some more changes. Plus, the exact deleveraging event is being tested in multiple other e2e tests. So, I think it's reasonable to omit here.

protocol/x/clob/keeper/liquidations_test.go
dydxwill: Bankruptcy price isn't exposed by the API. I could add it to the test params, but that would require some more changes. Plus, the exact deleveraging event is being tested in multiple other e2e tests. So, I think it's reasonable to omit here.

vincentwschau: Can you add comment with the above details?

protocol/x/clob/keeper/process_operations_test.go
vincentwschau: Sounds good, can we add a comment to that effect?

vincentwschau (approved):

protocol/x/clob/keeper/liquidations_test.go
dydxwill: done

protocol/x/clob/keeper/process_operations_test.go
dydxwill: done

PR #738, opened 2023-11-02
[CORE-666] - Implement liquidations healthcheck

clemire:
### Changelist
Implement the HealthCheck on the liquidations daemon, complete with refactors that both allow for easier mocking (for the sake of healthcheck unit tests) and prepare the daemon for implementing a Stop method in the future.

#### Daemon behavior
- daemon initializes to unhealthy
- daemon toggles to healthy whenever loop success occurs
 - daemon toggles to unhealthy whenever the task loop returns an error

#### Updates to timeBoundedHealthCheckable:
 - no change but FYI: this shared health tracking type still initializes as unhealthy and updates to healthy when the most recent report was successful
 - change: I removed the logic that toggles the HealthCheckable to unhealthy if the last success report was > 5 minutes old. We now log an error instead.

### Test Plan
Updated existing tests and added a new set of unit tests for health checks. Also added checks in unit tests for existing methods where health success is reported after a successful response of a paginated query.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -603,7 +603,7 @@ func New(
 			app.Server.ExpectLiquidationsDaemon(
 				daemonservertypes.MaximumAcceptableUpdateDelay(daemonFlags.Liquidation.LoopDelayMs),
 			)
-			app.LiquidationsClient = liquidationclient.NewClient()
+			app.LiquidationsClient = liquidationclient.NewClient(logger)
 			go func() {
 				if err := app.LiquidationsClient.Start(
 					// The client will use `context.Background` so that it can have a different context from
@@ -611,7 +611,6 @@ func New(
 					context.Background(),
 					daemonFlags,
 					appFlags,
-					logger,
 					&daemontypes.GrpcClientImpl{},
 				); err != nil {
 					panic(err)
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -288,7 +288,8 @@ type App struct {
 	// can correctly operate.
 	startDaemons func()
 
-	PriceFeedClient *pricefeedclient.Client
+	PriceFeedClient    *pricefeedclient.Client
+	LiquidationsClient *liquidationclient.Client
 }
 
 // assertAppPreconditions assert invariants required for an application to start.
@@ -602,8 +603,9 @@ func New(
 			app.Server.ExpectLiquidationsDaemon(
 				daemonservertypes.MaximumAcceptableUpdateDelay(daemonFlags.Liquidation.LoopDelayMs),
 			)
+			app.LiquidationsClient = liquidationclient.NewClient()
 			go func() {
-				if err := liquidationclient.Start(
+				if err := app.LiquidationsClient.Start(
 					// The client will use `context.Background` so that it can have a different context from
 					// the main application.
 					context.Background(),
--- a/protocol/app/app_test.go
+++ b/protocol/app/app_test.go
@@ -97,11 +97,17 @@ func TestAppIsFullyInitialized(t *testing.T) {
 			tApp.InitChain()
 			uninitializedFields := getUninitializedStructFields(reflect.ValueOf(*tApp.App))
 
-			// Note that the PriceFeedClient is currently hard coded as disabled in GetDefaultTestAppOptions.
-			// Normally it would be only disabled for non-validating full nodes or for nodes where the
-			// price feed client is explicitly disabled.
-			if idx := slices.Index(uninitializedFields, "PriceFeedClient"); idx >= 0 {
-				slices.Remove(&uninitializedFields, idx)
+			// Note that the daemon clients are currently hard coded as disabled in GetDefaultTestAppOptions.
+			// Normally they would be only disabled for non-validating full nodes or for nodes where any
+			// daemon is explicitly disabled.
+			expectedUninitializedFields := []string{
+				"PriceFeedClient",
+				"LiquidationsClient",
+			}
+			for _, field := range expectedUninitializedFields {
+				if idx := slices.Index(uninitializedFields, field); idx >= 0 {
+					slices.Remove(&uninitializedFields, idx)
+				}
 			}
 
 			require.Len(
--- a/protocol/daemons/liquidation/client/client.go
+++ b/protocol/daemons/liquidation/client/client.go
@@ -35,6 +35,7 @@ func NewClient(logger log.Logger) *Client {
 			&timelib.TimeProviderImpl{},
 			logger,
 		),
+		logger: logger,
 	}
 }
 
@@ -117,6 +118,7 @@ func StartLiquidationsDaemonTaskLoop(
 		select {
 		case <-ticker.C:
 			if err := s.RunLiquidationDaemonTaskLoop(
+				client,
 				ctx,
 				flags.Liquidation,
 				subaccountQueryClient,
--- a/protocol/daemons/liquidation/client/client.go
+++ b/protocol/daemons/liquidation/client/client.go
@@ -20,16 +20,20 @@ import (
 type Client struct {
 	// include HealthCheckable to track the health of the daemon.
 	daemontypes.HealthCheckable
+
+	// logger is the logger for the daemon.
+	logger log.Logger
 }
 
 // Ensure Client implements the HealthCheckable interface.
 var _ daemontypes.HealthCheckable = (*Client)(nil)
 
-func NewClient() *Client {
+func NewClient(logger log.Logger) *Client {
 	return &Client{
 		HealthCheckable: daemontypes.NewTimeBoundedHealthCheckable(
 			types.LiquidationsDaemonServiceName,
 			&timelib.TimeProviderImpl{},
+			logger,
 		),
 	}
 }
@@ -42,11 +46,10 @@ func (c *Client) Start(
 	ctx context.Context,
 	flags flags.DaemonFlags,
 	appFlags appflags.Flags,
-	logger log.Logger,
 	grpcClient daemontypes.GrpcClient,
 ) error {
 	// Log the daemon flags.
-	logger.Info(
+	c.logger.Info(
 		"Starting liquidations daemon with flags",
 		"LiquidationFlags", flags.Liquidation,
 	)
@@ -54,7 +57,7 @@ func (c *Client) Start(
 	// Make a connection to the Cosmos gRPC query services.
 	queryConn, err := grpcClient.NewTcpConnection(ctx, appFlags.GrpcAddress)
 	if err != nil {
-		logger.Error("Failed to establish gRPC connection to Cosmos gRPC query services", "error", err)
+		c.logger.Error("Failed to establish gRPC connection to Cosmos gRPC query services", "error", err)
 		return err
 	}
 	defer func() {
@@ -66,7 +69,7 @@ func (c *Client) Start(
 	// Make a connection to the private daemon gRPC server.
 	daemonConn, err := grpcClient.NewGrpcConnection(ctx, flags.Shared.SocketAddress)
 	if err != nil {
-		logger.Error("Failed to establish gRPC connection to socket address", "error", err)
+		c.logger.Error("Failed to establish gRPC connection to socket address", "error", err)
 		return err
 	}
 	defer func() {
@@ -93,7 +96,6 @@ func (c *Client) Start(
 		subaccountQueryClient,
 		clobQueryClient,
 		liquidationServiceClient,
-		logger,
 	)
 
 	return nil
@@ -110,7 +112,6 @@ func StartLiquidationsDaemonTaskLoop(
 	subaccountQueryClient satypes.QueryClient,
 	clobQueryClient clobtypes.QueryClient,
 	liquidationServiceClient api.LiquidationServiceClient,
-	logger log.Logger,
 ) {
 	for {
 		select {
@@ -123,7 +124,7 @@ func StartLiquidationsDaemonTaskLoop(
 				liquidationServiceClient,
 			); err != nil {
 				// TODO(DEC-947): Move daemon shutdown to application.
-				logger.Error("Liquidations daemon returned error", "error", err)
+				client.logger.Error("Liquidations daemon returned error", "error", err)
 				client.ReportFailure(err)
 			} else {
 				client.ReportSuccess()
--- a/protocol/daemons/liquidation/client/client.go
+++ b/protocol/daemons/liquidation/client/client.go
@@ -2,28 +2,43 @@ package client
 
 import (
 	"context"
+	"github.com/dydxprotocol/v4-chain/protocol/daemons/server/types"
+	timelib "github.com/dydxprotocol/v4-chain/protocol/lib/time"
 	"time"
 
-	"github.com/dydxprotocol/v4-chain/protocol/lib"
-
-	gometrics "github.com/armon/go-metrics"
 	"github.com/cometbft/cometbft/libs/log"
-	"github.com/cosmos/cosmos-sdk/telemetry"
-	"github.com/cosmos/cosmos-sdk/types/query"
 	appflags "github.com/dydxprotocol/v4-chain/protocol/app/flags"
 	"github.com/dydxprotocol/v4-chain/protocol/daemons/flags"
 	"github.com/dydxprotocol/v4-chain/protocol/daemons/liquidation/api"
 	daemontypes "github.com/dydxprotocol/v4-chain/protocol/daemons/types"
-	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
+
+// Client implements a daemon service client that periodically queries a gRPC server for all liquidatable subaccounts
+// and reports them back to the protocol via the daemon server.
+type Client struct {
+	// include HealthCheckable to track the health of the daemon.
+	daemontypes.HealthCheckable
+}
+
+// Ensure Client implements the HealthCheckable interface.
+var _ daemontypes.HealthCheckable = (*Client)(nil)
+
+func NewClient() *Client {
+	return &Client{
+		HealthCheckable: daemontypes.NewTimeBoundedHealthCheckable(
+			types.LiquidationsDaemonServiceName,
+			&timelib.TimeProviderImpl{},
+		),
+	}
+}
 
 // Start begins a job that periodically:
 // 1) Queries a gRPC server for all subaccounts including their open positions.
 // 2) Checks collateralization statuses of subaccounts with at least one open position.
 // 3) Sends a list of subaccount ids that potentially need to be liquidated to the application.
-func Start(
+func (c *Client) Start(
 	ctx context.Context,
 	flags flags.DaemonFlags,
 	appFlags appflags.Flags,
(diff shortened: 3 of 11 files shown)


protocol/daemons/liquidation/client/client.go
```diff
@@ -75,60 +119,105 @@ func Start(
...
 			logger.Error("Liquidations daemon returned error", "error", err)
+			client.RecordUpdateFailure(timeProvider, err)
+		} else {
+			client.RecordUpdateSuccess(timeProvider)
 		}
 	}
+}
 
-	return nil
+// CheckCollateralizationForSubaccounts queries a gRPC server using `AreSubaccountsLiquidatable`
```
clemire, line 129: There are no actual changes to the following methods. I just removed `RunLiquidationDaemonTaskLoop` from this file into `sub_task_runner.go` as a method on the `SubTaskRunner` interface. This is the same design as the pricefeed daemon and allows me to to mock the `RunLiquidationDaemonTaskLoop` method in testing.

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type SubTaskRunnerImpl struct{}
+
+// Ensure SubTaskRunnerImpl implements the SubTaskRunner interface.
+var _ SubTaskRunner = (*SubTaskRunnerImpl)(nil)
+
+// RunLiquidationDaemonTaskLoop contains the logic to communicate with various gRPC services
+// to find the liquidatable subaccount ids.
+func (s *SubTaskRunnerImpl) RunLiquidationDaemonTaskLoop(
```
clemire, line 33: This method should be a straight copy, no changes.

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,70 +85,152 @@ func Start(
...
+	s SubTaskRunner,
+	flags flags.DaemonFlags,
+	ticker *time.Ticker,
+	done <-chan bool,
 	subaccountQueryClient satypes.QueryClient,
 	clobQueryClient clobtypes.QueryClient,
 	liquidationServiceClient api.LiquidationServiceClient,
-) error {
+	logger log.Logger,
+) {
```
clemire, line 119: Added the `StartLiquidationsDaemonTaskLoop` as an entrypoint for testing that encapsulates the looping logic formerly in the `Start` method. This method takes a ticker as well as a newly introduced `done` channel that signals loop termination. 

This is similar to the design we use in the[ pricefeed daemon,](https://github.com/dydxprotocol/v4-chain/blob/755b0b928be793072d19eb3a1608e7a2503f396a/protocol/daemons/pricefeed/client/client.go#L205) which also supports terminating the daemon on command.

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +80,56 @@ func Start(
...
-		if err != nil {
-			return nil, err
-		}
-
-		for _, result := range results {
-			if result.IsLiquidatable {
-				liquidatableSubaccountIds = append(liquidatableSubaccountIds, result.SubaccountId)
+		select {
+		case <-ticker.C:
+			if err := s.RunLiquidationDaemonTaskLoop(
```
clemire, line 118: I removed RunLiquidationDaemonTaskLoop from this file into sub_task_runner.go as a method on the SubTaskRunner interface. This is the same design as the pricefeed daemon and allows me to to mock the RunLiquidationDaemonTaskLoop method in testing.

The subsequent methods below in client.go have been migrated wholesale into the sub_task_runner.go file since they are all called from the Run method. Associated tests have been moved to sub_task_runner_test.go.

protocol/daemons/liquidation/client/client_test.go
```diff
@@ -241,245 +245,94 @@ func TestRunLiquidationDaemonTaskLoop(t *testing.T) {
 	}
 }
 
-func TestGetAllSubaccounts(t *testing.T) {
```
clemire, line 244: Moved to sub_task_runner_test.go

protocol/daemons/liquidation/client/client_test.go
```diff
@@ -241,245 +245,94 @@ func TestRunLiquidationDaemonTaskLoop(t *testing.T) {
...
+	_ satypes.QueryClient,
+	_ clobtypes.QueryClient,
+	_ api.LiquidationServiceClient,
+) error {
+	f.called = true
+	return f.err
 }
 
-func TestCheckCollateralizationForSubaccounts(t *testing.T) {
+func TestHealthCheck_Mixed(t *testing.T) {
```
clemire, line 273: This diff is a bit painful! The test methods on the left were migrated out of this file. TestHealthCheck is entirely new.

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,276 @@
...
+		SubaccountIds: subaccountIds,
+	}
+	response, err := client.AreSubaccountsLiquidatable(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	return response.Results, nil
+}
+
+// SendLiquidatableSubaccountIds sends a list of unique and potentially liquidatable
```
clemire, line 111: Everything else below here is a copy.

protocol/app/app.go
```diff
@@ -603,7 +604,8 @@ func New(
 				daemonservertypes.MaximumAcceptableUpdateDelay(daemonFlags.Liquidation.LoopDelayMs),
 			)
 			go func() {
-				if err := liquidationclient.Start(
+				app.LiquidationsClient = liquidationclient.NewClient()
```
ttl33, line 607: should this live outside the `go func()` to be consistent w the price daemon [here](https://github.com/dydxprotocol/v4-chain/pull/738/files#diff-7148577e1c863205cc155a436a9c9b806e655073d37a5124da0af8b5e9370317R629)?

Also, as a follow up PR, can we create a similar method that starts the liquidation daemon like [this](https://github.com/dydxprotocol/v4-chain/blob/4bd3c34a0bf385e91bf5fb509115e19dcf9bd44a/protocol/daemons/pricefeed/client/client.go#L267)? We should log the liquidation daemon flags too

protocol/daemons/liquidation/client/client.go
```diff
@@ -2,28 +2,43 @@ package client
...
 	daemontypes "github.com/dydxprotocol/v4-chain/protocol/daemons/types"
-	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+// Client implements a daemon service client that periodically queries a gRPC server for all liquidatable subaccounts
+// and reports them back to the protocol via the daemon server.
+type Client struct {
+	// Compose HealthCheckableImpl to implement the HealthCheckable interface.
```
ttl33, line 21: update comment to be in sync w [this](https://github.com/dydxprotocol/v4-chain/pull/715/files#diff-d0cb3538fda4a5cbae8939680e81fdfeca343874dd974cc6295d8c8f4ebe507dR34)?

protocol/daemons/liquidation/client/client.go
```diff
@@ -2,28 +2,43 @@ package client
...
+		),
+	}
+}
+
 // Start begins a job that periodically:
 // 1) Queries a gRPC server for all subaccounts including their open positions.
 // 2) Checks collateralization statuses of subaccounts with at least one open position.
 // 3) Sends a list of subaccount ids that potentially need to be liquidated to the application.
-func Start(
+func (c *Client) Start(
```
ttl33, line 41: Can we make this into `StartNewClient` like price daemon [here](https://github.com/dydxprotocol/v4-chain/blob/4bd3c34a0bf385e91bf5fb509115e19dcf9bd44a/protocol/daemons/pricefeed/client/client.go#L267)? And make `Start` private?

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +80,56 @@ func Start(
...
+				flags.Liquidation,
+				subaccountQueryClient,
+				clobQueryClient,
+				liquidationServiceClient,
+			); err != nil {
+				// TODO(DEC-947): Move daemon shutdown to application.
+				logger.Error("Liquidations daemon returned error", "error", err)
+				client.ReportFailure(err)
+			} else {
+				client.ReportSuccess()
```
jayy04, line 129: for my understanding - is the client still reporting to the server? how is this implementation fundamentally different from the previous implementation?

If a task loop takes more than 5 minutes, we still report an error, right? this feels like we are effectively just setting the timeout from 3 * 1.6s = 4.8s to 5 minutes.

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +80,56 @@ func Start(
...
-func GetAllSubaccounts(
+// StartLiquidationsDaemonTaskLoop contains the logic to periodically run the liquidations daemon task.
+func StartLiquidationsDaemonTaskLoop(
 	ctx context.Context,
-	client satypes.QueryClient,
-	limit uint64,
-) (
-	subaccounts []satypes.Subaccount,
-	err error,
+	client *Client,
```
ttl33, line 105: nit: i think you added client at the top

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +80,56 @@ func Start(
...
+	s SubTaskRunner,
+	flags flags.DaemonFlags,
+	ticker *time.Ticker,
+	done <-chan bool,
```
ttl33, line 109: why do we need this? and what triggers this?

protocol/daemons/liquidation/client/client.go
clemire: Yes, if the task loop takes more than 5 minutes to complete, the daemon will enter into an unhealthy state until the update finishes - so if it extends to 6 minutes, then the daemon would be unhealthy for 1 minute. I added this implementation detail in order to capture cases where a daemon could hang/deadlock after a successful update.

I can take it out, but would ask:
- do we realistically expect the daemon to ever run for 5 minutes and, if so, would that have any deleterious effect on the protocol?
- is there a duration threshold that could work better here? It was meant to be a large enough number to be unrealistic but to capture the hanging case so that we don't silently fail, which would be a false positive for daemon health. 10 minutes or even half an hour is fine - I just want to capture this case, while also hopefully informing validators earlier than later if their nodes aren't able to make accurate proposals.

protocol/daemons/liquidation/client/client.go
clemire: `StartNewClient` itself is not necessarily the best design - we may wish to make the Start method blocking on the pricefeed daemon as well, and `StartNewClient` does not return any error result on start failure. I just didn't do this the last time I touched the pricefeed daemon's lifecycle in order to reduce scope and PR review-ability since I made so many changes.

protocol/daemons/liquidation/client/client.go
clemire: I think this design was originally to help us to detect when to halt the daemon based on when the stop channel is closed - in the daemon, we don't send a value to `stop`, we just close it and it emits a single nil value. But actually there is a syntax that lets you detect whether a channel is closed when you read a value:

`_, ok := <- ticker.C`

So it looks like I can refactor both here and the pricefeed daemon to get rid of the stop channel.

protocol/daemons/liquidation/client/client.go
clemire: Note too - if the daemon does become unhealthy, we are still adding a flag that allows users to extend the period of unhealth before the protocol panics from 5 minutes to indefinite.

protocol/daemons/liquidation/client/client.go
clemire: when a ticker stops ticking, it does not close it's tick channel, so we added another channel to directly signal the stop so that we can halt the daemons. done is triggered by whoever wants the infinite loop to halt - in this case, the test cases, but in the future, a Stop method on the daemon could also trigger the done channel.

protocol/app/app.go
clemire: See comment about `StartNewClient` below.

protocol/daemons/types/health_checkable.go
```diff
@@ -61,12 +62,20 @@ type timeBoundedHealthCheckable struct {
...
 	// creation and checking for update staleness during HealthCheck.
 	timeProvider libtime.TimeProvider
+
+	// logger is the logger used to log errors.
+	logger log.Logger
 }
 
 // NewTimeBoundedHealthCheckable creates a new HealthCheckable instance.
-func NewTimeBoundedHealthCheckable(serviceName string, timeProvider libtime.TimeProvider) HealthCheckable {
+func NewTimeBoundedHealthCheckable(
```
clemire, line 71: refactor to take a logger for logging if the last good response was stale instead of returning an error.

protocol/daemons/types/health_checkable.go
```diff
@@ -119,12 +128,15 @@ func (h *timeBoundedHealthCheckable) HealthCheck() error {
 
 	// If the last successful update was more than 5 minutes ago, report the specific error.
 	if h.timeProvider.Now().Sub(h.lastSuccessfulUpdate) > MaxAcceptableUpdateDelay {
-		return fmt.Errorf(
-			"last successful update occurred at %v, which is more than %v ago. Last failure occurred at %v with error '%w'",
-			h.lastSuccessfulUpdate,
-			MaxAcceptableUpdateDelay,
-			h.lastFailedUpdate.Timestamp(),
-			h.lastFailedUpdate.Error(),
+		h.logger.Error(
```
clemire, line 131: downgrade error to a log message.

protocol/daemons/pricefeed/client/client.go
```diff
@@ -52,19 +52,25 @@ type Client struct {
...
+
+	// logger is the logger for the daemon.
+	logger log.Logger
 }
 
 // Ensure Client implements the HealthCheckable interface.
 var _ daemontypes.HealthCheckable = (*Client)(nil)
 
-func newClient() *Client {
+func newClient(logger log.Logger) *Client {
```
clemire, line 63: update client to be created with a logger for the HealthCheckable initialization. Store it on the daemon for the `start` method.

protocol/daemons/liquidation/client/sub_task_runner_test.go
```diff
@@ -0,0 +1,284 @@
...
+				daemonClient,
+				grpc.Ctx,
+				queryClientMock,
+				df.Liquidation.SubaccountPageLimit,
+			)
+			if err != nil {
+				require.EqualError(t, err, tc.expectedError.Error())
+				// The daemon initializes as unhealthy.
+				// If a request fails, the daemon will not be toggled to healthy.
+				require.Error(t, daemonClient.HealthCheck())
```
clemire, line 114: The check here and 2 lines below is new - we validate daemon health based on whether or not the query returns an error.

protocol/daemons/liquidation/client/sub_task_runner.go
clemire: Edit: function signature changes to pass the client through to methods that toggle health status.

protocol/daemons/liquidation/client/sub_task_runner_test.go
```diff
@@ -0,0 +1,284 @@
...
+				grpc.Ctx,
+				queryClientMock,
+				tc.subaccountIds,
+			)
+
+			if err != nil {
+				require.EqualError(t, err, tc.expectedError.Error())
+				// The daemon initializes as unhealthy.
+				// If a request fails, the daemon will not be toggled to healthy.
+				require.Error(t, daemon.HealthCheck())
```
clemire, line 221: same here as above - added two lines to check daemon health based on returned error.

protocol/app/app_test.go
```diff
@@ -97,11 +97,17 @@ func TestAppIsFullyInitialized(t *testing.T) {
...
-			// Normally it would be only disabled for non-validating full nodes or for nodes where the
-			// price feed client is explicitly disabled.
-			if idx := slices.Index(uninitializedFields, "PriceFeedClient"); idx >= 0 {
-				slices.Remove(&uninitializedFields, idx)
+			// Note that the daemon clients are currently hard coded as disabled in GetDefaultTestAppOptions.
+			// Normally they would be only disabled for non-validating full nodes or for nodes where any
+			// daemon is explicitly disabled.
+			expectedUninitializedFields := []string{
+				"PriceFeedClient",
+				"LiquidationsClient",
```
clemire, line 105: Added LiquidationsClient as a field on app that is initialized when the gRPC server starts.

protocol/daemons/liquidation/client/client.go
clemire: As discussed offline, this PR removes the staleness check in the health-checkable and also reports success whenever paginated gRPC queries return successfully.

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,291 @@
...
+	query := &clobtypes.AreSubaccountsLiquidatableRequest{
+		SubaccountIds: subaccountIds,
+	}
+	response, err := client.AreSubaccountsLiquidatable(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+
+	// For the purposes of the health check, log the successful request as an indicator of daemon health.
+	daemon.ReportSuccess()
```
clemire, line 115: Note: success report occurs here

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,291 @@
...
+			limit,
+			nextKey,
+		)
+
+		if err != nil {
+			return nil, err
+		}
+
+		// For the purposes of the health check, log the successful request as an indicator of daemon health.
+		daemon.ReportSuccess()
```
clemire, line 215: Note: success report occurs here after paginated request above.

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +84,55 @@ func Start(
...
+				client,
+				ctx,
+				flags.Liquidation,
+				subaccountQueryClient,
+				clobQueryClient,
+				liquidationServiceClient,
+			); err != nil {
+				// TODO(DEC-947): Move daemon shutdown to application.
+				client.logger.Error("Liquidations daemon returned error", "error", err)
+				client.ReportFailure(err)
```
clemire, line 130: Failure report here. This should be called as soon as an error is returned, which should happen when any gRPC request in the daemon's task loop fails - so there should be little latency between a request failure and this report.

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +84,55 @@ func Start(
...
-		metrics.LiquidationDaemon,
-		time.Now(),
-		metrics.MainTaskLoop,
-		metrics.Latency,
-	)
+	done := make(chan bool)
 
-	// 1. Fetch all subaccounts from query service.
-	subaccounts, err := GetAllSubaccounts(
+	s := &SubTaskRunnerImpl{}
```
jayy04, line 89: nit: things like flags, query clients can just be part of this `SubTaskRunnerImpl` so that we don't need to pass it around 

up to you though

jayy04 (approved):

protocol/app/app.go
```diff
@@ -602,14 +603,14 @@ func New(
 			app.Server.ExpectLiquidationsDaemon(
 				daemonservertypes.MaximumAcceptableUpdateDelay(daemonFlags.Liquidation.LoopDelayMs),
 			)
+			app.LiquidationsClient = liquidationclient.NewClient(logger)
 			go func() {
```
ttl33, line 607: As a follow up PR, can we refactor so that `go func()` all live within `*Client.Start` or live in `app.go`?

I think we currently have a mixed bag of implementation between prices vs liq/bridge daemons

protocol/app/app.go
```diff
@@ -602,14 +603,14 @@ func New(
...
-				if err := liquidationclient.Start(
```
ttl33, line 606: nit: in a separate PR, can we be consistent with `StartNewClient` ([here](https://github.com/dydxprotocol/v4-chain/pull/738/files#diff-7148577e1c863205cc155a436a9c9b806e655073d37a5124da0af8b5e9370317L627)) vs `.*Client.Start`? 

I think we should consistently use `.*Client.Start`

protocol/daemons/liquidation/client/client.go
```diff
@@ -65,262 +84,55 @@ func Start(
...
-	clobQueryClient clobtypes.QueryClient,
-	liquidationServiceClient api.LiquidationServiceClient,
-) error {
-	defer telemetry.ModuleMeasureSince(
-		metrics.LiquidationDaemon,
-		time.Now(),
-		metrics.MainTaskLoop,
-		metrics.Latency,
-	)
+	done := make(chan bool)
```
ttl33, line 87:
```suggestion
	stop := make(chan bool)
```

protocol/daemons/liquidation/client/client.go
ttl33:
```suggestion
	stop <-chan bool,
```

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CheckCollateralizationForSubaccounts queries a gRPC server using `AreSubaccountsLiquidatable`
+// and returns a list of collateralization statuses for the given list of subaccount ids.
+func CheckCollateralizationForSubaccounts(
```
ttl33, line 90: why not make this a method with a receiver type?

```suggestion
func (c *Client) CheckCollateralizationForSubaccounts(
```

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	if response.Pagination != nil {
+		nextKey = response.Pagination.NextKey
+	}
+	return response.Subaccount, nextKey, nil
+}
+
+// GetAllSubaccounts queries a gRPC server and returns a list of subaccounts and
+// their balances and open positions.
+func GetAllSubaccounts(
```
ttl33, line 189: Same here, why not make this a method w receiver type?

If so, let's move this method up so that all the receiver types end up next to one another.

```suggestion
func (c *Client) GetAllSubaccounts(
```

protocol/daemons/liquidation/client/sub_task_runner.go
ttl33, line 115: To confirm, do we want to report success on each subaccount chunk? Or all the chunks? I think it's the former because we only care that we are making some incremental progress but wanted to confirm.

If so, let's add some comments around this

protocol/daemons/liquidation/client/sub_task_runner.go
ttl33, line 215: Similar to above, can you confirm that we want to do this for each page? I think the answer is yes but wanted to confirm. 

Also, let's add comments

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -0,0 +1,291 @@
...
+		metrics.GetAllSubaccounts,
+		metrics.Count,
+	)
+
+	return subaccounts, nil
+}
+
+// GetLiquidatableSubaccountIds verifies collateralization statuses of subaccounts with
+// at least one open position and returns a list of unique and potentially liquidatable subaccount ids.
+func GetLiquidatableSubaccountIds(
```
ttl33, line 237: Same as above, why not add the receiver type?

protocol/daemons/liquidation/client/sub_task_runner.go
ttl33, line 115: As currently written, there are 4 different places that `ReportSuccess` and they all share the same timestamp. Would it make sense to report success based on a particular loop so it's easier to debug which loop is making progress?

Something like `ReportSuccess(key string)` and have the `HealthCheck` func for liquidation daemon to iterate through all the loops and output healthy vs not-healthy result. You could provide some helper methods like "any recent loop healthy => healthy" vs "all loops must be healthy => healthy" as example ones

protocol/daemons/types/health_checkable.go
```diff
@@ -119,12 +128,15 @@ func (h *timeBoundedHealthCheckable) HealthCheck() error {
 
 	// If the last successful update was more than 5 minutes ago, report the specific error.
```
ttl33, line 129:
```suggestion
	// If the last successful update was more than 5 minutes ago, log the specific error.
	// <add comments why we are just logging instead of returning error>
```

protocol/daemons/types/health_checkable.go
ttl33: To confirm, we want this to take effect for all daemons right? (prices + liquidation + bridge)? 

I think I am aligned but wanted to call this will affect all daemons

protocol/daemons/liquidation/client/sub_task_runner.go
ttl33: Discussed this offline. Concluded that we will punt this work given that there are error logs related to failed loops

ttl33 (approved):

protocol/daemons/types/health_checkable.go
clemire: Yes!

protocol/daemons/liquidation/client/client.go
clemire: Filed a follow-up under daemon lifecycle, since this affects all daemons.

protocol/app/app.go
clemire: https://linear.app/dydx/issue/CORE-756/refactor-pricefeed-client-to-start-in-a-consistent-pattern-with-other

protocol/app/app.go
clemire: https://linear.app/dydx/issue/CORE-756/refactor-pricefeed-client-to-start-in-a-consistent-pattern-with-other

protocol/daemons/liquidation/client/sub_task_runner.go
clemire: My thought was, I don't really want to expose these methods to the app, which stores a reference to the client, since it's an implementation detail. But I'm open to it if you prefer it and think it's better.

protocol/daemons/liquidation/client/sub_task_runner.go
clemire: Yes, confirmed!

protocol/daemons/liquidation/client/sub_task_runner.go
clemire: answered above!

protocol/daemons/liquidation/client/client.go
```diff
@@ -2,44 +2,63 @@ package client
...
 	"github.com/dydxprotocol/v4-chain/protocol/daemons/flags"
 	"github.com/dydxprotocol/v4-chain/protocol/daemons/liquidation/api"
 	daemontypes "github.com/dydxprotocol/v4-chain/protocol/daemons/types"
-	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+// Client implements a daemon service client that periodically queries a gRPC server for all liquidatable subaccounts
+// and reports them back to the protocol via the daemon server.
```
ttl33, line 19:
```suggestion
// Client implements a daemon service client that periodically calculates and reports liquidatable subaccounts
// to the protocol.
```

protocol/daemons/liquidation/client/client.go
ttl33: yup, let's stick to `NewClient` + `Start` paradigm. I believe you already created a tix to follow up [here](https://linear.app/dydx/issue/CORE-756/refactor-pricefeed-client-to-start-in-a-consistent-pattern-with-other)

protocol/daemons/types/health_checkable.go
ttl33: Wait, if we change this from return an error to just logging an error, how would we enforce `[High] Panic on the protocol side if daemon is unhealthy for X minutes (aka grace period, X = 5 to 15 min TBD)`?

Wouldn't this change mean that we never report unhealthy back to the protocol side? Consider the scenario where the last success time was Y and the last failure time was Y-1 and Y was a long time ago and the daemon was not able to make any new successful or failure loop? Is it guaranteed that the loops would always report success or failure within a certain strict time window?

clemire: No, not gauranteed. We discussed this and considered the risk acceptable. The long-term fix is to add timeouts to gRPC queries, which is the most likely source of loop non-termination... But in practice we haven't seen the daemon hang on a gRPC call except during initialization.

ttl33 (approved):

PR #750, opened 2023-11-03
[CLOB-636] - update e2e stateful cancellations tests

jakob-dydx:


protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -69,11 +71,85 @@ func TestPlaceOrder_StatefulCancelFollowedByPlaceInSameBlockErrorsInCheckTx(t *t
...
+
+	// DeliverTx should fail for cancellation tx
+	ctx = tApp.AdvanceToBlock(3, testapp.AdvanceToBlockOptions{
+		ValidateDeliverTxs: func(
+			ctx sdktypes.Context,
+			request abcitypes.RequestDeliverTx,
+			response abcitypes.ResponseDeliverTx,
+			txIndex int,
+		) (haltChain bool) {
+			if lib.SlicesAreEqual(request.Tx, cancellationTx.Tx) {
```
jayy04, line 123: these are just byte arrays, right?

we can just use bytes.Equal. I think it's slightly more optimized and we don't have to write tests for `SlicesAreEqual`

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -188,14 +293,17 @@ func TestCancelStatefulOrder(t *testing.T) {
...
-			ctx := tApp.AdvanceToBlock(tc.checkResultsBlock, testapp.AdvanceToBlockOptions{})
+			_, exists := tApp.App.ClobKeeper.GetLongTermOrderPlacement(ctx, tc.expectations.orderId)
+			require.Equal(t, exists, tc.expectations.existsInState)
 			exists, fillAmount, _ := tApp.App.ClobKeeper.GetOrderFillAmount(
 				ctx,
-				tc.checkCancelledPlaceOrder.Order.OrderId,
+				tc.expectations.orderId,
 			)
-			require.False(t, exists)
+			require.Equal(t, false, exists)
```
jayy04, line 306: nit: require.False?

protocol/x/clob/e2e/long_term_orders_test.go
jayy04: maybe we can also use `txIndex`, which might be easier

jayy04 (approved):

protocol/x/clob/e2e/long_term_orders_test.go
jakob-dydx: oof, didn't know about bytes.Equal. Yeah I was thinking of using txIndex, I guess I can use that. Should reliably be able to get the same txIndex for the transaction I'm looking for

protocol/x/clob/e2e/long_term_orders_test.go
```diff
@@ -7,6 +7,8 @@ import (
 
 	"github.com/cometbft/cometbft/crypto/tmhash"
 
+	abcitypes "github.com/cometbft/cometbft/abci/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
```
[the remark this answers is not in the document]
jakob-dydx: they are used

PR #754, opened 2023-11-03
[IND-442] Don't send cancel messages if orders are fully filled in state.

vincentwschau:
### Changelist
Instead of using the filled amount of the order from the full-node (sent in order updates), use the filled amount in state (tracked in `redis` and set by `ender` when processing order fills) to determine:
- whether to send a websocket message indicating an order was removed
- whether to update the status of the order in Postgres

If the order is fully filled in state (filled amount >= size), do not send the canceled websocket messages or update the status.

This fixes a bug where orders that are fully-filled on the full-node but not fully-filled in state do not have websocket updates sent indicating the order was removed when a user cancels or the order expires.

### Test Plan
Tested in `dev` and with unit tests.

### Author/Reviewer Checklist


dydxwill (approved):

indexer/packages/redis/src/caches/state-filled-quantums-cache.ts
```diff
@@ -0,0 +1,38 @@
+import { RedisClient } from 'redis';
+
+import { getAsync, setexAsync } from '../helpers/redis';
+
+export const STATE_FILLED_QUANTUMS_CACHE_KEY_PREFIX: string = 'v4/state_filled_quantums/';
+export const STATE_FILLED_QUANTUMS_TTL_SECONDS: number = 300; // 5 minutes
+
+export async function updateStateFilledQuantums(
```
Christopher-Li, line 8: can you add comments to this? Are these optimistically filled quantums or onchain confirmed filled quantums?

Christopher-Li (approved):

indexer/packages/redis/src/caches/state-filled-quantums-cache.ts
vincentwschau: It's the second. `State` is equivalent to `on-chain confirmed` when referring to values on the network.

PR #762, opened 2023-11-06
GNU Affero General Public License Updates

jonfung-dydx:


jiajames (approved):
also add it to `indexer/LICENSE`

PR #764, opened 2023-11-07
[CORE-666] - Implement bridge daemon health checks

clemire:
### Changelist
Update bridge daemon to:
- have it's own struct
- compose it's type with a HealthCheckable field
- update that health checkable every time a task loop completes, with either success or failure.

### Test Plan
- Comprehensive Unit tests for healthchecks
- existing tests pass

### Author/Reviewer Checklist


protocol/daemons/bridge/client/client.go
```diff
@@ -2,37 +2,52 @@ package client
...
-	eth "github.com/ethereum/go-ethereum"
-	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
-	ethrpc "github.com/ethereum/go-ethereum/rpc"
 )
 
+// Client encapsulates the logic and interface for the bridge daemon. The bridge daemon periodically queries the
+// Ethereum blockchain for new bridge events and relays them to the Cosmos gRPC server.
+type Client struct {
+	daemontypes.HealthCheckable
```
clemire, line 24: Add a new `Client` type. Compose this struct with a HealthCheckable field that is initialized below.

protocol/daemons/bridge/client/client.go
```diff
@@ -2,37 +2,52 @@ package client
...
+func (c *Client) Start(
 	ctx context.Context,
 	flags daemonflags.DaemonFlags,
 	appFlags appflags.Flags,
-	logger log.Logger,
 	grpcClient daemontypes.GrpcClient,
 ) error {
 	// Log the daemon flags.
-	logger.Info(
+	c.logger.Info(
```
clemire, line 50: Thread the client through the Start call and access the logger stored on the client, which is needed for initializing the HealthCheckable field.

protocol/daemons/bridge/client/client.go
```diff
@@ -73,140 +88,59 @@ func Start(
...
 }
 
-// RunBridgeDaemonTaskLoop does the following:
-// 1) Fetches configuration information by querying the gRPC server.
-// 2) Fetches Ethereum events from a configured Ethereum client.
-// 3) Sends newly-recognized bridge events to the gRPC server.
-func RunBridgeDaemonTaskLoop(
+// StartBridgeDaemonTaskLoop operates the continuous loop that runs the bridge daemon. It receives as arguments
+// a ticker and a stop channel that are used to control and halt the loop.
+func StartBridgeDaemonTaskLoop(
```
clemire, line 115: `StartBridgeDaemonTaskLoop` takes a ticker and a stop channel, similar to how go subroutines for price daemon are designed, and uses the stop channel to abort the go routien. This is esp. useful for tests!

protocol/daemons/bridge/client/sub_task_runner.go
```diff
@@ -0,0 +1,148 @@
...
+
+type SubTaskRunnerImpl struct{}
+
+var _ SubTaskRunner = (*SubTaskRunnerImpl)(nil)
+
+// RunBridgeDaemonTaskLoop does the following:
+// 1) Fetches configuration information by querying the gRPC server.
+// 2) Fetches Ethereum events from a configured Ethereum client.
+// 3) Sends newly-recognized bridge events to the gRPC server.
+func (s *SubTaskRunnerImpl) RunBridgeDaemonTaskLoop(
```
clemire, line 39: Refactored the RunBridgeDaemonTaskLoop method to be a member of the SubTaskRunnerImpl struct so that I can mock this implementation for testing.

protocol/app/app.go
```diff
@@ -646,13 +646,13 @@ func New(
 			// environments.
 			// app.Server.ExpectBridgeDaemon(daemonservertypes.MaximumAcceptableUpdateDelay(daemonFlags.Bridge.LoopDelayMs))
 			go func() {
-				if err := bridgeclient.Start(
+				app.BridgeClient = bridgeclient.NewClient(logger)
+				if err := app.BridgeClient.Start(
```
ttl33, line 650: nit: not related to this PR, can we rename `pricefeedclient.StartNewClient` to `pricefeedclient.Start` to be consistent?

protocol/daemons/bridge/client/client.go
```diff
@@ -73,140 +88,59 @@ func Start(
...
-		Topics: [][]ethcommon.Hash{
-			{ethcommon.HexToHash(constants.BridgeEventSignature)},
-			eventIdHashes,
-		},
+) {
+	// Run the main task loop at an interval.
+	for {
+		select {
+		case <-ticker.C:
+			if err := s.RunBridgeDaemonTaskLoop(
```
ttl33, line 129: To confirm, this loop is sync and blocking right?

For example, if the ticker happens every 3 seconds and this task loop takes 1 second, then does that mean that the `RunBridgeDaemonTaskLoop` will run every 4 seconds? Is that correct or does this task run every 3 seconds somehow?

protocol/daemons/bridge/client/client.go
```diff
@@ -2,37 +2,52 @@ package client
...
+// Ethereum blockchain for new bridge events and relays them to the Cosmos gRPC server.
+type Client struct {
+	daemontypes.HealthCheckable
+
+	// logger is the logger used by the bridge daemon.
+	logger log.Logger
+}
+
+func NewClient(logger log.Logger) *Client {
+	logger = logger.With(sdklog.ModuleKey, "bridge-daemon")
```
ttl33, line 31:
```suggestion
	logger = logger.With(sdklog.ModuleKey, BridgeDaemonModuleName)
```

protocol/app/app.go
clemire: Yes, task filed based on comment from another PR!

protocol/daemons/bridge/client/client.go
clemire: The ticker does drop signals that are unconsumed, so yes, if the ticker channel is not drained, it will not add new ticks.

I can modify this by running the task loop in a go routine. I believe the same limitation applied to the [previous implementation](https://github.com/dydxprotocol/v4-chain/blob/98d817dbf570d5bda7eb2eb5019d0dbebd424576/protocol/daemons/bridge/client/client.go#L83) as well.

protocol/daemons/bridge/client/client.go
clemire: d'oh!

ttl33 (approved):

PR #765, opened 2023-11-07
Small short term place order refactor

jonfung-dydx:
`placeOrder` was only called in `func (k Keeper) PlaceShortTermOrder(`.

Combine these functions, as there is no point in having a thin wrapper function.

Functionally should have no change in behavior.

Testing: Looking for tests to pass.

### DIFF
--- a/protocol/x/clob/keeper/orders.go
+++ b/protocol/x/clob/keeper/orders.go
@@ -71,7 +71,7 @@ func (k Keeper) CancelShortTermOrder(
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), metrics.CancelShortTermOrder, metrics.Latency)
 	telemetry.IncrCounter(1, types.ModuleName, metrics.CancelShortTermOrder, metrics.Count)
 
-	// Perform all stateful validation on the short term.
+	// Perform all stateful validation on the short term order.
 	if err := k.PerformOrderCancellationStatefulValidation(ctx, msgCancelOrder, nextBlockHeight); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


protocol/x/clob/keeper/orders.go
```diff
@@ -107,12 +109,66 @@ func (k Keeper) PlaceShortTermOrder(
...
 
-	return k.placeOrder(ctx, msg, nextBlockHeight, k.MemClob)
+	order := msg.GetOrder()
+	order.OrderId.MustBeShortTermOrder()
+	orderLabels := order.GetOrderLabels()
+
+	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), metrics.PlaceOrder, metrics.Latency)
+	defer func() {
+		telemetry.IncrCounterWithLabels(
+			[]string{types.ModuleName, metrics.PlaceOrder, metrics.Count},
```
jonfung-dydx, line 122: In the future, I think we may want to consider changing PlaceOrder to PlaceShortTermOrder. 

`dydxprotocol.clob_place_order_count.count` lends itself to thinking the order is all order types.

protocol/x/clob/keeper/orders.go
```diff
@@ -71,22 +71,24 @@ func (k Keeper) CancelShortTermOrder(
...
 	}
-	return err
+
+	k.sendOffchainMessagesWithTxHash(
+		offchainUpdates,
+		tmhash.Sum(ctx.TxBytes()),
+		metrics.SendCancelOrderOffchainUpdates,
+	)
+
+	return nil
```
jiajames, line 91: Is `nil` return necessary in go? Been a while but I thought it wasn't needed

jiajames (approved):

protocol/x/clob/keeper/orders.go
jonfung-dydx: `missing returncompiler[MissingReturn](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#MissingReturn)`

PR #768, opened 2023-11-07
[IND-482]: Send timing fields for orders when possible

Christopher-Li:
### Changelist
Sends `createdAtHeight`, `updatedAtHeight`, and `updatedAt` for stateful order placements and cancellations. When an order is removed, send `updatedAtHeight` and `updatedAt` for all orders and `createdAtHeight` for stateful orders

### Test Plan
Tested with unit tests

### Author/Reviewer Checklist


indexer/services/vulcan/src/handlers/order-remove-handler.ts
```diff
@@ -276,16 +277,23 @@ export class OrderRemoveHandler extends Handler {
...
-      await runFuncWithTimingStat(
+      canceledOrder = await runFuncWithTimingStat(
         this.cancelOrderInPostgres(orderRemove),
         this.generateTimingStatsOptions('cancel_order_in_postgres'),
       );
+    } else {
+      await runFuncWithTimingStat(
+        OrderTable.findById(OrderTable.orderIdToUuid(orderRemove.removedOrderId!)),
+        this.generateTimingStatsOptions('find_order'),
+      );
```
vincentwschau, line 290: Is this call needed? I don't see the results being assigned to a variable.

indexer/services/vulcan/src/handlers/order-remove-handler.ts
```diff
@@ -276,16 +277,23 @@ export class OrderRemoveHandler extends Handler {
...
-      await runFuncWithTimingStat(
+      canceledOrder = await runFuncWithTimingStat(
         this.cancelOrderInPostgres(orderRemove),
         this.generateTimingStatsOptions('cancel_order_in_postgres'),
       );
+    } else {
+      canceledOrder = await runFuncWithTimingStat(
+        OrderTable.findById(OrderTable.orderIdToUuid(orderRemove.removedOrderId!)),
+        this.generateTimingStatsOptions('find_order'),
+      );
```
vincentwschau, line 290: Suggestion: This could probably be optimized by not querying for the order if there isn't an entry in the state filled quantums cache for the order. That would mean no fills for the order, and as it would be a short-term order that means it won't be in the DB.

vincentwschau (approved):

dydxwill (approved):

PR #772, opened 2023-11-08
[CLOB-858] Add liquidation tests with different block limits

jayy04:
### Changelist
- E2E tests to make sure liquidations respect position block limits and subaccount block limits.

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/e2e/liquidation_deleveraging_test.go
+++ b/protocol/x/clob/e2e/liquidation_deleveraging_test.go
@@ -332,6 +332,10 @@ func TestLiquidationConfig(t *testing.T) {
 			},
 
 			placedMatchableOrders: []clobtypes.MatchableOrder{
+				// Maker order at $50,500.
+				// This maker order is specifically chosen to be above the oracle price, to ensure that
+				// block limits use the notional value of the position (oracle price),
+				// and not the notional liquidated (match price)
 				&constants.Order_Carl_Num0_Id2_Clob0_Buy1BTC_Price50500_GTB10, // Order at $50,500
 			},
 			liquidatableSubaccountIds: []satypes.SubaccountId{constants.Dave_Num0},
@@ -401,7 +405,8 @@ func TestLiquidationConfig(t *testing.T) {
 
 				// Bankruptcy price is $50,499, and closing at $50,500 would require $1 from the insurance fund.
 				// First order would transfer $0.1 from the insurance fund and would succeed.
-				// Second order would require $0.9 from the insurance fund and would fail.
+				// Second order would require $0.9 from the insurance fund and would fail since subaccounts
+				// may only lose $0.5 per block.
 				&constants.Order_Dave_Num0_Id1_Clob0_Sell01BTC_Price50500_GTB10, // Order at $50,500
 				&constants.Order_Dave_Num0_Id0_Clob0_Sell1BTC_Price50500_GTB10,  // Order at $50,500
 			},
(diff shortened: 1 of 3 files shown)


protocol/x/clob/e2e/liquidation_deleveraging_test.go
```diff
@@ -20,6 +20,616 @@ import (
...
+			},
+			clobPairs: []clobtypes.ClobPair{constants.ClobPair_Btc},
+
+			expectedSubaccounts: []satypes.Subaccount{
+				{
+					Id: &constants.Carl_Num0,
+					AssetPositions: []*satypes.AssetPosition{
+						{
+							AssetId:  0,
+							Quantums: dtypes.NewInt(100_000_000_000 - 5_050_000_000),
```
jakob-dydx, line 361: Here I am seeing more than $5k leaving the asset position. To be clear, this IS expected because we chose to liquidate a number of quantums which corresponds to 5k in notional value? However, the actual price that we matched at (because fillable price can be worse than oracle price) resulted in the subaccount paying more than 5k

jakob-dydx: Was the price of the maker order selected to demonstrate this property? Would probably be more clear if the maker order was at the oracle price

protocol/x/clob/e2e/liquidation_deleveraging_test.go
```diff
@@ -20,6 +20,616 @@ import (
...
+				constants.Dave_Num0_1BTC_Long_50000USD,
+			},
+
+			placedMatchableOrders: []clobtypes.MatchableOrder{
+				// This order is irrelevant to the test, but is used to seed the insurance fund.
+				&constants.Order_Dave_Num0_Id2_Clob0_Sell1BTC_Price49500_GTB10, // Order at $49,500
+
+				// Bankruptcy price is $50,499, and closing at $50,500 would require $1 from the insurance fund.
+				// First order would transfer $0.1 from the insurance fund and would succeed.
+				// Second order would require $0.9 from the insurance fund and would fail.
```
jakob-dydx, line 404: Read this line before I saw the MaxQuantumsInsuranceLost value and was confused as to why the test would fail. Maybe add the detail that $.9 > (MaxQuantumsInsuranceLost - $.1)

protocol/x/clob/e2e/liquidation_deleveraging_test.go
```diff
@@ -20,6 +20,616 @@ import (
...
+				&constants.Order_Dave_Num0_Id1_Clob0_Sell01BTC_Price50500_GTB10, // Order at $50,500
+				&constants.Order_Dave_Num0_Id0_Clob0_Sell1BTC_Price50500_GTB10,  // Order at $50,500
```
jakob-dydx, line 406: interesting, so instead of partially matching the second order such that we do not exceed MaxQuantumsInsuranceLost, we instead don't match at all?

jakob-dydx (approved):

protocol/x/clob/e2e/liquidation_deleveraging_test.go
jayy04: yeah you are exactly right. we don't know what the order will get matched at when building the liquidation order, so we use the notional value (oracle price). for consistency, during matching, we also use the notional value (not the actual notional liquidated). otherwise, this inconsistency can lead to issues in DeliverTx.

jayy04: and yeah the maker order in this test was specifically chosen to demonstrate this. let me add a comment

protocol/x/clob/e2e/liquidation_deleveraging_test.go
jayy04: yup - partially matching will require a bunch of new logic in the order matching loop and probably not worth the engineering effort. but let's keep this in mind as this may happen 

I don't have numbers to back me up but my intuition tells me that orders on the book are going to be fairly small relative to our current `MaxInsuranceFundLost` value so this shouldn't be an issue for us.

jakob-dydx: true, thanks for clarifying

PR #783, opened 2023-11-09
[CORE-666] - Update update monitor to use health checks

clemire:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/app/app.go
```diff
@@ -290,6 +290,9 @@ type App struct {
 
 	PriceFeedClient    *pricefeedclient.Client
 	LiquidationsClient *liquidationclient.Client
+	BridgeClient       *bridgeclient.Client
+
+	HealthMonitor *daemonservertypes.HealthMonitor
```
clemire, line 295: Health Monitor does not need access to server responses, has been moved to the app.

protocol/app/app.go
```diff
@@ -676,9 +682,8 @@ func New(
 				}
 			}()
 			// Don't panic if metrics daemon loops are delayed. Use maximum value.
-			app.Server.ExpectMetricsDaemon(
-				daemonservertypes.MaximumAcceptableUpdateDelay(math.MaxUint32),
-			)
+			// TODO(CORE-666): Refactor metrics daemon and track health here
```
clemire, line 685: I'll need to do a fast-follow PR to take care of this.

protocol/app/app.go
```diff
@@ -675,10 +684,6 @@ func New(
 					)
 				}
 			}()
-			// Don't panic if metrics daemon loops are delayed. Use maximum value.
```
clemire, line 678: The metrics daemon is such a simple go-routine that I don't think it makes sense to re-engineer it at this time. We'll just remove the previous health check. The task loop does not produce any errors because the telemetry calls don't produce errors either, so this daemon would essentially never become unhealthy.

protocol/daemons/server/server.go
```diff
@@ -47,60 +40,14 @@ func NewServer(
...
 	}
 }
 
 // Stop stops the daemon server's gRPC service.
 func (server *Server) Stop() {
-	server.updateMonitor.Stop()
 	server.gsrv.Stop()
 }
 
-// DisableUpdateMonitoringForTesting disables the update monitor for testing purposes. This is needed in integration
```
clemire, line 60: Removed health monitor from daemon server and attached it to the app as a top level field, so this pass-through interface is no longer necessary.

protocol/daemons/server/types/util.go
clemire, line 1: This constant is not needed anymore.

protocol/daemons/server/types/update_monitor.go
clemire, line 1: file renamed to health_monitor.go

protocol/daemons/types/health_checkable.go
```diff
@@ -22,6 +22,8 @@ type HealthCheckable interface {
 	ReportFailure(err error)
 	// ReportSuccess records a successful update.
 	ReportSuccess()
+	// ServiceName returns the name of the service being monitored. This name is expected to be unique.
+	ServiceName() string
```
clemire, line 26: Add this field to easily key health checkables by service name.

protocol/app/app.go
```diff
@@ -183,6 +183,10 @@ import (
 var (
 	// DefaultNodeHome default home directories for the application daemon
 	DefaultNodeHome string
+
+	// MaximumDaemonUnhealthyDuration is the maximum amount of time that a daemon can be unhealthy before the
+	// application panics.
+	MaximumDaemonUnhealthyDuration = 5 * time.Minute
```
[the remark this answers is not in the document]
clemire: This TODO is addressed in the follow-up PR that's also out for review now, which I plan to merge together into main for the sake of keeping main sanitized.

protocol/app/app.go
```diff
@@ -589,6 +595,11 @@ func New(
 	bridgeEventManager := bridgedaemontypes.NewBridgeEventManager(timeProvider)
 	app.Server.WithBridgeEventManager(bridgeEventManager)
 
+	app.HealthMonitor = daemonservertypes.NewHealthMonitor(
```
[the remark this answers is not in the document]
clemire: Yes, I think so. Good point!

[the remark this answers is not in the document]
clemire: Filed just in case: https://linear.app/dydx/issue/CORE-801/refactor-app-daemon-initializiation-section-into-a-separate-method

protocol/app/app.go
```diff
@@ -637,16 +645,17 @@ func New(
 				constants.StaticExchangeDetails,
 				&pricefeedclient.SubTaskRunnerImpl{},
 			)
+			app.MonitorDaemon(app.PriceFeedClient, MaximumDaemonUnhealthyDuration)
```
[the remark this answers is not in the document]
clemire: Agree.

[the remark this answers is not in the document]
clemire: linked above comment [here](https://linear.app/dydx/issue/CORE-756/refactor-pricefeed-client-to-start-in-a-consistent-pattern-with-other).

protocol/daemons/server/liquidation.go
```diff
@@ -47,12 +36,6 @@ func (s *Server) LiquidateSubaccounts(
 		metrics.Received,
 		metrics.Count,
 	)
-	// If the daemon is unable to report a response, there is either an error in the registration of
```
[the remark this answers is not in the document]
clemire: The error here was emitted by the daemon monitor (which lived on the daemon server) if it failed to register the response. We don't report daemon responses to the monitor anymore; instead, we query the daemon for it's health directly.

protocol/daemons/server/server.go
```diff
@@ -47,60 +40,14 @@ func NewServer(
...
-			maximumAcceptableUpdateDelay,
-		)
-		panic(err)
-	}
-}
-
-// reportResponse reports a response from a daemon service with the update monitor. This is used to
-// ensure that the daemon continues to operate. If the update monitor does not see a response from a
-// registered daemon within the maximumAcceptableUpdateDelay, it will cause the protocol to panic.
-func (server *Server) reportResponse(
```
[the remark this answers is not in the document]
clemire: Hmmm, I see what you're saying. We can leave this in and remove the health monitor from this method.

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+	// maximumAcceptableUnhealthyDuration is the maximum acceptable duration for a health checkable service to
+	// remain unhealthy. If the service remains unhealthy for this duration, the monitor will execute the
+	// specified callback function.
+	maximumAcceptableUnhealthyDuration time.Duration
+}
+
+// Poll executes a health check for the health checkable service. If the service has been unhealthy for longer than the
+// maximum acceptable unhealthy duration, the callback function is executed.
+// This method is publicly exposed for testing.
+func (hc *healthChecker) Poll() {
```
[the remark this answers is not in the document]
clemire: No, it's not - great point!

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+	// pollFrequency is the frequency at which the health checkable service is polled.
+	pollFrequency time.Duration
+
+	// mostRecentSuccess is the timestamp of the most recent successful health check.
+	mostRecentSuccess time.Time
+
+	// firstFailureInStreak is the timestamp of the first error in the most recent streak of errors. It is set
+	// whenever the service toggles from healthy to an unhealthy state, and used to determine how long the daemon has
+	// been unhealthy. If this timestamp is nil, then the error streak ended before it could trigger a callback.
+	firstFailureInStreak timestampWithError
```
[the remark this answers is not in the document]
clemire: I think that I will leave it in order to avoid the potential issue of memory leaks for cases when someone disables the monitor by configuring it with a >> maximum unhealthy duration.

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+				"maximum acceptable unhealthy duration %v must be positive",
+			hc.ServiceName(),
+			maximumAcceptableUnhealthyDuration,
+		)
+	}
+
+	// Don't register daemon services if the monitor has been disabled.
+	if hm.disabled {
+		return nil
+	}
```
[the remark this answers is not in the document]
clemire: In this case, we want the registration to silently fail - returning an error would present a problem. The disable functionality was added for testing.

protocol/daemons/server/types/health_monitor.go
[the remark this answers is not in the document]
lcwik: We want to to keep track of the latest failure since we want to know if something has changed and the time at which the first failure occurred.

Storing only the first failure error will not allow us to see that the error condition has changed over time.

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Don't register additional daemon services if the monitor has already been stopped.
+	// This could be a concern for short-running integration test cases, where the network
+	// stops before all daemon services have been registered.
+	if hm.stopped {
+		return fmt.Errorf(
```
lcwik, line 219: I think we should stop the `HealthCheckable` and return `nil`.

This will prevent tests that register the daemons from panicking due to the failed registration if the timing of the test caused it to shutdown first and will also propagate the shutdown to the `HealthCheckable`.

Please also update the comment of `RegisterService`/`RegisterServiceWithCallback`

protocol/daemons/server/pricefeed.go
```diff
@@ -52,16 +45,10 @@ func (s *Server) UpdateMarketPrices(
 		metrics.Latency,
 	)
 
-	// If the daemon is unable to report a response, there is either an error in the registration of
-	// this daemon, or another one. In either case, the protocol should panic.
-	if err := s.reportResponse(servertypes.PricefeedDaemonServiceName); err != nil {
-		panic(err)
-	}
-
 	if s.marketToExchange == nil {
```
ttl33, line 48: nit: add comment why we would still panic here.

protocol/daemons/server/pricefeed.go
```diff
@@ -31,19 +31,12 @@ func (server *Server) WithPriceFeedMarketToExchangePrices(
...
-func (server *Server) ExpectPricefeedDaemon(maximumAcceptableUpdateDelay time.Duration) {
-	server.registerDaemon(servertypes.PricefeedDaemonServiceName, maximumAcceptableUpdateDelay)
-}
-
 // UpdateMarketPrices updates prices from exchanges for each market provided.
 func (s *Server) UpdateMarketPrices(
 	ctx context.Context,
 	req *api.UpdateMarketPricesRequest,
 ) (*api.UpdateMarketPricesResponse, error) {
+	s.reportResponse(types.PricefeedDaemonServiceName)
```
ttl33, line 39: Given that we don't know if the below logic will succeed or not at this point, should we move this `reportResponse` to right before the return statement?

protocol/daemons/server/liquidation.go
```diff
@@ -47,12 +37,7 @@ func (s *Server) LiquidateSubaccounts(
...
-	// this daemon, or another one. In either case, the protocol should panic.
-	if err := s.reportResponse(types.LiquidationsDaemonServiceName); err != nil {
-		s.logger.Error("Failed to report liquidations response to update monitor", "error", err)
-	}
-
+	s.reportResponse(types.LiquidationsDaemonServiceName)
```
ttl33, line 40: For both `ModuleSetGauge` and `reportResponse`: 

Given that we don't know if the below logic will succeed or not at this point, should we move these to right before the return statement?

protocol/daemons/server/bridge.go
```diff
@@ -40,12 +30,7 @@ func (s *Server) AddBridgeEvents(
 	*api.AddBridgeEventsResponse,
 	error,
 ) {
-	// If the daemon is unable to report a response, there is either an error in the registration of
-	// this daemon, or another one. In either case, the protocol should panic.
-	// TODO(CORE-582): Re-enable this check once the bridge daemon is fixed in local / CI environments.
-	//if err := s.reportResponse(types.BridgeDaemonServiceName); err != nil {
-	//	panic(err)
-	//}
+	s.reportResponse(types.BridgeDaemonServiceName)
```
ttl33, line 33: Given that we don't know if the below logic will succeed or not at this point, should we move this `reportResponse` to right before the return statement?

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// If the service has been unhealthy for longer than the maximum acceptable unhealthy duration, execute the
+	// callback function.
+	if !hc.firstFailureInStreak.IsZero() &&
+		now.Sub(hc.firstFailureInStreak.Timestamp()) >= hc.maximumAcceptableUnhealthyDuration {
+		hc.unhealthyCallback(hc.firstFailureInStreak.Error())
+	} else {
+		// If we do not execute the callback, schedule the next poll.
+		hc.timer.Reset(hc.pollFrequency)
+	}
```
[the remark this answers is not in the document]
ttl33: did we make this change?

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,293 @@
...
+	// healthCheckable is the health checkable service to be monitored.
+	healthCheckable types.HealthCheckable
+
+	// pollFrequency is the frequency at which the health checkable service is polled.
+	pollFrequency time.Duration
+
+	// maximumAcceptableUnhealthyDuration is the maximum acceptable duration for a health checkable service to
+	// remain unhealthy. If the service remains unhealthy for this duration, the monitor will execute the
+	// specified callback function.
+	maximumAcceptableUnhealthyDuration time.Duration
```
ttl33, line 57: nit:

```suggestion
	maxAcceptableUnhealthyDuration time.Duration
```

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,293 @@
...
+	// The first poll is scheduled after the startup grace period to allow the service to initialize.
+	checker.timer = time.AfterFunc(startupGracePeriod, checker.Poll)
+
+	return checker
+}
+
+// HealthMonitor monitors the health of daemon services, which implement the HealthCheckable interface. If a
+// registered health-checkable service sustains an unhealthy state for the maximum acceptable unhealthy duration,
+// the monitor will execute a callback function.
+type HealthMonitor struct {
```
ttl33, line 149: nit: we could break this file into `health_checker.go` and `health_monitor.go` but will leave it up to you!

protocol/daemons/server/types/health_monitor.go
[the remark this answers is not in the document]
clemire: Updated! Made some slight changes.

protocol/daemons/server/types/health_monitor.go
clemire: Not all daemons are stoppable at this time, but I added some code that conditionally stops daemons with a `Stop` method.

protocol/daemons/server/bridge.go
clemire: As of when we last touched this, the logic in reportResponse is only recording if the daemon successfully queries the server. We could definitely make a change here, but this is outside the scope of this work and, to be consistent, would probably require auditing other daemon gRPC query response code as well. I think it's better for now to leave the existing feature as is and stay focused on the health check feature.

protocol/daemons/server/liquidation.go
clemire: I think additional feature work here is out of scope for this PR.

protocol/daemons/server/pricefeed.go
clemire: See above.

protocol/daemons/server/pricefeed.go
clemire: This panic is removed.

protocol/daemons/server/types/health_monitor.go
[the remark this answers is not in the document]
clemire: ^ Apologies, I think you may have still been commenting when I addressed your last round of comments and I missed this the first time around.

protocol/daemons/server/types/health_monitor.go
lcwik: Is there a reason why we still want to return an error here?

ttl33:
`Update update monitor to use health checks` title seems a bit confusing. Would you be able to update the title? Perhaps: `Use health checks in daemon server`

protocol/daemons/server/pricefeed.go
clemire: Ahhh, IC. It looks to me like this panic would only occur if the client pinged the server before the prices cache was initialized on the server.

protocol/daemons/server/bridge.go
clemire: ICIC - if it's only a few lines, I can just move it to the end of each method. Does that fully capture the concern? I'm sorry about that.

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,152 @@
...
+	// mostRecentFailureStreakError tracks the timestamp of the first error in the most recent streak of errors, as well
+	// as the most recent error. It is updated on every error and reset every time the service sees a healthy response.
+	// This field is used to determine how long the daemon has been unhealthy. If this timestamp is nil, then either
+	// the service has never been unhealthy, or the most recent error streak ended before it could trigger a callback.
+	// Access to mostRecentFailureStreakError is synchronized.
+	mostRecentFailureStreakError timestampWithError
+
+	// timeProvider is used to get the current time. It is added as a field so that it can be mocked in tests.
+	// Access to timeProvider is synchronized.
+	timeProvider libtime.TimeProvider
```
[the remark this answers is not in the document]
clemire: ok, updated!

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Poll executes a health check for the health checkable service. If the service has been unhealthy for longer than the
+// maximum acceptable unhealthy duration, the callback function is executed.
+// This method is publicly exposed for testing. This method is synchronized.
+func (hc *healthChecker) Poll() {
+	hc.lock.Lock()
```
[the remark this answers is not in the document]
clemire: Makes sense!

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Capture healthy response.
+	if err == nil {
+		hc.mostRecentSuccess = now
+		// Whenever the service is healthy, reset the first failure in streak timestamp.
+		hc.mostRecentFailureStreakError.Reset()
+		return
+	}
+
+	hc.mostRecentFailureStreakError.Update(now, err)
```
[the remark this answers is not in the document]
clemire: I checked with Luke on slack and he confirmed that he was only asking for the most recent error.

protocol/daemons/server/liquidation.go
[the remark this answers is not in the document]
clemire: "...we never expect the response call to fail" - yes, since we've removed health monitoring code from this method, there's nothing here but telemetry and this call will never generate an error.

protocol/daemons/server/types/health_monitor.go
clemire: I thought of it as informative, but yes, it's not truly an error condition.

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,174 @@
...
+	// serviceToHealthChecker maps daemon service names to their update metadata.
+	serviceToHealthChecker map[string]*healthChecker
+	// stopped indicates whether the monitor has been stopped. Additional daemon services cannot be registered
+	// after the monitor has been stopped.
+	stopped bool
+	// disabled indicates whether the monitor has been disabled. This is used to disable the monitor in testApp
+	// tests, where app.New is not executed.
+	disabled bool
+	// lock is used to synchronize access to the monitor.
+	lock sync.Mutex
```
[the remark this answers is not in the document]
clemire: I walked through the code and confirmed that, according to my understanding, health monitor or health checker locks are being called in a nested fashion. I also reduced the critical section for the health monitor Stop method just in case.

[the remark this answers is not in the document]
clemire: Done!

protocol/app/app.go
```diff
@@ -1222,6 +1230,31 @@ func New(
 	return app
 }
 
+// MonitorDaemon registers a daemon service with the update monitor. If the daemon does not register, the method will
+// panic.
+func (app *App) MonitorDaemon(
```
ttl33, line 1235:
```suggestion
func (app *App) RegisterDaemonHealthMonitor(
```

protocol/daemons/server/bridge.go
```diff
@@ -23,30 +21,24 @@ func (server *Server) WithBridgeEventManager(
...
-	// If the daemon is unable to report a response, there is either an error in the registration of
-	// this daemon, or another one. In either case, the protocol should panic.
-	// TODO(CORE-582): Re-enable this check once the bridge daemon is fixed in local / CI environments.
-	//if err := s.reportResponse(types.BridgeDaemonServiceName); err != nil {
-	//	panic(err)
-	//}
-	if err := s.bridgeEventManager.AddBridgeEvents(req.BridgeEvents); err != nil {
+	// Capture valid responses in metrics.
+	defer func() {
+		if err == nil {
```
ttl33, line 35: Not 100% how the closure works, but will this value check the value assigned by [line 41](https://github.com/dydxprotocol/v4-chain/pull/783/files#diff-311868d8f30cb540f3ba861a80d88805bde4b5277b071aff85e4c8e5f401bd76R41)?

Generally, I think it's better to add this right above the return statement. If someone forgets the `err = ` and uses `err :=` then this wont work?

protocol/daemons/server/bridge.go
```diff
@@ -23,30 +21,24 @@ func (server *Server) WithBridgeEventManager(
...
+			s.reportValidResponse(types.PricefeedDaemonServiceName)
```
ttl33, line 36: Did you mean to use `BridgeDaemonServiceName` here?

protocol/daemons/server/bridge.go
```diff
@@ -23,30 +21,24 @@ func (server *Server) WithBridgeEventManager(
...
+		}
+	}()
+
+	s.reportValidResponse(types.BridgeDaemonServiceName)
```
ttl33, line 40: is this duplicate?

protocol/daemons/server/liquidation.go
```diff
@@ -26,32 +24,29 @@ func (server *Server) WithLiquidatableSubaccountIds(
...
 	req *api.LiquidateSubaccountsRequest,
-) (*api.LiquidateSubaccountsResponse, error) {
+) (
+	response *api.LiquidateSubaccountsResponse,
+	err error,
+) {
+	// Capture valid responses in metrics.
+	defer func() {
+		if err == nil {
+			s.reportValidResponse(types.PricefeedDaemonServiceName)
```
ttl33, line 39: same here, should this be liquidation daemon?

protocol/daemons/server/pricefeed.go
ttl33: gotcha, can you add some comments that "This is unexpected because xyz"

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+// timestampWithError couples a timestamp and error to make it easier to update them in tandem. The
+// timestampWithError will track the timestamp of the first error in a streak of errors, but keeps a record of the
+// most recent error. This is useful for determining how long a service has been unhealthy, as well as the current
+// state of the service.
+type timestampWithError struct {
+	timestamp time.Time
+	err       error
+}
+
+func (u *timestampWithError) Update(timestamp time.Time, err error) {
```
ttl33, line 20: Since `Update` suggests that both values will be updated, should update this to something that hints that it only updates err if the timestamp is set?

```suggestion
func (u *timestampWithError) UpdateLastErr(timestamp time.Time, err error) {
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	u.err = err
+}
+
+func (u *timestampWithError) Reset() {
+	u.timestamp = time.Time{}
+	u.err = nil
+}
+
+func (u *timestampWithError) IsZero() bool {
```
ttl33, line 33:
```suggestion
func (u *timestampWithError) IsUnset() bool {
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+	// mostRecentSuccess is the timestamp of the most recent successful health check.
+	// Access to mostRecentSuccess is synchronized.
+	mostRecentSuccess time.Time
+
+	// mostRecentFailureStreakError tracks the timestamp of the first error in the most recent streak of errors, as well
+	// as the most recent error. It is updated on every error and reset every time the service sees a healthy response.
+	// This field is used to determine how long the daemon has been unhealthy. If this timestamp is nil, then either
+	// the service has never been unhealthy, or the most recent error streak ended before it could trigger a callback.
+	// Access to mostRecentFailureStreakError is synchronized.
+	mostRecentFailureStreakError timestampWithError
```
ttl33, line 60:
```suggestion
	lastErrWithFirstErrTimestamp timestampWithError
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+
+// healthCheckerMutableState tracks the current health state of the HealthCheckable, encapsulating all mutable state
+// into a single struct for ease of synchronization.
+type healthCheckerMutableState struct {
+	// lock is used to synchronize access to mutable state fields.
+	lock sync.Mutex
+
+	// mostRecentSuccess is the timestamp of the most recent successful health check.
+	// Access to mostRecentSuccess is synchronized.
+	mostRecentSuccess time.Time
```
ttl33, line 53:
```suggestion
	lastSuccessTimestamp time.Time
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+		streakDuration := hc.mutableState.ReportFailure(now, err)
+		// If the service has been unhealthy for longer than the maximum acceptable unhealthy duration, execute the
+		// callback function.
+		if streakDuration >= hc.maxAcceptableUnhealthyDuration {
+			hc.unhealthyCallback(err)
+		}
+	}
+
+	// Schedule next poll.
+	hc.timer.Reset(hc.pollFrequency)
```
ttl33, line 137: Nice, thanks for simplifying this logic!

```suggestion
	if err == nil { // Capture healthy response.
		hc.mutableState.ReportSuccess(now)
	} else { // Capture error response.
		failureDuration := hc.mutableState.ReportFailure(now, err)
		// If the service has been unhealthy for longer than the maximum acceptable unhealthy duration, execute the
		// callback function.
		if failureDuration >= hc.maxAcceptableUnhealthyDuration {
			hc.unhealthyCallback(err)
		}
	}

	// Schedule next poll.
	hc.timer.Reset(hc.pollFrequency)
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Stop stops the health checker. This method is synchronized.
+func (hc *healthChecker) Stop() {
+	hc.timer.Stop()
```
ttl33, line 142: to confirm, is stopping only the timer sufficient?

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,207 @@
...
+
+// NewHealthMonitor creates a new health monitor.
+func NewHealthMonitor(
+	startupGracePeriod time.Duration,
+	pollingFrequency time.Duration,
+	logger log.Logger,
+) *HealthMonitor {
+	return &HealthMonitor{
+		mutableState:       NewHealthMonitorMutableState(),
+		logger:             logger.With(cosmoslog.ModuleKey, "health-monitor"),
```
ttl33, line 135:
```suggestion
		logger:             logger.With(cosmoslog.ModuleKey, "daemon-health-monitor"),
```

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,207 @@
...
+
+// RegisterServiceWithCallback registers a HealthCheckable with the health monitor. If the service
+// stays unhealthy every time it is polled during the maximum acceptable unhealthy duration, the monitor will
+// execute the callback function.
+// This method is synchronized. The method returns an error if the service was already registered or the
+// monitor has already been stopped. If the monitor has been stopped, this method will proactively stop the
+// health-checkable service before returning.
+func (hm *HealthMonitor) RegisterServiceWithCallback(
+	hc types.HealthCheckable,
+	maximumAcceptableUnhealthyDuration time.Duration,
```
ttl33, line 153: similar comment as in the other PR. should we keep this consistent w the flag name? `maxDaemonUnhealthyDuration` or something similar?

protocol/daemons/server/bridge.go
lcwik: Yes, the closure would check the `err` assigned in line 41 since that is what is in scope.

rant: `:=` should not allow creating a new symbol with the same name, it makes reading code very difficult.

lcwik, line 35: Optional: Note that if we change the report API to only have one method for both healthy (`err = nil`) and unhealthy (`err != nil`) then callers can just invoke the one API without needing an `if`.

protocol/app/app.go
clemire: I changed the method name slightly to `RegisterDaemonWithHealthMonitor` because the above rename seemed easy to mis-parse, i.e. we are not registering a daemon health monitor, but registering a daemon with the health monitor.

protocol/daemons/server/bridge.go
clemire: Yep! TY!

protocol/daemons/server/liquidation.go
clemire: Yes, I'm sorry about that.

protocol/daemons/server/types/health_checker.go
clemire: No, good point. If the timer is stopped during a poll, a new poll could be scheduled after the timer is stopped.

clemire: I moved the timer to the mutable state object and synchronize it with an internal `stopped` flag to prevent scheduling a poll if the health checker is stopped.

protocol/daemons/server/types/health_monitor.go
clemire: This is already addressed in the following PR, which is branched off of this branch, and will be merged into this branch before both are committed to main (in order to keep main in a releasable state.) For the sake of keeping things non-confusing and avoiding some rebasing churn, I'd like to keep this change in that PR if you don't mind!

protocol/daemons/server/bridge.go
clemire: In order to wrap this PR, I think I am going to leave as is unless there are strong feelings about it, since there's a good amount of back and forth and I need to cover some P0s before the release.

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,218 @@
...
+// This method is publicly exposed for testing. This method is synchronized.
+func (hc *healthChecker) Poll() {
+	err := hc.healthCheckable.HealthCheck()
+	now := hc.timeProvider.Now()
+
+	// Capture healthy response.
+
+	if err == nil {
+		hc.mutableState.ReportSuccess(now)
+	} else {
```
ttl33, line 177:
```suggestion
	if err == nil { // Capture healthy response.
		hc.mutableState.ReportSuccess(now)
	} else { // Capture unhealthy response.
```

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,218 @@
...
+		streakDuration := hc.mutableState.ReportFailure(now, err)
+		// If the service has been unhealthy for longer than the maximum acceptable unhealthy duration, execute the
+		// callback function.
+		if streakDuration >= hc.maxAcceptableUnhealthyDuration {
+			hc.unhealthyCallback(err)
+		}
+	}
+
+	// Schedule next poll.
+	hc.mutableState.SchedulePoll(hc.pollFrequency)
```
ttl33, line 187: To confirm, we want to schedule the next poll when the daemon is unhealthy right? Makes sense to me, but would be nice to add some comments why this is intended.

protocol/daemons/server/types/health_monitor.go
```diff
@@ -0,0 +1,223 @@
...
+	callback func(error),
+) error {
+	if maximumAcceptableUnhealthyDuration <= 0 {
+		return fmt.Errorf(
+			"health check registration failure for service %v: "+
+				"maximum acceptable unhealthy duration %v must be positive",
```
ttl33, line 162: nit: just to confirm, will you be updating this text too to match `maxDaemonUnhealthyDuration`?

ttl33 (approved):

protocol/daemons/server/types/health_checker.go
lcwik: I like `IsUnset`

lcwik (approved):

protocol/daemons/server/types/health_checker.go
```diff
@@ -0,0 +1,218 @@
...
+	// healthCheckable is the health-checkable service to be monitored.
+	healthCheckable types.HealthCheckable
+
+	// pollFrequency is the frequency at which the health-checkable service is polled.
+	pollFrequency time.Duration
+
+	// maxAcceptableUnhealthyDuration is the maximum acceptable duration for a health-checkable service to
+	// remain unhealthy. If the service remains unhealthy for this duration, the monitor will execute the
+	// specified callback function.
+	maxAcceptableUnhealthyDuration time.Duration
```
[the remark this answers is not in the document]
clemire: ^ it's renamed in that PR, which will be merged into this one before merging to main.

protocol/daemons/server/types/health_monitor.go
clemire: yes, it's updated [here](https://github.com/dydxprotocol/v4-chain/pull/802/files#diff-9dca63687afbddbc7f90992fa632c143333f1f33d9604b4af8445ee5a5d36342)

protocol/daemons/server/types/health_checker.go
clemire: I changed the type to `errorStreak` and call this `mostRecentErrorStreak`.

PR #792, opened 2023-11-20
[CLOB-1014] [CLOB-1004] Unify metrics library, un-modularize metrics keys

jonfung-dydx:
Provide a standard for metrics going forward.

First step to do this is to create a singular entrypoint method that we report metrics through.

We currently have multiple different ways we report telemetry, using these two libraries across all files in the codebase:
```
	gometrics "github.com/armon/go-metrics"
	"github.com/cosmos/cosmos-sdk/telemetry"
```

Barrel everything through metric functions in `metrics/lib.go`.
Files should be migrated one by one, so here's a start with migrating a few libraries.

There should be _no_ change in how metrics are reported.

After everything goes through the singular entrypoint method, we can look at unifying the metrics libraries.

Also, un-modularize the metrics keys instead.
instead of passing in `[]string{"clob", "place_order", "count"}`, we should just pass in `metrics.ClobPlaceOrderCount` instead. This will make our metrics and the naming system a lot more clear because of the 1:1 metric:name:dashboard ratio.

Other files will need to be modified too later.

### DIFF
--- a/protocol/lib/metrics/lib.go
+++ b/protocol/lib/metrics/lib.go
@@ -60,6 +60,8 @@ func AddSample(key string, val float32) {
 // ModuleMeasureSince provides a wrapper functionality for emitting a time measure
 // metric with global labels (if any).
 // Please try to use `AddSample` instead.
+// TODO(CLOB-1022) Roll our own calculations for timing on top of AddSample instead
+// of using MeasureSince.
 func ModuleMeasureSince(module string, key string, start time.Time) {
 	telemetry.ModuleMeasureSince(
 		module,
@@ -69,9 +71,10 @@ func ModuleMeasureSince(module string, key string, start time.Time) {
 }
 
 // ModuleMeasureSinceWithLabels provides a short hand method for emitting a time measure
-// metric for a module with labels. Global labels are not included
-// in this metric.
+// metric for a module with labels. Global labels are not included in this metric.
 // Please try to use `AddSampleWithLabels` instead.
+// TODO(CLOB-1022) Roll our own calculations for timing on top of AddSample instead
+// of using MeasureSince.
 func ModuleMeasureSinceWithLabels(
 	module string,
 	keys []string,
--- a/protocol/lib/metrics/lib.go
+++ b/protocol/lib/metrics/lib.go
@@ -50,7 +50,6 @@ func AddSampleWithLabels(key string, val float32, labels ...gometrics.Label) {
 // AddSample provides a wrapper functionality for emitting a sample
 // metric.
 func AddSample(key string, val float32) {
-	// TODO why the f is this a differnet library
 	gometrics.AddSampleWithLabels(
 		[]string{key},
 		val,
--- a/protocol/lib/metrics/lib.go
+++ b/protocol/lib/metrics/lib.go
@@ -7,42 +7,39 @@ import (
 	"github.com/cosmos/cosmos-sdk/telemetry"
 )
 
-// main entrypoint for logging in the v4 protocol
-// type Label struct {
-// 	Name  string
-// 	Value string
-// }
+// This file provides a main entrypoint for logging in the v4 protocol.
+// TODO(CLOB-1013) Drop both metrics libraries above for a library
+// that supports float64 (i.e hashicorp go-metrics)
 
 type Label = gometrics.Label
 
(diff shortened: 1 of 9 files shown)


jonfung-dydx:
https://linear.app/dydx/issue/CLOB-1014/enforce-standard-metrics-logging-scheme-going-forward

protocol/lib/metrics/lib.go
```diff
@@ -0,0 +1,90 @@
...
+		start,
+		key,
+	)
+}
+
+// ModuleMeasureSinceWithLabels provides a short hand method for emitting a time measure
+// metric for a module with labels. Global labels are not included
+// in this metric.
+// Please try to use `AddSampleWithLabels` instead.
+func ModuleMeasureSinceWithLabels(
```
jiajames, line 76: This comment is a bit confusing to me, since `AddSampleWithLabels` is not time related. Are you saying that its preferred to pass the calculating timing into `AddSampleWithLabels` instead of using this?

Also, `AddSampleWithLabels` also uses `gometrics`, so wouldn't that have missing global labels too?

protocol/lib/metrics/metric_keys.go
```diff
@@ -0,0 +1,37 @@
...
+	// Clob metric keys
+	// Currently Sorted by type of metric.
+	// TODO make sure the metric type is clear.
+
+	// stat
+	ClobExpiredStatefulOrders                         = "clob_expired_stateful_order_removed"
+	ClobPrepareCheckStateCannotDeleverageSubaccount   = "clob_prepare_check_state_cannot_deleverage_subaccount"
+	ClobDeleverageSubaccountTotalQuoteQuantums        = "clob_deleverage_subaccount_total_quote_quantums"
+	ClobDeleverageSubaccount                          = "clob_deleverage_subaccount"
+	LiqidationsPlacePerpetualLiquidationQuoteQuantums = "liquidations_place_perpetual_liquidation_quote_quantums"
```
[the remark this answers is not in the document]
jiajames: +1

protocol/lib/metrics/metric_keys.go
```diff
@@ -0,0 +1,39 @@
+package metrics
+
+// Metrics Keys Guidelines
+// 1. Be wary of length
+// 2. Prefix by module
+// 3. Suffix keys with a unit of measurement
+// 4. Delimit with '_'
+// 5. Information such as callback type should be added as tags, not in key names.
```
jiajames, line 8: This is helpful, and I would recommend extending a bit further:
1. Provide examples here (for ex. what is a reasonable length, what is an ex. module name (`x/clob` or `clob`), etc.
2. Would be helpful to formalize this into a [Technical Informational Doc](https://www.notion.so/dydx/44d99ab6a60c4278b9257516b6b75209?v=e14b73b5416f4a6780d44d53b54511dc&pvs=4) and add a link here perhaps

protocol/lib/metrics/metric_keys.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Clob Metrics Keys
+const (
+	// Stats
```
jiajames, line 12: Nit question but is there a benefit to formatting these `consts` by metrics type? Seems unnecessary IMO

jiajames (commented):
Might be overkill, but could be cool to also add a CI action to enforce that `gometrics` and `cosmos-sdk/telemetry` aren't being directly used, but not sure if possible

protocol/lib/metrics/metric_keys.go
jonfung-dydx: added examples

https://www.notion.so/dydx/Metrics-Guidelines-f31f02e792bc4914abac2309df02e3b0?pvs=4
started a doc with a todo here, will work on it later since its not P0

protocol/lib/metrics/metric_keys.go
jonfung-dydx: It's useful for when I migrate files over one by one since I can bucket the types, can't hurt to have it here for now. If you feel strongly about this I can combine later.

protocol/lib/metrics/lib.go
jiajames: nit: add TODO to code 😄

jiajames (approved):

PR #797, opened 2023-11-21
Statefully validate stateful orders in PrepareCheckState Replay

jonfung-dydx:
slack [link](https://dydx-team.slack.com/archives/C03SLFHC3L7/p1700586467347989?thread_ts=1700585921.453239&cid=C03SLFHC3L7)

We should validate stateful orders in PrepareCheckState Replay.

### DIFF
--- a/protocol/x/clob/types/mem_clob_keeper.go
+++ b/protocol/x/clob/types/mem_clob_keeper.go
@@ -60,7 +60,6 @@ type MemClobKeeper interface {
 	AddPreexistingStatefulOrder(
 		ctx sdk.Context,
 		order *Order,
-		blockHeight uint32,
 		memclob MemClob,
 	) (
 		orderSizeOptimisticallyFilledFromMatchingQuantums satypes.BaseQuantums,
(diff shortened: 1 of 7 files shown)


protocol/x/clob/types/mem_clob_keeper.go
```diff
@@ -57,6 +57,17 @@ type MemClobKeeper interface {
 		offchainUpdates *OffchainUpdates,
 		err error,
 	)
+	AddPreexistingStatefulOrder(
+		ctx sdk.Context,
+		order *Order,
+		blockHeight uint32,
```
jakob-dydx, line 63: nit: can we update this to not take in a blockHeight? As far as I can tell the blockHeight isn't actually used in this code path. This is being passed in because the `PerformStatefulOrderValidation` function takes in a blockHeight but it is only used for short-term orders. The other place in the code where we call `AddPreexistingStatefulOrder` always passes in zero as blockHeight

jakob-dydx (approved):

PR #808, opened 2023-11-27
[CORE-517] add integration tests for prices governance messages

tqin7:
### Changelist
add integration tests for `MsgUpdateMarketParam`

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testing/e2e/gov/prices_test.go
+++ b/protocol/testing/e2e/gov/prices_test.go
@@ -27,9 +27,9 @@ var (
 	}
 
 	MODIFIED_MARKET_PARAM = pricestypes.MarketParam{
-		Id:                 0,
+		Id:                 GENESIS_MARKET_PARAM.Id,
 		Pair:               "eth-adv4tnt",
-		Exponent:           -8,
+		Exponent:           GENESIS_MARKET_PARAM.Exponent, // exponent cannot be updated
 		MinExchanges:       3,
 		MinPriceChangePpm:  2_002,
 		ExchangeConfigJson: `{"exchanges":[{"exchangeName":"Bitfinex","ticker":"tBTCUSD"}]}`,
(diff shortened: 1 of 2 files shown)


protocol/testing/e2e/gov/prices_test.go
```diff
@@ -0,0 +1,200 @@
...
+		Exponent:           -8,
+		MinExchanges:       2,
+		MinPriceChangePpm:  1_000,
+		ExchangeConfigJson: "{}",
+	}
+
+	MODIFIED_MARKET_PARAM = pricestypes.MarketParam{
+		Id:                 0,
+		Pair:               "eth-adv4tnt",
+		Exponent:           -8,
```
clemire, line 32: nit: you can use `GENESIS_MARKET_PARAM.Exponent` here instead of -8 with a comment that exponents cannot change.

clemire: ... you could also use `GENESIS_MARKET_PARAM.Id` to populate the `Id` field.

protocol/testing/e2e/gov/prices_test.go
```diff
@@ -0,0 +1,200 @@
...
+					ExchangeConfigJson: MODIFIED_MARKET_PARAM.ExchangeConfigJson,
+				},
+			},
+			expectedProposalStatus: govtypesv1.ProposalStatus_PROPOSAL_STATUS_FAILED,
+		},
+		"Failure: empty pair": {
+			msg: &pricestypes.MsgUpdateMarketParam{
+				Authority: lib.GovModuleAddress.String(),
+				MarketParam: pricestypes.MarketParam{
+					Id:                 GENESIS_MARKET_PARAM.Id,
```
clemire, line 86: Question: if you are using `GENESIS_MARKET_PARAM.Id` instead of `MODIFIED_MARKET_PARAM.Id` here (presumably to show the id doesn't change,) does it also make sense to do the same thing with the exponent field? It seems like the treatment of these two fields is inconsistent.

tqin7: good call. modified to `MODIFIED_MARKET_PARAM.Id` to be consistent

clemire (approved):

PR #812, opened 2023-11-28
Add v3.0.0 upgrade handler with module account initialization

roy-dydx:
### Changelist
- Add v3.0.0 upgrade handler to main
- Initialize module accounts in handler to allow v1.0.1 -> v3.0.0 upgrade path for testnet

### Test Plan
N/A. Copied directly from v2.0.0 upgrade handler.

### Author/Reviewer Checklist


protocol/app/upgrades_test.go
```diff
@@ -1,17 +1,10 @@
 package app_test
 
-import (
-	"errors"
-	"testing"
+/*
```
jayy04, line 3: should we just remove these tests? these should already be covered in `x/upgrade` so I don't think these tests are too useful

roy-dydx: yup I agree

PR #827, opened 2023-11-30
[IND-481] Use a single SQL function to process a block.

lcwik:
### Changelist
[IND-481] Use a single SQL function to process a block.

Added support to use named parameterized queries to postgres package.

### Test Plan
Updated existing tests. Moved away from using mocks in several cases.

### Author/Reviewer Checklist


indexer/services/ender/__tests__/handlers/markets/market-create-handler.test.ts
```diff
@@ -137,13 +137,9 @@ describe('marketCreateHandler', () => {
...
     expect(market.minPriceChangePpm).toEqual(50);
 
-    expect(loggerError).toHaveBeenCalledWith(expect.objectContaining({
-      at: 'MarketCreateHandler#logAndThrowParseMessageError',
-      message: 'Market in MarketCreate already exists',
-    }));
     expect(loggerCrit).toHaveBeenCalledWith(expect.objectContaining({
-      at: 'onMessage#onMessage',
-      message: 'Error: Unable to parse message, this must be due to a bug in V4 node',
+      at: expect.stringContaining('PL/pgSQL function dydx_market_create_handler('),
```
lcwik, line 141: Note that the main log location will now also include the stack trace from postgres which contains function names (and now correct line numbers)

indexer/services/ender/__tests__/lib/block-processor.test.ts
```diff
@@ -90,18 +92,18 @@ describe('block-processor', () => {
       transactionIndex0,
       eventIndex0,
     ),
-    createIndexerTendermintEvent(
-      DydxIndexerSubtypes.ASSET,
-      defaultAssetEventBinary,
-      transactionIndex0,
-      eventIndex1,
-    ),
     createIndexerTendermintEvent(
```
lcwik, line 95: The order of these for the mock based testing was wrong since the market needs to exist before we create the asset.

indexer/services/ender/__tests__/lib/sync-handlers.test.ts
```diff
@@ -1,155 +1,42 @@
...
-
-    afterEach(async () => {
-      await dbHelpers.clearData();
-    });
-
-    afterAll(async () => {
-      await dbHelpers.teardown();
-    });
-
     it('successfully adds handler', async () => {
```
lcwik, line 10: Swapped to use mocks for the handlers explicitly here.

indexer/packages/kafka/src/batch-kafka-producer.ts
```diff
@@ -28,7 +28,7 @@ export class BatchKafkaProducer {
   constructor(
     topic: KafkaTopics,
     producer: Producer,
-    maxBatchSizeBytes: number = config.KAFKA_MAX_BATCH_WEBSOCKET_MESSAGE_SIZE_BYTES,
```
lcwik, line 31: default parameters are resolved when the module is loaded so moved it to pass in the value during instantiation to allow for tests to easily modify the value.

indexer/services/ender/src/validators/validator.ts
```diff
@@ -23,6 +26,16 @@ export abstract class Validator<T> {
    */
   public abstract validate(): void;
 
+  /**
+   * Returns the decoded event and any additional details the SQL function needs to process.
+   */
+  // TODO(IND-513): Convert handlers to have a 1-1 relationship with each validator by merging
+  // the order fill handlers together into a single handler (and the respective SQL function)
+  // and push this method down into the handler itself.
+  public getEventForBlockProcessor(): Promise<object> {
```
Christopher-Li, line 35: Is there a way to return a typed object here?

indexer/services/ender/src/validators/order-fill-validator.ts
```diff
@@ -32,6 +37,30 @@ export class OrderFillValidator extends Validator<OrderFillEventV1> {
     }
   }
 
+  public async getEventForBlockProcessor(): Promise<object> {
+    if (this.event.order) {
```
Christopher-Li, line 41: nit: add comment for why we only pull from CanceledOrdersCache if event.order is populated

indexer/services/ender/src/lib/helper.ts
```diff
@@ -91,6 +91,7 @@ export function dateToDateTime(
  * @returns
  */
 export function indexerTendermintEventToEventProtoWithType(
+  blockEventIndex: number,
```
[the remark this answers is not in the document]
lcwik: Each event in the block is stored on the handler so that the corresponding row index is passed to said handler.

Is there a better way I could do this mapping?

indexer/services/ender/src/validators/validator.ts
lcwik: It is `T` + some additional fields (really only the order cancellation status for the order fill handler uses this currently). Is there a good way to represent that?

indexer/services/ender/src/scripts/dydx_block_processor.sql
```diff
@@ -0,0 +1,39 @@
...
+    PERFORM dydx_create_initial_rows_for_tendermint_block(block_height, block_time, block->'txHashes', block->'events');
+
+    /** In genesis, handle sync events first, then batched events. In other blocks, handle batched events first, then sync events. */
+    IF NOT block_height = 0 THEN
+        rval = dydx_block_processor_batched_handlers(block);
+        rval_to_merge = dydx_block_processor_sync_handlers(block);
+    ELSE
+        rval = dydx_block_processor_sync_handlers(block);
+        rval_to_merge = dydx_block_processor_batched_handlers(block);
+    END IF;
```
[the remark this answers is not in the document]
lcwik: Yeah, this is definitely one of the areas that we could clean-up. The only caveat was around why sync / batch handlers are performed in a different order for block height 0.

indexer/services/ender/src/scripts/dydx_block_processor.sql
```diff
@@ -0,0 +1,39 @@
...
+
+    /** Note that arrays are 1-indexed in PostgreSQL and empty arrays return NULL for array_length. */
+     FOR i in 1..coalesce(array_length(rval, 1), 0) LOOP
```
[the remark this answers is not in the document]
lcwik: They are technically indexable over any range but they naturally are from 1...N

indexer/services/ender/src/lib/block-processor.ts
```diff
@@ -187,14 +217,55 @@ export class BlockProcessor {
...
+          dataBytes: values[i],
+        };
+      }
+    });
+
+    const start: number = Date.now();
+    let success = false;
+    let resultRow: pg.QueryResultRow;
+    try {
+      const result: pg.QueryResult = await storeHelpers.rawQuery(
```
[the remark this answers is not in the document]
lcwik: Yes, we pass in the `txId` as an option.

indexer/services/ender/src/validators/validator.ts
Christopher-Li: we could just use `T` for now and add a comment that OrderFill id different in the OrderFillValidator and in validator.ts. I don't think that would cause an error, but if does then we can leave it like this and file a bug ticket

indexer/services/ender/src/lib/helper.ts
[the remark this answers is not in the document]
vincentwschau: Makes sense, can you add a comment describing this property?

indexer/services/ender/src/scripts/dydx_block_processor_unordered_handlers.sql
```diff
@@ -0,0 +1,69 @@
...
+                            'order',
+                            dydx_order_fill_handler_per_order('order', block_height, block_time, event_data, event_index, transaction_index, jsonb_array_element_text(block->'txHashes', transaction_index), 'TAKER', 'LIMIT', USDC_ASSET_ID, event_data->>'takerCanceledOrderStatus'));
+                ELSE
+                    rval[i] = jsonb_build_object(
+                            'makerOrder',
+                            dydx_liquidation_fill_handler_per_order('makerOrder', block_height, block_time, event_data, event_index, transaction_index, jsonb_array_element_text(block->'txHashes', transaction_index), 'MAKER', 'LIQUIDATION', USDC_ASSET_ID),
+                            'liquidationOrder',
+                            dydx_liquidation_fill_handler_per_order('liquidationOrder', block_height, block_time, event_data, event_index, transaction_index, jsonb_array_element_text(block->'txHashes', transaction_index), 'TAKER', 'LIQUIDATED', USDC_ASSET_ID));
```
vincentwschau, line 50: nit: Add comments indicating the first case is for normal fills, and the second is for liquidations.

indexer/services/ender/src/validators/validator.ts
vincentwschau: Could you have a second type in the generic declaration? Like:
```
export abstract class Validator<T, V extends T>  {
...
getExtendedEvent(event: T): V {
  return event;
}
...
public getEventForBlockProcessor(): V {
  return getExtendedevent(Promise.resolve(this.event));
}
```

and then in the order fill handler you have V be the type with the extra fields and override `getExtendedEvent`?

I don't think this is too important though, the above approach would add a lot of extra type declarations to every other handler that doesn't need it.

If we don't add it, perhaps just add a comment that this is `object` because of the logic in order fill handler?

lcwik: I got:
```TS2322: Type  T  is not assignable to type  V 
 T  is assignable to the constraint of type  V , but  V  could be instantiated with a different subtype of constraint  object```
 
 Swapped to use `T` since that is technically true since it is a sub-type of `T`.

Christopher-Li (approved):

PR #829, opened 2023-12-01
[CLOB-1009] - Allow FINAL_SETTLEMENT, and cancel open stateful orders

jakob-dydx:
- enables the transition of clob pair status to `FINAL_SETTLEMENT` to indicate market close
- forcefully cancels open stateful orders, sending indexer events for order removals
- removes untriggered stateful orders from the clob keeper in-memory map

Should not be merged until we are comfortable with `main` supporting this status transition

NOTE: at this point in time (Dec 20th), this PR has become the feature branch for the changes for this project. All PRs able to be merged to main have been merged to main but ones that depend on this PR and need to only be merged into main once this feature is complete and ready to ship are now sitting in this PR. Merging this will complete the P0 protocol work for the project.


protocol/x/clob/keeper/clob_pair.go
```diff
@@ -468,13 +468,19 @@ func (k Keeper) UpdateClobPair(
...
 	oldStatus := oldClobPair.Status
 	newStatus := clobPair.Status
-	if oldStatus != newStatus && !types.IsSupportedClobPairStatusTransition(oldStatus, newStatus) {
-		return errorsmod.Wrapf(
-			types.ErrInvalidClobPairStatusTransition,
-			"Cannot transition from status %+v to status %+v",
-			oldStatus,
-			newStatus,
-		)
+	if oldStatus != newStatus {
```
jayy04, line 471: nit: move `oldStatus == newStatus` check to `types.IsSupportedClobPairStatusTransition`.  

transitioning to the same status is still a transition. if in the future we want to disallow transition to the state status for `INITIALIZING` or `FINAL_SETTLEMENT`, they should all be defined inside this transition function

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,70 @@
...
+	safelyRemoveStatefulOrder := func(ctx sdk.Context, orderId types.OrderId) {
+		defer func() {
+			if r := recover(); r != nil {
+				k.Logger(ctx).Error(
+					"mustCancelStatefulOrdersForFinalSettlement: Failed to remove stateful order with OrderId %+v: %v",
+					orderId,
+					r,
+				)
+			}
+		}()
```
jayy04, line 39: for my understanding - when would a panic occur?

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,70 @@
...
+		k.MustRemoveStatefulOrder(ctx, orderId)
+	}
+
+	for _, order := range statefulOrders {
```
jayy04, line 43: any performance concerns here to iterate over all stateful orders?

ideally maybe we can update the keys from e.g. `SO/` to `SO/<clob_pair_id>` so that we only iterate over stateful orders for that market. 

no need to block on this though. we can optimize separately. maybe add a TODO or something

jayy04 (approved):

protocol/x/clob/keeper/final_settlement.go
jakob-dydx: a panic could occur if for some reason the order failed to be found in state (unlikely since we just fetched the whole list of stateful orders), but also from this block. 
```
if len(longTermOrdersExpiringAtTime) != len(updatedStatefulOrdersExpiringAtTime)+1 {
		panic(
			fmt.Sprintf(
				"MustRemoveStatefulOrder: order ID %v is not in state for time %v",
				orderId,
				goodTilBlockTime,
			),
		)
	}
```

I think these are unlikely scenarios, but I had a brief convo with Lucas about this and think it makes sense to recover from panics here so that EndBlocker doesn't panic if a stateful order simply fails to be removed from a final settlement market. If a bug is introduced into that code path we don't want EndBlocker panicking

protocol/x/clob/keeper/final_settlement.go
jakob-dydx: yeah not super concerned here simply because this is a one time operation. Will add a todo

protocol/x/clob/keeper/final_settlement.go
jayy04: I see. 

In that case, we shouldn't be calling `MustRemoveStatefulOrder` and recover from panics. We should define a `MaybeRemoveStatefulOrder` and not panic at all. 

Slight preference for a new function for cleaner code, but will let you make the call

proto/dydxprotocol/indexer/protocol/v1/clob.proto
```diff
@@ -176,6 +176,7 @@ enum ClobPairStatus {
   // CLOB_PAIR_STATUS_FINAL_SETTLEMENT represents a clob pair that has been
   // deactivated. Clob pairs in this state do not accept new orders and trading
   // is blocked. All open positions are closed by the protocol when the clob
-  // pair gains this status.
+  // pair gains this status. All open stateful orders are closed and short-term
```
lucas-dydx, line 179: Suggestion, up to you:
```
  // is blocked. All open positions are closed and open stateful orders canceled by the protocol when the clob
  // pair transitions to this status. All open short-term
```

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,71 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	indexerevents "github.com/dydxprotocol/v4-chain/protocol/indexer/events"
+	"github.com/dydxprotocol/v4-chain/protocol/indexer/indexer_manager"
+	indexershared "github.com/dydxprotocol/v4-chain/protocol/indexer/shared"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// MustEnterFinalSettlement holds logic executed when a market transitions to FINAL_SETTLEMENT status.
+// This function will forcefully cancel all stateful open orders for the clob pair.
+func (k Keeper) mustEnterFinalSettlement(ctx sdk.Context, clobPairId types.ClobPairId) {
```
lucas-dydx, line 13: Suggestion, up to you:
```suggestion
func (k Keeper) mustTransitionToFinalSettlement(ctx sdk.Context, clobPairId types.ClobPairId) {
```

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,71 @@
...
+	// TODO(CLOB-1053): Iterate over stateful orders for only specified clob pair
+	for _, order := range statefulOrders {
+		if order.GetClobPairId() == clobPairId {
+			// Remove from state, recovering from panic if necessary
+			safelyRemoveStatefulOrder(ctx, order.OrderId)
+
+			// Append to RemovedStatefulOrderIds so this order gets removed
+			// from the memclob in PrepareCheckState during the PurgeInvalidMemclobState step
+			processProposerMatchesEvents.RemovedStatefulOrderIds = append(
+				processProposerMatchesEvents.RemovedStatefulOrderIds,
```
lucas-dydx, line 52: To clarify, it's not an issue if removed untriggered conditional orders are added to this data structure right?

I don't think it is based on [this logic](https://github.com/dydxprotocol/v4-chain/blob/a795a81bd3bf52dcc6e2d4b184d42d6be7eeeeeb/protocol/x/clob/memclob/memclob.go#L1270-L1278), but might be worth noting that in a comment within that logic (at first glance it's non-obvious imo)

protocol/x/clob/types/clob_pair.go
```diff
@@ -10,9 +10,15 @@ import (
...
+	ClobPair_STATUS_ACTIVE: {
+		ClobPair_STATUS_FINAL_SETTLEMENT: struct{}{},
+	},
 	ClobPair_STATUS_INITIALIZING: {
-		ClobPair_STATUS_ACTIVE: struct{}{},
+		ClobPair_STATUS_ACTIVE:           struct{}{},
+		ClobPair_STATUS_FINAL_SETTLEMENT: struct{}{},
+	},
+	ClobPair_STATUS_FINAL_SETTLEMENT: {
+		ClobPair_STATUS_INITIALIZING: struct{}{},
```
lucas-dydx, line 21: Why is this a valid state transition?

For simplicity, is it fine to assume that the final settlement state is terminal for markets? Seems safer to just re-initialize a new CLOB pair.

Note if we allow this, then it's possible for a final settlement market to eventually begin trading again which is difficult to test (requires winding down market then bringing it up again which would require multiple gov proposals right?)

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// MustEnterFinalSettlement holds logic executed when a market transitions to FINAL_SETTLEMENT status.
+// This function will forcefully cancel all stateful open orders for the clob pair.
+func (k Keeper) mustEnterFinalSettlement(ctx sdk.Context, clobPairId types.ClobPairId) {
+	// Forcefully cancel all stateful orders from state for this clob pair
+	k.mustCancelStatefulOrdersForFinalSettlement(ctx, clobPairId)
+
+	// Delete untriggered conditional orders for this clob pair from memory
```
lucas-dydx, line 17: Small suggestion:
```suggestion
	// Forcefully cancel all stateful orders from state for this clob pair.
	k.mustCancelStatefulOrdersForFinalSettlement(ctx, clobPairId)

	// Delete untriggered conditional orders for this clob pair from memory.
```

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,71 @@
...
+	delete(k.UntriggeredConditionalOrders, clobPairId)
+}
+
+// mustCancelStatefulOrdersForFinalSettlement forcefully cancels all stateful orders
+// for the provided ClobPair. These orders will be removed from the memclob in PrepareCheckState.
```
lucas-dydx, line 22: Suggestion: is this more accurate?
```suggestion
// for the provided ClobPair. Open orders will be removed from the memclob in PrepareCheckState.
```

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,71 @@
...
+					r,
+				)
+			}
+		}()
+		k.MustRemoveStatefulOrder(ctx, orderId)
+	}
+
+	// TODO(CLOB-1053): Iterate over stateful orders for only specified clob pair
+	for _, order := range statefulOrders {
+		if order.GetClobPairId() == clobPairId {
```
lucas-dydx, line 45: Suggestion: simpler to just continue if they don't match to avoid nesting if statements

```suggestion
		if order.GetClobPairId() != clobPairId {
		    continue
		}
```

lucas-dydx (approved):

protocol/x/clob/types/clob_pair.go
jakob-dydx: I took the stance that the flow for re-enabling trading for a market would be to transition back into initializing mode, and then to active mode. 

The reasoning here is that this project does not cover cleaning up the clob pair from state (removing from state). This is because we don't want to allow the clob pair to be deleted and then a new clob pair to be created with the same id as the deleted one. This would have implications, particularly for indexer. 

Because we are going to leave the clob pair in state, it makes sense to allow trading to be re-enabled with a MsgUpdateClobPair message instead of creating a new clob pair

protocol/x/clob/keeper/final_settlement.go
jakob-dydx: no because only the stateful open orders are removed from the memclob in PCS, the short-term ones aren't. So "these" refers to "these stateful orders"

protocol/x/clob/keeper/final_settlement.go
jakob-dydx: i think I will leave this as is, defining the MaybeRemoveStatefulOrder function would basically involve copy and pasting the entire MustRemoveStatefulOrder function. Seems fine to just recover from the panic honestly

protocol/x/clob/abci.go
```diff
@@ -197,8 +197,26 @@ func PrepareCheckState(
...
 
 	// 6. Get all potentially liquidatable subaccount IDs and attempt to liquidate them.
-	subaccountIds := daemonLiquidationInfo.GetLiquidatableSubaccountIds()
-	if err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds); err != nil {
+	liquidatableSubaccountIds := daemonLiquidationInfo.GetLiquidatableSubaccountIds()
+	subaccountsToDeleverage, err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, liquidatableSubaccountIds)
+	if err != nil {
+		panic(err)
+	}
+	subaccountPositionInfo := daemonLiquidationInfo.GetSubaccountsWithPositions()
```
jayy04, line 205: heads up - I am planning to update `GetSubaccountsWithPosition` to take a `perpetualId` and also `isLong` as parameters. 

currently GetSubaccountsWithPositions makes a copy of the entire map, which is not necessary in most cases. 

non blocking, feel free to merge and I can update this in my PR

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,23 +532,88 @@ func (k Keeper) ProcessDeleveraging(
...
+	subaccountsToDeleverage []subaccountToDeleverage,
+) error {
+	defer func() {
+		telemetry.MeasureSince(
+			time.Now(),
+			types.ModuleName,
+			metrics.LiquidateSubaccounts_Deleverage,
+			metrics.Latency,
+		)
+	}()
```
jayy04, line 600: I don't think this works. time.Now() will get evaluated when defer function runs

protocol/x/clob/keeper/final_settlement.go
jayy04: Sure, functional-wise it's fine to panic and recover. but recovering from panics that may potentially occur is a code smell and tech debt imo, and may mask some real panics. 

not sure about Go, but IIRC in languages like Java, it's also quite expensive to catch/handle exceptions, so there might be performance implications. 

No need to block on this, let's go ahead and merge this PR. I will probably put out a PR to clean this up.

jayy04 (approved):

protocol/x/clob/types/clob_pair.go
lucas-dydx: >The reasoning here is that this project does not cover cleaning up the clob pair from state (removing from state). This is because we don't want to allow the clob pair to be deleted and then a new clob pair to be created with the same id as the deleted one. This would have implications, particularly for indexer.

Why not just create a new CLOB pair for trading the market, and leave the other one permanently in state?

I personally don't think it makes sense to support this functionality since it overcomplicates things. I think we should keep it simple as possible to minimize the surface area of this change.

Note that when markets are de-listed, they are essentially only wound back-up with significant changes (for example LUNA => LUNA classic). It is probably fine to create a new CLOB pair for this usecase.

LMK wyt.

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -589,6 +637,19 @@ func (k Keeper) validateInternalOperationAgainstClobPairStatus(
 			clobPairId,
 			types.ClobPair_STATUS_INITIALIZING,
 		)
+	case types.ClobPair_STATUS_FINAL_SETTLEMENT:
+		// Only allow deleveraging events. This allows the protocol to close out open
+		// positions in the market. All trading is blocked.
```
lucas-dydx, line 642: Suggestion: is this more accurate?

```suggestion
		// Only allow deleveraging events. This allows the protocol to close out open
		// positions in the market. All other operations are not allowed.
```

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -33,13 +33,17 @@ func (k Keeper) MaybeDeleverageSubaccount(
...
+	shouldDeleverageAtBankruptcyPrice, shouldDeleverageAtOraclePrice, err := k.CanDeleverageSubaccount(
+		ctx,
+		subaccountId,
+		perpetualId,
+	)
 	if err != nil {
 		return new(big.Int), err
 	}
 
 	// Early return to skip deleveraging if the subaccount can't be deleveraged.
```
lucas-dydx, line 45: Wdyt of the following? IMO slightly more accurate but a bit verbose and maybe too tightly coupled to the `CanDeleverageSubaccount` implementation, up to you.
```suggestion
	// Early return to skip deleveraging if the subaccount doesn't have negative equity or a position in a final settlement market.
```

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -152,16 +165,24 @@ func (k Keeper) CanDeleverageSubaccount(
...
+	clobPairId, err := k.GetClobPairIdForPerpetual(ctx, perpetualId)
+	if err != nil {
+		return false, false, err
+	}
+	clobPair := k.mustGetClobPair(ctx, clobPairId)
+
+	// Negative TNC, deleverage at bankruptcy price.
+	if bigNetCollateral.Sign() == -1 {
+		return true, false, nil
 	}
```
lucas-dydx, line 180: nit: move the fetching of the CLOB pair to after the `bigNetCollateral` sign check (since it's not used if you early return)?

protocol/x/clob/keeper/deleveraging.go
lucas-dydx: Why can't this just be the following (so func args are evaluated right away)?

```suggestion
	defer telemetry.MeasureSince(
			time.Now(),
			types.ModuleName,
			metrics.LiquidateSubaccounts_Deleverage,
			metrics.Latency,
		)
	}
```

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -0,0 +1,73 @@
...
+// mustCancelStatefulOrdersForFinalSettlement forcefully cancels all stateful orders
+// for the provided ClobPair. These orders will be removed from the memclob in PrepareCheckState.
+func (k Keeper) mustCancelStatefulOrdersForFinalSettlement(ctx sdk.Context, clobPairId types.ClobPairId) {
+	statefulOrders := k.GetAllStatefulOrders(ctx)
+	processProposerMatchesEvents := k.GetProcessProposerMatchesEvents(ctx)
+
+	// This logic is executed in EndBlocker and should not panic. This would be unexpected,
+	// but if it happens we would rather recover and continue if an order fails to be removed from state
+	// rather than halt the chain.
+	safelyRemoveStatefulOrder := func(ctx sdk.Context, orderId types.OrderId) {
```
lucas-dydx, line 30: suggestion: `safelyRemoveStatefulOrder` is a bit of a misnomer, maybe you could instead call it `removeStatefulOrderWithoutPanicing` (or something like that)?

protocol/x/clob/memclob/memclob.go
```diff
@@ -694,6 +694,7 @@ func (m *MemClobPriceTimePriority) DeleverageSubaccount(
 	subaccountId satypes.SubaccountId,
 	perpetualId uint32,
 	deltaQuantums *big.Int,
+	isFinalSettlement bool,
```
lucas-dydx, line 697: suggestion: update func comment to note if `isFinalSettlement` is true it will be the oracle price? Also maybe note that error will be returned if the subaccount can't be deleveraged?

lucas-dydx (approved):

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: yeah that's my bad, Lucas' suggestion is right

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: I'll update this, might as well make it as clear as possible

protocol/x/clob/types/clob_pair.go
jakob-dydx: I see what you're saying, and this was my initial thought as well but during spec review process I think I was nudged the other direction.

To support your suggestion we would need to remove the current clob pair id from the clob keeper's `PerpetualIdToClobPairId` in-memory mapping of perpetual id -> clob pair. This is because we currently only support one clob pair per perpetual. If we want to be able to create a new one then the old one needs to be removed.

We would also need to update the hydration logic on startup to not populate Clob Pairs with final settlement status into this in-memory mapping.

I think this is kind of strange because the in-memory data structure (1 clob pair) would be a modified view of what is actually in state (2 clob pairs)

jakob-dydx: also in slack you listed the concern of what happens if all positions haven't been closed by the time the market re-opens, but this problem still stands if we create a new clob pair because the clob pairs would in theory be pointing to the same perpetual so any trading in the new clob pair would be altering position size of the final settlement positions from the other clob pair

PR #832, opened 2023-12-01
feat(mev): enable sending BpMevMatches and ProposalReceiveTime to MEV indexer

Zygimantass:
### Changelist
Adds a `bp_mev_matches` and `proposal_receive_time` field to the `MEVDatapoint` structure. This enables Skip and others to perform additional debugging on MEV telemetry data sent from nodes.

### Test Plan
* Ran nodes with this patch on both mainnet and testnet

### Author/Reviewer Checklist
- [x] If the PR has breaking postgres changes to the indexer add the `indexer-postgres-breaking` label.


proto/dydxprotocol/clob/mev.proto
```diff
@@ -57,4 +57,6 @@ message ValidatorMevMatches {
 message MevNodeToNodeMetrics {
   ValidatorMevMatches validator_mev_matches = 1;
   repeated ClobMidPrice clob_mid_prices = 2 [ (gogoproto.nullable) = false ];
+  ValidatorMevMatches bp_mev_matches = 3;
+  uint64 proposal_receive_time = 4;
```
lucas-dydx, line 61: suggestion: update field name to indicate this is in seconds

lucas-dydx (approved):

proto/dydxprotocol/clob/mev.proto
Zygimantass: we already use this field in production, so if this is a nit, I'd prefer to keep it as-is.

PR #833, opened 2023-12-01
[CLOB-1015][CLOB-1016] Update MatchPerpetualDeleveraging and DeleveragingEventV1 to have final settlement flag

jakob-dydx:
- modify the `MatchPerpetualDeleveraging` operations queue type to have an `is_final_settlement` flag. The oracle price is used for final settlement instead of the bankruptcy price, indicating to the protocol how to process this operation in DeliverTx.
- the same flag has been added to the DeleveragingEventV1 proto for the purposes of indicating to indexer the difference between a final settlement deleverage event and a regular deleverage event


proto/dydxprotocol/indexer/events/events.proto
```diff
@@ -184,6 +184,9 @@ message DeleveragingEventV1 {
   uint64 price = 5;
   // `true` if liquidating a short position, `false` otherwise.
   bool is_buy = 6;
+  // `true` if the deleveraging event is for final settlement, indicating
+  // the match occurred at the oracle price rather than bankruptcy price.
```
lucas-dydx, line 188: Not: also note it's the bankruptcy price of the deleveraged subaccount (not offsetting)

lucas-dydx (approved):

jayy04 (approved):

PR #834, opened 2023-12-04
[CLOB-1021] final settlement DeliverTx validation to block trading

jakob-dydx:
- this PR disallows all non deleveraging operations in the ops queue for clob pairs in `FINAL_SETTLEMENT`
- the idea here is that the protocol should be able to insert standard deleveraging and final settlement deleveraging matches (to close out all open positions in a market) into the ops queue but all other trading should be blocked 
- the tests included in this PR also confirm that all edge cases are checked surrounding the use of the `isFinalSettlement` flag


protocol/x/clob/keeper/clob_pair.go
```diff
@@ -595,6 +595,21 @@ func (k Keeper) validateInternalOperationAgainstClobPairStatus(
...
 			types.ClobPair_STATUS_INITIALIZING,
 		)
+	case types.ClobPair_STATUS_FINAL_SETTLEMENT:
+		// Only allow deleveraging events. This allows the protocol to close out open
+		// positions in the market. All trading is blocked.
+		if match, ok := internalOperation.Operation.(*types.InternalOperation_Match); ok {
+			if _, ok := match.Match.Match.(*types.ClobMatch_MatchPerpetualDeleveraging); ok {
+				return nil
+			}
+		}
```
jayy04, line 605: nit: slightly cleaner to do 
```suggestion
		if match := internalOperation.GetMatch(); match != nil {
			if match.GetMatchPerpetualDeleveraging() != nil {
				return nil
			}
		}
```

protocol/x/clob/keeper/process_operations_test.go
```diff
@@ -1426,6 +1426,289 @@ func TestProcessProposerOperations(t *testing.T) {
 			},
 			expectedError: types.ErrInvalidOrderRemoval,
 		},
+		"Fails with order removal for market in final settlement": {
```
jayy04, line 1429: for my understanding - didn't we add a order removal reason for this?

protocol/x/clob/keeper/process_operations_test.go
```diff
@@ -1426,6 +1426,289 @@ func TestProcessProposerOperations(t *testing.T) {
...
+							FillAmount: 10,
+							MakerOrderId: constants.LongTermOrder_Alice_Num0_Id1_Clob0_Sell20_Price10_GTBT10.OrderId,
+						},
+					},
+				),
+			},
+			expectedError: types.ErrOperationConflictsWithClobPairStatus,
+		},
+		// Liquidations are disallowed for markets in final settlement because they may result
+		// in a position increasing in size. This is not allowed for markets in final settlement.
```
jayy04, line 1482: for my understanding - total OI for the market is going to either stay the same or decrease, right? so technically we can allow liquidations to go through. probably just a product decision?

protocol/x/clob/keeper/process_operations_test.go
```diff
@@ -1426,6 +1426,289 @@ func TestProcessProposerOperations(t *testing.T) {
...
+				constants.ClobPair_Btc,
+			},
+			subaccounts: []satypes.Subaccount{
+				// liquidatable: MMR = $5000, TNC = $499
+				constants.Carl_Num0_1BTC_Short_50499USD,
+				constants.Dave_Num0_1BTC_Long_50000USD,
+			},
+			marketIdToOraclePriceOverride: map[uint32]uint64{
+				constants.BtcUsd.MarketId: 5_050_000_000, // $50,500 / BTC
+			},
```
jayy04, line 1615: nit: we can also just use a subaccount that has a negative TNC at $50k for simplicity. up to you

protocol/x/clob/keeper/process_operations_test.go
```diff
@@ -1426,6 +1426,289 @@ func TestProcessProposerOperations(t *testing.T) {
...
+				constants.Carl_Num0: constants.Usdc_Asset_50_000.GetBigQuantums().Int64(),
+				constants.Dave_Num0: constants.Usdc_Asset_100_000.GetBigQuantums().Int64(),
+			},
+			expectedPerpetualPositions: map[satypes.SubaccountId][]*satypes.PerpetualPosition{
+				constants.Carl_Num0: {},
+				constants.Dave_Num0: {},
+			},
+		},
+		// This throws an error because the CanDeleverageSubaccount function will return false for
+		// shouldFinalSettlePosition, but the IsFinalSettlement flag is set to true.
```
jayy04, line 1598: can we add a test for 
- `CanDeleverageSubaccount` return true for shouldFinalSettlePosition
- IsFinalSettlement flag is false

jayy04 (approved):

protocol/x/clob/keeper/process_operations_test.go
jakob-dydx: we added an indexer order removal reason (for when stateful orders are forcefully cancelled), there is no ops queue OrderRemoval for this though

All stateful orders will be cancelled in EndBlocker when gov proposal is processed, so we don't need an OrderRemoval type for this in the ops queue. They will never need to be proposed

protocol/x/clob/keeper/process_operations_test.go
jakob-dydx: I can ask a product person about this... my thinking was that it is not good to have anyone's position size increasing after the market enters final settlement. If we wanted to allow for liquidations then we could also consider not cancelling stateful orders to increase the chances that the liquidation matches with a maker

jakob-dydx: also we might cancel all short-term orders at some point in the future, so a liquidation order wouldn't even be possible to fill at that point. Don't think we should bother supporting this.

Also if we did support this it would mean we would need to allow short term order placements in the ops queue as well to allow the liquidation order to match against short-term makers. Don't think this is worth the complexity, nor do I think it is helpful in the first place

jayy04: yeah I was just curious. probably not a big deal like you said.

no need to ping product since they were also part of the spec review. as long as you are comfortable with this design decision, let's stick to it 🔥

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -595,6 +595,21 @@ func (k Keeper) validateInternalOperationAgainstClobPairStatus(
 			clobPairId,
 			types.ClobPair_STATUS_INITIALIZING,
 		)
+	case types.ClobPair_STATUS_FINAL_SETTLEMENT:
+		// Only allow deleveraging events. This allows the protocol to close out open
+		// positions in the market. All trading is blocked.
+		if match := internalOperation.GetMatch(); match != nil {
+			if match.GetMatchPerpetualDeleveraging() != nil {
```
lucas-dydx, line 602: Suggestion: cleaner to put on one line? Up to you

```suggestion
		if match := internalOperation.GetMatch(); match != nil && match.GetMatchPerpetualDeleveraging() != nil {
```

lucas-dydx (approved):

PR #835, opened 2023-12-04
[CLOB-1028] allow settlement at oracle price in ProcessDeleveraging

jakob-dydx:
- updates ProcessDeleveraging to take in `deltaQuoteQuantums`
- this allows us to decide on the price of the deleveraging event and to pass this in to the function
- I have created a helper method which fetches the price - bankruptcy for regular deleveraging and oracle price for final settlement deleveraging. This is determined by the status of the clob pair linked with the perpetual

No tests for OffsetSubaccountPerpetualPosition have been included for final settlement clob pairs yet, because the final settlement status is still not allowed on `main`. Tests will be added for this in a separate PR for when the status is enabled. Holding off on enabling the status to properly gate these changes


protocol/x/clob/keeper/deleveraging.go
```diff
@@ -241,6 +242,7 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
 				*offsettingSubaccount.Id,
 				perpetualId,
 				deltaQuantums,
+				false,
```
jayy04, line 245: nit: I would pull the branching logic out of `ProcessDeleveraging` and just pass in `fillPriceDeltaQuoteQuantums` to this function. 

personally not a fan of using booleans as function arguments in general since it kind of ruins code readability. wdyt?

protocol/x/clob/keeper/deleveraging_test.go
```diff
@@ -1095,6 +1096,211 @@ func TestProcessDeleveraging(t *testing.T) {
...
+				require.ErrorContains(t, err, tc.expectedErr.Error())
+			}
+		})
+	}
+}
+
+func TestProcessDeleveragingAtOraclePrice(t *testing.T) {
+	tests := map[string]struct {
+		// Setup.
+		liquidatedSubaccount satypes.Subaccount
```
jayy04, line 1127: nit: let's use a different name since we are not really liquidating this subaccount.

protocol/x/clob/keeper/deleveraging_test.go
```diff
@@ -1095,6 +1096,211 @@ func TestProcessDeleveraging(t *testing.T) {
...
+		offsettingSubaccount satypes.Subaccount
+		deltaQuantums        *big.Int
+
+		// Expectations.
+		expectedLiquidatedSubaccount satypes.Subaccount
```
jayy04, line 1132: ditto

protocol/x/clob/keeper/deleveraging_test.go
```diff
@@ -1095,6 +1096,211 @@ func TestProcessDeleveraging(t *testing.T) {
...
+		expectedOffsettingSubaccount satypes.Subaccount
+		expectedErr                  error
+	}{
+		"Liquidated: well-collateralized, offsetting: well-collateralized": {
```
jayy04, line 1136: ditto

jayy04 (approved):

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: hmm, I'd argue it is a little cleaner from a code perspective to push this branching down into `ProcessDeleveraging`. If we have to pass in the fill price to the function then we need to calculate this anywhere we call `ProcessDeleveraging`. Right now that is in both PrepareCheckState and DeliverTx (for persisting the deleveraging match). This will mean the code for calculating fill price or bankruptcy price will be written twice in the code base instead of just once in the `ProcessDeleveraging` function. Also makes those top level functions that call `ProcessDeleveraging` more clear to read as more of the deleveraging logic is abstracted away

jakob-dydx: curious how strong your opinion on this is

jakob-dydx: i guess deltaQuantums is currently passed in to ProcessDeleveraging though. Could see this going either way.

protocol/x/clob/keeper/deleveraging_test.go
jayy04: sure - this is test only so will let you make the call here

protocol/x/clob/keeper/deleveraging.go
jayy04: Slight preference for pulling this out. Having the calculation not dependent on a boolean flag is also more future proof in case we need to settle at a price other than bankruptcy/oracle price. 

Your argument about having the logic twice makes sense. Maybe we can define a helper function to minimize duplicated code, something like `GetDeleveragingSettlementPrice(bool)`?

We might not even need that boolean since if TNC < 0 we will use bankruptcy, and oracle otherwise. but we don't need to worry about this now

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -236,13 +236,37 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
 			}
 
+			// Fetch delta quote quantums. Calculated at bankruptcy price for standard
+			// deleveraging and at oracle price for final settlement deleveraging.
+			deltaQuoteQuantums, err := k.getDeleveragingQuoteQuantumsDelta(
+				ctx,
+				perpetualId,
+				liquidatedSubaccountId,
+				deltaQuantums,
+			)
```
jakob-dydx, line 246: errors returned from bankruptcy price calculation were previously caught within ProcessDeleveraging and recovered from in OffsetSubaccountPerpetualPosition, continuing onto the next iteration without breaking. I am opting to continue this pattern, but will log a debug msg

protocol/x/clob/keeper/process_operations.go
```diff
@@ -651,29 +651,41 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
 	for _, fill := range matchDeleveraging.GetFills() {
-		deltaQuantums := new(big.Int).SetUint64(fill.FillAmount)
-		if deltaQuantumsIsNegative {
-			deltaQuantums.Neg(deltaQuantums)
+		deltaBaseQuantums := new(big.Int).SetUint64(fill.FillAmount)
+		if deltaBaseQuantumsIsNegative {
+			deltaBaseQuantums.Neg(deltaBaseQuantums)
+		}
+
+		deltaQuoteQuantums, err := k.getDeleveragingQuoteQuantumsDelta(
```
jakob-dydx, line 662: this is where we call ProcessDeleveraging in DeliverTx, we call the helper function to get the delta quote quantums to pass in

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -305,22 +329,50 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
+// This returns the bankruptcy price for standard deleveraging operations, and the oracle price for
+// final settlement deleveraging operations. The type of deleveraging event is determined by the
+// clob pair status of the clob pair associated with the provided perpetual.
+func (k Keeper) getDeleveragingQuoteQuantumsDelta(
+	ctx sdk.Context,
+	perpetualId uint32,
+	subaccountId satypes.SubaccountId,
+	deltaQuantums *big.Int,
+) (deltaQuoteQuantums *big.Int, err error) {
+	clobPair := k.mustGetClobPairForPerpetualId(ctx, perpetualId)
```
jakob-dydx, line 342: Question: is it okay if we panic here? This logic is executed in PrepareCheckState. Would we rather recover from this and continue? I think it is fine to panic but just want to double check.

jayy04: yeah should be fine to panic - if clob pair is missing, then something is really really wrong

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -305,22 +329,50 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
+	isFinalSettlement := clobPair.Status == types.ClobPair_STATUS_FINAL_SETTLEMENT
+
+	if isFinalSettlement {
+		deltaQuoteQuantums, err = k.perpetualsKeeper.GetNetNotional(ctx, perpetualId, deltaQuantums)
+	} else {
+		deltaQuoteQuantums, err = k.GetBankruptcyPriceInQuoteQuantums(
```
jayy04, line 348: nit: we can just `return k.perpetualsKeeper.GetNetNotional` or `return k.GetBankruptcyPriceInQuoteQuantums`. up to you

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -236,13 +236,37 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
+			// deleveraging and at oracle price for final settlement deleveraging.
+			deltaQuoteQuantums, err := k.getDeleveragingQuoteQuantumsDelta(
+				ctx,
+				perpetualId,
+				liquidatedSubaccountId,
+				deltaQuantums,
+			)
+			if err != nil {
+				liquidatedSubaccount := k.subaccountsKeeper.GetSubaccount(ctx, liquidatedSubaccountId)
+				k.Logger(ctx).Debug(
```
jayy04, line 249: should this be `Debug` or `Error`?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -236,13 +236,37 @@ func (k Keeper) OffsetSubaccountPerpetualPosition(
...
+					"Encountered error when getting quote quantums for deleveraging",
+					"error", err,
+					"blockHeight", ctx.BlockHeight(),
+					"checkTx", ctx.IsCheckTx(),
```
jayy04, line 253: nit: `k.Logger()` already has this, so we can remove this here

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: true, will update

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: will change to `Error`

protocol/x/clob/keeper/deleveraging_test.go
```diff
@@ -1095,6 +1136,215 @@ func TestProcessDeleveraging(t *testing.T) {
 				*tc.offsettingSubaccount.GetId(),
 				uint32(0),
 				tc.deltaQuantums,
+				bankruptcyPriceQuoteQuantums,
```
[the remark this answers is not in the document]
jakob-dydx: yeah I needed to push this into the if statement though, because the `GetBankruptcyPriceInQuoteQuantums` returns an error for my test cases where tc.expectedErr != nil. So the test code was erroring out before executing the ProcessDeleveraging logic. 

I am intentionally skipping this calculation for error cases.

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: Just realized this function isn't doing exactly what I want for usability. The purpose of this function is to allow the deleveraging flow to be reusable such that non-negative TNC subaccounts with positions in a final settlement market can be closed at the oracle price.

Currently this function is written such that ALL subaccounts (including negative TNC) for the closed market will be settled at oracle price. This isn't what we want. I think the easiest way to updated this would be to do a collateral check and use the oracle price iff `isFinalSettlement && nonNegativeTNC`. 

I think what I'll do here is use `CanDeleverageSubaccount` to determine if the account has non negative TNC and only call this if the market is in final settlement.

PR #836, opened 2023-12-05
[DEC-2196] deprecate non-linear margin requirements

jayy04:
### Changelist
- Deprecate scaling (non-linear) margin requirements

### Test Plan

### Author/Reviewer Checklist


proto/dydxprotocol/perpetuals/perpetual.proto
```diff
@@ -96,7 +96,7 @@ message LiquidityTier {
   // The maximum position size at which the margin requirements are
   // not increased over the default values. Above this position size,
   // the margin requirements increase at a rate of sqrt(size).
-  uint64 base_position_notional = 5;
```
teddyding, line 99: Is there an upgrade handler needed for this change? Or will the protocol just ignore this value after the upgrade?

proto/dydxprotocol/indexer/events/events.proto
```diff
@@ -374,7 +374,7 @@ message LiquidityTierUpsertEventV1 {
   // The maximum position size at which the margin requirements are
```
teddyding, line 374: Might be helpful to note this is deprecated since some version in the comment, similar to Cosmos SDK [style](https://github.com/search?q=org%3Adydxprotocol+%22%2F%2F+Deprecated+post+v0.46.x%22&type=code)

I think here we should say `Deprecated since v3.x`?

protocol/x/clob/e2e/liquidation_deleveraging_test.go
```diff
@@ -694,134 +694,6 @@ func TestPlacePerpetualLiquidation_Deleveraging(t *testing.T) {
 				},
 			},
 		},
-		`Can place and match liquidation order for subaccount with non-linear margin requirement`: {
```
teddyding, line 697: It seems like we are mostly deleting unit tests. Do we have a test where a subaccount would've been liquidated under non-linear margin requirements, but is collateralized after this change?

proto/dydxprotocol/perpetuals/perpetual.proto
jayy04: No, since the field is only marked as deprecated so it's backwards compatible.

protocol/x/perpetuals/types/liquidity_tier_test.go
```diff
@@ -213,115 +199,39 @@ func TestLiquidityTierGetMaxAbsFundingClampPpm(t *testing.T) {
...
 		},
 		"initial margin 56.7243%, quote quantums = 123_456, should round up to 70_030": {
-			initialMarginPpm:     uint32(567_243), // 56.7243%
-			bigQuoteQuantums:     big.NewInt(123_456),
-			basePositionNotional: uint64(1_000_000),
+			initialMarginPpm: uint32(567_243), // 56.7243%
+			bigQuoteQuantums: big.NewInt(123_456),
 			// initial margin * margin adjustment * quote quantums
 			// = 56.7243% * 100% * 123_456
 			// ~= 70029.5518 -> round up to 70030
```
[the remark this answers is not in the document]
jayy04: are you taking about this test specifically? from the comment below - the adjustment was 100% so the margin requirements are the same before and after 

Removed the comment since margin adjustment is no longer applicable, but pasted here as an FYI
```
			// initial margin * margin adjustment * quote quantums
			// = 56.7243% * 100% * 123_456
			// ~= 70029.5518 -> round up to 70030
```

There are a few tests that got updated I believe.

teddyding (approved):

PR #842, opened 2023-12-06
[CLOB-1010] final settlement checktx validation

jakob-dydx:
- reject all orders in CheckTx for markets in final settlement

### DIFF
--- a/protocol/x/clob/keeper/clob_pair.go
+++ b/protocol/x/clob/keeper/clob_pair.go
@@ -373,6 +373,13 @@ func (k Keeper) validateOrderAgainstClobPairStatus(
 				clobPair.Status,
 			)
 		}
+	case types.ClobPair_STATUS_FINAL_SETTLEMENT:
+		return errorsmod.Wrapf(
+			types.ErrOrderConflictsWithClobPairStatus,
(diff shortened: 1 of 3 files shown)


jayy04 (approved):

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -373,6 +373,13 @@ func (k Keeper) validateOrderAgainstClobPairStatus(
 				clobPair.Status,
 			)
 		}
+	case types.ClobPair_STATUS_FINAL_SETTLEMENT:
```
lucas-dydx, line 376: Somewhat unrelated, but why is it a valid state transition for a final settlement CLOB pair to transition into `INITIALIZING`? [source](https://github.com/dydxprotocol/v4-chain/blob/0f0d706c80054911482720426bd17caf11294c97/protocol/x/clob/types/clob_pair.go#L20-L22)

lucas-dydx (approved):

protocol/x/clob/keeper/clob_pair.go
jakob-dydx: answer here: https://github.com/dydxprotocol/v4-chain/pull/829#discussion_r1430742315

PR #847, opened 2023-12-06
Implement `GetNegativeTncSubaccountSeenAtBlock` and `SetNegativeTncSubaccountSeenAtBlock` [CLOB-1026]

lucas-dydx:
### Changelist
Implement `GetNegativeTncSubaccountSeenAtBlock` and `SetNegativeTncSubaccountSeenAtBlock`.

### Test Plan
Unit tests only, didn't write E2E tests since this isn't used in any top-level functionality yet.

### Author/Reviewer Checklist
- [x] If the PR has breaking postgres changes to the indexer add the `indexer-postgres-breaking` label.

### DIFF
--- a/protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
+++ b/protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
@@ -1,0 +1,63 @@
+package keeper
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	gogotypes "github.com/cosmos/gogoproto/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+)
+
+// GetNegativeTncSubaccountSeenAtBlock gets the last block height a negative TNC subaccount was
+// seen in state. Note this defaults to 0 if it has not been set.
+func (k Keeper) GetNegativeTncSubaccountSeenAtBlock(
+	ctx sdk.Context,
+) uint32 {
+	store := ctx.KVStore(k.storeKey)
+	return k.getNegativeTncSubaccountSeenAtBlock(store)
+}
+
+// getNegativeTncSubaccountSeenAtBlock is a helper function that takes a store and returns the last
+// block height a negative TNC subaccount was seen in state.
+func (k Keeper) getNegativeTncSubaccountSeenAtBlock(
+	store sdk.KVStore,
+) uint32 {
+	b := store.Get(
+		[]byte(types.NegativeTncSubaccountSeenAtBlockKey),
+	)
+	blockHeight := gogotypes.UInt32Value{Value: 0}
+	if b != nil {
+		k.cdc.MustUnmarshal(b, &blockHeight)
+	}
+
+	return blockHeight.Value
+}
+
+// SetNegativeTncSubaccountSeenAtBlock sets a block number in state where a negative TNC subaccount
+// was seen. This function will overwrite previous values at this key.
+// This function will panic if the old block heigh is greater than the new block height.
+func (k Keeper) SetNegativeTncSubaccountSeenAtBlock(
+	ctx sdk.Context,
+	blockHeight uint32,
+) {
+	store := ctx.KVStore(k.storeKey)
+
+	// Panic if the new block height is less than the current block height.
+	currentValue := k.getNegativeTncSubaccountSeenAtBlock(store)
+	if blockHeight < currentValue {
+		panic(
+			fmt.Sprintf(
+				"SetNegativeTncSubaccountSeenAtBlock: new block height (%d) is less than the current block height (%d)",
+				blockHeight,
+				currentValue,
+			),
+		)
+	}
+
+	blockHeightValue := gogotypes.UInt32Value{Value: 0}
+	blockHeightValue.Value = blockHeight
+	store.Set(
+		[]byte(types.NegativeTncSubaccountSeenAtBlockKey),
+		k.cdc.MustMarshal(&blockHeightValue),
+	)
+}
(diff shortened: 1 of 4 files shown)


protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -0,0 +1,62 @@
...
+	if b != nil {
+		k.cdc.MustUnmarshal(b, &blockHeight)
+	}
+
+	return blockHeight.Value
+}
+
+// SetNegativeTncSubaccountSeenAtBlock sets a block number in state where a negative TNC subaccount
+// was seen. This function will overwrite previous values at this key.
+// This function will panic if the old block heigh is greater than the new block height.
```
jonfung-dydx, line 38:
```suggestion
// This function will panic if the old block height is greater than the new block height.
```

protocol/x/subaccounts/types/keys.go
```diff
@@ -11,6 +11,9 @@ const (
 
 // State
 const (
-	// SubaccountKeyPrefix is the prefix to retrieve all Subaccount
+	// SubaccountKeyPrefix is the prefix to retrieve all Subaccounts
 	SubaccountKeyPrefix = "SA:"
+	// NegativeTncSubaccountSeenAtBlockKey is the store key that stores the last
+	// block a negative TNC subaccount was seen in state.
+	NegativeTncSubaccountSeenAtBlockKey = "NegSA:"
```
jonfung-dydx, line 18:
```suggestion
	NegativeTncSubaccountSeenAtBlockKey = "LastNegSA:"
```

protocol/x/subaccounts/keeper/negative_tnc_subaccount_test.go
```diff
@@ -0,0 +1,146 @@
...
+				require.Equal(
+
+					t,
+					uint32(10),
+					k.GetNegativeTncSubaccountSeenAtBlock(ctx),
+				)
+			},
+
+			expectedMultiStoreWrites: []string{
+				"NegSA:",
```
jonfung-dydx, line 84: Meganit use the variable

protocol/x/subaccounts/keeper/negative_tnc_subaccount_test.go
```diff
@@ -0,0 +1,146 @@
...
+				"NegSA:",
+				"NegSA:",
+			},
+		},
+		"Block height can be updated to same block height": {
+			setupTestAndPerformAssertions: func(ctx sdk.Context, k keeper.Keeper) {
+				k.SetNegativeTncSubaccountSeenAtBlock(ctx, 0)
+				require.Equal(
+
```
jonfung-dydx, line 94: extra newline. ditto with all other newlines

jonfung-dydx (approved):

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -0,0 +1,62 @@
...
+
+// getNegativeTncSubaccountSeenAtBlock is a helper function that takes a store and returns the last
+// block height a negative TNC subaccount was seen in state.
+func (k Keeper) getNegativeTncSubaccountSeenAtBlock(
+	store sdk.KVStore,
+) uint32 {
+	b := store.Get(
+		[]byte(types.NegativeTncSubaccountSeenAtBlockKey),
+	)
+	blockHeight := gogotypes.UInt32Value{Value: 0}
```
jayy04, line 28: I almost wonder if we should be pessimistic here and return a non zero default value 🤔

protocol/x/subaccounts/keeper/negative_tnc_subaccount_test.go
```diff
@@ -0,0 +1,146 @@
...
+			},
+
+			expectedMultiStoreWrites: []string{
+				"NegSA:",
+				"NegSA:",
+				"NegSA:",
+				"NegSA:",
+			},
+		},
+		"Block height can be updated to same block height": {
```
jayy04, line 90: for my understanding - when would we need to set this to the same height?

protocol/x/subaccounts/types/keys.go
jayy04: Let's just keep it short. We wouldn't need to look at this unless we are using `iaviewer` so it's fine for this to be less human readable

jayy04 (approved):

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
lucas-dydx: Honestly I prefer returning 0, since then it won't cause issues when re-starting the network from genesis (although I'll admit the current value will gate withdrawals in the first 50 blocks after re-starting from genesis, which I personally didn't think was an issue)

protocol/x/subaccounts/keeper/negative_tnc_subaccount_test.go
lucas-dydx: I don't think this would ever be necessary, but I figured it's fine as long as it's not decreasing

Would you like me to add validation that the block height is always increasing?

protocol/x/subaccounts/types/keys.go
lucas-dydx: Agreed with Jay!

protocol/x/subaccounts/keeper/negative_tnc_subaccount_test.go
jayy04: nah I think this is correct if this information eventually would come from the daemon. 

if block time > daemon loop time, then I guess this could technically happen?

PR #848, opened 2023-12-06
[CLOB-1017]  final settlement deleveraging step in PrepareCheckState

jakob-dydx:
This PR adds a new step in PrepareCheckState to deleverage subaccounts with open positions in final settlement markets. It does this by leveraging the new field from the liquidations daemon which contains a mapping of perpetual -> position side -> subaccount id which allows us to know which subaccounts have open positions in specific markets. We re-use the `MaxDeleveragingAttemptsPerBlock` clob flag to limit the number of attempts. The regular liquidations/deleveraging flow now returns a total number of deleveraging attempts so we don't go over the limit. 

Note: this PR needs to be updated once the liquidations daemon actually outputs this information. Right now a placeholder variable of the expected type has been used and initialized to be empty.


protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,23 +529,50 @@ func (k Keeper) ProcessDeleveraging(
...
-		),
-	)
+	return nil
+}
+
+// DeleverageSubaccountsInFinalSettlementMarkets uses the subaccountOpenPositionInfo returned from the
+// liquidations daemon to deleverage subaccounts with open positions in final settlement markets. Note
+// this function will deleverage both negative TNC and non-negative TNC subaccounts. The deleveraging code
+// uses the bankruptcy price for the former and the oracle price for the latter.
+func (k Keeper) DeleverageSubaccountsInFinalSettlementMarkets(
```
jakob-dydx, line 539: this is the function called in PrepareCheckState to deleverage positions in closed markets

protocol/x/clob/keeper/liquidations.go
```diff
@@ -105,7 +106,18 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
 			}
 
 			// Return unexpected errors.
-			return err
+			return 0, err
+		}
+
+		// Skip liquidation if the ClobPair for the liquidation order is in final settlement. There is a separate flow
```
jakob-dydx, line 112: We skip attempting to place the liquidation if the market is in final settlement, as liquidation matches are not allowed in closed markets. We do append the order to unfilledLiquidations so that the deleveraging flow may naturally pick up negative TNC subaccounts in the market and deleverage them as part of the standard deleveraging flow

protocol/x/clob/keeper/process_operations.go
```diff
@@ -619,27 +619,43 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
-				"PersistMatchDeleveragingToState: Failed to determine if subaccount can be deleveraged. "+
-					"SubaccountId %+v, error %+v",
-				liquidatedSubaccountId,
-				err,
-			),
-		)
-	} else if !canDeleverageSubaccount {
-		// TODO(CLOB-853): Add more verbose error logging about why deleveraging failed validation.
+	// If `IsFinalSettlement` flag on deleveraging match is set to true, verify that the market is in final settlement.
+	if matchDeleveraging.IsFinalSettlement && clobPair.Status != types.ClobPair_STATUS_FINAL_SETTLEMENT {
```
jakob-dydx, line 626: This piece of validation is required so that block proposers cannot propose final settlement deleverage events for markets which are not in final settlement.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -619,27 +619,43 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
+			"MatchPerpetualDeleveraging %+v is a final settlement match, but the clob pair %+v is not in final settlement",
+			matchDeleveraging,
+			clobPair,
 		)
 	}
 
-	perpetualId := matchDeleveraging.GetPerpetualId()
+	// Skip the below validation for final settlement deleveraging matches. Final settlement matches do
+	// not require the subaccount to have negative TNC to be considered deleveragable.
+	if !matchDeleveraging.IsFinalSettlement {
```
jakob-dydx, line 637: only need to verify that the subaccount has negative TNC if this is not a final settlement match

protocol/x/clob/abci.go
```diff
@@ -198,7 +199,14 @@ func PrepareCheckState(
...
 	// 6. Get all potentially liquidatable subaccount IDs and attempt to liquidate them.
 	subaccountIds := liquidatableSubaccountIds.GetSubaccountIds()
-	if err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds); err != nil {
+	numDeleveragingAttempts, err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds)
+	if err != nil {
+		panic(err)
+	}
+
+	// 7. Deleverage subaccounts with open positions in final settlement markets.
+	subaccountOpenPositionInfo := make(map[uint32]map[bool]map[subaccounttypes.SubaccountId]struct{})
```
jakob-dydx, line 208: This variable is a placeholder for the real information the liquidations daemon will be retrieving

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,23 +529,50 @@ func (k Keeper) ProcessDeleveraging(
...
+			perpetualId := clobPair.MustGetPerpetualId()
+			finalSettlementPerpetualIds[perpetualId] = struct{}{}
+		}
+	}
+
+	// Deleverage subaccounts with open positions in final settlement markets.
+	for perpetualId := range finalSettlementPerpetualIds {
+		for isBuy := range subaccountOpenPositionInfo[perpetualId] {
+			for subaccountId := range subaccountOpenPositionInfo[perpetualId][isBuy] {
+				if numDeleveragingAttempts >= int(k.Flags.MaxDeleveragingAttemptsPerBlock) {
```
lucas-dydx, line 557: Should we use a separate flag to limit the number of final settlement deleveraging ops? This looks through all subaccounts with open positions in the market so it's bounded right?

protocol/x/clob/keeper/liquidations.go
lucas-dydx: Why not pick another perpetual ID to liquidate?

What if the subaccount is liquidatable but has multiple positions? Closing their open but final settled position at the oracle price won't change their collateralization, so IMO could make sense to just liquidate another position

However I could see us leaning towards the simplest approach and just waiting another block since I _think_ it picks a random position to liquidate

Wdyt?

lucas-dydx (commented):
WIP review, will finish rest later

protocol/x/clob/keeper/liquidations.go
jakob-dydx: yeah admittedly was going for simplicity here. Picking a new position to liquidate would mean introducing a new flow here where we selectively re-generate liquidation orders depending on clob pair status and I just don't think that is worth it for now. If you feel strongly, we could implement this later but for now I think this is fine

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: I thought about this, but opted for re-using the flag in the interest of respecting the current performance implications of proposing deleveraging operations.

When you say this operation is bounded, what do you mean by that? In theory there is no limit to the number of subaccounts that can havae open positions in a market so it isn't bounded

protocol/x/clob/keeper/liquidations.go
```diff
@@ -105,7 +106,18 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+		// for triggering deleveraging/final settlement for positions in closed markets.
+		if _, found := clobPairStatuses[liquidationOrder.GetClobPairId()]; !found {
+			clobPair := k.mustGetClobPair(ctx, liquidationOrder.GetClobPairId())
+			clobPairStatuses[liquidationOrder.GetClobPairId()] = clobPair.Status
+		}
+		if clobPairStatuses[liquidationOrder.GetClobPairId()] == types.ClobPair_STATUS_FINAL_SETTLEMENT {
+			unfilledLiquidations = append(unfilledLiquidations, *liquidationOrder)
+			continue
```
jayy04, line 120: nit: For normal orders, trading is disabled within the `keeper.PlaceOrder` flow(they would fail clob pair status transition validation). for consistency, this logic should also live inside `keeper.PlacePerpetualLiquidation`

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,23 +521,50 @@ func (k Keeper) ProcessDeleveraging(
...
+		if clobPair.Status == types.ClobPair_STATUS_FINAL_SETTLEMENT {
+			perpetualId := clobPair.MustGetPerpetualId()
+			finalSettlementPerpetualIds[perpetualId] = struct{}{}
+		}
+	}
+
+	// Deleverage subaccounts with open positions in final settlement markets.
+	for perpetualId := range finalSettlementPerpetualIds {
+		for isBuy := range subaccountOpenPositionInfo[perpetualId] {
+			for subaccountId := range subaccountOpenPositionInfo[perpetualId][isBuy] {
```
jayy04, line 548: this is going to introduce a ton of MEV (although deleveraging is currently not included in MEV calculation IIRC), but still, we should probably iterate over these deterministically.

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -33,13 +32,24 @@ func (k Keeper) MaybeDeleverageSubaccount(
...
 
-	// Early return to skip deleveraging if the subaccount can't be deleveraged.
-	if !canPerformDeleveraging {
+	// Generate final settlement deleveraging matches (at oracle price) if subaccount has non-negative TNC and
+	// the market is in final settlement. This means we will regularly deleverage subaccounts with negative TNC
+	// in markets that are in final settlement.
+	isFinalSettlement := clobPair.Status == types.ClobPair_STATUS_FINAL_SETTLEMENT && !canPerformDeleveraging
+
+	// Early return to skip deleveraging if the subaccount can't be deleveraged. Final settlement deleveraging.
+	if !isFinalSettlement && !canPerformDeleveraging {
```
jayy04, line 52: this is slightly confusing to read since you have `!canPerformDeleveraging` at the end of both

jayy04: also nit: I feel like we should move this to `keeper.CanDeleverageSubaccount`. We use the word `Deleverage` to mean both -ve TNC deleveraging and final settlement deleveraging in pretty much all deleveraging functions, so doesn't make sense for `CanDeleverageSubaccount` to be -ve TNC deleveraging only

protocol/x/clob/abci.go
```diff
@@ -198,7 +200,27 @@ func PrepareCheckState(
 
 	// 6. Get all potentially liquidatable subaccount IDs and attempt to liquidate them.
 	subaccountIds := liquidatableSubaccountIds.GetSubaccountIds()
-	if err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds); err != nil {
+	numDeleveragingAttempts, err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds)
+	if err != nil {
+		panic(err)
+	}
+
+	// 7. Deleverage subaccounts with open positions in final settlement markets.
```
jayy04, line 208: organizational suggestion: 

instead of having step 6 being liquidate + -ve TNC deleveraging, let's make step 6 liquidation only and return []subaccounts to deleverage instead of `numDeleveragingAttempts`

step 7 can be deleveraging for both -v TNC and final settlement.  wdyt?

protocol/x/clob/abci.go
```diff
@@ -198,7 +200,27 @@ func PrepareCheckState(
...
+	subaccountPositionInfo := map[uint32]map[bool]map[satypes.SubaccountId]struct{}{
```
lucas-dydx, line 209: Should we have a pre-processing step that verifies the `subaccountPositionInfo` map is accurate (it could contain entries for subaccounts that have closed positions due to daemon lag right)?

protocol/x/clob/keeper/liquidations.go
jakob-dydx: `PerformStatefulOrderValidation` is where we call `validateOrderAgainstClobPairStatus`, the validation you are referring to. We don't currently call stateful validation when placing liquidation orders, instead we skip straight to calling `matchOrder` because the liquidation order is protocol generated and therefore cannot / does not need to be validated in the same way a regular order does.

The `validateOrderAgainstClobPairStatus` fn expects an `Order` which cannot be fetched from a liquidation order. 

I feel like keeping this as is would be simpler. If I have the `PlacePerpetualLiquidation` fn return an error depending on clob pair status then we also need to properly recover from this error  in `LiquidateSubaccountsAgainstOrderbook` when we could have just skipped the function call in the first place

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: noted, working on updating this

protocol/x/clob/abci.go
jakob-dydx: hmm, this would only be useful if the preprocessing was pretty fast right? The preprocessing would be able to save us from attempting to final settle deleverage subaccounts with no open position. I don't think it is very costly for the deleveraging operation to simply fail due to no position existing, so personally don't think we need to preprocess

protocol/x/clob/abci.go
jakob-dydx: actually this might be a little weird because of how the deleveraging attempts from step 6 come from a list of unfilled liquidation orders. Will think about this and make the change if I think it makes sense. I see your point

jakob-dydx: decided to follow your advice, updating now

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -137,13 +137,16 @@ func (k Keeper) GetInsuranceFundBalance(
...
 // CanDeleverageSubaccount returns true if a subaccount can be deleveraged.
-// Specifically, this function returns true if both of the following are true:
-// - The subaccount's total net collateral is negative.
-// This function returns an error if `GetNetCollateralAndMarginRequirements` returns an error.
+// This function returns two booleans, canDeleverageSubaccount and shouldFinalSettlePosition.
+// - canDeleverageSubaccount is true if the subaccount can be deleveraged.
+// - shouldFinalSettlePosition is true if the position should be final settled (this occurs at oracle price)
+// This function returns an error if `GetNetCollateralAndMarginRequirements` returns an error or if there is
+// an error when fetching the clob pair for the provided perpetual.
 func (k Keeper) CanDeleverageSubaccount(
```
jakob-dydx, line 145: Major change here is that this function now returns a `shouldFinalSettlePosition` boolean to tell the protocol whether or not the subaccount should be final settle deleveraged or regularly deleveraged

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +520,79 @@ func (k Keeper) ProcessDeleveraging(
...
-			),
-		),
+	return nil
+}
+
+// GetSubaccountsWithOpenPositionsInFinalSettlementMarkets uses the subaccountOpenPositionInfo returned from the
+// liquidations daemon to fetch subaccounts with open positions in final settlement markets. These subaccounts
+// will be deleveraged either at the oracle price if non-negative TNC or at bankruptcy price if negative TNC during
+// PrepareCheckState.
+func (k Keeper) GetSubaccountsWithOpenPositionsInFinalSettlementMarkets(
```
jakob-dydx, line 530: This function is called in PrepareCheckState to fetch subaccounts with open positions in final settlement markets, so we can combine with the set of subaccounts to deleverage returned from step 6 (liquidations)

protocol/daemons/server/liquidation.go
```diff
@@ -25,6 +26,15 @@ func (server *Server) WithDaemonLiquidationInfo(
 	return server
 }
 
+// SetSubaccountOpenPositions stores the list of subaccount open positions in a go-routine safe map.
+// Placeholder to allow for testing.
+func (s *Server) SetSubaccountOpenPositions(
```
jakob-dydx, line 31: @jayy04 I've added this as a placeholder so I can modify this datastructure for testing. I imagine there will be a new function created for this in one of your PRs

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -152,16 +155,22 @@ func (k Keeper) CanDeleverageSubaccount(
...
-	if bigNetCollateral.Sign() >= 0 {
-		return false, nil
+	clobPairId, err := k.GetClobPairIdForPerpetual(ctx, perpetualId)
+	if err != nil {
+		return false, false, err
 	}
+	clobPair := k.mustGetClobPair(ctx, clobPairId)
+
+	// Can deleverage subaccount if net collateral is negative or if the market is in final settlement.
+	canDeleverageSubaccount = bigNetCollateral.Sign() == -1 || clobPair.Status == types.ClobPair_STATUS_FINAL_SETTLEMENT
```
jayy04, line 168: nit: these two `bool` should be mutually exclusive. let's move this to line 160 (before `k.GetClobPairIdForPerpetual` to save us from 1 state read if TNC < 0

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +520,79 @@ func (k Keeper) ProcessDeleveraging(
...
+	subaccountOpenPositionInfo map[uint32]*types.SubaccountOpenPositionInfo,
+) (subaccountsToDeleverage []subaccountToDeleverage) {
+	// Gather perpetualIds for perpetuals whose clob pairs are in final settlement.
+	finalSettlementPerpetualIds := make(map[uint32]struct{})
+	for _, clobPair := range k.GetAllClobPairs(ctx) {
+		if clobPair.Status == types.ClobPair_STATUS_FINAL_SETTLEMENT {
+			perpetualId := clobPair.MustGetPerpetualId()
+			finalSettlementPerpetualIds[perpetualId] = struct{}{}
+		}
+	}
```
jayy04, line 541: nit: let's use `lib.FilterSlice(...)` here. slightly simpler and also iterating over map might have MEV implications so let's try to preserve ordering if possible.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -637,10 +638,19 @@ func (k Keeper) PersistMatchDeleveragingToState(
 			"Subaccount %+v failed deleveraging validation",
 			liquidatedSubaccountId,
 		)
+	} else if matchDeleveraging.IsFinalSettlement != shouldFinalSettlePosition {
```
jayy04, line 641: there is a bug here - in line 634, it's now valid to have `canDeleverageSubaccount == false`, and `shouldFinalSettlePosition == true`

the two bools are technically mutually exclusive, and I think we should rename these variables to make this clear

protocol/x/clob/keeper/liquidations.go
```diff
@@ -105,7 +113,20 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
+		// Skip liquidation if the ClobPair for the liquidation order is in final settlement.
+		if _, found := clobPairStatuses[liquidationOrder.GetClobPairId()]; !found {
+			clobPair := k.mustGetClobPair(ctx, liquidationOrder.GetClobPairId())
+			clobPairStatuses[liquidationOrder.GetClobPairId()] = clobPair.Status
+		}
+		if clobPairStatuses[liquidationOrder.GetClobPairId()] == types.ClobPair_STATUS_FINAL_SETTLEMENT {
+			subaccountsToDeleverage = append(subaccountsToDeleverage, subaccountToDeleverage{
+				SubaccountId: liquidationOrder.GetSubaccountId(),
+				PerpetualId:  liquidationOrder.MustGetLiquidatedPerpetualId(),
+			})
```
jayy04, line 128: we don't need to append here, right? since this is going to get picked up by `GetSubaccountsWithOpenPositionsInFinalSettlementMarkets`

protocol/x/clob/keeper/liquidations.go
jayy04: Putting it in `PlacePerpetualLiquidation` will make it easier to write a unit test, e.g. if we decide to write a regression test or something. Error handling should be trivial.

but yeah idk. `LiquidateSubaccountsAgainstOrderbook` is a top level function, so I feel like this implementation detail should live in a downstream function, such as `keeper.PlacePerpetualLiquidation`. If you want this to fail earlier or get skipped more upstream, this logic can also be inside `GetPerpetualPositionToLiquidate` (additional benefit is that we would be picking another perpetual position to liquidate like Lucas mentioned).

Will let you make the call here.

protocol/x/clob/abci.go
```diff
@@ -197,8 +197,24 @@ func PrepareCheckState(
 	}
 
 	// 6. Get all potentially liquidatable subaccount IDs and attempt to liquidate them.
-	subaccountIds := daemonLiquidationInfo.GetLiquidatableSubaccountIds()
-	if err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, subaccountIds); err != nil {
+	liquidatableSubaccountIds := daemonLiquidationInfo.GetLiquidatableSubaccountIds()
+	subaccountsToDeleverage, err := keeper.LiquidateSubaccountsAgainstOrderbook(ctx, liquidatableSubaccountIds)
```
jayy04, line 201: thinking about this more - `subaccountsToDeleverage` here should mostly contain subaccounts with unfilled liquidations and need to be deleveraged at bankruptcy price (-ve TNC), right?

we do have `DaemonLiquidationInfo.NegativeTncSubaccounts`, maybe we can just use that? this will de-couple step 6&7 completely. wdyt?

protocol/x/clob/keeper/deleveraging.go
jayy04, line 168: nit: maybe move this to above `k.GetClobPairIdForPerpetual`. this save us 1 state read if TNC < 0.

also curious to get your thoughts on making these two bools mutually exclusive. something like `shouldDeleverageAtBankruptcyPrice` and `shouldDeleverageAtOraclePrice` to improve code readability?

jayy04 (approved):

protocol/x/clob/keeper/liquidations.go
jakob-dydx: I can push it downstream. I don't want it upstream because I actually do want the liquidations flow to pick up negative TNC subaccounts in markets that are in final settlement and deleverage them normally (at bankruptcy price). So would prefer to not skip these and find a different position

protocol/x/clob/keeper/liquidations.go
jakob-dydx: my motivation in still appending here is that this negative TNC deleveraging attempt will occur earlier if I append it here rather than wait for the other function to pick it up. It's more important that we deleverage negative TNC subaccounts than it is that we do the final settlement deleveraging for non-negative TNC subaccounts

protocol/x/clob/abci.go
jakob-dydx: yes it does contain -ve TNC subaccounts, paired with a perpetual id that we tried to liquidate with. I think it probably makes sense to use `DaemonLiquidationInfo.NegativeTncSubaccounts` as you suggested. Unsure if we want to deleverage using the same perpetual we tried to liquidate with though, if so we would need to add some extra logic for that

jayy04: yeah maybe we can do this in a separate PR. no need to block on this

jayy04: maybe create a ticket and add a TODO?

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: hmm, not a bad idea

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +530,73 @@ func (k Keeper) ProcessDeleveraging(
...
+			return err
+		}
+	}
+
+	telemetry.MeasureSince(
+		startDeleverageSubaccounts,
+		types.ModuleName,
+		metrics.LiquidateSubaccounts_Deleverage,
+		metrics.Latency,
 	)
```
lucas-dydx, line 600: nit: this can just be a defer statement now right?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +530,73 @@ func (k Keeper) ProcessDeleveraging(
...
+
+// DeleverageSubaccounts deleverages a slice of subaccounts paired with a perpetual position to deleverage with.
+func (k Keeper) DeleverageSubaccounts(
+	ctx sdk.Context,
+	subaccountsToDeleverage []subaccountToDeleverage,
+) error {
+	// For each unfilled liquidation, attempt to deleverage the subaccount.
+	startDeleverageSubaccounts := time.Now()
+	var i int
+	for i = 0; i < int(k.Flags.MaxDeleveragingAttemptsPerBlock) && i < len(subaccountsToDeleverage); i++ {
```
lucas-dydx, line 580: nit: can you do this instead?

```suggestion
	for i := 0; i < int(k.Flags.MaxDeleveragingAttemptsPerBlock) && i < len(subaccountsToDeleverage); i++ {
```

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -304,6 +304,42 @@ func (k Keeper) GetAllClobPairs(ctx sdk.Context) (list []types.ClobPair) {
...
+	}
+
+	if !types.IsSupportedClobPairStatus(clobPair.Status) {
+		panic(
+			fmt.Sprintf(
+				"validateLiquidationAgainstClobPairStatus: clob pair status %v is not supported",
+				clobPair.Status,
+			),
+		)
+	}
```
lucas-dydx, line 329: Hmm for my understanding wouldn't this logic allow a liquidation in an `INITIALIZING` market?

Why would that be valid? Shouldn't liquidations only be valid for `ACTIVE` markets?

protocol/x/clob/keeper/clob_pair.go
```diff
@@ -304,6 +304,42 @@ func (k Keeper) GetAllClobPairs(ctx sdk.Context) (list []types.ClobPair) {
 	return
 }
 
+// validateLiquidationAgainstClobPairStatus returns an error if placing the provided
+// liquidation order would conflict with the clob pair's current status.
+func (k Keeper) validateLiquidationAgainstClobPairStatus(
```
lucas-dydx, line 309: Question: should we unit test this function?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -152,16 +165,22 @@ func (k Keeper) CanDeleverageSubaccount(
...
+	clobPair := k.mustGetClobPair(ctx, clobPairId)
 
-	// The subaccount's total net collateral is negative, so deleveraging can be performed.
-	return true, nil
+	// Negative TNC, deleverage at bankruptcy price.
+	if bigNetCollateral.Sign() == -1 {
+		return true, false, nil
+	}
+
+	// Non-negative TNC, deleverage at oracle price if market is in final settlement.
```
lucas-dydx, line 182: suggestion: worth explaining that if a subaccount's TNC is non-negative, then it should always be a valid state transition to deleverage them at the current oracle price?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +530,73 @@ func (k Keeper) ProcessDeleveraging(
...
+		}
+		for _, subaccountId := range positionInfo.SubaccountsWithShortPosition {
+			subaccountsToDeleverage = append(subaccountsToDeleverage, subaccountToDeleverage{
+				SubaccountId: subaccountId,
+				PerpetualId:  perpetualId,
+			})
+		}
+	}
+
+	return subaccountsToDeleverage
```
lucas-dydx, line 569: Would it make sense to stat number of subaccounts with positions in final settlement markets here?

protocol/x/clob/keeper/deleveraging.go
lucas-dydx, line 600: nit: this can just be defer'd right?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -499,22 +530,73 @@ func (k Keeper) ProcessDeleveraging(
...
+}
+
+// DeleverageSubaccounts deleverages a slice of subaccounts paired with a perpetual position to deleverage with.
```
lucas-dydx, line 572: suggestion: note it returns an error if deleveraging fails

protocol/x/clob/keeper/liquidations.go
```diff
@@ -258,6 +244,10 @@ func (k Keeper) PlacePerpetualLiquidation(
 		metrics.PlacePerpetualLiquidation,
 	)
 
+	if err := k.validateLiquidationAgainstClobPairStatus(ctx, liquidationOrder); err != nil {
```
lucas-dydx, line 247: Worth updating func comment that perp liquidations can return an error if placed on a non-active market?

protocol/x/clob/keeper/process_operations.go
```diff
@@ -630,17 +635,26 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
+		// of use of the isFinalSettlement flag. The isFinalSettlement flag should be set to true if-and-only-if the
+		// subaccount has non-negative TNC and the market is in final settlement. Otherwise, it must be false.
+		return errorsmod.Wrapf(
+			types.ErrDeleveragingIsFinalSettlementFlagMismatch,
+			"MatchPerpetualDeleveraging %+v has isFinalSettlement flag (%v), expected (%v)",
+			matchDeleveraging,
+			matchDeleveraging.IsFinalSettlement,
+			shouldDeleverageAtOraclePrice,
+		)
 	}
```
lucas-dydx, line 656: Suggestion: can we split these out into different if statements? IMO `else if` statements are more difficult to reason about

protocol/x/clob/keeper/process_operations.go
```diff
@@ -630,17 +635,26 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
 		return errorsmod.Wrapf(
 			types.ErrInvalidDeleveragedSubaccount,
 			"Subaccount %+v failed deleveraging validation",
 			liquidatedSubaccountId,
 		)
+	} else if matchDeleveraging.IsFinalSettlement != shouldDeleverageAtOraclePrice {
+		// Throw error if the isFinalSettlement flag does not match the expected value. This prevents misuse or lack
+		// of use of the isFinalSettlement flag. The isFinalSettlement flag should be set to true if-and-only-if the
+		// subaccount has non-negative TNC and the market is in final settlement. Otherwise, it must be false.
+		return errorsmod.Wrapf(
```
lucas-dydx, line 649: Suggestion: update func comment with new error conditions

protocol/x/clob/types/errors.go
```diff
@@ -201,6 +201,11 @@ var (
 		43,
 		"Order has remaining size",
 	)
+	ErrLiquidationConflictsWithClobPairStatus = errorsmod.Register(
+		ModuleName,
+		44,
+		"Liquidation conflicts with ClobPair status",
+	)
```
lucas-dydx, line 208: nit: isn't this a liquidations error?

protocol/x/clob/keeper/liquidations.go
```diff
@@ -105,21 +112,26 @@ func (k Keeper) LiquidateSubaccountsAgainstOrderbook(
...
 
 			// Return unexpected errors.
-			return err
+			return nil, err
 		}
 
 		optimisticallyFilledQuantums, _, err := k.PlacePerpetualLiquidation(ctx, *liquidationOrder)
-		if err != nil {
+		// Exception for liquidation which conflicts with clob pair status. This is expected for liquidations generated
+		// for subaccounts with open positions in final settlement markets.
```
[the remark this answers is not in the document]
jakob-dydx: this won't be removed. Basically this exception (catching the `ErrLiquidationConflictsWithClobPairStatus` error) allows the regular liquidations flow to block liquidation orders for final settlement markets, while still allowing the function to return the subaccount as deleveragable.

Even if we separate steps 6 and 7 completely and make `LiquidateSubaccountsAgainstOrderbook` not return any data, we will want to pass over this exception.

protocol/x/clob/keeper/deleveraging.go
jakob-dydx: yeah sorry this was a vestige from some old version of the code

protocol/x/clob/keeper/clob_pair.go
jakob-dydx: will change this to only allow if status is active

protocol/x/clob/keeper/clob_pair.go
jakob-dydx: adding test cases to PlacePerpetualLiquidation for this

protocol/x/clob/types/errors.go
jakob-dydx: yep will move, my b

PR #858, opened 2023-12-08
[CLOB-1001] Add flag to format tag values for datadog error tracking

jonfung-dydx:
For datadog error tracking, we want each error-based log field to have:
```
error: { stack: <stacktrace>, kind: <error_kind>, message: <error_message>} 
```

In order to do this, we introduce a custom error formatter to the zerolog logger that is used by `cometbft`, `cosmos-sdk`, and `v4-chain`. This error formatter is a global variable, so we will set it here in `PersistentPreRunE`.

This formatter will run on any log event `Key:Value` tag value of type errors and massage it to be of the above form.
Thus, any error logs that have `"error": error` will be automatically datadog error tracked.

Note that `cosmos-sdk` emits error logs with the key `"err"` instead of `"error"`. Zerolog logging library is not powerful enough to support custom error event logging for stack traces in their custom `ErrorStackMarshallingFunc` [here](https://github.com/rs/zerolog/blob/master/fields.go#L78-L90). We could PR into the library, but an easier solution we have done is to set up a datadog log pipeline that remaps all `err` keys into `error` keys. That way, we can datadog error track `cosmos-sdk` errors.
 
 
 TL;DR all `error` and `err` keys are now properly datadog error tracked.


protocol/cmd/dydxprotocold/cmd/root.go
```diff
@@ -119,7 +127,28 @@ func NewRootCmdWithInterceptors(
...
 			}
 
-			serverCtxInterceptor(server.GetServerContextFromCmd(cmd))
+			serverCtx := server.GetServerContextFromCmd(cmd)
+			if ddErrorTrackingFormatterEnabled :=
+				serverCtx.Viper.Get(protocolflags.DdErrorTrackingFormat); ddErrorTrackingFormatterEnabled != nil {
+				if enabled, err := cast.ToBoolE(ddErrorTrackingFormatterEnabled); err == nil && enabled {
+					// Error fields are default set under `error`
+					// Extract + add the kind and message field
+					zerolog.ErrorMarshalFunc = func(err error) interface{} {
```
lcwik, line 136: can we use a sync.Once global to ensure that this is only ever done once otherwise any tests that launch multiple instances of this app in parallel will hit a data race.

jonfung-dydx: done

protocol/app/datadog.go
```diff
@@ -111,3 +115,42 @@ func initDatadogProfiler(logger log.Logger, ddAgentHost string, ddAgentPort uint
...
+}
+
+var (
+	zerologFormatterOnce sync.Once
+)
+
+// SetZerologDatadogErrorTrackingFormat sets custom error formatting for log tag
+// values that are errors for the zerolog library. Converts them to a format that
+// is compatible with datadog error tracking.
+func SetZerologDatadogErrorTrackingFormat() {
```
lcwik, line 138: Please add a test that captures the logs from zerolog for an error to ensure that the stack traces doesn't lose information (or add useless details) since the stack depth can change and it is dependent on the libraries implementation.

lcwik (changes requested):

protocol/app/datadog.go
jonfung-dydx: As discussed, linear ticket [here](https://linear.app/dydx/issue/CLOB-1049/add-test-to-datadog-error-tracking-formatter-stack-trace-truncater) and added TODO.

lcwik (approved):

PR #863, opened 2023-12-08
[CORE-813] - Add pricefeed config tomls back into the repository

clemire:
### Changelist
Since the project work for adding a daemon config file will not be ready by the next release, add the pricefeed configuration toml back into the protocol for the next release so that the pricefeed daemon does not lose configurability.

This PR is a manually constructed reversion of [this PR](https://github.com/dydxprotocol/v4-chain/pull/627/files#diff-44c5af6acd7eab6d96bc260e8ffbbe134fa36e1f7e765d37a780357a1c2958a8).

### Test Plan
- restored missing test cases
- make localnet-start
- modified localnet to start with a pricefeed config for only 1 exchange and confirmed that only one exchange was used via log statements

### Author/Reviewer Checklist

### DIFF
--- a/protocol/daemons/configs/default_pricefeed_exchange_config_test.go
+++ b/protocol/daemons/configs/default_pricefeed_exchange_config_test.go
@@ -102,6 +102,11 @@ const (
 	TimeoutMs = 3000
 	MaxQueries = 1
 	[[exchanges]]
+	ExchangeId = "TestFixedPriceExchange"
+	IntervalMs = 2000
+	TimeoutMs = 3000
+	MaxQueries = 3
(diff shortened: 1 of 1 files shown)


protocol/daemons/configs/default_pricefeed_exchange_config.go
clemire, line 1: The following two files were added back in wholesale with the following caveat: the old `ExchangeStartupConfig` type was renamed to `ExchangeQueryConfig`, so I replaced `Startup` with `Query` everywhere in this file and the corresponding test file.

tqin7 (approved):

PR #867, opened 2023-12-08
[CORE-538] Upgrade to Cosmos 0.50.1 and CometBFT 0.38

lcwik:
### Changelist
Update to Cosmos 0.50.1 and CometBFT 0.38

### Test Plan
Updated existing tests, some tests have been skipped or partially disabled with TODO tags added.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/msgs/internal_msgs.go
+++ b/protocol/app/msgs/internal_msgs.go
@@ -56,8 +56,10 @@ var (
 		"/cosmos.crisis.v1beta1.MsgUpdateParamsResponse": nil,
 
 		// distribution
-		"/cosmos.distribution.v1beta1.MsgUpdateParams":         &distribution.MsgUpdateParams{},
-		"/cosmos.distribution.v1beta1.MsgUpdateParamsResponse": nil,
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend":         &distribution.MsgCommunityPoolSpend{},
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse": nil,
+		"/cosmos.distribution.v1beta1.MsgUpdateParams":               &distribution.MsgUpdateParams{},
+		"/cosmos.distribution.v1beta1.MsgUpdateParamsResponse":       nil,
 
 		// gov
 		"/cosmos.gov.v1.MsgExecLegacyContent":         &gov.MsgExecLegacyContent{},
--- a/protocol/app/msgs/normal_msgs.go
+++ b/protocol/app/msgs/normal_msgs.go
@@ -51,8 +51,6 @@ var (
 
 		// distribution
 		"/cosmos.distribution.v1beta1.CommunityPoolSpendProposal":             nil,
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend":                  &distr.MsgCommunityPoolSpend{},
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse":          nil,
 		"/cosmos.distribution.v1beta1.MsgDepositValidatorRewardsPool":         &distr.MsgDepositValidatorRewardsPool{},
 		"/cosmos.distribution.v1beta1.MsgDepositValidatorRewardsPoolResponse": nil,
 		"/cosmos.distribution.v1beta1.MsgFundCommunityPool":                   &distr.MsgFundCommunityPool{},
@@ -185,11 +183,13 @@ var (
 		"/ibc.core.client.v1.MsgRecoverClientResponse":      nil,
 		"/ibc.core.client.v1.MsgSubmitMisbehaviour":         &ibcclient.MsgSubmitMisbehaviour{}, //nolint:staticcheck
 		"/ibc.core.client.v1.MsgSubmitMisbehaviourResponse": nil,
-		"/ibc.core.client.v1.MsgUpdateClient":               &ibcclient.MsgUpdateClient{},
-		"/ibc.core.client.v1.MsgUpdateClientResponse":       nil,
-		"/ibc.core.client.v1.MsgUpgradeClient":              &ibcclient.MsgUpgradeClient{},
-		"/ibc.core.client.v1.MsgUpgradeClientResponse":      nil,
-		"/ibc.core.client.v1.UpgradeProposal":               nil,
+		// TODO(CORE-538): Move MsgUpdateClient and MsgUpgradeClient to unsupported_msgs once upgrade has been added
+		// and verified to function.
+		"/ibc.core.client.v1.MsgUpdateClient":          &ibcclient.MsgUpdateClient{},
+		"/ibc.core.client.v1.MsgUpdateClientResponse":  nil,
+		"/ibc.core.client.v1.MsgUpgradeClient":         &ibcclient.MsgUpgradeClient{},
+		"/ibc.core.client.v1.MsgUpgradeClientResponse": nil,
+		"/ibc.core.client.v1.UpgradeProposal":          nil,
 
 		// ibc.core.commitment
 		"/ibc.core.commitment.v1.MerklePath":   nil,
--- a/protocol/app/msgs/unsupported_msgs.go
+++ b/protocol/app/msgs/unsupported_msgs.go
@@ -14,13 +14,14 @@ var (
 		"/cosmos.gov.v1beta1.MsgSubmitProposalResponse": nil,
 
 		// ICA Controller messages - these are not used since ICA Controller is disabled.
-		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParams": &icacontrollertypes.
-			MsgUpdateParams{},
-		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParamsResponse": nil,
-		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTx":               &icacontrollertypes.MsgSendTx{},
-		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTxResponse":       nil,
 		"/ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccount": &icacontrollertypes.
 			MsgRegisterInterchainAccount{},
 		"/ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccountResponse": nil,
+		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTx": &icacontrollertypes.
+			MsgSendTx{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTxResponse": nil,
+		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParams": &icacontrollertypes.
+			MsgUpdateParams{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParamsResponse": nil,
 	}
 )
--- a/protocol/app/msgs/internal_msgs_test.go
+++ b/protocol/app/msgs/internal_msgs_test.go
@@ -38,6 +38,8 @@ func TestInternalMsgSamples_Gov_Key(t *testing.T) {
 		"/cosmos.crisis.v1beta1.MsgUpdateParamsResponse",
 
 		// distribution
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend",
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse",
 		"/cosmos.distribution.v1beta1.MsgUpdateParams",
 		"/cosmos.distribution.v1beta1.MsgUpdateParamsResponse",
 
--- a/protocol/app/msgs/normal_msgs_test.go
+++ b/protocol/app/msgs/normal_msgs_test.go
@@ -40,8 +40,6 @@ func TestNormalMsgs_Key(t *testing.T) {
 
 		// distribution
 		"/cosmos.distribution.v1beta1.CommunityPoolSpendProposal",
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend",
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse",
 		"/cosmos.distribution.v1beta1.MsgDepositValidatorRewardsPool",
 		"/cosmos.distribution.v1beta1.MsgDepositValidatorRewardsPoolResponse",
 		"/cosmos.distribution.v1beta1.MsgFundCommunityPool",
--- a/protocol/lib/ante/internal_msg.go
+++ b/protocol/lib/ante/internal_msg.go
@@ -47,6 +47,7 @@ func IsInternalMsg(msg sdk.Msg) bool {
 		*crisis.MsgUpdateParams,
 
 		// distribution
+		*distribution.MsgCommunityPoolSpend,
 		*distribution.MsgUpdateParams,
 
 		// gov


protocol/testutil/constants/messages.go
```diff
@@ -68,6 +68,16 @@ var (
 			Amount:    500_000_000, // $500
 		},
 	}
+	Msg_Transfer_Invalid_SameSenderAndRecipient = &sendingtypes.MsgCreateTransfer{
```
lcwik, line 71: ValidateBasic is now expected to not be implemented and users should be simulating their transactions before submitting them which is why I swapped to a message where we control the message type explicitly.

protocol/x/assets/keeper/asset.go
```diff
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	storetypes "cosmossdk.io/store/types"
```
jayy04, line 4: nit: move this down

protocol/x/clob/abci.go
```diff
@@ -119,6 +119,10 @@ func PrepareCheckState(
 	keeper *keeper.Keeper,
 	daemonLiquidationInfo *liquidationtypes.DaemonLiquidationInfo,
 ) {
+	// TODO(CORE-538): Is there a cleaner way to check to see if this is genesis?
+	if ctx.BlockHeight() == 0 || ctx.BlockHeight() == 1 {
+		return
+	}
```
jayy04, line 125: for my understanding - why do we need to do this?

lcwik: The issue that I ran into is for the first block is that there is no `ProcessProposerMatches` events during `PrepareCheckState`. Cosmos 0.47 didn't run `PrepareCheckState` for the genesis block while Cosmos 0.50 does.

So I keep the original behavior by skipping PrepareCheckState for the genesis block.

jayy04 (approved):

protocol/app/ante/sigverify.go
```diff
@@ -2,30 +2,33 @@ package ante
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	sdkante "github.com/cosmos/cosmos-sdk/x/auth/ante"
 	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
 	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+	gometrics "github.com/hashicorp/go-metrics"
 )
 
-// Verify all signatures for a tx and return an error if any are invalid. Note,
+// SigVerificationDecorator verifies all signatures for a tx and return an error if any are invalid. Note,
```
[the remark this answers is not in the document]
lcwik: https://github.com/dydxprotocol/cosmos-sdk/blob/b48250d4e0efacbd40f3e0b980c1e1004d29e351/x/auth/ante/sigverify.go#L215

protocol/app/process/other_msgs.go
```diff
@@ -59,8 +59,10 @@ func DecodeOtherMsgsTx(decoder sdk.TxDecoder, txBytes []byte) (*OtherMsgsTx, err
 // Validate returns an error if one of the underlying msgs fails `ValidateBasic`.
 func (omt *OtherMsgsTx) Validate() error {
 	for _, msg := range omt.msgs {
-		if err := msg.ValidateBasic(); err != nil {
-			return getValidateBasicError(msg, err)
+		if m, ok := msg.(sdk.HasValidateBasic); ok {
```
ttl33, line 62: When do we NOT expect `HasValidateBasic` to be the case? 

Should we handle else case?

protocol/app/testdata/default_genesis_state.json
```diff
@@ -417,4 +427,4 @@
       }
     ]
   }
-}
+}
```
ttl33, line 430: nit: add new line

protocol/x/sending/types/message_withdraw_from_subaccount.go
```diff
@@ -26,16 +26,6 @@ func NewMsgWithdrawFromSubaccount(
 	}
 }
 
-// GetSigners specifies that the sender of the message must sign.
```
teddyding, line 29: For my understanding, going forward how does this message indicate that `msg.Sender.Owner` is the expected signer?

proto/dydxprotocol/sending/transfer.proto
```diff
@@ -48,6 +48,8 @@ message MsgDepositToSubaccount {
 // MsgWithdrawFromSubaccount represents a single transfer from an
 // `x/subaccounts` subaccount to an `x/bank` account.
 message MsgWithdrawFromSubaccount {
+  option (cosmos.msg.v1.signer) = "sender";
```
teddyding, line 51: I understand that we are specifying the sender should be `sender.owner` in `CustomGetSigner` [here](https://github.com/dydxprotocol/v4-chain/pull/867/files#r1431040322). 

Should we remove this option, given that `sender` is not a valid signer (i.e. it's not an address string)?

protocol/app/upgrades/v3.0.0/upgrade.go
```diff
@@ -98,9 +99,10 @@ func CreateUpgradeHandler(
 	configurator module.Configurator,
 	ak authkeeper.AccountKeeper,
 ) upgradetypes.UpgradeHandler {
```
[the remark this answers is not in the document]
teddyding: Can also add a ticket in appropriate P0 project. I'm following up on the default values [here](https://dydx-team.slack.com/archives/C03FP255RRU/p1702996806918859?thread_ts=1702681066.948759&cid=C03FP255RRU)

proto/dydxprotocol/sending/transfer.proto
lcwik: The cosmos SDK was meant to support these recursively but didn't do it quite right since `sender` is supposed to go into that message and then use `owner` to find the appropriate address string. I filed https://github.com/cosmos/cosmos-sdk/issues/18722 and they have fixed this but it won't be available till a future Cosmos SDK release.

I'd like to leave this for now and once able we can remove the `CustomGetSigners` implementation.

protocol/app/process/other_msgs.go
lcwik: Many of the Cosmos SDK messages have migrated away from using `ValidateBasic` which I filed https://github.com/cosmos/cosmos-sdk/issues/18750 about. Cosmos SDK wants users to simulate their transactions to validate them instead of relying on `ValidateBasic` during `CheckTx` which is a change in expected user behavior going forward.

This is one of the points I brought up that we want to discuss in slack since this has ramifications to user experience.

protocol/app/ante/sigverify_test.go
```diff
@@ -1,26 +1,54 @@
...
 	suite := testante.SetupTestSuite(t, true)
+	suite.TxBankKeeper.EXPECT().DenomMetadata(
+		gomock.Any(),
+		gomock.Any(),
+	).Return(&banktypes.QueryDenomMetadataResponse{}, nil).AnyTimes()
+
+	// signing.SignMode_SIGN_MODE_TEXTUAL and signing.SignMode_SIGN_MODE_LEGACY_AMINO_JSON are disabled
+	// from the test suite since they include the sequence number as part of the signature preventing
+	// us from skipping sequence validation for certain message types.
+	enabledSignModes := []signing.SignMode{signing.SignMode_SIGN_MODE_DIRECT}
```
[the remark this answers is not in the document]
lcwik: We don't need them and they also don't support skipping sequence numbers since the sequence number is part of the bytes used for signing/signature verification preventing us from practically validating them.

protocol/app/ante/sigverify_test.go
```diff
@@ -1,26 +1,54 @@
...
+
+	// Since TEXTUAL is not enabled by default, we create a custom TxConfig
+	// here which includes it.
+	txConfigOpts := authtx.ConfigOptions{
+		TextualCoinMetadataQueryFn: txmodule.NewGRPCCoinMetadataQueryFn(suite.ClientCtx),
+		EnabledSignModes:           enabledSignModes,
+	}
```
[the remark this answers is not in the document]
lcwik: I tried to minimize the number of changes to simplify how much work is needed when porting over the code from Cosmos 0.50.

I wish there was a way to not have to have copies of this code/tests since it is cumbersome to maintain.

protocol/app/ante/sigverify_test.go
```diff
@@ -33,30 +61,42 @@ func TestSigVerification(t *testing.T) {
...
 	}
 
 	feeAmount := testdata.NewTestFeeAmount()
 	gasLimit := testdata.NewTestGasLimit()
 
 	spkd := sdkante.NewSetPubKeyDecorator(suite.AccountKeeper)
-	svd := customante.NewSigVerificationDecorator(
-		suite.AccountKeeper,
-		suite.ClientCtx.TxConfig.SignModeHandler(),
+	txConfigOpts = authtx.ConfigOptions{
```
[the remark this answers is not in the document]
lcwik: I tried to minimize the number of changes to simplify how much work is needed when porting over the code from Cosmos SDK so that diffing the original and the forked version is simpler next time when porting over the code.

I wish there was a way to not have to have copies of this code/tests since it is cumbersome to maintain.

protocol/app/ante/sigverify_test.go
```diff
@@ -171,45 +211,51 @@ func TestSigVerification(t *testing.T) {
...
+					badSig, _ := tc.privs[0].Sign([]byte("unrelated message"))
+					txSigs[0] = signing.SignatureV2{
+						PubKey: tc.privs[0].PubKey(),
+						Data: &signing.SingleSignatureData{
+							SignMode:  defaultSignMode,
+							Signature: badSig,
+						},
+						Sequence: tc.accSeqs[0],
+					}
+					require.NoError(t, suite.TxBuilder.SetSignatures(txSigs...))
```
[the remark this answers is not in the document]
lcwik: Blame the linter requirements.

protocol/app/app.go
```diff
@@ -423,30 +429,34 @@ func New(
 	// add keepers
 	app.AccountKeeper = authkeeper.NewAccountKeeper(
 		appCodec,
-		keys[authtypes.StoreKey],
+		runtime.NewKVStoreService(keys[authtypes.StoreKey]),
 		authtypes.ProtoBaseAccount,
 		maccPerms,
-		sdk.Bech32MainPrefix,
+		addresscodec.NewBech32Codec(sdk.GetConfig().GetBech32AccountAddrPrefix()),
```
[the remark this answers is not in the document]
lcwik: I don't think the redirection helps with code readability since the lookup is quite clear as to what is being fetched.

protocol/app/app.go
```diff
@@ -406,8 +408,12 @@ func New(
...
 	// set the BaseApp's parameter store
 	app.ConsensusParamsKeeper = consensusparamkeeper.NewKeeper(
-		appCodec, keys[upgradetypes.StoreKey], lib.GovModuleAddress.String())
-	bApp.SetParamStore(&app.ConsensusParamsKeeper)
+		appCodec,
+		runtime.NewKVStoreService(keys[upgradetypes.StoreKey]),
+		lib.GovModuleAddress.String(),
+		app.event,
+	)
+	bApp.SetParamStore(&app.ConsensusParamsKeeper.ParamsStore)
```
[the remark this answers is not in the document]
lcwik: That is correct.

This is how Cosmos is choosing to expose the generic collection based interfaces to various data stores.

protocol/app/app.go
```diff
@@ -498,16 +519,24 @@ func New(
 	// See: https://github.com/cosmos/cosmos-sdk/blob/release/v0.46.x/x/gov/spec/01_concepts.md#proposal-messages
 	govRouter := govv1beta1.NewRouter()
 	govRouter.AddRoute(govtypes.RouterKey, govv1beta1.ProposalHandler).
-		AddRoute(paramproposal.RouterKey, params.NewParamChangeProposalHandler(app.ParamsKeeper)).
-		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(app.UpgradeKeeper))
+		AddRoute(paramproposal.RouterKey, params.NewParamChangeProposalHandler(app.ParamsKeeper))
+	/** TODO(CORE-538): Migrate software upgrade type to Cosmos 0.50.
```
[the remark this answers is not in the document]
lcwik: It is because these types don't exist anymore and I haven't had time to integrate the new way to do upgrades yet.

There are a few TODOs in the codebase of this nature.

protocol/app/app.go
```diff
@@ -742,7 +784,7 @@ func New(
 		},
 	)
 	blockTimeModule := blocktimemodule.NewAppModule(appCodec, app.BlockTimeKeeper)
-
+	txConfig.SigningContext()
```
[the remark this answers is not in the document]
lcwik: It isn't. Removed.

protocol/app/basic_manager/basic_manager.go
```diff
@@ -57,10 +57,6 @@ var (
 		gov.NewAppModuleBasic(
 			[]govclient.ProposalHandler{
 				paramsclient.ProposalHandler,
-				upgradeclient.LegacyProposalHandler,
```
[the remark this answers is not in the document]
lcwik: They don't exist anymore.

protocol/app/module/interface_registry.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if InterfaceRegistry, err = NewInterfaceRegistry(
+		sdk.GetConfig().GetBech32AccountAddrPrefix(),
+		sdk.GetConfig().GetBech32ValidatorAddrPrefix(),
+	); err != nil {
+		panic(err)
+	}
+}
+
+func NewInterfaceRegistry(addrPrefix string, valAddrPrefix string) (types.InterfaceRegistry, error) {
```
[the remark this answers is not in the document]
lcwik: They have coverage via e2e test framework since the framework signs all messages that go through the whole ante handler stack.

Also, this will go away with Cosmos ?0.50.3? since nesting has been [fixed](https://github.com/cosmos/cosmos-sdk/issues/18722) and we can migrate to just using the annotations in all situations.

protocol/app/msgs/app_injected_msgs_test.go
```diff
@@ -51,7 +51,9 @@ func TestAppInjectedMsgSamples_GetSigners(t *testing.T) {
 		_ = testTxBuilder.SetMsgs(sample.Msg)
 		sigTx, ok := testTxBuilder.GetTx().(authsigning.SigVerifiableTx)
 		require.True(t, ok)
-		require.Empty(t, sigTx.GetSigners())
+		signers, err := sigTx.GetSigners()
+		require.Error(t, err)
```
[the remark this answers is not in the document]
lcwik: Fixed.

That was supposed to be NoError and it was causing an error since the MakeTestEncodingConfig wasn't using the correct InterfaceRegistry since it was using the default Cosmos one that doesn't have the custom signers defined.

protocol/testutil/ante/testutil.go
```diff
@@ -109,7 +122,7 @@ func (suite *AnteTestSuite) CreateTestAccounts(numAccs int) []TestAccount {
 	for i := 0; i < numAccs; i++ {
 		priv, _, addr := testdata.KeyTestPubAddr()
 		acc := suite.AccountKeeper.NewAccountWithAddress(suite.Ctx, addr)
-		if err := acc.SetAccountNumber(uint64(i)); err != nil {
+		if err := acc.SetAccountNumber(uint64(i + 1000)); err != nil {
```
roy-dydx, line 125: Any reason for +1000?

roy-dydx (commented):
protocol/testutil and protocol/testing look good to me

protocol/testutil/ante/testutil.go
lcwik: It was a holdover from me debugging something. I swapped it back.

jakob-dydx (approved):

protocol/app/msgs/internal_msgs.go
```diff
@@ -52,10 +52,12 @@ var (
 		"/cosmos.crisis.v1beta1.MsgUpdateParamsResponse": nil,
 
 		// distribution
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend":         &distribution.MsgCommunityPoolSpend{},
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse": nil,
-		"/cosmos.distribution.v1beta1.MsgUpdateParams":               &distribution.MsgUpdateParams{},
-		"/cosmos.distribution.v1beta1.MsgUpdateParamsResponse":       nil,
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend":                  &distribution.MsgCommunityPoolSpend{},
+		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpendResponse":          nil,
+		"/cosmos.distribution.v1beta1.MsgDepositValidatorRewardsPool":         &distribution.MsgDepositValidatorRewardsPool{},
```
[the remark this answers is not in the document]
lcwik: I would concur and reached out to Alexander as well.

protocol/app/msgs/normal_msgs.go
```diff
@@ -139,9 +141,11 @@ var (
 		"/dydxprotocol.sending.MsgWithdrawFromSubaccountResponse": nil,
 
 		// ibc.applications
-		"/ibc.applications.transfer.v1.MsgTransfer":           &ibctransfer.MsgTransfer{},
-		"/ibc.applications.transfer.v1.MsgTransferResponse":   nil,
-		"/ibc.applications.transfer.v1.TransferAuthorization": nil,
+		"/ibc.applications.transfer.v1.MsgTransfer":             &ibctransfer.MsgTransfer{},
+		"/ibc.applications.transfer.v1.MsgTransferResponse":     nil,
+		"/ibc.applications.transfer.v1.MsgUpdateParams":         &ibctransfer.MsgUpdateParams{},
```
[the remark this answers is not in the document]
lcwik: The Cosmos 0.50 upgrade documentation says the same thing.

jayy04 (approved):

protocol/app/msgs/normal_msgs.go
```diff
@@ -173,10 +177,16 @@ var (
...
-		"/ibc.core.client.v1.MsgSubmitMisbehaviour":         &ibcclient.MsgSubmitMisbehaviour{},
+		"/ibc.core.client.v1.MsgIBCSoftwareUpgrade":         &ibcclient.MsgIBCSoftwareUpgrade{},
+		"/ibc.core.client.v1.MsgIBCSoftwareUpgradeResponse": nil,
+		"/ibc.core.client.v1.MsgRecoverClient":              &ibcclient.MsgRecoverClient{},
+		"/ibc.core.client.v1.MsgRecoverClientResponse":      nil,
+		"/ibc.core.client.v1.MsgSubmitMisbehaviour":         &ibcclient.MsgSubmitMisbehaviour{}, //nolint:staticcheck
 		"/ibc.core.client.v1.MsgSubmitMisbehaviourResponse": nil,
 		"/ibc.core.client.v1.MsgUpdateClient":               &ibcclient.MsgUpdateClient{},
 		"/ibc.core.client.v1.MsgUpdateClientResponse":       nil,
+		"/ibc.core.client.v1.MsgUpdateParams":               &ibcclient.MsgUpdateParams{},
```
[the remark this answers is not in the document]
lcwik: Moved to internal

protocol/app/msgs/normal_msgs.go
```diff
@@ -198,7 +208,8 @@ var (
 		"/ibc.core.connection.v1.MsgConnectionOpenInitResponse":    nil,
 		"/ibc.core.connection.v1.MsgConnectionOpenTry":             &ibcconn.MsgConnectionOpenTry{},
 		"/ibc.core.connection.v1.MsgConnectionOpenTryResponse":     nil,
-		"/ibc.core.connection.v1.Version":                          nil,
+		"/ibc.core.connection.v1.MsgUpdateParams":                  &ibcconn.MsgUpdateParams{},
```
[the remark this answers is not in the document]
lcwik: Moved to internal

protocol/app/msgs/internal_msgs.go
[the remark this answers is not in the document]
lcwik: Alexander replied and said that these should be submittable by anyone so moved to `normal_msg`.

protocol/app/msgs/normal_msgs.go
```diff
@@ -173,10 +177,16 @@ var (
 		"/ibc.core.client.v1.Height":                        nil,
 		"/ibc.core.client.v1.MsgCreateClient":               &ibcclient.MsgCreateClient{},
 		"/ibc.core.client.v1.MsgCreateClientResponse":       nil,
-		"/ibc.core.client.v1.MsgSubmitMisbehaviour":         &ibcclient.MsgSubmitMisbehaviour{},
+		"/ibc.core.client.v1.MsgIBCSoftwareUpgrade":         &ibcclient.MsgIBCSoftwareUpgrade{},
+		"/ibc.core.client.v1.MsgIBCSoftwareUpgradeResponse": nil,
+		"/ibc.core.client.v1.MsgRecoverClient":              &ibcclient.MsgRecoverClient{},
+		"/ibc.core.client.v1.MsgRecoverClientResponse":      nil,
```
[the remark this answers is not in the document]
lcwik: I never heard back so using the documentation as described above and leaving these in `normal_msgs`.

protocol/app/app.go
```diff
@@ -245,6 +248,7 @@ type App struct {
 	ParamsKeeper     paramskeeper.Keeper
 	// IBC Keeper must be a pointer in the app, so we can SetRouter on it correctly
 	IBCKeeper             *ibckeeper.Keeper
+	ICAControllerKeeper   icacontrollerkeeper.Keeper
```
teddyding, line 251: Why are we adding `ICAControllerKeeper`, for my understanding? For supporting interchain accounts we specifically wanted to avoid the controllerkeeper ([context](https://www.notion.so/dydx/Supporting-Interchain-Accounts-Liquid-Staking-33e7b09f0bcf44a5b3d1d6931c58d99c?pvs=4#a83746557b00456592390583a6641cdc)).

lcwik: The ICAControllerKeeper is still disabled since `Enabled` is [false in the upgrade](https://github.com/dydxprotocol/v4-chain/blob/3e2c518826ea9d76bb80bdb50f9d0ee955d354bd/protocol/app/upgrades/v3.0.0/upgrade.go#L141) (note the `Enabled` field is taking on the default value).

The [simulation](https://github.com/cosmos/ibc-go/blob/2551dea41cd3c512845007ca895c8402afa9b79f/modules/apps/27-interchain-accounts/simulation/proposals.go#L23) in IBC v8 expects that the controller module to exist.

teddyding: Instead of going with this approach, is it possible to simply exclude ICA module from simulation testing? We've found that simulation testing is not very high signal for the dYdX chain, let alone a third-party module that is tested  elsewhere. 

We specifically prefer to explicitly disable (in code, instead of through gov-updated params) the controller sub-module altogether to reduce surface of risk. Simulation testing only does not justify altering the original approach, IMO. Lmk if the above is possible and happy to huddle!

lcwik: I swapped to exclude the `interchainaccounts` module from simulation as suggested.

teddyding (approved):

protocol/app/app.go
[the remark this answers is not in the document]
ttl33: Would merging this PR make the `main` branch to be in a broken state? For example, if the route for the upgrade is not registered, then upgrade via gov wouldn't work.

If there are no easy ways to resolve this in this PR, then I am okay. Just wanted to call out that we wanted to keep `main` always deployable and releasable.

protocol/app/app.go
```diff
@@ -945,7 +990,7 @@ func New(
 	// must be passed by reference here.
 	app.ModuleManager = module.NewManager(
 		genutil.NewAppModule(
-			app.AccountKeeper, app.StakingKeeper, app.BaseApp.DeliverTxShouldLock,
```
ttl33, line 948: To confirm, why are we removing this?

protocol/app/app.go
```diff
@@ -1000,13 +1046,16 @@ func New(
...
+		upgradetypes.ModuleName,
+	)
+
 	// During begin block slashing happens after distr.BeginBlocker so that
 	// there is nothing left over in the validator fee pool, so as to keep the
 	// CanWithdrawInvariant invariant.
 	// NOTE: staking module is required if HistoricalEntries param > 0
 	app.ModuleManager.SetOrderBeginBlockers(
 		blocktimemoduletypes.ModuleName, // Must be first
-		upgradetypes.ModuleName,
```
ttl33, line 1009: To confirm: why are we removing this? Is this because this is now part of the preblocker?

protocol/app/msgs/internal_msgs.go
```diff
@@ -52,10 +56,8 @@ var (
 		"/cosmos.crisis.v1beta1.MsgUpdateParamsResponse": nil,
 
 		// distribution
-		"/cosmos.distribution.v1beta1.MsgCommunityPoolSpend":         &distribution.MsgCommunityPoolSpend{},
```
ttl33, line 55: Blocking Comment: why are we moving this from internal to normal? Based on [this comment](https://github.com/cosmos/cosmos-sdk/blob/b729e981f1302d2490244b32aba7874a490d80ed/proto/cosmos/distribution/v1beta1/tx.proto#L45-L51), unless we adopt the new `x/protocolpool` module, this msg should still be executed only by `gov`. Therefore, this should stay as internal.

protocol/app/msgs/unsupported_msgs.go
```diff
@@ -15,8 +14,13 @@ var (
...
 		// ICA Controller messages - these are not used since ICA Controller is disabled.
-		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTx": &icacontrollertypes.MsgSendTx{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParams": &icacontrollertypes.
+			MsgUpdateParams{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParamsResponse": nil,
+		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTx":               &icacontrollertypes.MsgSendTx{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgSendTxResponse":       nil,
 		"/ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccount": &icacontrollertypes.
 			MsgRegisterInterchainAccount{},
+		"/ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccountResponse": nil,
```
ttl33, line 24: nit: can we update these to be alphabetic?

protocol/app/msgs/normal_msgs.go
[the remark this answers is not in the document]
ttl33: For `MsgRecoverClient`: yes, I think we should keep this in `normal`. This is because anyone should be able to submit the gov proposal to recover a client. See [doc/comments here](https://github.com/cosmos/ibc-go/pull/4620/files#diff-e0868b09c0b7df02b3339ed6f3bedef5abd9168009812f784d6b80b94982f095R78-R93)

For `MsgIBCSoftwareUpgrade`,  yes, I also think we should keep this in `normal`. This is because it's replacing the `MsgUpgradeClient` which was the proposal submitted by users to upgrade.

Should we then move `MsgUpdateClient` and `MsgUpgradeClient` (note update vs upgrade here) to unsupported ones?

protocol/app/process/other_msgs.go
ttl33: Blocking comment: what's the conclusion of this discussion?

protocol/app/module/interface_registry.go
```diff
@@ -0,0 +1,100 @@
...
+	return types.NewInterfaceRegistryWithOptions(types.InterfaceRegistryOptions{
+		ProtoFiles: gogoproto.HybridResolver,
+		SigningOptions: signing.Options{
+			AddressCodec:          addresscodec.NewBech32Codec(addrPrefix),
+			ValidatorAddressCodec: addresscodec.NewBech32Codec(valAddrPrefix),
+			// TODO(CORE-840) cosmos.msg.v1.signer annotation doesn't supported nested messages beyond a depth of 1
```
ttl33, line 74:
```suggestion
			// TODO(CORE-840): cosmos.msg.v1.signer annotation doesn't supported nested messages beyond a depth of 1
```

protocol/app/app.go
lcwik: This is part of the Cosmos upgrade guide to move it to preblocker.

protocol/app/app.go
lcwik: Since the lock we added to Cosmos SDK for `CheckTx` parallelism also needed to be added to `DeliverTx` but there were use cases where `DeliverTx` is re-entrant and hence already held the lock.

Now that Cosmos SDK uses `FinalizeBlock`, the lock that was elevated to `FinalizeBlock` and now the it doesn't matter that `DeliverTx` is re-entrant since it no longer acquires the lock.

protocol/app/process/other_msgs.go
lcwik: The conclusion was to simulate the transaction.

protocol/app/msgs/unsupported_msgs.go
lcwik: Done

protocol/app/msgs/internal_msgs.go
lcwik: I have been moving things back and forth too many times and moved this one by mistake.

lcwik:
> Would merging this PR make the main branch to be in a broken state? For example, if the route for the upgrade is not > registered, then upgrade via gov wouldn't work.
> 
> If there are no easy ways to resolve this in this PR, then I am okay. Just wanted to call out that we wanted to keep main always deployable and releasable.

We should be upgrading from the release branch with patches and backporting any fixes to it that need to be released immediately. There is no easy way to resolve this without implementing even more and this PR has been delayed due to RC0 and peoples vacation schedules so I want to get it in finally and not waste more time rebasing constantly.

> Blocking comment: what's the conclusion of this discussion?

To rely on simulating transactions since CheckTx only does some of the work to validate things.

> For MsgRecoverClient: yes, I think we should keep this in normal. This is because anyone should be able to submit the gov proposal to recover a client. See [doc/comments here](https://github.com/cosmos/ibc-go/pull/4620/files#diff-e0868b09c0b7df02b3339ed6f3bedef5abd9168009812f784d6b80b94982f095R78-R93)
> 
> For MsgIBCSoftwareUpgrade, yes, I also think we should keep this in normal. This is because it's replacing the MsgUpgradeClient which was the proposal submitted by users to upgrade.
> 
> Should we then move MsgUpdateClient and MsgUpgradeClient (note update vs upgrade here) to unsupported ones?

I added a TODO to handle this once I figure out how to get the upgrades done.

protocol/app/app.go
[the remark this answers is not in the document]
ttl33: Luke replied here: https://github.com/dydxprotocol/v4-chain/pull/867#issuecomment-1877416949

protocol/app/msgs/normal_msgs.go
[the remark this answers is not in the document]
ttl33: Luke replied here: https://github.com/dydxprotocol/v4-chain/pull/867#issuecomment-1877416949

protocol/app/process/other_msgs.go
ttl33: Gotcha, nit: would you be able to add a comment explaining this?

protocol/app/testdata/default_genesis_state.json
```diff
@@ -262,10 +261,22 @@
       "threshold": "0.500000000000000000",
       "veto_threshold": "0.334000000000000000",
       "min_initial_deposit_ratio": "0.000000000000000000",
+      "proposal_cancel_ratio": "0.500000000000000000",
```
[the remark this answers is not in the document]
ttl33: To match the suggested staging/prod value, can you use the following?

```suggestion
      "proposal_cancel_ratio": "1.0"
      "proposal_cancel_dest": "",
      "expedited_voting_period": "86400s",
      "expedited_threshold": "0.75",
      "expedited_min_deposit": [
        {
          "denom": "stake",
          "amount": "50000000"
        }
      ],
      "burn_vote_quorum": false,
      "burn_proposal_deposit_prevote": false,
      "burn_vote_veto": true,
      "min_deposit_ratio": "0.20"
    },
    "constitution": ""
  },
```

ttl33 (approved):

PR #872, opened 2023-12-11
[CLOB-1020] [CLOB-1048] Logging Library with Contextual Tags

jonfung-dydx:
1. Introduce a new logging entrypoint functions in `protocol/lib/log/utils.go`. 
- These functions take in the context `ctx` object and grab the logger from the context, which carries a set of k:v tag values for logs.
- New method `AddPersistentTagsToLogger` returns a new context `ctx` object with specified k:v tags attached onto the logger.
- All subsequent logging calls using the specified ctx object will contain previous tags added with `AddPersistentTagsToLogger`
2. `protocol/lib/log/constants.go` constants for metric tag keys
3. Migrate callsites in [protocol/x/clob/keeper/](https://github.com/dydxprotocol/v4-chain/pull/872/files#diff-ae3a9d0d6029ff89d66f4bf9a308ce4d067f532bf0b5ee25d9bb0ebc9a5f197f) to use new library


.github/workflows/protocol-build-and-push-snapshot.yml
```diff
@@ -3,6 +3,7 @@ name: Protocol Build & Push Image to AWS ECR
 on:  # yamllint disable-line rule:truthy
   push:
     branches:
+      - jonfung/**
```
jonfung-dydx, line 6: will remove these before merge

protocol/x/clob/abci.go
```diff
@@ -119,6 +130,11 @@ func PrepareCheckState(
 	keeper *keeper.Keeper,
 	liquidatableSubaccountIds *liquidationtypes.LiquidatableSubaccountIds,
 ) {
+	ctx = log.AddPersistentTagsToLogger(ctx,
```
jonfung-dydx, line 133: all abci.go functions have tags on them

protocol/x/clob/ante/clob.go
```diff
@@ -62,8 +63,20 @@ func (cd ClobDecorator) AnteHandle(
 	msgs := tx.GetMsgs()
 	var msg = msgs[0]
 
+	// Set request-level logging tags
+	ctx = log.AddPersistentTagsToLogger(ctx,
```
jonfung-dydx, line 67: Note we add persistent tags for checktx here

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -24,6 +24,16 @@ func (k msgServer) CancelOrder(
 ) (resp *types.MsgCancelOrderResponse, err error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	// Attach various logging tags relative to this request. These should be static with no changes.
+	ctx = log.AddPersistentTagsToLogger(ctx,
```
jonfung-dydx, line 28: delivertx persistent tags

protocol/lib/log/constants.go
```diff
@@ -0,0 +1,52 @@
...
+	StackTrace          = "stack_trace"
+	Proposer            = "proposer"
+	PrunableBlockHeight = "prunable_block_height"
+
+	OrderSizeOptimisticallyFilledFromMatchingQuantums = "order_size_optimistically_filled_from_matching_quantums"
+	NewLocalValidatorOperationsQueue                  = "new_local_validator_operations_queue"
+	LocalValidatorOperationsQueue                     = "local_validator_operations_queue"
+)
+
+// Tag values
```
lucas-dydx, line 35: why do these tag values use underscores (or backslashes)?

why are module names like `x/clob` defined in same const def as callbacks?

protocol/lib/log/utils.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// InfoLog reports msg as an info level log with specified key vals.
+func InfoLog(ctx sdk.Context, msg string, keyvals ...interface{}) {
+	ctx.Logger().Info(msg, keyvals...)
```
lucas-dydx, line 11: is it worth validating that `keyvals` is an even length?

What happens if it's an odd-length? Should we note this in the comment?

protocol/lib/log/utils.go
```diff
@@ -0,0 +1,37 @@
...
+
+// ErrorLogWithError reports msg as a error log with specified key vals,
+// as well as attaching the error object to the log for datadog error tracking.
+func ErrorLogWithError(ctx sdk.Context, msg string, err error, keyvals ...interface{}) {
+	ctx.Logger().Error(msg, append(keyvals, Error, err))
+}
+
+// ErrorLog reports msg as a error log. It constructs an error object on the fly with
+// the given message object.
+// Please try to use define a new message and use `ErrorLogWithError` instead.
```
lucas-dydx, line 27: Assuming u meant this

```suggestion
// Please try to define a new message and use `ErrorLogWithError` instead.
```

protocol/x/clob/abci.go
```diff
@@ -20,6 +21,11 @@ func BeginBlocker(
 	ctx sdk.Context,
 	keeper types.ClobKeeper,
 ) {
+	ctx = log.AddPersistentTagsToLogger(ctx,
+		log.Handler, log.BeginBlocker,
+		log.BlockHeight, ctx.BlockHeight(),
+	)
+
```
lucas-dydx, line 28: nit: each arg on it's own line

protocol/x/clob/abci.go
```diff
@@ -117,6 +128,11 @@ func PrepareCheckState(
 	ctx sdk.Context,
 	keeper *keeper.Keeper,
 ) {
+	ctx = log.AddPersistentTagsToLogger(ctx,
+		log.Handler, log.PrepareCheckState,
+		log.BlockHeight, ctx.BlockHeight(),
```
lucas-dydx, line 133: Shouldn't this be block height + 1 since it's `PrepareCheckState` for the next block?

protocol/x/clob/ante/clob.go
```diff
@@ -62,8 +63,20 @@ func (cd ClobDecorator) AnteHandle(
...
+		log.Module, log.Clob,
+		log.Callback, lib.TxMode(ctx),
+		log.BlockHeight, ctx.BlockHeight(),
```
lucas-dydx, line 70: isn't this also block height + 1 since it's not deliver tx?

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -24,6 +24,16 @@ func (k msgServer) CancelOrder(
...
+		log.Module, log.Clob,
+		log.ProposerConsAddress, sdk.ConsAddress(ctx.BlockHeader().ProposerAddress),
+		log.Callback, lib.TxMode(ctx),
+		log.BlockHeight, ctx.BlockHeight(),
+		log.Handler, log.CancelOrder,
+		log.Msg, msg,
```
lucas-dydx, line 34: is `msg` just a pointer in the resulting log? or do the fields get logged properly

lucas-dydx (approved):

protocol/lib/log/constants.go
jonfung-dydx: To retain backwards compatibility with the existing tags. Existing tags are in underscore delimited except for the module names, which are `x/clob` and `x/vesting`.

protocol/lib/log/utils.go
jonfung-dydx: I think we should adopt the same mentality as the underlying library, which doesn't perform validation. In an odd number case, the underlying library + this function discards the last value. I think Logging/metrics libraries should not be able to affect application behavior, so in this case I think we shouldn't panic.

protocol/x/clob/abci.go
jonfung-dydx: i think it's a bit more clear here since it's `key, val` alternating here

protocol/x/clob/keeper/msg_server_cancel_orders.go
jonfung-dydx: Looks fine to me
Screenshot 2024-01-04 at 12 35 44 AM

protocol/lib/log/constants.go
```diff
@@ -0,0 +1,54 @@
...
+const (
+	// Module tag values are prefixed with `x/`
+	Clob = "x/clob"
+
+	CheckTx   = "check_tx"
+	RecheckTx = "recheck_tx"
+	DeliverTx = "deliver_tx"
+)
+
+// Tag values that should be camelcased (i.e function names)
```
jakob-dydx, line 45: nit, but isn't this PascalCase, not camelCase ;)

protocol/lib/log/utils.go
```diff
@@ -0,0 +1,43 @@
...
+// `keyvals` should be even number in length and be of alternating types (string, interface{}).
+func ErrorLog(ctx sdk.Context, msg string, keyvals ...interface{}) {
+	err := errors.New(msg)
+	ErrorLogWithError(ctx, msg, err, keyvals...)
+}
+
+// AddPersistentTagsToLogger returns a new sdk.Context with a logger that has new persistent
+// tags that are added to all logs emitted.
+// `keyvals` should be even number in length and be of alternating types (string, interface{}).
+func AddPersistentTagsToLogger(ctx sdk.Context, keyvals ...interface{}) sdk.Context {
```
jakob-dydx, line 41: Not super opinionated about this, but have you considered alternative names for this function?

Was thinking something like `PutTagsToLogger` could be helpful, since if this function is called twice it is possible to overwrite previous tag values. The idea of a put operation is common and could make the internals of this extra clear. But yeah, your call

jakob-dydx (approved):

PR #878, opened 2023-12-11
[feature] Block withdrawals if a negative equity subaccount was recently seen in state

lucas-dydx:
### Changelist
Block withdrawals if a negative equity subaccount was recently seen within `WITHDRAWAL_AND_TRANSFERS_BLOCKED_AFTER_NEGATIVE_TNC_SUBACCOUNT_SEEN_BLOCKS`.

### Test Plan

### Author/Reviewer Checklist


protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -355,14 +357,15 @@ func (k Keeper) CanUpdateSubaccounts(
 		time.Now(),
 		metrics.CanUpdateSubaccounts,
 		metrics.Latency,
+		// TODO: Add UpdateType as label
```
lucas-dydx, line 360: Ideally can add this as a label to the existing metrics.
- [ ] Add `UpdateType` as a label within this PR / create ticket to do so later.

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -9,28 +9,30 @@ import (
...
 	ctx sdk.Context,
-) uint32 {
+) (uint32, bool) {
 	store := ctx.KVStore(k.storeKey)
 	return k.getNegativeTncSubaccountSeenAtBlock(store)
 }
 
 // getNegativeTncSubaccountSeenAtBlock is a helper function that takes a store and returns the last
-// block height a negative TNC subaccount was seen in state.
+// block height a negative TNC subaccount was seen in state and a boolean for whether it exists in state.
```
lucas-dydx, line 21: As an FYI, updated this function to avoid defaulting to 0 so that withdrawals aren't blocked when block height is less than `WITHDRAWAL_AND_TRANSFERS_BLOCKED_AFTER_NEGATIVE_TNC_SUBACCOUNT_SEEN_BLOCKS` (happens after chain restarts from genesis, unit tests, etc).

protocol/x/subaccounts/keeper/subaccount_test.go
```diff
@@ -2260,6 +2261,430 @@ func TestUpdateSubaccounts(t *testing.T) {
 	}
 }
 
+func TestUpdateSubaccounts_WithdrawalsBlocked(t *testing.T) {
```
lucas-dydx, line 2264: A few notes here:
- I'm planning on only adding tests for `UpdateSubaccounts` here, since the only logic is modified in `internalCanUpdateSubaccounts` (which is shared between `UpdateSubaccounts` and `CanUpdateSubaccounts`).
    - Note: this means that technically this logic isn't tested for things like add-to-orderbook collateralization checks, which _should_ be fine, but LMK if anyone disagrees.
- I still need to add tests for transfers and matches (they'll look very similar to the current tests for withdrawals and deposits respectively).
- I am not adding E2E tests in this PR for withdrawal gating, and will do that when the feature is fully implemented (requires `PrepareCheckState` changes for adding the zero-fill deleveraging matches and `DeliverTx` changes for writing the block number to state when negative TNC subaccount seen).

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -42,9 +44,9 @@ func (k Keeper) SetNegativeTncSubaccountSeenAtBlock(
 ) {
 	store := ctx.KVStore(k.storeKey)
 
-	// Panic if the new block height is less than the current block height.
-	currentValue := k.getNegativeTncSubaccountSeenAtBlock(store)
-	if blockHeight < currentValue {
+	// Panic if the new block height exists and is less than the current block height.
```
jakob-dydx, line 47:
```suggestion
	// Panic if the stored block height value exists and is greater than the new block height value.
```

protocol/x/subaccounts/keeper/subaccount_test.go
lucas-dydx: Added tests for transfers and matches!

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -515,6 +522,39 @@ func (k Keeper) internalCanUpdateSubaccounts(
 	success = true
 	successPerUpdate = make([]types.UpdateResult, len(settledUpdates))
 
+	// Block all withdrawals and transfers if there was a negative TNC subaccount seen within the
+	// last `WITHDRAWAL_AND_TRANSFERS_BLOCKED_AFTER_NEGATIVE_TNC_SUBACCOUNT_SEEN_BLOCKS`.
+	if updateType == types.Withdrawal || updateType == types.Transfer {
```
jayy04, line 527: nit: can we move this to a helper function?

lucas-dydx: LMK if you hard disagree here, but this logic is pretty simple for now so I prefer to leave it in the main function for readability

I agree once we add more reasons for blocking withdrawals / trades / etc then all of that should be abstracted into a helper function

LMK if I'm missing your point though

jayy04 (approved):

PR #887, opened 2023-12-12
Add upgradable testnet to container test

roy-dydx:
### Changelist
Expose a `NewTestnetWithPreupgradeGenesis()` function which starts a network on the last version. It can be upgraded to the current version using gov proposal. This is done by including the old genesis + binary in the docker image and adding an alternative entrypoint to configure the image for preupgrade genesis.

The genesis.json is generated at the commit of the old binary. It is taken from one of the containers.

### Test Plan
Manually tried out existing tests with the preupgrade genesis to make sure it works.

### Author/Reviewer Checklist


protocol/testing/containertest/preupgrade_entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+#   - Set the preupgrade binary (downloaded in containertest.sh) to be the genesis binary
+#   - Set the binary at the current commit to  be the ugprade binary
+#   - Set the genesis to be the preupgrade genesis
+
+if [[ -z "${UPGRADE_TO_VERSION}" ]]; then
+    echo >&2 "UPGRADE_TO_VERSION must be set"
+    exit 1
+fi
+
+MONIKERS=(
```
roy-dydx, line 15: I could pass in moniker as an env variable as well. Choosing not to do so for now.

protocol/testing/containertest/containertest.sh
```diff
@@ -7,6 +7,7 @@ set -eo pipefail
 source "./genesis.sh"
 
 CHAIN_ID="localdydxprotocol"
+PREUPGRADE_VERSION="v2.0.0"
```
teddyding, line 10: Wondering how we will manage this version string? As of now wre we thinking just manually updated this for every new major release?

protocol/testing/containertest/preupgrade_genesis.json
```diff
@@ -0,0 +1,2164 @@
+{
```
teddyding, line 1: Questions:
1. Could you document in this PR how this particular `.json` file was generated?
2. This file needs to for every major version right? So if we are testing `v3 -> v4` upgrade then it will need a genesis file compatible with v3. How will we manage this in the long run?

teddyding (approved):

protocol/testing/containertest/containertest.sh
roy-dydx: Yup those two will need to be changed. Probably in the same PR as adding new upgrade handler

protocol/testing/containertest/preupgrade_genesis.json
roy-dydx: Added better comment and PR description.

Yeah it's not ideal, but it's not the worst since we don't have to do this often. The other solution I was thinking was to check out the files at that commit, but that's also not that nice.

roy-dydx: For example, osmosis gets the old binary and genesis via docker hub containers that they build.

PR #894, opened 2023-12-15
Add upgrade test

roy-dydx:
### Changelist
- Temporarily remove other container tests (previously disabled through due to flakiness). Reasonably certain the flakiness happens due to multiple tests, so having only the upgrade test for now is fine.
- Increase test node wait times to 60s.
- Reinitialize connection per query in test node. Required because nodes shut down during upgrade.
- Set voting period for 15s in preupgrade entrypoint.
- Add upgrade test.
- Add ratelimit store upgrade + todo to initialize params to desired state.

### Test Plan
Run the test

### Author/Reviewer Checklist


protocol/app/upgrades/v3.0.0/upgrade.go
```diff
@@ -101,6 +101,8 @@ func CreateUpgradeHandler(
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		ctx.Logger().Info("Running %s Upgrade...", UpgradeName)
 		InitializeModuleAccs(ctx, ak)
+
+		// TODO: Initialize ratelimit module params to desired state.
```
roy-dydx, line 105: @teddyding For you to add do when you get to it

protocol/testing/containertest/testnet_test.go
```diff
@@ -80,6 +84,9 @@ func expectProto(t *testing.T, tag string, message proto.Message) bool {
 }
 
 func TestPlaceOrder(t *testing.T) {
+	if os.Getenv("SKIP_DISABLED") != "" {
```
teddyding, line 87: Can we add some documentation around why these specific tests are disabled? Doesn't seem like this specific to upgrade testing?

protocol/testing/containertest/testnet_test.go
```diff
@@ -80,6 +84,9 @@ func expectProto(t *testing.T, tag string, message proto.Message) bool {
...
+		t.Skip("Skipping disabled test")
```
teddyding, line 88: Would it be helpful to log out test name being skipped? Or does the test output already make that clear

protocol/testing/containertest/testnet_test.go
```diff
@@ -262,3 +275,60 @@ func TestMarketPrices(t *testing.T) {
...
+
+	require.NoError(t, BroadcastTx(
+		node,
+		proposal,
+		constants.AliceAccAddress.String(),
+	))
+	err = node.Wait(2)
+	require.NoError(t, err)
+
+	addresses := []string{
```
teddyding, line 313: Nit: Seems useful to define this list as a common variable for the file

protocol/testing/containertest/testnet_test.go
```diff
@@ -262,3 +275,60 @@ func TestMarketPrices(t *testing.T) {
...
+		constants.AliceAccAddress.String(),
+		constants.BobAccAddress.String(),
+		constants.CarlAccAddress.String(),
+		constants.DaveAccAddress.String(),
+	}
+
+	for _, address := range addresses {
+		require.NoError(t, BroadcastTx(
```
teddyding, line 321: My understanding is this synchronous or asynchronous broadcast? What would cause `BroadcastTx` to return an error?

protocol/testing/containertest/testnet_test.go
```diff
@@ -262,3 +275,60 @@ func TestMarketPrices(t *testing.T) {
...
+				ProposalId: 1,
+				Voter:      address,
+				Option:     gov.VoteOption_VOTE_OPTION_YES,
+			},
+			address,
+		))
+	}
+
+	err = node.WaitUntilBlockHeight(12)
+	require.NoError(t, err)
```
teddyding, line 333: Will we run other "baseline" tests after upgrade, like placing order, etc?

protocol/app/upgrades/v3.0.0/upgrade.go
teddyding: Could you reference this ticket: CORE-824, for tracking purpose

protocol/testing/containertest/testnet_test.go
roy-dydx: It'll default to sync here. I'm bootstrapping by reusing some broadcast functions and those can error in various places before and during the actual broadcast. Realistically this will probably never error in practice though.

protocol/testing/containertest/testnet_test.go
roy-dydx: Not for now. I don't think it is that helpful for us because we'd need 100% coverage to catch anything meaningful. I would rather attempt to improve TestApp to allow upgrades and reuse existing tests (similar to the nondeterminism check Luke added) rather than try to get full coverage in container test.

teddyding (approved):

PR #897, opened 2023-12-18
[CORE-836] Implement `GetBaseline`

teddyding:
### Changelist
Pretty straightforward logic change. Mostly adding/updating tests

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/ratelimit/keeper/keeper.go
+++ b/protocol/x/ratelimit/keeper/keeper.go
@@ -16,16 +16,11 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/x/ratelimit/types"
 )
 
-var (
-	// TempTVLPlacerholder is a placeholder value for TVL.
-	// TODO(CORE-836): Remove this after `GetBaseline` is fully implemented.
-	TempTVLPlacerholder = big.NewInt(20_000_000_000_000)
-)
-
 type (
 	Keeper struct {
-		cdc      codec.BinaryCodec
-		storeKey storetypes.StoreKey
+		cdc        codec.BinaryCodec
+		storeKey   storetypes.StoreKey
+		bankKeeper types.BankKeeper
 
 		// TODO(CORE-824): Implement `x/ratelimit` keeper
 
@@ -37,11 +32,13 @@ type (
 func NewKeeper(
 	cdc codec.BinaryCodec,
 	storeKey storetypes.StoreKey,
+	bankKeeper types.BankKeeper,
 	authorities []string,
 ) *Keeper {
 	return &Keeper{
 		cdc:         cdc,
 		storeKey:    storeKey,
+		bankKeeper:  bankKeeper,
 		authorities: lib.UniqueSliceToSet(authorities),
 	}
 }
@@ -130,8 +127,8 @@ func (k Keeper) GetBaseline(
 	limiter types.Limiter,
 ) *big.Int {
 	// Get the current TVL.
-	// TODO(CORE-836): Query bank Keeper to get current supply of the token.
-	currentTVL := TempTVLPlacerholder
+	supply := k.bankKeeper.GetSupply(ctx, denom)
+	currentTVL := supply.Amount.BigInt()
 
 	return lib.BigMax(
 		limiter.BaselineMinimum.BigInt(),
--- a/protocol/x/ratelimit/keeper/keeper_test.go
+++ b/protocol/x/ratelimit/keeper/keeper_test.go
@@ -1,8 +1,12 @@
 package keeper_test
 
 import (
+	"math/big"
 	"testing"
 
+	cometbfttypes "github.com/cometbft/cometbft/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/dydxprotocol/v4-chain/protocol/dtypes"
 	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
 	"github.com/dydxprotocol/v4-chain/protocol/x/ratelimit/types"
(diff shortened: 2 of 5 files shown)


protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
 }
 
 func TestSetGetLimitParams_Success(t *testing.T) {
```
teddyding, line 60: Updated test method to use `tests` instead of a single test

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -130,8 +127,8 @@ func (k Keeper) GetBaseline(
 	limiter types.Limiter,
 ) *big.Int {
 	// Get the current TVL.
-	// TODO(CORE-836): Query bank Keeper to get current supply of the token.
-	currentTVL := TempTVLPlacerholder
+	supply := k.bankKeeper.GetSupply(ctx, denom)
+	currentTVL := supply.Amount.BigInt()
```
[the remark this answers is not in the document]
ttl33: I don't think this comment is relevant since both the current and new CosmosSDK versions won't return errors. In case of internal error, the `GetSupply` returns a 0 coin. 

Of course, panic is a different story

Old: https://github.com/dydxprotocol/cosmos-sdk/blob/bdf96fddf421e90574e53af383c1468fdb33d7a3/x/bank/keeper/keeper.go#L218

New: https://github.com/dydxprotocol/cosmos-sdk/blob/b48250d4e0efacbd40f3e0b980c1e1004d29e351/x/bank/keeper/keeper.go#L201

protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
...
+					BaselineMinimum: dtypes.NewInt(100_000_000_000), // 100k tokens (assuming 6 decimals)
+					BaselineTvlPpm:  10_000,                         // 1%
+				},
+				{
+					PeriodSec:       86_400,
+					BaselineMinimum: dtypes.NewInt(1_000_000_000_000), // 1m tokens (assuming 6 decimals)
+					BaselineTvlPpm:  100_000,                          // 10%
+				},
+			},
+			expectedCapacityList: []dtypes.SerializableInt{
```
ttl33, line 82: for my own understanding, will there be capacity per denom?

protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
...
+					BaselineMinimum: dtypes.NewInt(100_000_000_000), // 100k tokens (assuming 6 decimals)
+					BaselineTvlPpm:  10_000,                         // 1%
+				},
+				{
+					PeriodSec:       86_400,
+					BaselineMinimum: dtypes.NewInt(1_000_000_000_000), // 1m tokens (assuming 6 decimals)
+					BaselineTvlPpm:  100_000,                          // 10%
+				},
+			},
+			balances: []banktypes.Balance{
```
ttl33, line 101: 1. can we add a test case where there's some rounding down or rounding up happens?
2. corner cases: can we add test cases where baseline min and the ppm are 0 or maxmium? 

for example, 10M TVL and the base line ppm are 3% and 30% or something like that?

protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
...
+				},
+			},
+			limiter: types.Limiter{
+				PeriodSec:       3_600,
+				BaselineMinimum: dtypes.NewInt(100_000_000_000), // 100k token
+				BaselineTvlPpm:  10_000,                         // 1%
+			},
+			expectedBaseline: big.NewInt(150_000_000_000), // 150k token (1% of 15m)
+		},
+		"max(10% of TVL, $1mm), TVL = 20M token": {
```
ttl33, line 266: why do we use `$1mm` here? is it supposed to be 1m token?

protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
...
+				},
+			},
+			limiter: types.Limiter{
+				PeriodSec:       3_600,
+				BaselineMinimum: dtypes.NewInt(100_000_000_000), // 1m token
+				BaselineTvlPpm:  100_000,                        // 10%
+			},
+			expectedBaseline: big.NewInt(2_000_000_000_000), // 2m token (10% of 20m)
+		},
+		"max(10% of TVL, $1mm), TVL = 8M token": {
```
ttl33, line 304: same here for `$1mm`

protocol/x/ratelimit/keeper/keeper_test.go
```diff
@@ -51,72 +58,310 @@ func TestSetGetDenomCapacity(t *testing.T) {
...
 
 func TestGetBaseline(t *testing.T) {
-	// TODO(CORE-836): Add test for GetBaseline.
+	tests := map[string]struct {
+		denom            string
+		balances         []banktypes.Balance
+		limiter          types.Limiter
+		expectedBaseline *big.Int
+	}{
+		"max(1% of TVL, 100k token), TVL = 5M token": {
```
ttl33, line 199: similar to above, 
1. can you add test cases that do not divide evenly and would cause rounding down and rounding up?
2. as a corner case, can we add test cases that results in 0 baseline?

ttl33 (approved):

protocol/x/ratelimit/keeper/keeper_test.go
teddyding: This list is per denom. See test body below - 
```
			// Expected `DenomCapacity` is initialized such that each capacity is equal to the baseline.
			expectedDenomCapacity := types.DenomCapacity{
				Denom:        tc.denom,
				CapacityList: tc.expectedCapacityList,
			}
```

protocol/x/ratelimit/keeper/keeper_test.go
teddyding: 1. Added unit test for rounding.

protocol/x/ratelimit/keeper/keeper_test.go
teddyding: 1. Added such cases
2. As mentioned above, we will not allow `0` baseline or `ppm`

PR #898, opened 2023-12-19
[IND-546] Prevent IOC/FOK orders from changing order book levels / sending updates.

vincentwschau:
### Changelist
Previously, it was possible that an IOC/FOK order is updated to have non-zero remaining amount, and the logic in the order handlers would then add a price level for that IOC/FOK order (e.g. a buy order with price $50 would lead to the creation of a $50 price level on the buy side). This is incorrect as the IOC/FOK order should never lead to the order book having a new price level as it never rests on the book.

This change update the `order-update-handler` to not change orderbook price levels when the remaining amount of FOK/IOC orders change. Logic around decreasing price levels when orders are removed which exist in the `order-place-handler` (for replacing orders) and `order-remove-handler` also needed to be changed to exclude FOK/IOC orders.
This will require some clean-up of orderbook levels that are affected by IOC/FOK orders after upgrading to a version with this change, as there may be some crossed orderbooks.

### Test Plan
Unit tests, will test in dev/staging.

### Author/Reviewer Checklist

### DIFF
--- a/indexer/packages/v4-proto-parser/src/order-helpers.ts
+++ b/indexer/packages/v4-proto-parser/src/order-helpers.ts
@@ -24,7 +24,7 @@ export function isStatefulOrder(orderFlag: number | String): boolean {
   return numberOrderFlag === ORDER_FLAG_CONDITIONAL || numberOrderFlag === ORDER_FLAG_LONG_TERM;
 }
 
-export function isIOC(tif: IndexerOrder_TimeInForce): boolean {
+export function requiresImmediateExecution(tif: IndexerOrder_TimeInForce): boolean {
   return (
     tif === IndexerOrder_TimeInForce.TIME_IN_FORCE_FILL_OR_KILL ||
     tif === IndexerOrder_TimeInForce.TIME_IN_FORCE_IOC
--- a/indexer/packages/v4-proto-parser/src/order-helpers.ts
+++ b/indexer/packages/v4-proto-parser/src/order-helpers.ts
@@ -1,6 +1,6 @@
 import { createHash } from 'crypto';
 
-import { IndexerOrderId } from '@dydxprotocol-indexer/v4-protos';
+import { IndexerOrderId, IndexerOrder_TimeInForce } from '@dydxprotocol-indexer/v4-protos';
 
 import { ORDER_FLAG_CONDITIONAL, ORDER_FLAG_LONG_TERM } from './constants';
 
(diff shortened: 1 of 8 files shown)


indexer/packages/v4-proto-parser/src/order-helpers.ts
```diff
@@ -23,3 +23,10 @@ export function isStatefulOrder(orderFlag: number | String): boolean {
   }
   return numberOrderFlag === ORDER_FLAG_CONDITIONAL || numberOrderFlag === ORDER_FLAG_LONG_TERM;
 }
+
+export function isIOC(tif: IndexerOrder_TimeInForce): boolean {
```
Christopher-Li, line 27: why does isIOC return true if the order if fill or kill? Do we want to differentiate or is it because fok is a subset of ioc?

indexer/services/vulcan/src/handlers/order-update-handler.ts
```diff
@@ -145,27 +145,33 @@ export class OrderUpdateHandler extends Handler {
...
-      .getPerpetualMarketFromTicker(updateResult.order!.ticker);
-    if (perpetualMarket === undefined) {
-      logger.error({
-        at: 'OrderUpdateHandler#handle',
-        message: `Received order update for order with unknown perpetual market, ticker ${
-          updateResult.order!.ticker}`,
-      });
-      return;
-    }
+    if (!isIOC(updateResult.order!.order!.timeInForce)) {
```
Christopher-Li, line 149: can we reverse this logic?
```
if (isIOC) {
  return
}
other logic
```

This reduces the code complexity

indexer/packages/v4-proto-parser/src/order-helpers.ts
vincentwschau: Updated to `requiresImmediateExecution` matching the naming of the same [method](https://github.com/dydxprotocol/v4-chain/blob/ece92b4216ce61c6e28090124aae1c4f1144ec2f/protocol/x/clob/types/order.go#L156) in the protocol.

indexer/services/vulcan/src/handlers/order-update-handler.ts
vincentwschau: I don't entirely agree, as this only would work for errors or if the order book calculations are the last piece of logic to run. If more logic is added after this block of code, we could miss that this earlier block has an early return and not realize the logic won't run for orders with immediate execution.

Christopher-Li (approved):

PR #904, opened 2023-12-21
[feat][protocol] Write block number to state when encountering zero-fill deleveraging op [CLOB-1030]

lucas-dydx:
### Changelist
When encountering a zero-fill deleveraging operation in `DeliverTx`, write the current block number to state.

Note the zero-fill deleveraging operation is meant to be a sentinel value to indicate a negative TNC subaccount exists in state at the time of processing the deleveraging operation.

### Test Plan
Unit tests. Will add E2E tests when the feature is fully implemented.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/process_operations.go
+++ b/protocol/x/clob/keeper/process_operations.go
@@ -675,6 +675,14 @@ func (k Keeper) PersistMatchDeleveragingToState(
 	}
 	deltaBaseQuantumsIsNegative := position.GetIsLong()
 
+	// TODO: This logic assumes that the subaccount is negative equity. Should account for final settlement markets.
+	// If there are zero-deleveraging fills, this is a sentinel value to indicate a subaccount could not be liquidated or deleveraged
+	// and still has negative equity. Mark the current block number in state to indicate a negative TNC subaccount was seen.
+	if len(matchDeleveraging.GetFills()) == 0 {
+		k.subaccountsKeeper.SetNegativeTncSubaccountSeenAtBlock(ctx, lib.MustConvertIntegerToUint32(ctx.BlockHeight()))
+		return nil
+	}
+
 	for _, fill := range matchDeleveraging.GetFills() {
 		deltaBaseQuantums := new(big.Int).SetUint64(fill.FillAmount)
 		if deltaBaseQuantumsIsNegative {
--- a/protocol/x/clob/keeper/process_operations.go
+++ b/protocol/x/clob/keeper/process_operations.go
@@ -691,13 +691,11 @@ func (k Keeper) PersistMatchDeleveragingToState(
 			)
 		}
 
-		telemetry.IncrCounterWithLabels(
-			[]string{types.ModuleName, metrics.ProcessOperations, metrics.NegativeTncSubaccountSeen, metrics.Count},
-			1,
-			[]metrics.Label{
-				metrics.GetLabelForIntValue(metrics.PerpetualId, int(perpetualId)),
-				metrics.GetLabelForBoolValue(metrics.IsLong, position.GetIsLong()),
-			},
+		metrics.IncrCountMetricWithLabels(
+			types.ModuleName,
+			metrics.SubaccountsNegativeTncSubaccountSeen,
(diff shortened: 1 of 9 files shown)


protocol/x/clob/keeper/process_operations.go
```diff
@@ -675,6 +675,26 @@ func (k Keeper) PersistMatchDeleveragingToState(
 	}
 	deltaBaseQuantumsIsNegative := position.GetIsLong()
 
+	// TODO: This logic assumes that the subaccount is negative equity. Should account for final settlement markets.
+	// If there are zero-deleveraging fills, this is a sentinel value to indicate a subaccount could not be liquidated
+	// or deleveraged and still has negative equity. Mark the current block number in state to indicate a negative TNC
+	// subaccount was seen.
+	if len(matchDeleveraging.GetFills()) == 0 {
```
lucas-dydx, line 682: Question: should validation be added that this only happens once per block?

I punted on this to keep things simple and there doesn't seem to be much benefit to adding this validation 🤔

protocol/x/clob/keeper/process_operations.go
```diff
@@ -675,6 +675,25 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
+		if !shouldDeleverageAtBankruptcyPrice {
+			panic(
+				fmt.Sprintf(
+					"PersistMatchDeleveragingToState: zero-fill deleveraging operation included for subaccount %+v"+
+						" and perpetual %d but cannot be be deleveraged at bankruptcy price",
+					liquidatedSubaccountId,
+					perpetualId,
+				),
+			)
```
jakob-dydx, line 690: Is there a reason we are panicking here instead of throwing an error?

Does it make sense to incr a counter for the error to track how often this is happening?
Does it make sense to add telemetry for the SetNegativeTncSubaccountSeenAtBlock fn so we know how often the success (valid zero fill deleverage) case happens?

jakob-dydx (approved):

protocol/x/clob/keeper/process_operations.go
lucas-dydx: Will update this to throw an error + add telemetry

jakob-dydx (approved):

protocol/x/clob/keeper/process_operations.go
```diff
@@ -675,6 +675,34 @@ func (k Keeper) PersistMatchDeleveragingToState(
 	}
 	deltaBaseQuantumsIsNegative := position.GetIsLong()
 
+	// If there are zero-fill deleveraging operations, this is a sentinel value to indicate a subaccount could not be
+	// liquidated or deleveraged and still has negative equity. Mark the current block number in state to indicate a
+	// negative TNC subaccount was seen.
+	if len(matchDeleveraging.GetFills()) == 0 {
```
jayy04, line 681: for my understanding, what happens when a -ve TNC subaccount's position is only partially deleveraged? do we have a normal deleveraging operation with len(fills) > 0 followed by a empty deleveraging operation?

jayy04 (approved):

protocol/x/clob/keeper/process_operations.go
```diff
@@ -675,6 +675,34 @@ func (k Keeper) PersistMatchDeleveragingToState(
...
+				fmt.Sprintf(
+					"PersistMatchDeleveragingToState: zero-fill deleveraging operation included for subaccount %+v"+
+						" and perpetual %d but subaccount isn't negative TNC",
+					liquidatedSubaccountId,
+					perpetualId,
+				),
+			)
+		}
+
+		telemetry.IncrCounterWithLabels(
```
jonfung-dydx, line 694: can you use the new metrics entrypoint [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/lib/metrics/util.go) instead?

protocol/x/clob/keeper/process_operations.go
jonfung-dydx: +1 on question, also good to have a test for this case if true

lucas-dydx: Answered synchronously, but yeah

PR #915, opened 2024-01-02
Update BlockTable.getLatest to throw an error and simplify logic

Christopher-Li:
### Changelist
Update BlockTable.getLatest to throw an error and simplify logic

### Test Plan
Update unit tests

### Author/Reviewer Checklist


dydxwill (commented):
Is the reason to do this to simplify the code?
Won't this introduce timing constraints between Indexer services? i.e. we need Ender to process a block before starting Roundtable/Comlink/etc.

dydxwill (approved):

PR #918, opened 2024-01-02
Lukecore538.3

lcwik:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/app/upgrades/v3.0.0/upgrade.go
```diff
@@ -1,169 +0,0 @@
...
-	mm *module.Manager,
-	configurator module.Configurator,
-	ak authkeeper.AccountKeeper,
-) upgradetypes.UpgradeHandler {
-	return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
-		sdkCtx := sdk.UnwrapSDKContext(ctx)
-		sdkCtx.Logger().Info("Running %s Upgrade...", UpgradeName)
-		InitializeModuleAccs(sdkCtx, ak)
-
-		// TODO(CORE-824): Initialize ratelimit module params to desired state.
```
roy-dydx, line 163: Can you move this TODO to v4.0.0 handler? ratelimit module is not implemented fully yet, but it will have genesis params and those need to be initialized properly.

protocol/testing/containertest/containertest.sh
```diff
@@ -154,27 +153,22 @@ setup_cosmovisor() {
 	done
 }
 
-download_preupgrade_binary() {
+copy_preupgrade_binary_arch() {
```
roy-dydx, line 156: Can start using v3.0.0-rc0 binaries (building right now and will be published soon for testnet).

protocol/app/upgrades/v3.0.0/upgrade.go
```diff
@@ -1,169 +0,0 @@
-package v_3_0_0
```
roy-dydx, line 1: Can we keep this old upgrade handler around? If I understand correctly it doesn't cost anything.

 If we want to cherry-pick something that needs to modify the upgrade handler, it will be nice to have still.

protocol/testing/containertest/containertest.sh
lcwik: Can we just swap them for the 3.0.0 binaries once they are qualified. I don't suspect that this swap will cause many issues and then we will only need to do it the one time.

protocol/app/upgrades/v3.0.0/upgrade.go
lcwik: I can keep but I suspect the merge pain of developing here and then porting it to 0.47 would be annoying when cherry-picking on to 3.0.0 release branch.

roy-dydx: Oh good point. Then I don't care either way.

roy-dydx (approved):

PR #926, opened 2024-01-04
[CLOB-1055] - final settlement e2e tests

jakob-dydx:
This PR contains an e2e test suite for generating, passing, and executing a governance proposal which winds down a market. The test suite also verifies that open stateful orders in the market are closed, positions are closed via deleveraging / final settlement deleveraging, and that the proper indexer events are emitted


protocol/testing/e2e/gov/wind_down_market_test.go
```diff
@@ -0,0 +1,287 @@
...
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+)
+
+func TestWindDownMarketProposal(t *testing.T) {
+	tests := map[string]struct {
+		subaccounts    []satypes.Subaccount
+		statefulOrders []clobtypes.MsgPlaceOrder
+
+		expectedSubaccounts []satypes.Subaccount
+	}{
```
jayy04, line 32: can we add a few more test cases 
- trading is disabled (new messages rejected)
- trading is disabled during PCS replay (not sure if there is a good way to test this with e2e framework)
- liquidation is disabled

jayy04 (approved):

protocol/testing/e2e/gov/wind_down_market_test.go
jakob-dydx: i think some of these tests are things that I can break out into a different e2e test which doesn't need to include the passing of the gov proposal in the test suite. The purpose of this test suite in particular was to verify things that happen as side effects of the proposal passing (open orders cancelled, open positions closed). Will break these out

PR #936, opened 2024-01-05
[feat][protocol]: gate withdrawals if negative TNC subaccount encountered after liquidation and deleveraging steps

lucas-dydx:
### Changelist
Gate withdrawals if a negative TNC subaccount is encountered after the liquidation and deleveraging steps in `PrepareCheckState`.

Note that all potentially liquidatable subaccounts specified by the daemon will be checked for simplicity.

### Test Plan
Wrote E2E tests verifying that when there's negative TNC subaccounts found in state, that withdrawals are blocked.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/e2e/withdrawal_gating_test.go
+++ b/protocol/x/clob/e2e/withdrawal_gating_test.go
@@ -1,0 +1,224 @@
+package clob_test
+
+import (
+	"testing"
+
+	abcitypes "github.com/cometbft/cometbft/abci/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/dydxprotocol/v4-chain/protocol/daemons/liquidation/api"
+
+	"github.com/cometbft/cometbft/types"
+
+	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
+	clobtest "github.com/dydxprotocol/v4-chain/protocol/testutil/clob"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	assettypes "github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+	feetiertypes "github.com/dydxprotocol/v4-chain/protocol/x/feetiers/types"
+	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
+	prices "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
+	sendingtypes "github.com/dydxprotocol/v4-chain/protocol/x/sending/types"
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithdrawalGating(t *testing.T) {
+	tests := map[string]struct {
+		// State.
+		subaccounts                   []satypes.Subaccount
+		marketIdToOraclePriceOverride map[uint32]uint64
+
+		// Parameters.
+		placedMatchableOrders     []clobtypes.MatchableOrder
+		liquidatableSubaccountIds []satypes.SubaccountId
+		negativeTncSubaccountIds  []satypes.SubaccountId
+
+		// Configuration.
+		liquidationConfig clobtypes.LiquidationsConfig
+		liquidityTiers    []perptypes.LiquidityTier
+		perpetuals        []perptypes.Perpetual
+		clobPairs         []clobtypes.ClobPair
+
+		// Expectations.
+		expectedSubaccounts                      []satypes.Subaccount
+		expectedWithdrawalsGated                 bool
+		expectedNegativeTncSubaccountSeenAtBlock uint32
+		expectedErr                              string
+	}{
+		`Can place a liquidation order that is unfilled and cannot be deleveraged due to
+		non-overlapping bankruptcy prices, withdrawals are gated`: {
+			subaccounts: []satypes.Subaccount{
+				constants.Carl_Num0_1BTC_Short_49999USD,
+				constants.Dave_Num0_1BTC_Long_50000USD_Short,
+			},
+
+			placedMatchableOrders: []clobtypes.MatchableOrder{
+				// Carl's bankruptcy price to close 1 BTC short is $49,999, and closing at $50,000
+				// would require $1 from the insurance fund. Since the insurance fund is empty,
+				// deleveraging is required to close this position.
+				&constants.Order_Dave_Num0_Id1_Clob0_Sell025BTC_Price50000_GTB11,
+			},
+			liquidationConfig:         constants.LiquidationsConfig_FillablePrice_Max_Smmr,
+			liquidatableSubaccountIds: []satypes.SubaccountId{constants.Carl_Num0},
+			negativeTncSubaccountIds:  []satypes.SubaccountId{constants.Carl_Num0},
+
+			liquidityTiers: constants.LiquidityTiers,
+			perpetuals: []perptypes.Perpetual{
+				constants.BtcUsd_20PercentInitial_10PercentMaintenance,
+			},
+			clobPairs: []clobtypes.ClobPair{constants.ClobPair_Btc},
+
+			expectedSubaccounts: []satypes.Subaccount{
+				// Deleveraging fails.
+				// Dave's bankruptcy price to close 1 BTC long is $50,000, and deleveraging can not be
+				// performed due to non overlapping bankruptcy prices.
+				constants.Carl_Num0_1BTC_Short_49999USD,
+				constants.Dave_Num0_1BTC_Long_50000USD_Short,
+			},
+			expectedWithdrawalsGated:                 true,
+			expectedNegativeTncSubaccountSeenAtBlock: 4,
+			expectedErr:                              "WithdrawalsAndTransfersBlocked: failed to apply subaccount updates",
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			tApp := testapp.NewTestAppBuilder(t).WithGenesisDocFn(func() (genesis types.GenesisDoc) {
+				genesis = testapp.DefaultGenesis()
+				testapp.UpdateGenesisDocWithAppStateForModule(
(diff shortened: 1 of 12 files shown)


protocol/x/clob/keeper/deleveraging.go
```diff
@@ -188,6 +188,50 @@ func (k Keeper) CanDeleverageSubaccount(
...
+		// If the subaccount has negative TNC, mark that a negative TNC subaccount was found.
+		if bigNetCollateral.Sign() == -1 {
+			foundNegativeTncSubaccount = true
+			negativeTncSubaccountId = subaccountId
+			break
+		}
+	}
+
+	if !foundNegativeTncSubaccount {
+		return nil
```
lucas-dydx, line 221: Planning to add more metrics here.
- [ ] Add counters for total calls to this method.
- [ ] Add counters for halted withdrawals (can add alerts on this).
- [ ] Add latency metrics to this method.

protocol/x/clob/e2e/withdrawal_gating_test.go
```diff
@@ -0,0 +1,224 @@
...
+				testapp.AdvanceToBlockOptions{
+					ValidateFinalizeBlock: func(
+						ctx sdktypes.Context,
+						request abcitypes.RequestFinalizeBlock,
+						response abcitypes.ResponseFinalizeBlock,
+					) (haltchain bool) {
+						// Note the first TX is MsgProposeOperations, the second is all other TXs.
+						execResult := response.TxResults[1]
+						require.True(t, execResult.IsErr())
+						require.Equal(t, satypes.ErrFailedToUpdateSubaccounts.ABCICode(), execResult.Code)
```
lucas-dydx, line 216: Planning to add the following additional tests:
- Transfers and withdrawals become unblocked after `WITHDRAWAL_AND_TRANSFERS_BLOCKED_AFTER_NEGATIVE_TNC_SUBACCOUNT_SEEN_BLOCKS` blocks
- Both transfers and withdrawals fail
- Transfers and withdrawals are not blocked if subaccounts specified in negative TNC subaccounts have non-negative TNC after liquidations and deleveraging steps

protocol/x/clob/e2e/withdrawal_gating_test.go
```diff
@@ -0,0 +1,231 @@
...
+		clobPairs         []clobtypes.ClobPair
+
+		// Expectations.
+		expectedSubaccounts                      []satypes.Subaccount
+		expectedWithdrawalsGated                 bool
+		expectedNegativeTncSubaccountSeenAtBlock uint32
+		expectedErr                              string
+	}{
+		`Can place a liquidation order that is unfilled and cannot be deleveraged due to
+		non-overlapping bankruptcy prices, withdrawals are gated`: {
```
jayy04, line 50: with the current implementation, liquidation order doesn't need to be unfilled and deleverage can still happen for withdrawals to be gated, right? since this is happening in step 8 which is after liquidations/deleveraging steps. 

can we add tests for these?

lucas-dydx: yeah I'll add tests for all variants of the below possible scenarios that can happen for verifying withdrawals are gated / not gated:
- liquidation order unfilled, partial deleveraging
- liquidation order filled, deleveraging fails
- liquidation order filled, partial deleveraging

protocol/x/clob/e2e/withdrawal_gating_test.go
```diff
@@ -0,0 +1,231 @@
...
+				tApp.App,
+				testapp.MustMakeCheckTxOptions{
+					AccAddressForSigning: constants.Carl_Num0.Owner,
+					Gas:                  1000000,
+					FeeAmt:               constants.TestFeeCoins_5Cents,
+				},
+				&withdrawMsg,
+			) {
+				resp := tApp.CheckTx(checkTx)
+				require.Conditionf(t, resp.IsOK, "Expected CheckTx to succeed. Response: %+v", resp)
```
jayy04, line 203: Is there a reason why we can't reject withdrawals / transfers here? is it because we don't actually process the transfer until `DeliverTx`?

lucas-dydx: yeah iirc txs only execute their handler in `DeliverTx` right? https://github.com/dydxprotocol/cosmos-sdk/blob/b48250d4e0efacbd40f3e0b980c1e1004d29e351/baseapp/baseapp.go#L1106-L1119

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -188,6 +188,56 @@ func (k Keeper) CanDeleverageSubaccount(
...
+	subaccount := k.subaccountsKeeper.GetSubaccount(ctx, negativeTncSubaccountId)
+	perpetualPositions := subaccount.GetPerpetualPositions()
+	if len(perpetualPositions) == 0 {
+		return errorsmod.Wrapf(
+			types.ErrNoPerpetualPositionsToLiquidate,
+			"GateWithdrawalsIfNegativeTncSubaccountSeen: subaccount has no open positions: (%s)",
+			lib.MaybeGetJsonString(subaccount),
+		)
+	}
+	perpetualId := subaccount.PerpetualPositions[0].PerpetualId
```
jonfung-dydx, line 235: Is there any case a subaccount can have negative TNC and have a no perpetual positions? ie price slips fast enough and we match/liquidate at a less favorable price than the bankruptcy price?

In this case do we want to default to some value if this doesn't exist? I don't think the perpetual id is used for zero fill deleveraging operations, so having a default would be good to protect against the error when we access a nonexisting array element.

protocol/x/clob/e2e/withdrawal_gating_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}{
+		`Can place a liquidation order that is unfilled and cannot be deleveraged due to
+		non-overlapping bankruptcy prices, withdrawals are gated`: {
+			subaccounts: []satypes.Subaccount{
+				constants.Carl_Num0_1BTC_Short_49999USD,
+				constants.Dave_Num0_1BTC_Long_50000USD_Short,
+			},
+
+			placedMatchableOrders: []clobtypes.MatchableOrder{
+				// Carl's bankruptcy price to close 1 BTC short is $49,999, and closing at $50,000
```
jonfung-dydx, line 57: is this dave

protocol/x/clob/keeper/deleveraging.go
lucas-dydx: > Is there any case a subaccount can have negative TNC and have a no perpetual positions? ie price slips fast enough and we match/liquidate at a less favorable price than the bankruptcy price?

shouldn't be possible since subaccounts shouldn't be able to close out all of their positions and still have negative tnc

if this was possible, it would require a subaccount state transition to occur at a worse price than subaccount's bankruptcy price (price that subaccount tnc goes to zero), and that should be invalid

i agree it's still good to think through this though in case invariant gets violated. want me to create an m7 ticket to handle this? also maybe worth discussing with rest of CLOB team to ensure it never happens

subaccounts shouldn't be able to close out all of their positions and still have negative tnc. it shouldn't be possible to close a position and still have negative TNC also note subaccounts [aren't even liquidatable if they have zero positions](https://github.com/dydxprotocol/v4-chain/blob/6014ec261478d509e1ffa6742b40fa5967e970a9/protocol/x/clob/keeper/liquidations.go#L376), so i guess that means they wouldn't be deleverage-able either? so if it's possible for subaccounts to be in this state, then i dont think theres any way out of it (besides transferring the subac

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -188,6 +188,56 @@ func (k Keeper) CanDeleverageSubaccount(
...
+	k.MemClob.InsertZeroFillDeleveragingIntoOperationsQueue(ctx, negativeTncSubaccountId, perpetualId)
+
+	return nil
```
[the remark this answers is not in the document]
lucas-dydx: honestly would prefer to avoid this. the current approach is simpler since it avoids modifying existing critical paths to add on a hack (this is prob not long-term withdrawal gating solution)

for example if we modify `memclob.DeleverageSubaccount`, then wouldn't we have to update the logic to handle cases where zero-fills are returned from [`OffsetSubaccountPerpetualPosition`](https://github.com/dydxprotocol/v4-chain/blob/6014ec261478d509e1ffa6742b40fa5967e970a9/protocol/x/clob/keeper/deleveraging.go#L223)? maybe it's fine to add these to the op queue, but imo just makes things more complicated for little gain

given this is not the long-term solution, would rather add this new function so it's easy to rip out later once we build long-term solution for withdrawal gating

protocol/x/clob/e2e/withdrawal_gating_test.go
lucas-dydx: u r dave

protocol/x/clob/keeper/deleveraging.go
jonfung-dydx: i think quick fix we should just add a defeault value of 0, and make a ticket to think through edge cases later

jonfung-dydx (approved):

protocol/x/clob/e2e/withdrawal_gating_test.go
jayy04: hmm yeah not sure if it's worth building an ante handler for this 🤔 nbd for now

protocol/x/clob/keeper/deleveraging.go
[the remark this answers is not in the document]
jayy04: yeah sounds good

jayy04 (approved):

protocol/x/clob/e2e/withdrawal_gating_test.go
lucas-dydx: yeah i dont think it is personally since this is an edge case

ideally frontend's query whether withdrawals are blocked before submitting in vast majority of cases

protocol/x/clob/keeper/deleveraging.go
lucas-dydx: synced on slack and we're not doing this for simplicity. created ticket here: https://linear.app/dydx/issue/CLOB-1061/use-default-perpetual-id-for-withdrawal-gating

PR #938, opened 2024-01-08
[CLOB-1054] update state R/W methods

jayy04:
### Changelist
- Update state R/W methods for improved conditional order triggering

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/match_state.go
+++ b/protocol/x/clob/keeper/match_state.go
@@ -23,8 +23,8 @@ func (k Keeper) GetTradePricesForPerpetual(
 		return 0, 0, false
 	}
 
-	maxTradePriceSubticks, _ = k.getMaxTradePriceForPerpetual(ctx, perpetualId)
-	return minTradePriceSubticks, maxTradePriceSubticks, true
+	maxTradePriceSubticks, found = k.getMaxTradePriceForPerpetual(ctx, perpetualId)
+	return minTradePriceSubticks, maxTradePriceSubticks, found
 }
 
 // getMinTradePriceForPerpetual gets the min trade price for a perpetual.
--- a/protocol/x/clob/keeper/match_state.go
+++ b/protocol/x/clob/keeper/match_state.go
@@ -7,7 +7,9 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 )
 
-// GetTradePricesForPerpetual gets the last trade price for a perpetual.
+// GetTradePricesForPerpetual gets the maximum and minimum traded prices for a perpetual for the
+// current block.
+// These prices are intended to be used for improved conditional order triggering in the EndBlocker.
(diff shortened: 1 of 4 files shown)


protocol/x/clob/keeper/match_state.go
```diff
@@ -24,17 +59,26 @@ func (k Keeper) GetLastTradePriceForPerpetual(
 	return types.Subticks(result.Value), true
 }
 
-// SetLastTradePriceForPerpetual sets the last trade price for a perpetual.
-func (k Keeper) SetLastTradePriceForPerpetual(
+// SetTradePricesForPerpetual sets the trade prices for a perpetual.
+func (k Keeper) SetTradePricesForPerpetual(
```
jonfung-dydx, line 63: For my understanding, where do you expect to be calling this in order to set the trade prices? In the memclob matching loop or do we populate this from the operations queue?

protocol/x/clob/keeper/match_state.go
```diff
@@ -24,17 +59,26 @@ func (k Keeper) GetLastTradePriceForPerpetual(
 	return types.Subticks(result.Value), true
 }
 
-// SetLastTradePriceForPerpetual sets the last trade price for a perpetual.
-func (k Keeper) SetLastTradePriceForPerpetual(
+// SetTradePricesForPerpetual sets the trade prices for a perpetual.
```
jonfung-dydx, line 62: Can we add some color in the docstring for
- what is a trade price?
- sets in transient state store
- where trade price is used
- where this is supposed to be called in?

jonfung-dydx: maybe `SetLastTradePriceForPerpetual` too

jonfung-dydx (approved):

protocol/x/clob/keeper/match_state.go
jayy04: in process_operations when we actually process these matches.

protocol/x/clob/keeper/match_state.go
```diff
@@ -7,12 +7,49 @@ import (
...
+	maxTradePriceSubticks types.Subticks,
+	found bool,
+) {
+	minTradePriceSubticks, found = k.getMinTradePriceForPerpetual(ctx, perpetualId)
+	if !found {
+		return 0, 0, false
+	}
+
+	maxTradePriceSubticks, _ = k.getMaxTradePriceForPerpetual(ctx, perpetualId)
+	return minTradePriceSubticks, maxTradePriceSubticks, true
```
jakob-dydx, line 27: Don't think this is likely to occur because of the way you're writing to the store for min/max price trades, but the way this function is written we get a weird result if `getMaxTradePriceForPerpetual`'s found value is false, but the min trade price is found.

We would end up returning `minTradePriceSubticks, 0, true`

jakob-dydx (approved):

PR #941, opened 2024-01-09
[CORE-824] `x/ratelimit`: Implement `UpdateCapacityEndBlocker`

teddyding:
### Changelist
This PR implements the capacity update logic in the `x/ratelimit` EndBlocker. See more context in tech [spec](https://www.notion.so/dydx/x-ratelimit-Global-Withdrawal-Speed-Limit-f7afcda6413f4298a42700f0e4a360ca?pvs=4#3de2b1984e8744b09c2e316fa5673ff9)

### Test Plan
Unit tests

### Author/Reviewer Checklist


protocol/app/app.go
```diff
@@ -588,10 +588,22 @@ func New(
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
 	transferIBCModule := transfer.NewIBCModule(app.TransferKeeper)
 
+	app.BlockTimeKeeper = *blocktimemodulekeeper.NewKeeper(
```
teddyding, line 591: This code block is moved from below to before the `app.RateLimitKeeper` initialization.

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -122,17 +128,13 @@ func (k Keeper) ProcessDeposit(
 //	baseline = max(baseline_minimum, baseline_tvl_ppm * current_tvl)
 func (k Keeper) GetBaseline(
 	ctx sdk.Context,
-	denom string,
+	currentSupply *big.Int,
```
teddyding, line 131: Change function signature, because whereever `GetBaseline` is called for a limiter, that caller function could've called `bankKeeper.GetSupply()` for all limiters of a denom.

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+
+		if new(big.Rat).Abs(bigRatcapacityMinusBaseline).Cmp(capacityDiff) < 0 {
+			// if `abs(capacity - baseline) < capacity_diff` then `capacity = baseline``
+			newCapacityList[i] = dtypes.NewIntFromBigInt(baseline)
+		} else if capacityList[i].BigInt().Cmp(baseline) < 0 {
+			// else if `capacity < baseline` then `capacity += capacity_diff`
+			newCapacityList[i] = dtypes.NewIntFromBigInt(
+				new(big.Int).Add(
+					capacityList[i].BigInt(),
+					lib.BigRatRound(capacityDiff, false), // rounds down `capacity_diff`
```
ttl33, line 347: nit: didnt look at the unit tests closely, can you confirm that we have unit tests that verifies the round down behaviors for this method?

ttl33 (approved):
Overall lgtm. Will rely on Brendan to double check the math

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+				metrics.GetLabelForStringValue(metrics.RateLimitDenom, limitParams.Denom),
+				metrics.GetLabelForIntValue(metrics.LimiterIndex, i),
+			},
+		)
+	}
+
+	k.SetDenomCapacity(ctx, types.DenomCapacity{
+		Denom:        limitParams.Denom,
+		CapacityList: newCapacityList,
+	})
```
BrendanChou, line 374: Instead of having a function that:
- Determines the new value
- Emits telemetry
- Writes the new value to state

I suggest the following to be more clear, modular, and unit-testable:
- One function to determine the new value, but not update it in state (call the state update outside of this function)
- Move the telemetry to `SetDenomCapacity` such that it automatically does the telemetry on writes (and you don't have to have telemetry before every time you call `SetDenomCapacity`

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+	}
+
+	// Convert timestamps to milliseconds for algebraic operations.
+	blockTimeMilli := ctx.BlockTime().UnixMilli()
+	prevBlockInfo := k.blockTimeKeeper.GetPreviousBlockInfo(ctx)
+	prevBlockTimeMilli := prevBlockInfo.Timestamp.UnixMilli()
+	timeSinceLastBlockMilli := new(big.Int).Sub(
+		big.NewInt(blockTimeMilli),
+		big.NewInt(prevBlockTimeMilli),
+	)
```
BrendanChou, line 285: nit: this could be left as `int64` instead of `Big.Int`. Later can use `new(big.Rat).SetInt64(...)`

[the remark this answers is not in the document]
BrendanChou: Could just be a keeper function on the `blockTimeKeeper` as well

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+
+		// Calculate right operand: `time_since_last_block / period`
+		periodMilli := new(big.Int).Mul(
+			new(big.Int).SetUint64(uint64(limiter.PeriodSec)),
+			big.NewInt(1000),
+		)
+		operandR := new(big.Rat).Quo(
+			new(big.Rat).SetInt(timeSinceLastBlockMilli),
+			new(big.Rat).SetInt(periodMilli),
+		)
```
BrendanChou, line 328:
```suggestion
		operandR := new(big.Rat).SetFrac(
			timeSinceLastBlockMilli,
			periodMilli,
		)
```

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+			operandL,
+			operandR,
+		)
+
+		bigRatcapacityMinusBaseline := new(big.Rat).SetInt(capacityMinusBaseline)
+
+		if new(big.Rat).Abs(bigRatcapacityMinusBaseline).Cmp(capacityDiff) < 0 {
+			// if `abs(capacity - baseline) < capacity_diff` then `capacity = baseline``
+			newCapacityList[i] = dtypes.NewIntFromBigInt(baseline)
+		} else if capacityList[i].BigInt().Cmp(baseline) < 0 {
```
BrendanChou, line 342:
```suggestion
		} else if capacityMinusBaseline.Sign() < 0 {
```

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+// This is basically saying that the capacity returns to the baseline over the course of the `period`.
+// Usually in a linear way, but if the `capacity` is more than twice the `baseline`, then in an exponential way.
+//
+//	`capacity =`
+//	    if `abs(capacity - baseline) < capacity_diff` then `capacity = baseline`
+//	    else if `capacity < baseline` then `capacity += capacity_diff`
+//	    else `capacity -= capacity_diff`
+//
+// On a high level, `capacity` trends towards `baseline` by `capacity_diff` but does not “cross” it.
+func (k Keeper) UpdateCapacityForLimitParams(
```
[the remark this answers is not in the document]
teddyding: Updated to private

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+	ctx sdk.Context,
+	limitParams types.LimitParams,
```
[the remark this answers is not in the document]
teddyding: In general, I'm not a fan of passing values from function A to function B and then doing validations in function B. Similarly, `LimitParams` is an input to `updateCapacityForLimitParams` and therefore it is the responsibility of caller function to pass in correct value. I will document this assumption in function docstring.

We do not assume `DenomCapacity` to exist in this function. If it does not, the `GetDenomCapacity` will return an empty list, causing the code to throw an error in if condition `if len(capacityList) != len(limitParams.Limiters) `.

protocol/x/ratelimit/keeper/keeper.go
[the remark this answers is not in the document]
teddyding: Added `blockTimeKeeper.GetTimeSinceLastBlock`

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -218,6 +221,159 @@ func (k Keeper) SetDenomCapacity(
...
+			lib.BigMax(
+				baseline,
+				capacityMinusBaseline,
+			),
+		)
+
+		// Calculate right operand: `time_since_last_block / period`
+		periodMilli := new(big.Int).Mul(
+			new(big.Int).SetUint64(uint64(limiter.PeriodSec)),
+			big.NewInt(1000),
```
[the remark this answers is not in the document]
teddyding: I think either could work. Since we eventually need to cast to `big.Int`, I actually like the current way better, which doesn't give possibility for int overflow. Let me know if you disagree!

protocol/x/ratelimit/keeper/keeper.go
teddyding: Yes wo do!

protocol/x/ratelimit/util/capacity.go
```diff
@@ -0,0 +1,96 @@
...
+
+		// Calculate right operand: `time_since_last_block / period`
+		periodMilli := new(big.Int).Mul(
+			new(big.Int).SetUint64(uint64(limiter.PeriodSec)),
+			big.NewInt(1000),
+		)
+		operandR := new(big.Rat).SetFrac(
+			new(big.Int).SetInt64(timeSinceLastBlock.Milliseconds()),
+			periodMilli,
+		)
```
BrendanChou, line 62: nit: if we do use the `time.Duration` type (see above comment) then this can just use `SetFrac64`

protocol/x/ratelimit/util/capacity.go
```diff
@@ -0,0 +1,96 @@
...
+		// Calculate: `capacity_diff = max(baseline, capacity-baseline) * (time_since_last_block / period)`
+		// Since both operands > 0, `capacity_diff` is positive or zero (due to rounding).
+		capacityDiff := new(big.Rat).Mul(
+			operandL,
+			operandR,
+		)
+
+		bigRatcapacityMinusBaseline := new(big.Rat).SetInt(capacityMinusBaseline)
+
+		if new(big.Rat).Abs(bigRatcapacityMinusBaseline).Cmp(capacityDiff) < 0 {
```
BrendanChou, line 73:
```suggestion
		capacityDiffRat := new(big.Rat).Mul(operandL, operandR)
		capacityDiff := lib.BigRatRound(capacityDiffRat, false) // rounds down

		if new(big.Int).Abs(capacityMinusBaseline).Cmp(capacityDiff) <= 0 {
```

BrendanChou: Also remove the `lib.BigRatRound`s below

protocol/x/ratelimit/util/capacity.go
```diff
@@ -0,0 +1,96 @@
...
+) (newCapacityList []dtypes.SerializableInt) {
+	// Declare new capacity list to be populated.
+	newCapacityList = make([]dtypes.SerializableInt, len(prevCapapcityList))
+
+	for i, limiter := range limitParams.Limiters {
+		// For each limiter, calculate the current baseline.
+		baseline := GetBaseline(bigTvl, limiter)
+
+		capacityMinusBaseline := new(big.Int).Sub(
+			prevCapapcityList[i].BigInt(), // array access is safe because of input invariant
```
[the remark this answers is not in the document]
teddyding: At second thought, I believe this invariant should be maintained in `UpdateCapacityEndBlocker`, because `SetLimitParams` atomically updates the `DenomCapacity` list as well. 

So I move the invariant check from caller function to this function, and returns the error if invariant failed.

protocol/x/ratelimit/util/capacity.go
```diff
@@ -0,0 +1,96 @@
...
+				baseline,
+				capacityMinusBaseline,
+			),
+		)
+
+		// Calculate right operand: `time_since_last_block / period`
+		periodMilli := new(big.Int).Mul(
+			new(big.Int).SetUint64(uint64(limiter.PeriodSec)),
+			big.NewInt(1000),
+		)
```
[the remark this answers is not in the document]
teddyding: Good point, `period` should be in `time.Duration`. Updated proto and implementation accordingly.

protocol/x/ratelimit/util/capacity.go
teddyding: Done.

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -259,26 +259,25 @@ func (k Keeper) updateCapacityForLimitParams(
 	tvl := k.bankKeeper.GetSupply(ctx, limitParams.Denom)
 
 	capacityList := k.GetDenomCapacity(ctx, limitParams.Denom).CapacityList
-	if len(capacityList) != len(limitParams.Limiters) {
```
teddyding: Moved this check into `CalculateNewCapacityList` which throws an error

protocol/x/ratelimit/keeper/keeper.go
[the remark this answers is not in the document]
teddyding: Update: checking this under in `CalculateNewCapacityList`

BrendanChou (approved):

PR #960, opened 2024-01-11
[CORE-794] enable authz module to allow granting privileges

jayy04:
### Changelist
- Enable authz module to allow granting privileges

### Test Plan

### Author/Reviewer Checklist


protocol/app/app.go
```diff
@@ -346,10 +350,20 @@ func New(
 	bApp.SetTxEncoder(txConfig.TxEncoder())
 
 	keys := storetypes.NewKVStoreKeys(
-		authtypes.StoreKey, banktypes.StoreKey, stakingtypes.StoreKey, crisistypes.StoreKey,
-		distrtypes.StoreKey, slashingtypes.StoreKey,
-		govtypes.StoreKey, paramstypes.StoreKey, consensusparamtypes.StoreKey, upgradetypes.StoreKey, feegrant.StoreKey,
-		ibcexported.StoreKey, ibctransfertypes.StoreKey,
+		authtypes.StoreKey,
+		authzkeeper.StoreKey,
```
BrendanChou, line 354: weird this one is not set up in the same way... no `authztypes`?

protocol/app/app.go
```diff
@@ -429,6 +443,14 @@ func New(
 		sdk.GetConfig().GetBech32AccountAddrPrefix(),
 		lib.GovModuleAddress.String(),
 	)
+
+	app.AuthzKeeper = authzkeeper.NewKeeper(
+		runtime.NewKVStoreService(keys[authzkeeper.StoreKey]),
+		appCodec,
```
BrendanChou, line 449: the order of these params is also reversed vs most `NewKeeper` methods?

protocol/app/upgrades/v4.0.0/constants.go
```diff
@@ -23,6 +24,9 @@ var Upgrade = upgrades.Upgrade{
 			// Add new ICA stores that are needed by ICA host types as of v8.
 			icacontrollertypes.StoreKey,
 			icahosttypes.StoreKey,
+
+			// Add authz module to allow granting arbitrary privileges from one account to another acocunt.
```
teddyding, line 28: Is there anything else needed in the upgrade to initialize `authz` state? Is [this](https://github.com/cosmos/cosmos-sdk/blob/eaf92c225cfa2420946ff738cdcf05cf8b402040/x/authz/genesis.go#L11) inherently left empty with no `Authorization`?

protocol/app/msgs/nested_msgs.go
```diff
@@ -2,12 +2,17 @@ package msgs
...
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 	gov "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
 )
 
 var (
 	// NestedMsgSamples are msgs that have can include other arbitrary messages inside.
 	NestedMsgSamples = map[string]sdk.Msg{
+		// authz
```
teddyding, line 12: Haven't dug into how nested messages are verified - will someone be able to execute an otherwise disallowed messages as a nested message under `MsgExec`? Could you reply to @ttl33 's relevant question [here](https://linear.app/dydx/issue/CORE-794/investigate-and-add-xauthz#comment-f99b5032)?

protocol/app/app.go
jayy04: yeah not sure how `authz` was bootstrapped initially, but there is no `authz/types` and the key is under `keeper/keys.go` 😕

jayy04: looked into this quickly. `types` got removed in [this PR](https://github.com/cosmos/cosmos-sdk/pull/9042). still not sure why though

protocol/app/app.go
jayy04: yeah `authz` always have store key before codec, not sure how it was bootstrapped

protocol/app/upgrades/v4.0.0/constants.go
jayy04: yeah I think this should be empty. there is currently no authorizations since it was not enabled previously

Each `Authorization` consists of (granter, grantee, msgURL) and permissions needs to be granted explicitly by the granter

protocol/app/msgs/nested_msgs.go
jayy04: so this PR also updates nested messages verification (see this [commit](https://github.com/dydxprotocol/v4-chain/pull/960/commits/e1b924f7481f848f1bf24a76019a5e1a9013d21c#diff-1a216abd7dbddbbc9b8c42908b044ccf7c982ac2daba515ea6ef10135358aecd)). 

see `validateInnerMsg` for [more details ](https://github.com/dydxprotocol/v4-chain/blob/0c4e1d9771d388fd547e8f1fb2c5a1b41b98d79b/protocol/lib/ante/nested_msg.go#L53C1-L74). Essentially we disallow inner messages to be 
1. unsupported (some ICA controller messages)
2. app injected
3. another nested message

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
 		// For "internal msgs", we allow them, because they are designed to be nested.
 	}
 	return nil
 }
+
+// getInnerMsgs returns the inner msgs of the given msg.
+func getInnerMsgs(msg sdk.Msg) ([]sdk.Msg, error) {
+	switch msg := msg.(type) {
+	case
+		*gov.MsgSubmitProposal:
```
BrendanChou, line 93: style nit: put on the same line

protocol/app/app.go
```diff
@@ -1051,6 +1074,7 @@ func New(
 	// NOTE: staking module is required if HistoricalEntries param > 0
 	app.ModuleManager.SetOrderBeginBlockers(
 		blocktimemoduletypes.ModuleName, // Must be first
+		authz.ModuleName,                // Delete expired grants.
```
ttl33, line 1077: 💯

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
 	"fmt"
+	"strings"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	authz "github.com/cosmos/cosmos-sdk/x/authz"
 	gov "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 )
 
+const DYDX_MSG_PREFIX = "/dydxprotocol"
```
ttl33, line 13: mega nit: might be better to use the [AppName](https://github.com/dydxprotocol/v4-chain/blob/4357938804d73e46bf86ab16c28bb983720c2bcd/protocol/app/constants.go#L4)

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
 			return fmt.Errorf("Invalid nested msg: app-injected msg type")
 		}
 
 		// 3. double-nested msgs.
-		if IsNestedMsg(msg) {
+		if IsNestedMsg(inner) {
 			return fmt.Errorf("Invalid nested msg: double-nested msg type")
 		}
 
+		// 4.Reject nested dydxprotocol messages in `MsgExec`.
```
ttl33, line 72:
```suggestion
		// 4. Reject nested dydxprotocol messages in `MsgExec`.
```

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
+		if _, ok := msg.(*authz.MsgExec); ok {
+			metrics.IncrCountMetricWithLabels(
```
ttl33, line 74: 💯

protocol/lib/ante/nested_msg_test.go
```diff
@@ -85,6 +85,18 @@ func TestValidateNestedMsg(t *testing.T) {
 			msg:         testmsgs.MsgSubmitProposalWithDoubleNestedInner,
 			expectedErr: invalidInnerMsgErr_Nested,
 		},
+		"Invalid MsgExec: unsupported inner msg": {
```
ttl33, line 88: should we add a testcase for the dydx custom msg types?

protocol/lib/ante/nested_msg_test.go
```diff
@@ -85,6 +85,18 @@ func TestValidateNestedMsg(t *testing.T) {
...
+			msg:         &testmsgs.MsgExecWithAppInjectedInner,
+			expectedErr: invalidInnerMsgErr_AppInjected,
+		},
+		"Invalid MsgExec: double-nested inner msg": {
+			msg:         &testmsgs.MsgExecWithDoubleNestedInner,
+			expectedErr: invalidInnerMsgErr_Nested,
+		},
 		"Valid: empty inner msg": {
 			msg:         testmsgs.MsgSubmitProposalWithEmptyInner,
 			expectedErr: nil,
```
ttl33, line 102: nit: should we add test cases for `IsDydxMsg` method?

ttl33 (approved):

protocol/lib/ante/nested_msg_test.go
teddyding: Agree this would be valuable to have 👍

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
+				metrics.Ante,
+				metrics.MsgExec,
+				metrics.GetLabelForStringValue(metrics.InnerMsg, sdk.MsgTypeURL(inner)),
+			)
+			if IsDydxMsg(inner) {
```
teddyding, line 79: Does this mean we don't allow `MsgExec` to call any of the dYdX messages (incl. order placing, subaccount transfer)? I can see why that would make sense to limit security risk. 

Is there any context/discussion behind this? If so could you document briefly?

protocol/lib/ante/nested_msg.go
```diff
@@ -2,73 +2,99 @@ package ante
...
+	authz "github.com/cosmos/cosmos-sdk/x/authz"
 	gov "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 )
 
+const DYDX_MSG_PREFIX = "/dydxprotocol"
+
 // IsNestedMsg returns true if the given msg is a nested msg.
 func IsNestedMsg(msg sdk.Msg) bool {
 	switch msg.(type) {
```
teddyding, line 17: Question: do we need `MsgDelayMsg` here?

protocol/lib/ante/nested_msg.go
jayy04: ah there was a comment discussion in the authz one pager that got resolved. I added a short section in the doc saying that dydx messages will be explicitly disallowed for now and an allowlist can be maintained in the future if there are use cases.

protocol/lib/ante/nested_msg.go
jayy04: good callout. I think so since it also wraps other messages. 

Let's add this in a separate PR since it's unrelated. I also have less context on delay messages so maybe one of the OTE eng can pick this up?

protocol/testing/e2e/authz/authz_test.go
```diff
@@ -0,0 +1,416 @@
...
+				constants.Bob_Num0_100_000USD,
+			},
+
+			msgGrant: nil,
+
+			msgExec: &authz.MsgExec{
+				Grantee: constants.BobAccAddress.String(),
+				Msgs: []*codectypes.Any{
+					newAny(
+						&clobtypes.MsgProposedOperations{},
```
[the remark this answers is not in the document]
jayy04: it's both a app injected and also dydx message. we check for app injected before dydx messages so it will get rejected before even reaching dydx message check.

PR #962, opened 2024-01-12
[CORE-834] Wire x/ratelimit methods to IBC transfer logic

teddyding:
### Changelist
- Implement x/ratelimit IBC Middleware which wraps around native IBC transfer module
- Implement additional keeper methods for dealing with IBC packets

The overall architecture of the IBC middleware resembles closely to Stride's `x/ratelimit` [implementation](https://github.com/Stride-Labs/stride/blob/main/x/ratelimit/README.md). The `x/ratelimit` module [spec](https://www.notion.so/dydx/x-ratelimit-Global-Withdrawal-Speed-Limit-f7afcda6413f4298a42700f0e4a360ca?pvs=4#5908494b6b7147be9a113a3aaa9a24d3) includes a high level summary.

Many functions were re-used or re-written from Stride implementation. I left TODO items to for giving generally better attributions. 

### Test Plan
Some unit tests.
More tests to come in below tickets:
- https://linear.app/dydx/issue/CORE-856/unit-test-coverage-for-packetgo Improve unit test coverage
- https://linear.app/dydx/issue/CORE-855/e2e-testing-of-ratelimit-ibc-middleware E2E testing of IBC Middleware

### Author/Reviewer Checklist


protocol/x/ratelimit/util/ibc_test.go
```diff
@@ -0,0 +1,232 @@
+package util_test
```
teddyding, line 1: Unit tests most re-used from https://github.com/Stride-Labs/stride/blob/121f2ac5d2e5f8e406f89999410a49ea4277a552/x/ratelimit

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
+package util
```
teddyding, line 1: Utilities mostly re-used from https://github.com/Stride-Labs/stride/blob/121f2ac5d2e5f8e406f89999410a49ea4277a552/x/ratelimit

protocol/x/ratelimit/types/expected_keepers.go
```diff
@@ -15,3 +18,23 @@ type BankKeeper interface {
 type BlockTimeKeeper interface {
 	GetTimeSinceLastBlock(ctx sdk.Context) time.Duration
 }
+
+// ICS4Wrapper defines the expected ICS4Wrapper for middleware
+type ICS4Wrapper interface {
```
ttl33, line 23: noob question: does this interface exist somewhere already like ibc-go? if so, why do we need to re-define it here?

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
...
+// Parse the denom from the Recv Packet that will be used by the rate limit module
+// The denom that the rate limiter will use for a RECEIVE packet depends on whether it was a source or sink
+//
+//	Sink:   The token moves forward, to a chain different than its previous hop
+//	        The new port and channel are APPENDED to the denom trace.
+//	        (e.g. A -> B, B is a sink) (e.g. A -> B -> C, C is a sink)
+//
+//	Source: The token moves backwards (i.e. revisits the last chain it was sent from)
+//	        The port and channel are REMOVED from the denom trace - undoing the last hop.
+//	        (e.g. A -> B -> A, A is a source) (e.g. A -> B -> C -> B, B is a source)
```
ttl33, line 55: Just to make sure I understand this correctly, the `source` here is the source of the last hop?

For example, `A -> B -> C -> D -> E -> F`, the source is `D`? Or is this example of a `Sink`? 

I think thinking the latter but wasnt sure tbh

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
...
+//
+//	If the chain is acting as a SINK: We add on the Stride port and channel and hash it
```
ttl33, line 57: nit: did we still want to have `Stride` in the comment? understand this if this is all copy-paste

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
...
+
+		// Hash the denom trace
+		denomTrace := ibctransfertypes.ParseDenomTrace(prefixedDenom)
+		denom = denomTrace.IBCDenom()
+	}
+
+	return denom
+}
+
+func ParsePacketInfo(
```
ttl33, line 109: should we add the comments?

https://github.com/Stride-Labs/stride/blob/eb3564c70cc4024caf73b3b8f7506a2716617573/x/ratelimit/keeper/packet.go#L115-L126

protocol/x/ratelimit/util/ibc.go
ttl33: Can you separate these util functions into two files:
1. exact copy from Stride
2. methods that made some modifications

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
...
+// Parse the denom from the Send Packet that will be used by the rate limit module
+// The denom that the rate limiter will use for a SEND packet depends on whether
+// it was a NATIVE token (e.g. ustrd, stuatom, etc.) or NON-NATIVE token (e.g. ibc/...)...
+//
+// We can identify if the token is native or not by parsing the trace denom from the packet
+// If the token is NATIVE, it will not have a prefix (e.g. ustrd),
+// and if it is NON-NATIVE, it will have a prefix (e.g. transfer/channel-2/uosmo)
+//
+// For NATIVE denoms, return as is (e.g. ustrd)
+// For NON-NATIVE denoms, take the ibc hash (e.g. hash "transfer/channel-2/usoms" into "ibc/...")
```
ttl33, line 30: IIUC, if the denom was IBC'd multiple hops, it has multiple prefixes. This is my interpretation from [this doc](https://tutorials.cosmos.network/academy/3-ibc/7-token-transfer.html)
> We can send assets (or their IBC voucher representation) in multiple hops across multiple chains. Every single time the path will be prepended with the port/channel-id/... prefix. When sending this IBC denom (having had multiple hops) back to its source chain, for every hop back one port/channel-id/... prefix will be taken off. This results in a return to the original denom if all the hops are reversed.

Does this remove all the prefixes for multi-hops? If not, does that cause an issue where the same tokens are considered different if the IBC path is different from capacity calculation perspective?

protocol/x/ratelimit/util/ibc.go
```diff
@@ -0,0 +1,191 @@
...
+		ChannelID: channelID,
+		Denom:     denom,
+		Amount:    amount,
+	}
+
+	return packetInfo, nil
+}
+
+// UnpackAcknowledgementResponseForTransfer unmarshals Acknowledgements for IBC transfers, determines the status of the
+// acknowledgement (success or failure), and, if applicable, assembles the message responses
```
ttl33, line 151:
```suggestion
// acknowledgement (success or failure).
```

protocol/x/ratelimit/util/ibc.go
teddyding: Done - moved exact copies and tests in to file `parse_denom.go` and `parse_denom_test.go`

protocol/app/app.go
```diff
@@ -604,20 +591,39 @@ func New(
...
 		// set the governance and delaymsg module accounts as the authority for conducting upgrades
 		[]string{
 			lib.GovModuleAddress.String(),
 			delaymsgmoduletypes.ModuleAddress.String(),
 		},
 	)
 
-	// TODO(CORE-834): Add ratelimitKeeper to the IBC transfer stack.
+	// Create Transfer Keepers
+	app.TransferKeeper = ibctransferkeeper.NewKeeper(
```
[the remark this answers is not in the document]
teddyding: The `TransferKeeper` needs to take in `app.RateLimitKeeper` so we needed to move its declaration after `RateLimitKeeper`.

protocol/app/app.go
```diff
@@ -604,20 +591,39 @@ func New(
...
+		app.BankKeeper,
+		scopedIBCTransferKeeper,
+		lib.GovModuleAddress.String(),
+	)
+	transferModule := transfer.NewAppModule(app.TransferKeeper)
+	transferIBCModule := transfer.NewIBCModule(app.TransferKeeper)
+
+	// Wrap the x/ratelimit middlware over the IBC Transfer module
+	var transferStack ibcporttypes.IBCModule = transferIBCModule
+	transferStack = ratelimitmodule.NewIBCMiddleware(app.RatelimitKeeper, transferStack)
```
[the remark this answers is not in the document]
teddyding: Prefer `transferStack`. It refers to the collection of base transfer IBC module + all middleware on top of it. As an example, see `ibc-go` example [here](https://github.com/cosmos/ibc-go/blob/04531d83bf51d949f4bf2d61e88e419d1620ea32/docs/docs/04-middleware/02-callbacks/06-gas.md?plain=1#L29-L32) where this stack contains multiple middlewares

protocol/x/ratelimit/ibc_middleware.go
```diff
@@ -0,0 +1,203 @@
+package ratelimit
+
+// TODO(CORE-854): Improve attribution message.
```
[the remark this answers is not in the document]
teddyding: Generally need to do attribution better since a lot of code is reused from Stride: https://en.wikipedia.org/wiki/Attribution_(copyright)

I have a call with legal team next week to align on what needs to be done here

protocol/x/ratelimit/ibc_middleware.go
```diff
@@ -0,0 +1,203 @@
...
+func (im IBCMiddleware) OnChanOpenInit(ctx sdk.Context,
+	order channeltypes.Order,
+	connectionHops []string,
+	portID string,
+	channelID string,
+	channelCap *capabilitytypes.Capability,
+	counterparty channeltypes.Counterparty,
+	version string,
+) (string, error) {
+	return im.app.OnChanOpenInit(
```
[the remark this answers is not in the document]
teddyding: Because rate-limiting logic doesn't require customized logic at channel initialization/opening. I added comments to clarify this.

protocol/x/ratelimit/ibc_middleware.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		im.keeper.Logger(ctx).Error(fmt.Sprintf("ICS20 receive packet could not be parsed: %s", err.Error()))
+		return channeltypes.NewErrorAcknowledgement(err)
+	}
+
+	// Process deposit of incoming asset.
+	// If the `Recv` fails on the underlying transfer stack (e.g. Reciver is invalid), the state
+	// change is not committed.
+	// TODO(CORE-855): Add an E2E test for this.
+	im.keeper.ProcessDeposit(ctx, ibcTransferPacketInfo.Denom, ibcTransferPacketInfo.Amount)
```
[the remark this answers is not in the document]
teddyding: > IIUC, this simply updates the capacity. Where does the capacity check happen?

`OnRecvPacket` only processed inbound transfer so we only increase capacity here, and will never rate-limit inbound transfers. 

> Also, can you explain why the state change is not committed?

`OnRecvPacket` is invoked during message handling of `[MsgRecvPacket](https://github.com/cosmos/ibc-go/blob/04531d83bf51d949f4bf2d61e88e419d1620ea32/modules/core/keeper/msg_server.go#L447)` which is a txn sent by the relayer. If txn fails, states are not committed just like any other txns.

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.PendingSendPacketPrefix))
+	key := types.GetPendingSendPacketKey(channelId, sequence)
+
+	return store.Has(key)
+}
+
+// Middleware implementation for OnAckPacket
+// It is called on the sender chain when a relayer relays back the acknowledgement from the receiver chain.
+// On the dYdX chain, this includes the “response” of the receiver chain for outbound transfer from dYdX.
+func (k Keeper) AcknowledgeIBCTransferPacket(
```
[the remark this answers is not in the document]
teddyding: `Ack` is the final step of an IBC transfer - see full life cycle [here](https://tutorials.cosmos.network/academy/3-ibc/7-token-transfer.html#transfer-packet-flow).

If the acknowledgement indicates a failure on the receiver chain then we revert the capacity decrease. Otherwise, we simply complete the flow by removing "pending" packet. 

Similar to `OnRecvPacket`, `Acknowledge` is called when relayer submits an acknowledgement as part of a transaction. All state changes are committed only if the underlying `Acknowledge` logic of the transfer application succeeds as well

protocol/x/ratelimit/ibc_middleware.go
```diff
@@ -0,0 +1,203 @@
...
+
+// TODO(CORE-855): Add e2e testing for the x/ratelimit IBC Middlware
+func NewIBCMiddleware(k keeper.Keeper, app porttypes.IBCModule) IBCMiddleware {
+	return IBCMiddleware{
+		app:    app,
+		keeper: k,
+	}
+}
+
+// OnChanOpenInit implements the IBCMiddleware interface
```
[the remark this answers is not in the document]
teddyding: Added more detailed description on these functions.

protocol/x/ratelimit/types/expected_keepers.go
teddyding: The interface does, but here we only define the subset of the interface used by the `ratelimitKeeper`. It's the same pattern as all other keeper interfaces defined within this `expected_keepers.go` file

protocol/x/ratelimit/util/ibc.go
teddyding: Done

protocol/x/ratelimit/util/ibc.go
teddyding: For each IBC transfer from `A -> B`, exactly one of them acts as source chain and the other acts as sink chain. 
The `ibc-go` repo has a more clear explanation that I like: https://github.com/cosmos/ibc-go/blob/04531d83bf51d949f4bf2d61e88e419d1620ea32/modules/apps/transfer/keeper/relay.go#L23-L54

teddyding: I edited the documentation to clarify

protocol/x/ratelimit/util/ibc.go
teddyding: Think it makes more sense to adapt documentation in the context of dYdX chain. Obviously, as part of attribution/licensing we will still make it clear these functions came from Stride implementation

protocol/x/ratelimit/util/ibc.go
teddyding: Not exactly sure you mean by "Does this remove all the prefixes for multi-hops", but I believe the following is relevant to your question:


Consider the two different paths that `uusdc` travel to dYdX chain:

1. Noble -> dYdX Chain 
2. Noble -> Osmosis -> dYdX Chain

The resulting IBC denom on dYdX Chain is different, and therefore non-fungible, by design (more discussion [here](https://dydx-team.slack.com/archives/C063PG2P551/p1699377562092449?thread_ts=1699377113.925409&cid=C063PG2P551)). 
- Only path 1 uusdc will be accepted as collateral/received by users from trading. 
- Path 2 uusdc has NO significance on the dYdX chain. 

Therefore in the context of rate-limiting, only path 1 uusdc relevant for the ratelimit.

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+	channelId string,
+	sequence uint64,
+	denom string,
+	amount *big.Int,
+) {
+	if !k.HasPendingSendPacket(ctx, channelId, sequence) {
+		return
+	}
+	// Undo'ing a withdrawal is equivalent to processing a deposit.
+	k.ProcessDeposit(ctx, denom, amount)
```
BrendanChou, line 104: Does calling this function have any negative downstream effects? For example logging or metrics that indicate a deposit happened?

Wonder if it's better to have an explicit `UndoWithdrawal` function to prevent these issues (now or in the future if code is modified/refactored)

BrendanChou (approved):

protocol/x/ratelimit/keeper/packet_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require.False(t, k.HasPendingSendPacket(ctx, "non-existent-channel", testSequence))
+	require.False(t,
+		k.HasPendingSendPacket(
+			ctx, testChannelId,
+			42, // non-existent sequence number
+		),
+	)
+}
+
+// TODO(CORE-856): Improve coverage for remaining functions in packet.go
```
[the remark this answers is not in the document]
teddyding: Fixed

protocol/x/ratelimit/keeper/packet.go
teddyding: Good point, I renamed `ProcessDeposit` to `IncrementCapacitiesForDenom` and wrapped `UndoWithdrawal` around it (and added metrics)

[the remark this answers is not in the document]
teddyding: Makes sense - see reply [here](https://github.com/dydxprotocol/v4-chain/pull/962/files#r1463516314)

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("ICS20 packet send was denied: %s", err.Error()))
+		return 0, err
+	}
+	return sequence, err
+}
+
+// Middleware implementation for SendPacket with rate limiting
+// Checks whether the rate limit has been exceeded - and if it hasn't, sends the packet
+func (k Keeper) SendRateLimitedPacket(ctx sdk.Context, packet channeltypes.Packet) error {
```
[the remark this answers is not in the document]
teddyding: Intention is to cover this function with unit tests (CORE-856)

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+func (k Keeper) SendPacket(
+	ctx sdk.Context,
+	channelCap *capabilitytypes.Capability,
+	sourcePort string,
+	sourceChannel string,
+	timeoutHeight clienttypes.Height,
+	timeoutTimestamp uint64,
+	data []byte,
+) (sequence uint64, err error) {
+	// The packet must first be sent up the stack to get the sequence number from the channel keeper
```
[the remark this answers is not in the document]
teddyding: "Up" is indeed correct here. There are two directions of control flow in IBC: 
1) base application -> middlewares -> IBC core (e.g. SendPacket)
2) IBC core -> middlewares -> base application (callbacks like `OnRecvPacket`) 
Conventionally base application is considered the bottom layer. See a helpful chart [here](https://ibc.cosmos.network/main/ibc/middleware/overview.html)

So "up the stack here" means sending the packet up via IBC core

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+	sequence, err = k.ics4Wrapper.SendPacket(
```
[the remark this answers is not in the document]
teddyding: For each `MsgTransfer`, this function is called exactly once. The code just needs to call IBC core (top layer in this stack) to get this sequence number and then process middleware related logic. This is a bit unintuitive, but all these state changes happen atomically.

protocol/x/ratelimit/types/ibc.go
```diff
@@ -0,0 +1,34 @@
...
+	AckResponseStatus_NOT_SPECIFIED AckResponseStatus = iota
+	AckResponseStatus_SUCCESS
+	AckResponseStatus_TIMEOUT
+	AckResponseStatus_FAILURE
+)
+
+// AcknowledgementResponse contains information about an acknowledgement of IBC transfer packet.
+type AcknowledgementResponse struct {
+	Status AckResponseStatus
+	Error  string
```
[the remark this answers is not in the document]
teddyding: Because this field is intended to track the IBC [Acknowledgement error field,](https://github.com/cosmos/ibc-go/blob/97ea045855e76177bd92f6112209fb4d038086cd/proto/ibc/core/channel/v1/channel.proto#L169) which is a string.

BrendanChou (approved):

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+	channelId string,
+	sequence uint64,
+	denom string,
+	amount *big.Int,
+) {
+	if !k.HasPendingSendPacket(ctx, channelId, sequence) {
+		return
+	}
+	// Undo'ing a withdrawal is equivalent to processing a deposit.
+	k.IncrementCapacitiesForDenom(ctx, denom, amount)
```
ttl33, line 104: nit: should we call `UndoWithdrawal` instead?

protocol/x/ratelimit/keeper/packet.go
```diff
@@ -0,0 +1,181 @@
...
+
+// Remove a pending packet sequence number from the store
+// Used after the ack or timeout for a packet has been received
+func (k Keeper) RemovePendingSendPacket(ctx sdk.Context, channelId string, sequence uint64) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.PendingSendPacketPrefix))
+	key := types.GetPendingSendPacketKey(channelId, sequence)
+	store.Delete(key)
+}
+
+// Sets the sequence number of a packet that was just sent
```
ttl33, line 30:
```suggestion
// Sets a pending packet sequence number in the store
// Used when packet is received
```

ttl33 (approved):

PR #963, opened 2024-01-12
Indexer account for reduce only resize errors

jonfung-dydx:
Datadog error [here](https://app.datadoghq.com/logs?query=env%3Adev%20status%3Aerror&agg_q=status%2Cservice&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&sort_m=%2C&sort_t=%2C&stream_sort=desc&top_n=10%2C10&top_o=top%2Ctop&viz=pattern&x_missing=true%2Ctrue&from_ts=1704995802250&to_ts=1705082202250&live=true) after running load tester

Account for the specific error type


Roockeeeet (commented):
.diff

PR #970, opened 2024-01-17
Remove basePositionNotional from Indexer APIs

dydxwill:
### Changelist
Remove basePositionNotional from Indexer APIs

### Test Plan
unit tested.

### Author/Reviewer Checklist


indexer/packages/postgres/src/db/migrations/migration_files/20240117120410_remove_unused_liquidity_tiers_columns.ts
```diff
@@ -0,0 +1,17 @@
...
+    .alterTable('liquidity_tiers', (table) => {
+      table.dropColumn('basePositionNotional');
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+  return knex
+    .schema
+    .alterTable('liquidity_tiers', (table) => {
+      table.decimal('basePositionNotional', null).notNullable();
```
vincentwschau, line 15: I don't think this would work on a table with existing rows.

dydxwill: hmm it seems like there's precedent for not adding a default here: https://github.com/dydxprotocol/v4-chain/blob/main/indexer/packages/postgres/src/db/migrations/migration_files/20230905160352_remove_min_order_base_quantums.ts and in other places.

I think without adding any default logic, the default is auto set to 0.0?

dydxwill: added a default anyways.

vincentwschau (approved):

PR #984, opened 2024-01-18
[CORE-538] Reduce usage of basic_manager.ModuleBasics in favor of using module.NewBasicManagerFrom...

lcwik:
### Changelist
[CORE-538] Reduce usage of basic_manager.ModuleBasics in favor of using module.NewBasicManagerFrom...

### Test Plan
Minor fixes to existing tests.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/cmd/dydxprotocold/cmd/root.go
+++ b/protocol/cmd/dydxprotocold/cmd/root.go
@@ -35,11 +35,9 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/auth/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/cosmos/cosmos-sdk/x/crisis"
-	"github.com/cosmos/cosmos-sdk/x/genutil"
 	genutilcli "github.com/cosmos/cosmos-sdk/x/genutil/client/cli"
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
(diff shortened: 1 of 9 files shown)


protocol/cmd/dydxprotocold/cmd/root.go
```diff
@@ -85,12 +83,24 @@ func NewRootCmdWithInterceptors(
 	appConfigInterceptor func(string, *DydxAppConfig) (string, *DydxAppConfig),
 	appInterceptor func(app *dydxapp.App) *dydxapp.App,
 ) *cobra.Command {
-	encodingConfig := dydxapp.GetEncodingConfig()
+	tempApp := dydxapp.New(
```
jonfung-dydx, line 86: why is there a temp app?

jonfung-dydx (approved):

protocol/cmd/dydxprotocold/cmd/root.go
lcwik: This is how Cosmos recommends to build the v1 app https://docs.cosmos.network/v0.50/build/migrations/upgrading#autocli
https://github.com/cosmos/cosmos-sdk/blob/v0.50.0-beta.0/simapp/app.go#L636-L655

Effectively you have to build an instance of the app to build another instance of the app again.

It seems as though if we used depinject (aka v2 type app) we wouldn't have to do this.

PR #993, opened 2024-01-19
[CORE-538] Remove genesis check that isn't needed since the issue was with how testapp was simulating block processing.

lcwik:
### Changelist
[CORE-538] Remove genesis check that isn't needed since the issue was with how testapp was simulating block processing.

Also clean-up TODO comments that aren't important and are low value.

### Test Plan
No changes to tests, existing tests pass.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/abci.go
+++ b/protocol/x/clob/abci.go
@@ -136,10 +136,6 @@ func PrepareCheckState(
 		log.BlockHeight, ctx.BlockHeight()+1,
 	)
 
-	// TODO(CORE-538): Is there a cleaner way to check to see if this is genesis?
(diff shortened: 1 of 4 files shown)


protocol/x/clob/abci.go
```diff
@@ -136,10 +136,6 @@ func PrepareCheckState(
 		log.BlockHeight, ctx.BlockHeight()+1,
 	)
 
-	// TODO(CORE-538): Is there a cleaner way to check to see if this is genesis?
```
lcwik, line 139: I added this genesis check to get some tests working during the upgrade to cosmos 0.50 but it isn't necessary.

PR #996, opened 2024-01-22
[CLOB-1063] optimize conditional order trigger runtime

jayy04:
### Changelist
- implement max and min heap for untriggered conditional orders

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/untriggered_conditional_orders.go
+++ b/protocol/x/clob/keeper/untriggered_conditional_orders.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"container/heap"
 	"fmt"
 	"math/big"
 
@@ -25,12 +26,12 @@ type UntriggeredConditionalOrders struct {
 	// All untriggered take profit buy orders and stop loss sell orders sorted by time priority.
 	// These orders will be triggered when the oracle price goes lower than or equal to the trigger price.
 	// This array functions like a max heap.
-	OrdersToTriggerWhenOraclePriceLTETriggerPrice []types.Order
+	OrdersToTriggerWhenOraclePriceLTETriggerPrice types.MaxConditionalOrderHeap
 
 	// All untriggered take profit sell orders and stop loss buy orders sorted by time priority.
 	// These orders will be triggered when the oracle price goes greater than or equal to the trigger price.
 	// This array functions like a min heap.
-	OrdersToTriggerWhenOraclePriceGTETriggerPrice []types.Order
+	OrdersToTriggerWhenOraclePriceGTETriggerPrice types.MinConditionalOrderHeap
 }
 
 func (k Keeper) NewUntriggeredConditionalOrders() *UntriggeredConditionalOrders {
@@ -94,13 +95,13 @@ func (untriggeredOrders *UntriggeredConditionalOrders) AddUntriggeredConditional
 
 	if order.IsTakeProfitOrder() {
 		if order.IsBuy() {
-			untriggeredOrders.OrdersToTriggerWhenOraclePriceLTETriggerPrice = append(
-				untriggeredOrders.OrdersToTriggerWhenOraclePriceLTETriggerPrice,
+			heap.Push(
+				&untriggeredOrders.OrdersToTriggerWhenOraclePriceLTETriggerPrice,
 				order,
 			)
 		} else {
-			untriggeredOrders.OrdersToTriggerWhenOraclePriceGTETriggerPrice = append(
-				untriggeredOrders.OrdersToTriggerWhenOraclePriceGTETriggerPrice,
+			heap.Push(
+				&untriggeredOrders.OrdersToTriggerWhenOraclePriceGTETriggerPrice,
 				order,
 			)
 		}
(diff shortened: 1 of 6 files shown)


protocol/x/clob/types/untriggered_conditional_orders.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (h MinConditionalOrderHeap) Len() int {
+	return len(h)
+}
+
+func (h MinConditionalOrderHeap) Less(i, j int) bool {
+	x, y := h[i], h[j]
+
+	// If the trigger subticks are the same, sort by order hash.
+	// This is required for determinism in the case of multiple orders with the same trigger subticks.
```
jayy04, line 20: I think technically we don't need this since orders will get added deterministically in the same order, but I added this to be 100% safe. Let me know if anyone disagrees

protocol/x/clob/keeper/untriggered_conditional_orders.go
```diff
@@ -25,12 +26,12 @@ type UntriggeredConditionalOrders struct {
 	// All untriggered take profit buy orders and stop loss sell orders sorted by time priority.
 	// These orders will be triggered when the oracle price goes lower than or equal to the trigger price.
 	// This array functions like a max heap.
-	OrdersToTriggerWhenOraclePriceLTETriggerPrice []types.Order
+	OrdersToTriggerWhenOraclePriceLTETriggerPrice types.MaxConditionalOrderHeap
 
 	// All untriggered take profit sell orders and stop loss buy orders sorted by time priority.
 	// These orders will be triggered when the oracle price goes greater than or equal to the trigger price.
 	// This array functions like a min heap.
```
jonfung-dydx, line 33: Update comments given that this is a heap?

protocol/x/clob/types/untriggered_conditional_orders.go
```diff
@@ -0,0 +1,83 @@
+package types
```
jonfung-dydx, line 1: `untriggered_conditional_orders_heap.go`?

protocol/x/clob/types/untriggered_conditional_orders.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (h *MaxConditionalOrderHeap) Push(x interface{}) {
+	*h = append(*h, x.(Order))
+}
+
+func (h *MaxConditionalOrderHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
```
jonfung-dydx, line 81: the official documentation for `container/heap` has some safety checks here for memory:
```
func (pq *PriorityQueue) Pop() any {
	old := *pq
	n := len(old)
	item := old[n-1]
	old[n-1] = nil  // avoid memory leak
	item.index = -1 // for safety
	*pq = old[0 : n-1]
	return item
}
```

Should we do this?

jonfung-dydx: Also, when we poll the order we look at the 0th element in the queue, as seen [here](https://github.com/dydxprotocol/v4-chain/pull/996/files#diff-fc56c9715292486eb043611b0cfee5372bc9ca67921ffac388d175e504b4ac01R254). However, for both the min and max heap we are returning the last element in the array (old[n-1]). Is that correct? Don't we want to poll/pop the same element?

BrendanChou: For your second comment, I was similarly confused.

The confusing part is that `heap.Pop(h)` does not simply call `h.Pop()` under the hood. Whereas `h.Pop()` returns the lowest priority element, `heap.Pop(h)` returns the highest priority element. This is what is extremely confusing. The logic of `heap.Pop(h)` is inessence:
1. Call `h.Pop()` to remove the lowest priority element
1. Replace the highest priority element with the popped element
1. Call `heap.Fix(h, 0)` to re-sort the replaced item
1. Return the original highest-priority element that you removed

[Here's](https://cs.opensource.google/go/go/+/master:src/container/heap/heap.go) the source code in case you want to read line-by-line and understand.

BrendanChou: You could consider adding an `h.Peek()` method which simply returns `h[0]`

protocol/x/clob/types/untriggered_conditional_orders_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+		h := types.MaxConditionalOrderHeap(orders)
+		heap.Init(&h)
+
+		heap.Push(&h, constants.ConditionalOrder_Alice_Num0_Id0_Clob0_Buy1BTC_Price50000_GTBT10_SL_50300)
+		heap.Push(&h, constants.ConditionalOrder_Alice_Num0_Id0_Clob0_Buy1BTC_Price50000_GTBT10_TP_49700)
+
+		actual := make([]types.Order, 0)
+		for h.Len() > 0 {
+			actual = append(actual, heap.Pop(&h).(types.Order))
```
BrendanChou, line 80: you probably want a test that `heap.Pop` throws an error when the length of the structure is zero

protocol/x/clob/keeper/untriggered_conditional_orders.go
```diff
@@ -215,6 +216,10 @@ func (untriggeredOrders *UntriggeredConditionalOrders) RemoveUntriggeredConditio
 		}
 	}
 	untriggeredOrders.OrdersToTriggerWhenOraclePriceGTETriggerPrice = newOrdersToTriggerWhenOraclePriceGTETriggerPrice
+
```
lucas-dydx, line 219: should funcs like these have timing stats? since they're now heap-sorting all conditional orders (although it should be fast since in-memory)

PR #1003, opened 2024-01-23
Migrate more logs to new logging lib, add `lib.UnwrapSDKContext`

jonfung-dydx:
1. Introduce `lib.UnwrapSDKContext`. This should be used instead of the cosmos sdk context unwrapper because it tags all logs with `x/<modulename>` if logs are generated with the new logging library.
2. Migrate some instances of old logging to new logging library. The migrated callsites that use `ErrorLogWithError` will now be properly data dog error tracked.

Note that now, instead of obtaining the logger with `keeper.Logger()`, in order to have log tags that contain  `x/<modulename>`, it is imperative that all code entrypoints that use `UnwrapSDKContext` use this new function.


protocol/lib/context.go
```diff
@@ -1,12 +1,35 @@
...
 
 func GetTxHash(tx []byte) TxHash {
 	return TxHash(fmt.Sprintf("%X", tmhash.Sum(tx)))
 }
+
+// UnwrapSDKContext is a thin wrapper around cosmos sdk's unwrap function
+// that extracts the cosmos context from the standard golang context.
+// If moduleName is provided, it appends the persistent log tag with
+// the module name to the logger in the context.
+func UnwrapSDKContext(
```
jonfung-dydx, line 22: This function is intended to replace the cosmos sdk `UnwrapSDKContext` function. It accepts an additional parameter for module name and any logs generated with the context will have be tagged with module name.

protocol/x/clob/keeper/final_settlement.go
```diff
@@ -29,11 +30,14 @@ func (k Keeper) mustCancelStatefulOrdersForFinalSettlement(ctx sdk.Context, clob
...
+					ctx,
+					"mustCancelStatefulOrdersForFinalSettlement: Failed to remove stateful order",
+					"orderId",
 					orderId,
-					r,
+					"error",
+					err,
 				)
 			}
 		}()
```
jayy04, line 43: unrelated to this PR - but seems like we do this a lot in many places. don't have numbers to back me up but these can be costly if they are in hot code paths. 

created [this](https://linear.app/dydx/issue/CLOB-1065/see-if-recovering-from-panics-in-defer-func-is-costly-and-potentially) ticket to do some analysis and audit panics.

protocol/x/epochs/module.go
```diff
@@ -145,7 +147,8 @@ func (AppModule) ConsensusVersion() uint64 { return 1 }
 
 // BeginBlock executes all ABCI BeginBlock logic respective to the epochs module.
 func (am AppModule) BeginBlock(ctx context.Context) error {
+	sdkCtx := lib.UnwrapSDKContext(ctx, types.ModuleName)
```
jayy04, line 150: nit: move this to below the latency metric

jayy04 (approved):

protocol/x/epochs/module.go
jonfung-dydx: Will do in second pr, got another one coming up

PR #1016, opened 2024-01-26
Implement x/govplus MsgSlashValidator

roy-dydx:
### Changelist
- Implement MsgSlashValidator for community enabled slashing

### Test Plan
- Wrote tests for correct validation of some fields.
- Tested on a local network

### Author/Reviewer Checklist


protocol/x/govplus/keeper/keeper_test.go
```diff
@@ -1,851 +0,0 @@
-package keeper_test
```
roy-dydx, line 1: Delete. Accidentally copied from x/ratelimit earlier.

protocol/x/govplus/types/keys.go
```diff
@@ -6,5 +6,5 @@ const (
 	ModuleName = "govplus"
 
 	// StoreKey defines the primary module store key
-	StoreKey = ModuleName
+	StoreKey = "dydxgovplus"
```
roy-dydx, line 9: Store keys can't be prefixes of another apparently, so I can't use govplus with gov.

proto/dydxprotocol/govplus/tx.proto
```diff
@@ -1,7 +1,58 @@
...
+  // adjusted by power reduction to get consensus power.
+  bytes power_at_infraction_height = 4 [
+    (gogoproto.customtype) =
+        "github.com/dydxprotocol/v4-chain/protocol/dtypes.SerializableInt",
+    (gogoproto.nullable) = false
+  ];
+
+  // Multiplier for how much of the validator's stake should be slashed.
+  // slash_factor * power = amount slashed
+  string slash_factor = 5 [
```
teddyding, line 49: For my understanding why is the extra options needed, besides `[(cosmos_proto.scalar) = "cosmos.Dec"]` ? The gov v1 [fields](https://github.com/dydxprotocol/cosmos-sdk/blob/bdf96fddf421e90574e53af383c1468fdb33d7a3/proto/cosmos/gov/v1/gov.proto#L174C21-L174C59) only have this one option, while v1beta fields have more options ([example](https://github.com/dydxprotocol/cosmos-sdk/blob/bdf96fddf421e90574e53af383c1468fdb33d7a3/proto/cosmos/staking/v1beta1/staking.proto#L34-L37))

proto/dydxprotocol/govplus/tx.proto
```diff
@@ -1,7 +1,58 @@
...
+}
+
+// MsgSlashValidator is the Msg/SlashValidator request type.
+message MsgSlashValidator {
+  // The address that controls the module (likely the gov module account).
+  option (cosmos.msg.v1.signer) = "authority";
+  string authority = 1;
+
+  // Address of the validator to slash
+  string validator_address = 2;
```
teddyding, line 25: Explicitly name this as `validator_consensus_address` for disambiguation

proto/dydxprotocol/govplus/tx.proto
```diff
@@ -1,7 +1,58 @@
...
-service Msg {}
+service Msg {
+  // SlashValidator is exposed to allow slashing of a misbehaving validator via
+  // governance.
+  rpc SlashValidator(MsgSlashValidator) returns (MsgSlashValidatorResponse);
+}
+
+// MsgSlashValidator is the Msg/SlashValidator request type.
+message MsgSlashValidator {
+  // The address that controls the module (likely the gov module account).
```
teddyding, line 20:
```suggestion
  // The address that controls the module (the gov module account).
```

proto/dydxprotocol/govplus/tx.proto
teddyding: Should also use relevant options for this field. [Example](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/proto/cosmos/staking/v1beta1/query.proto#L167)?

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,40 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
 }
 
 var _ types.MsgServer = msgServer{}
+
+func (k msgServer) SlashValidator(
+	goCtx context.Context,
+	msg *types.MsgSlashValidator,
```
teddyding, line 27: Are we choosing not to implement `ValidateBasic` for `MsgSlashValidator`? Understand this is optional post 0.50, but are we not doing any validation on the fields? For example, `PowerAtInfractionHeight` should be non-negative

proto/dydxprotocol/govplus/tx.proto
```diff
@@ -1,7 +1,58 @@
...
+  // Colloquially, the height at which the validator is deemed to have
+  // misbehaved. In practice, this is the height used to determine the targets
+  // of the slash. For example, undelegating after this height will not escape
+  // slashing. This height should be set to a recent height at the time of the
+  // proposal to prevent delegators from undelegating during the vote period.
+  //
+  // NB: At the time this message is applied, this height must have occured
+  // equal to or less than an unbonding period in the past in order for the
+  // slash to be effective.
+  uint32 infraction_height = 3;
```
teddyding, line 36: Can we more explicitly document the suggested value for `infraction_height` in a proposal? Is it: `estimated_proposal_tally_height - unbonding_period < infraction_height < proposal_submission_height`, where the lowerbound is needed because `At the time this message is applied, this height must have occured equal to or less than an unbonding period in the past in order for the slash to be effective`.

Also in practice can we just suggest `infraction_height = proposal_submission_height`? Is there any use case where this should not be true?

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,40 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	if err != nil {
+		return nil, types.ErrValidatorAddress
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	_, err = k.stakingKeeper.Slash(
+		ctx,
+		consAddr,
+		int64(msg.InfractionHeight),
```
teddyding, line 47: Nit: Add comment for why casting is safe

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,40 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+		msg.PowerAtInfractionHeight.BigInt().Int64(),
+		msg.SlashFactor,
+	)
+	if err != nil {
+		log.ErrorLogWithError(
+			ctx,
+			"error occured when slashing validator",
+			err,
+		)
+		panic(err)
```
teddyding, line 57: Why panic here instead of returning an error? 
Since `x/gov` message execution doesn't recover from panic, wouldn't this lead to consensus failure?

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,40 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+		return nil, types.ErrValidatorAddress
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	_, err = k.stakingKeeper.Slash(
+		ctx,
+		consAddr,
+		int64(msg.InfractionHeight),
+		msg.PowerAtInfractionHeight.BigInt().Int64(),
```
teddyding, line 48: It seems pretty bareboned that proposer would have to input the voting power themselves. Can we at least calculating the voting power from the delegation coin amount?

From `Slash()` [implementation](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/x/staking/keeper/slash.go#L46-L47), it calculates the amount to check with no additional validation on this figure. If proposer mistakenly put a larger than actual amount, would this lead to any delegator being slashed more than its delegation amount? In general, it seems like we need some E2Es test to cover possible scenarios.

teddyding: The `Slash()` function in `x/staking` was not meant to be exposed for usage by human, and it has a couple of input [contracts](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/x/staking/keeper/slash.go#L20-L36) listed. Can we explicitly check these contracts (if not checked within the function) to make sure there are possible weird cases?

proto/dydxprotocol/govplus/tx.proto
```diff
@@ -1,7 +1,58 @@
...
+  // SlashValidator is exposed to allow slashing of a misbehaving validator via
+  // governance.
+  rpc SlashValidator(MsgSlashValidator) returns (MsgSlashValidatorResponse);
+}
+
+// MsgSlashValidator is the Msg/SlashValidator request type.
+message MsgSlashValidator {
+  // The address that controls the module (likely the gov module account).
+  option (cosmos.msg.v1.signer) = "authority";
+  string authority = 1;
```
jayy04, line 22: Looking at other protos, can we also add `[ (cosmos_proto.scalar) = "cosmos.AddressString" ]` to this field?

proto/dydxprotocol/govplus/tx.proto
roy-dydx: If someone knows they are getting slashed and undelegated before proposal.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: Only errors if catastrophic. I.e if we used the slashing module and it errored equivalently, it would also panic.

roy-dydx: It won’t panic if it’s just a bad address or something.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: You cannot because you can only get the delegation coin amount at the time of execution. If someone sees a proposal to slash them and knows they can avoid it by manipulating their delegation amount, it becomes worthless

protocol/x/govplus/types/keys.go
teddyding: Looks likes break the convention that `StoreKey` = `ModuleName`. Should we just update module name to `dydxgovplus` (and update directory names accordingly)?

proto/dydxprotocol/govplus/tx.proto
teddyding: > If someone knows they are getting slashed and undelegated before proposal.

Makes sense. Could you add some documentation here for how `infraction_height` should be set, including an interval like suggested above?

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,40 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	}
+
+	consAddr, err := sdk.ConsAddressFromBech32(msg.ValidatorAddress)
+	if err != nil {
+		return nil, types.ErrValidatorAddress
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	_, err = k.stakingKeeper.Slash(
```
jayy04, line 44: curious - why do we call `stakingKeeper.Slash` here and not `slashingKeeper.Slash`? slashing calls staking.Slash but also emits some events

protocol/x/govplus/keeper/msg_server.go
jayy04: doesnt slashing module just [return errors](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/x/slashing/keeper/keeper.go#L105-L108)?

teddyding: Wouldn't this cause a panic and halt the chain if [Slash()](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/x/staking/keeper/slash.go#L37) returns any kind of error? For example: negative slash factor, validator is unbonded, etc

teddyding: Actually, in 0.50 gov does recover from [panic](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/x/gov/abci.go#L186), so at least this shouldn't cause a chain halt

protocol/x/govplus/keeper/msg_server.go
teddyding: > Can we at least calculating the voting power from the delegation coin amount?
I meant having user input the total delegated coin amount, as opposed to voting power. The conversion is [trivial](https://github.com/cosmos/cosmos-sdk/blob/b947b02dc63db0a8011a96e217197bf4559dea92/types/staking.go#L35) and can just be done on protocol.

protocol/x/govplus/types/keys.go
BrendanChou: May be worth adding a comment here to reflect that

protocol/x/govplus/keeper/msg_server.go
BrendanChou, line 47: should the proto itself just use `int64/sint64`?

protocol/x/govplus/types/keys.go
roy-dydx: Feel like it's not a particularly worthwhile convention to keep. For example not all cosmos modules even follow the convention: https://github.com/dydxprotocol/cosmos-sdk/blob/dydx-fork-v0.50.3/x/auth/types/keys.go#L12

I'd rather have a nicer module name than try to keep everything the same.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: We use uint32 elsewhere for height.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: No particular reason. We didn't have any use for that event.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: Just [in genesis state.](https://github.com/dydxprotocol/cosmos-sdk/blob/dydx-fork-v0.50.3/proto/cosmos/staking/v1beta1/genesis.proto#L27). I can change this to tokens instead if you still prefer it that way. Let me know

protocol/x/govplus/keeper/msg_server.go
roy-dydx: Implemented this stuff in ValidateBasic

proto/dydxprotocol/govplus/tx.proto
roy-dydx: cosmos_protos are just annotations. gogoproto.customtype adds some codegen to help using that type. gogoproto.nullable affects the codegen (pointer or value)

protocol/x/govplus/keeper/msg_server.go
roy-dydx: Thanks for double checking, for some reason I remembered the slashing module panic instead. I changed this to error.

protocol/x/govplus/keeper/msg_server.go
roy-dydx: > Can we explicitly check these contracts (if not checked within the function)?
All parts of the contract are checked in the Slash function and will error out. I've added to ValidateBasic as well.

(Except "Infraction was committed equal to or less than an unbonding period in the past, so all unbonding delegations and redelegations from that height are stored". It is up to the community to slash before the tokens are gone.)

protocol/x/govplus/keeper/msg_server.go
```diff
@@ -15,3 +21,39 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	)
+	if err != nil {
+		log.ErrorLogWithError(
+			ctx,
+			"error occured when slashing validator",
+			err,
+		)
+		return nil, err
+	}
+	return &types.MsgSlashValidatorResponse{}, nil
```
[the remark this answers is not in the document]
teddyding: This sounds helpful. Can we add some info logs here for debugging purpose?

teddyding (approved):

protocol/x/govplus/keeper/msg_server.go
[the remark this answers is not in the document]
roy-dydx: I used to have info logs here. But it was just redundant cuz the underlying method has a bunch of logging already.

PR #1018, opened 2024-01-26
[IND-545]: Batch Process Vulcan messages

Christopher-Li:
### Changelist
Batch Process Vulcan messages. It doesn't seem like this change makes much of an impact on processing, I believe it's because Vulcan is reading from multiple partitions and the benefit would be improved if there the same number of partitions as instances of vulcan.

See research [here](https://www.notion.so/dydx/Vulcan-Batch-Processing-646a032e6d2c4a939fe6016f4d8e6516?pvs=4)

### Test Plan
Tested in Staging and internal mainnet

### Author/Reviewer Checklist

### DIFF
--- a/indexer/services/vulcan/src/config.ts
+++ b/indexer/services/vulcan/src/config.ts
@@ -7,7 +7,6 @@ import {
   parseSchema,
   baseConfigSchema,
   parseBoolean,
-  ONE_SECOND_IN_MILLISECONDS,
 } from '@dydxprotocol-indexer/base';
 import {
   kafkaConfigSchema,
@@ -26,7 +25,7 @@ export const configSchema = {
     default: 20,
   }),
   KAFKA_BATCH_PROCESSING_COMMIT_FREQUENCY_MS: parseNumber({
-    default: 3 * ONE_SECOND_IN_MILLISECONDS,
+    default: 3_000,
   }),
   FLUSH_KAFKA_MESSAGES_INTERVAL_MS: parseNumber({
     default: 10,
--- a/indexer/services/vulcan/src/config.ts
+++ b/indexer/services/vulcan/src/config.ts
@@ -7,6 +7,7 @@ import {
   parseSchema,
   baseConfigSchema,
   parseBoolean,
+  ONE_SECOND_IN_MILLISECONDS,
 } from '@dydxprotocol-indexer/base';
 import {
   kafkaConfigSchema,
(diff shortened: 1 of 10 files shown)


indexer/services/vulcan/src/lib/on-batch.ts
```diff
@@ -0,0 +1,92 @@
...
+  stats.timing(
+    'vulcan.batch_time_in_queue',
+    batchTimeInQueue,
+    metricTags,
+  );
+
+  let lastCommitTime: number = startTime;
+  for (let i = 0; i < batch.messages.length; i++) {
+    const message: KafkaMessage = batch.messages[i];
+    await onMessage(message);
```
BrendanChou, line 53: One thing we did in v3 I think is use a [worker pool ](https://github.com/dydxprotocol/archivist/blob/26de6a4e4f1631b36a0912f2fe8296d96ce2537d/src/lib/on-batch.ts#L67)to process in parallel, but then to do it in chunks so that we could commit to an offset knowing that we had processed up to a current offset

Would we want to use the same strategy here?

Christopher-Li: I don't think it works the same. The reason I designed the v3 solution to work in chunks was because archivist could handle messages out of order, but vulcan needs all messages to be processed in order, so there cannot be parallelization between messages of the same partition without additional knowledge of the message

BrendanChou (approved):

dydxwill (approved):

indexer/services/vulcan/src/config.ts
```diff
@@ -20,6 +20,13 @@ export const configSchema = {
   ...kafkaConfigSchema,
   ...redisConfigSchema,
 
+  BATCH_PROCESSING_ENABLED: parseBoolean({ default: true }),
+  KAFKA_BATCH_PROCESSING_COMMIT_FREQUENCY: parseNumber({
```
haoxu07, line 24: Nit： seems not used in this change. Is this for minimum # within 1 second?

Christopher-Li: good chance, removed!

indexer/services/vulcan/src/lib/on-batch.ts
```diff
@@ -0,0 +1,92 @@
...
+  stats.timing(
+    'vulcan.batch_processing_time',
+    batchProcessingTime,
+    metricTags,
+  );
+  stats.gauge(
+    'vulcan.batch_size',
+    batch.messages.length,
+    metricTags,
+  );
```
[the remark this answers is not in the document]
Christopher-Li: did some research, I think timing is the best option we have

PR #1022, opened 2024-01-29
Fix Aggregate Historical Trading rewards and always use generated id in tables

Christopher-Li:
### Changelist
Fix Aggregate Historical Trading rewards and always use generated id in tables
- Since we did not query for existing `trading_reward_aggregations` with `startedAtHeight`, any existing trading aggregation for that address and period was found and we would attempt to update a `trading_reward_aggregations` that did not exist rather than create a new aggregation.
- Always use generated ids in tables instead of defaulting to the object that is passed in. This will prevent errors where the object contains an id that is incorrect is sent to the table. Instead of overwriting the incorrect id with the generated id, postgres would use the passed in id.

### Test Plan
Unit tests, will test in staging before merging

### Author/Reviewer Checklist

### DIFF
--- a/indexer/packages/postgres/src/stores/candle-table.ts
+++ b/indexer/packages/postgres/src/stores/candle-table.ts
@@ -111,8 +111,8 @@ export async function create(
   return CandleModel.query(
     Transaction.get(options.txId),
   ).insert({
-    id: uuid(candle.startedAt, candle.ticker, candle.resolution),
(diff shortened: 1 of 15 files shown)


vincentwschau (approved):
Non-blocking nit: Add tests for the logic to use the generated id rather than the id in the object.

dydxwill (approved):

PR #1062, opened 2024-02-09
Make app runnable without creating default home dir

roy-dydx:
### Changelist
- Create a temp dir instead as a quick hack.

### Test Plan
- Run localnet, inspect the containers and see that no default home dir was created during any step.
- Run a command `./build/dydxprotocold  query subaccounts show-subaccount dydx1q628jg3e8vzy6xcns6fxls7k7kplqgxr0gy87t 0 --home ~/.dydxprotocol2`, with existing custom home dir, ensure it works properly and doesn't create default home dir.
  - Now run it without custom home dir, ensure it creates default home dir and uses it
  - Now run it with custom home dir again, ensure it's still using the custom home dir

### Author/Reviewer Checklist

### DIFF
--- a/protocol/cmd/dydxprotocold/cmd/root.go
+++ b/protocol/cmd/dydxprotocold/cmd/root.go
@@ -165,6 +165,9 @@ func NewRootCmdWithInterceptors(
 	if err != nil {
 		panic(err)
 	}
+	// [ Home Dir Temp Fix ] (also see protocol/cmd/dydxprotocold/main.go)
+	// Unset the temp home dir. This must be done after `ReadFromClientConfig`, otherwise it will
+	// create a temp dir in cwd.
 	initClientCtx.HomeDir = ""
 	if err := autoCliOpts(tempApp, initClientCtx).EnhanceRootCommand(rootCmd); err != nil {
 		panic(err)
--- a/protocol/cmd/dydxprotocold/cmd/root.go
+++ b/protocol/cmd/dydxprotocold/cmd/root.go
@@ -165,6 +165,7 @@ func NewRootCmdWithInterceptors(
 	if err != nil {
 		panic(err)
 	}
+	initClientCtx.HomeDir = ""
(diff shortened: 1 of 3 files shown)


vincentwschau (approved):

protocol/cmd/dydxprotocold/main.go
```diff
@@ -23,3 +23,13 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+var tempDir = func() string {
+	dir, err := os.MkdirTemp("", "dydxprotocol")
+	if err != nil {
+		dir = app.DefaultNodeHome
+	}
+	defer os.RemoveAll(dir)
```
lcwik, line 32: won't defer delete `dir` when function is returning and this would clear the `DefaultNodeHome` from any existing state if creating a temp dir creation is not possible?

protocol/cmd/dydxprotocold/cmd/root.go
```diff
@@ -165,6 +165,7 @@ func NewRootCmdWithInterceptors(
 	if err != nil {
 		panic(err)
 	}
+	initClientCtx.HomeDir = ""
```
lcwik, line 168: is there a reason why it is set here instead of above on line 109?

If so can you document why with a comment?

protocol/cmd/dydxprotocold/main.go
roy-dydx: Yeah, I just copied this straight from cosmos. But it doesn't matter because the directories are recreated created in `ReadFromClientConfig`.

protocol/cmd/dydxprotocold/cmd/root.go
roy-dydx: If we set this before `ReadFromClientConfig`, the config will be created in cwd instead.

lcwik (approved):

PR #1073, opened 2024-02-13
Fix unnecessary home directory creation issues

roy-dydx:
### Changelist
- Revert https://github.com/dydxprotocol/v4-chain/pull/1062 temporary fix and bring in official cosmos fix for it
- Pass in a temp dir in client tempApp to prevent creation of `data` directory in the cwd

### Test Plan
- Tested query commands to ensure correct home directory was being used in default and non-default cases. Checked to make sure default one wasn't created when not intended.
- Ran local net and inspected containers to make sure data directory isn't created in `/dydxprotocol/`.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/cmd/dydxprotocold/cmd/root.go
+++ b/protocol/cmd/dydxprotocold/cmd/root.go
@@ -28,6 +28,7 @@ import (
 	runtimeservices "github.com/cosmos/cosmos-sdk/runtime/services"
 	"github.com/cosmos/cosmos-sdk/server"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	simtestutil "github.com/cosmos/cosmos-sdk/testutil/sims"
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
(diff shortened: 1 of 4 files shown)


protocol/cmd/dydxprotocold/cmd/root.go
```diff
@@ -90,7 +91,7 @@ func NewRootCmdWithInterceptors(
 		dbm.NewMemDB(),
 		nil,
 		true,
-		viper.New(),
```
lcwik, line 93: Can this be replaced with:
```
initAppOptions := viper.New()
initAppOptions.Set(flags.FlagHome, tempDir())
tempApp(..., initAppOptions)
```

Was hoping to keep away from test utilities in Cosmos SDK as part of the production code.

lcwik (approved):
There is an existing ticket about get command line tests for coverage of these edge cases since it is difficult to know what the initial client context and config should be since the Cosmos SDK documentation is lacking as to how those values are needed for Auto CLI opts. Is the additional tests you can add in a follow-up?

lcwik:
There is a lint issue with an import of the same statement twice and tests are failing (one seems to be due to a flake)

roy-dydx:
@Mergifyio backport release/protocol/v4.x

PR #1090, opened 2024-02-22
Upgrade client/v2 to fix autocli signing issues

roy-dydx:
### Changelist
Use client/v2 commit from cosmos-sdk v0.50.4. This package released separately, but the latest fixes aren't released yet.

Specifically, we need the deleted block in msg.go to fix the bug: https://github.com/cosmos/cosmos-sdk/pull/19219/files#diff-5560b4d40a3e412c5265a379f1e495c01249be7b08bdb6931555088832fa34d6

Without this change, sign mode handlers are overwritten and we can't sign for pure autocli messages (e.g. `slashing unjail`) (with direct mode).

Note: We don't actually support or intend to support sign mode textual.

### Test Plan
Tested `slashing unjail`. Without this change we get the error `unsuppored sign mode SIGN_MODE_TEXTUAL` when trying to run that command.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/go.mod
+++ b/protocol/go.mod
@@ -411,8 +411,8 @@ require (
 )
 
 replace (
-	// Ideally we rely on a released version (we don't make any changes in our cosmos-sdk fork). In this case the
-	// latest signing mode fixes aren't tagged as a release yet.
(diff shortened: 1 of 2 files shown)


lcwik (approved):
Ideally add a TODO with a linear ticket to revert back once we don't need an override.

PR #1100, opened 2024-02-27
Ensure that all matching status are returned in list orders endpoint

Christopher-Li:
### Changelist
When listing orders, we were previously only querying the latest ${limit} orders, and not matching by status. So if the latest 100 orders were all filled, and a user was attempting to untriggered orders, they would not receive any orders.

### Test Plan
add a test to ensure that status is properly filtered for

### Author/Reviewer Checklist

### DIFF
--- a/indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
@@ -86,13 +86,9 @@ class OrdersController extends Controller {
       goodTilBlockTimeBeforeOrAt,
     };
     if (!_.isEmpty(status)) {
-      // BEST_EFFORT_OPENED status not included in the filter, orders in postgres cannot be
-      // BEST_EFFORT_OPENED. An order is only BEST_EFFORT_OPENED if it exists in redis and not
-      // in postgres.
-      orderQueryConfig.statuses = _.filter(
-        status,
-        (s: APIOrderStatus) => s !== APIOrderStatusEnum.BEST_EFFORT_OPENED,
-      ) as OrderStatus[];
+      // BEST_EFFORT_OPENED status is not filtered out, because it's a minor optimization,
+      // is more confusing, and is not going to affect the result of the query.
+      orderQueryConfig.statuses = status as OrderStatus[];
     }
     const ordering: Ordering = returnLatestOrders !== undefined
       ? returnLatestOrdersToOrdering(returnLatestOrders)
--- a/indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
@@ -85,8 +85,11 @@ class OrdersController extends Controller {
       goodTilBlockBeforeOrAt: goodTilBlockBeforeOrAt?.toString(),
       goodTilBlockTimeBeforeOrAt,
     };
-    if (status !== undefined && !_.isEmpty(status)) {
-      orderQueryConfig.status = _.filter(
+    if (!_.isEmpty(status)) {
+      // BEST_EFFORT_OPENED status not included in the filter, orders in postgres cannot be
+      // BEST_EFFORT_OPENED. An order is only BEST_EFFORT_OPENED if it exists in redis and not
+      // in postgres.
+      orderQueryConfig.statuses = _.filter(
         status,
         (s: APIOrderStatus) => s !== APIOrderStatusEnum.BEST_EFFORT_OPENED,
       ) as OrderStatus[];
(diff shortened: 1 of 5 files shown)


indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
```diff
@@ -74,6 +76,21 @@ class OrdersController extends Controller {
...
+    const orderQueryConfig: OrderQueryConfig = {
+      subaccountId: [SubaccountTable.uuid(address, subaccountNumber)],
+      limit,
+      clobPairId,
+      side,
+      type,
+      goodTilBlockBeforeOrAt: goodTilBlockBeforeOrAt?.toString(),
+      goodTilBlockTimeBeforeOrAt,
+    };
+    if (status !== undefined && !_.isEmpty(status)) {
```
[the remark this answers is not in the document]
BrendanChou: nit: so if `status = []` (filter out all statuses) you are setting `orderQueryConfig.status = undefined` instead, which doesn't filter at all. I would probably not do that as it seems unexpected behavior

indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
```diff
@@ -114,7 +123,28 @@ class OrdersController extends Controller {
...
+      (redisOrder: RedisOrder) => {
+        return OrderTable.orderIdToUuid(redisOrder.order!.orderId!);
+      },
+    );
+
+    // Fetch additional orders from Postgres that are not in the initial query, because they
+    // might not conform to the status filter above. For example, if the user is querying for
+    // `status: [BEST_EFFORT_OPENED]`, we need to fetch all orders from Postgres, because if the
+    // order in postgres is 'OPENED', then we do not want to return this order to the user as
+    // 'BEST_EFFORT_OPENED'.
```
vincentwschau, line 137: nit: The main reason for this is because postgres is regarded as the source of truth, and redis and postgres may differ in the status of an order. Can you call this out specifically in the comment?

indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
```diff
@@ -114,7 +123,28 @@ class OrdersController extends Controller {
...
+    const additionalPostgresOrders: OrderFromDatabase[] = await OrderTable.findAll({
```
vincentwschau, line 138: nit: This query only needs to be done if the user is querying for orders with status including `BEST_EFFORT_OPENED` right? There's no need to do this extra SQL query if the user is querying for any other status, as they will definitely exist in postgres.

Christopher-Li: No need to filter by status. `RedisOrder`s do not contain a status field, we assume they are BEST_EFFORT_OPENED unless there is a corresponding postgres order that says otherwise

indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
```diff
@@ -74,6 +76,21 @@ class OrdersController extends Controller {
...
+      clobPairId,
+      side,
+      type,
+      goodTilBlockBeforeOrAt: goodTilBlockBeforeOrAt?.toString(),
+      goodTilBlockTimeBeforeOrAt,
+    };
+    if (status !== undefined && !_.isEmpty(status)) {
+      orderQueryConfig.status = _.filter(
+        status,
+        (s: APIOrderStatus) => s !== APIOrderStatusEnum.BEST_EFFORT_OPENED,
```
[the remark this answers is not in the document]
Christopher-Li: Added an additional comment to explain this weird logic:
`BEST_EFFORT_OPENED status not included in the filter, orders in postgres cannot be BEST_EFFORT_OPENED. An order is only BEST_EFFORT_OPENED if it exists in redis and not in postgres.`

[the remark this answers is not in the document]
Christopher-Li: decided now to just remove this filter for readability

indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
```diff
@@ -129,17 +128,24 @@ class OrdersController extends Controller {
...
-      ...DEFAULT_POSTGRES_OPTIONS,
-    });
+    // Postgres is regarded as the source of truth, so for any redis orders not returned from the
+    // initial postgres query, we need to fetch them from Postgres to ensure we have the most
+    // accurate status. For example, if the user is querying for `status: [BEST_EFFORT_OPENED]`,
+    // we need to fetch all orders from Postgres, because if the order in postgres is 'OPENED',
+    // then we do not want to return this order to the user as 'BEST_EFFORT_OPENED'.
+    let additionalPostgresOrders: OrderFromDatabase[] = [];
+    if (!_.isEmpty(postgresOrderIdsToFetch)) {
+      additionalPostgresOrders = await OrderTable.findAll({
```
BrendanChou, line 143: nit: does it make sense to filter on `statuses` here? Just because there could be a race condition where some order exists in the redis cache, but by the time you go to get it from the database, it's in some unexpected status like `CANCELED`

vincentwschau: I think in this case, it would be correct to use the status found in postgres for the order, e.g. if the user was looking for all open / best-effort opened orders and by the time we fetch from postgres, the order is canceled, then we shouldn't return the order.

vincentwschau (approved):

indexer/services/comlink/src/controllers/api/v4/orders-controller.ts
Christopher-Li: While redis is not the source of truth for status, we do use it as source of truth for other values mentioned here:
```
    size: redisOrder.size,
    price: redisOrder.price,
    goodTilBlock: protocolTranslations.getGoodTilBlock(redisOrder.order!)?.toString() ?? undefined,
    goodTilBlockTime: protocolTranslations.getGoodTilBlockTime(redisOrder.order!) ?? undefined,
    timeInForce: apiTranslations.orderTIFToAPITIF(redisOrderTIF),
    postOnly: apiTranslations.isOrderTIFPostOnly(redisOrderTIF),
    reduceOnly: redisOrder.order!.reduceOnly,
    ```

PR #1101, opened 2024-02-27
[CT-520] Validatebasic for new MsgBatchCancel

jonfung-dydx:
MsgBatchCancel stateless validation.

### DIFF
--- a/protocol/x/clob/types/constants.go
+++ b/protocol/x/clob/types/constants.go
@@ -8,6 +8,10 @@ import (
 // `MsgPlaceOrder` or `MsgCancelOrder` message will be considered valid by the validator.
 const ShortBlockWindow uint32 = 20
 
+// MaxMsgBatchCancelBatchSize represents the maximum number of cancels that a MsgBatchCancel
+// can have in one Msg.
+const MaxMsgBatchCancelBatchSize uint32 = 100
+
 // StatefulOrderTimeWindow represents the maximum amount of time in seconds past the current block time that a
 // long-term/conditional `MsgPlaceOrder` message will be considered valid by the validator.
 const StatefulOrderTimeWindow time.Duration = 95 * 24 * time.Hour // 95 days.
--- a/protocol/x/clob/types/errors.go
+++ b/protocol/x/clob/types/errors.go
@@ -209,7 +209,7 @@ var (
 	ErrInvalidBatchCancel = errorsmod.Register(
 		ModuleName,
 		45,
-		"Invalid Batch Cancel",
+		"Invalid batch cancel message",
 	)
 
 	// Liquidations errors.
--- a/protocol/x/clob/types/errors.go
+++ b/protocol/x/clob/types/errors.go
@@ -206,6 +206,11 @@ var (
 		44,
 		"invalid time in force",
 	)
+	ErrInvalidBatchCancel = errorsmod.Register(
+		ModuleName,
+		45,
+		"Invalid Batch Cancel",
+	)
 
 	// Liquidations errors.
 	ErrInvalidLiquidationsConfig = errorsmod.Register(
--- a/protocol/x/clob/types/message_batch_cancel.go
+++ b/protocol/x/clob/types/message_batch_cancel.go
@@ -10,7 +10,11 @@ import (
 var _ sdk.Msg = &MsgBatchCancel{}
 
 // NewMsgBatchCancel constructs a MsgBatchCancel.
-func NewMsgBatchCancel(subaccountId satypes.SubaccountId, cancelBatch []OrderBatch, goodTilBlock uint32) *MsgBatchCancel {
+func NewMsgBatchCancel(
+	subaccountId satypes.SubaccountId,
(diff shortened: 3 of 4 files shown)


protocol/x/clob/types/errors.go
```diff
@@ -206,6 +206,11 @@ var (
 		44,
 		"invalid time in force",
 	)
+	ErrInvalidBatchCancel = errorsmod.Register(
+		ModuleName,
+		45,
+		"Invalid Batch Cancel",
```
jayy04, line 212:
```suggestion
		"Invalid batch cancel message",
```

protocol/x/clob/types/constants.go
```diff
@@ -8,6 +8,10 @@ import (
 // `MsgPlaceOrder` or `MsgCancelOrder` message will be considered valid by the validator.
 const ShortBlockWindow uint32 = 20
 
+// MaxMsgBatchCancelBatchSize represents the maximum number of cancels that a MsgBatchCancel
+// can have in one Msg.
+const MaxMsgBatchCancelBatchSize uint32 = 100
```
jayy04, line 13: for my understanding - did we decide to set to to 100 or 1000?

protocol/x/clob/types/message_batch_cancel.go
```diff
@@ -0,0 +1,54 @@
...
+func NewMsgBatchCancel(subaccountId types.SubaccountId, cancelBatch []OrderBatch, goodTilBlock uint32) *MsgBatchCancel {
+	return &MsgBatchCancel{
+		SubaccountId:     subaccountId,
+		ShortTermCancels: cancelBatch,
+		GoodTilBlock:     goodTilBlock,
+	}
+}
+
+// ValidateBasic performs stateless validation for the `MsgBatchCancel` msg.
+func (msg *MsgBatchCancel) ValidateBasic() (err error) {
```
jayy04, line 22: so ValidateBasic is actually optional as of v0.50 - I think it's still supported for backwards compatibility but can you maybe spend 10-15 minutes to see how it works w/ the new version and what was the motivation? kind of curious about this

jayy04: just wondering if transactions that are obviously wrong (e.g. failing ValidateBasic) can now enter mempool and be included in a block

protocol/x/clob/types/message_batch_cancel_test.go
```diff
@@ -0,0 +1,123 @@
...
+			),
+			err: types.ErrInvalidBatchCancel,
+		},
+		"over 100 cancels split over two clob pair id": {
+			msg: *types.NewMsgBatchCancel(
+				constants.Alice_Num0,
+				[]types.OrderBatch{
+					{
+						ClobPairId: 0,
+						ClientIds:  oneOverMax[:52],
```
jayy04, line 56: meganit: instead of hardcoding 52 here, this should be a function of MaxShortTermCancels which is subject to change

protocol/x/clob/types/message_batch_cancel_test.go
```diff
@@ -0,0 +1,123 @@
...
+			),
+			err: types.ErrInvalidBatchCancel,
+		},
+		"success: two clob pair id, 100 cancels": {
+			msg: *types.NewMsgBatchCancel(
+				constants.Alice_Num0,
+				[]types.OrderBatch{
+					{
+						ClobPairId: 0,
+						ClientIds:  oneOverMax[:50],
```
jayy04, line 73: ditto

protocol/x/clob/types/message_batch_cancel_test.go
```diff
@@ -0,0 +1,123 @@
...
+						ClobPairId: 0,
+						ClientIds: []uint32{
+							0, 1, 2, 3, 1,
+						},
+					},
+				},
+				10,
+			),
+			err: types.ErrInvalidBatchCancel,
+		},
```
jayy04, line 111: add cases where `len(short_term_cancel) == 0` and `len(client_ids) == 0`?

jayy04 (approved):

protocol/x/clob/types/constants.go
jonfung-dydx: https://dydx-team.slack.com/archives/C06JZMTFACV/p1708985170576359?thread_ts=1708970027.055389&cid=C06JZMTFACV

We can change this as needed later.

protocol/x/clob/types/message_batch_cancel.go
jonfung-dydx: https://github.com/cosmos/cosmos-sdk/issues/15648

Seems like they have moved all validation to the msg server.

Example PRS
https://github.com/cosmos/cosmos-sdk/pull/15832

Based on this comment
https://github.com/cosmos/cosmos-sdk/pull/15782/files#r1163771773
it seems like we lose the functionality of verifying if the proposal message is correct before execution. Apparently simulation covers that case. However, a bit confused as users can still submit messages without simulating first.

PR #1110, opened 2024-02-28
[CT-514] Clob `MsgBatchCancel` functionality

jonfung-dydx:
Changes included in this PR:
- `go.mod` replace included to use new changes [here](https://github.com/dydxprotocol/cometbft/pull/28)
- Clob Ante Handler -> Keeper `BatchCancel` flow implemented
- Single e2e test for this flow, more to come
- Various locations in the code (i.e InterfaceRegistries) now account for `MsgBatchCancel`
- Proto updates: In the `OrderBatch` object, `repeated fixed32` was having issues deserializing/serializing. Changed to `repeated uint32`. Context [here](https://clement-jean.github.io/packed_vs_unpacked_repeated_fields/)


jonfung-dydx:
[CT-513](https://linear.app/dydx/issue/CT-513/clob-antehandler-updates)
[CT-510](https://linear.app/dydx/issue/CT-510/modify-msgbatchcancel-to-skip-signature-verification)

proto/dydxprotocol/clob/tx.proto
```diff
@@ -123,7 +123,9 @@ message OrderBatch {
   // The Clob Pair ID all orders in this order batch belong to.
   uint32 clob_pair_id = 1;
   // List of client ids in this order batch.
-  repeated fixed32 client_ids = 2;
+  // Note that this is serialized as a uint32 instead of a fixed32 in the
+  // regular OrderId object to save space.
```
jayy04, line 127:
```suggestion
  // Note that this is serialized as uint32 instead of fixed32 to 
  // avoid issues when decoding repeated packed fixed32.
```

jayy04: might need to fix lint ^

protocol/lib/log/constants.go
```diff
@@ -46,9 +46,10 @@ const (
 	// Module tag values are prefixed with `x/`
 	Clob = "x/clob"
 
-	CheckTx   = "check_tx"
-	RecheckTx = "recheck_tx"
-	DeliverTx = "deliver_tx"
+	CheckTx        = "check_tx"
+	RecheckTx      = "recheck_tx"
+	DeliverTx      = "deliver_tx"
+	MsgBatchCancel = "msg_batch_cancel"
```
jayy04, line 52: is there a better group for this? doesn't seem related to CheckTx, DeliverTx, etc

protocol/x/clob/keeper/orders.go
```diff
@@ -46,6 +46,83 @@ func (k Keeper) GetOperations(ctx sdk.Context) *types.MsgProposedOperations {
...
+	// Note that we add `+1` here to account for the fact that `ctx.BlockHeight()` is technically the
+	// previously mined block, not the next block that will be proposed. This is due to the fact that
+	// this function is only ever called during `CheckTx`.
+	nextBlockHeight := lib.MustConvertIntegerToUint32(ctx.BlockHeight() + 1)
+
+	// Statefully validate the GTB and the clob pair ids.
+	goodTilBlock := msg.GetGoodTilBlock()
+	if err := k.validateGoodTilBlock(goodTilBlock, nextBlockHeight); err != nil {
+		return err
+	}
```
jayy04, line 71: don't we do this in `CancelShortTermOrder`?

jonfung-dydx: we do but i'd rather just error out here than have each one error out slowly in each CancelShortTermOrder call and then return the user a `BatchCancelAllFailed with error: invalid GTB`, a top level `invalid gtb` seems better

protocol/x/clob/keeper/orders.go
```diff
@@ -46,6 +46,83 @@ func (k Keeper) GetOperations(ctx sdk.Context) *types.MsgProposedOperations {
...
+// `goodTilBlock` in the memclob for all specified orders.
+// This message is not atomic. It will optimistically call `CancelShortTermOrder` for every order in the batch.
+// If any of the orders error, the error will be silently logged. This msg will only error if:
+// - Stateful validation fails
+// - Every cancel in the batch fails.
+// This method assumes the provided MsgBatchCancel has already passed ValidateBasic in CheckTx.
+func (k Keeper) BatchCancelShortTermOrder(
+	ctx sdk.Context,
+	msg *types.MsgBatchCancel,
+) error {
```
jayy04, line 60: thought about this a little bit more - what do you think about 

1. returning (succeeded []uint32, failed []uint32) here
2. modify `MsgBatchCancelResponse` to return this information to the user (currently it's empty)

This way we don't have to add the logic for `oneCancelSucceeded` and code would be much cleaner, and at the same time, users know exactly which cancels went through. we can also include `[]error` in addition to `failed []uint32`

protocol/x/clob/keeper/orders.go
```diff
@@ -46,6 +46,83 @@ func (k Keeper) GetOperations(ctx sdk.Context) *types.MsgProposedOperations {
...
+	for _, batchOrder := range msg.GetShortTermCancels() {
+		clobPairId := batchOrder.GetClobPairId()
+		if _, found := k.GetClobPair(ctx, types.ClobPairId(clobPairId)); !found {
+			return errorsmod.Wrapf(
+				types.ErrInvalidClobPairParameter,
+				"Invalid clob pair id %+v",
+				clobPairId,
+			)
+		}
+	}
```
jayy04, line 81: ditto - do we not have duplicated validation in `CancelShortTermOrder`?

protocol/x/clob/keeper/orders.go
```diff
@@ -46,6 +46,83 @@ func (k Keeper) GetOperations(ctx sdk.Context) *types.MsgProposedOperations {
...
+					SubaccountId: subaccountId,
+					OrderFlags:   types.OrderIdFlags_ShortTerm,
+					ClobPairId:   clobPairId,
+					ClientId:     clientId,
+				},
+				GoodTilOneof: &types.MsgCancelOrder_GoodTilBlock{
+					GoodTilBlock: goodTilBlock,
+				},
+			}
+			// Run the short term order. If it errors, just log silently.
```
jayy04, line 98: instead of logging silently we also have the option to wrap and return it back to the user 

something like 

```
var wrapped error
for {
    ...
    if err != nil {
        wrapped = errorsmod.Wrapf(wrapped, error, ...)
    }
}
```

jayy04: ah I guess we made the decision to only return error when all cancellations fail?

protocol/x/clob/ante/clob.go
```diff
@@ -130,6 +130,26 @@ func (cd ClobDecorator) AnteHandle(
...
+	case *types.MsgBatchCancel:
+		// MsgBatchCancel currently only processes short-term cancels right now.
+		// No need to process short term orders on `ReCheckTx`.
+		if ctx.IsReCheckTx() {
+			return next(ctx, tx, simulate)
+		}
+
+		ctx = log.AddPersistentTagsToLogger(ctx,
+			log.Handler, log.MsgBatchCancel,
+		)
```
jayy04, line 142:
```suggestion
		ctx = log.AddPersistentTagsToLogger(
		    ctx,
			log.Handler, log.MsgBatchCancel,
		)
```

protocol/x/clob/ante/clob.go
```diff
@@ -130,6 +130,26 @@ func (cd ClobDecorator) AnteHandle(
...
+
+		err := cd.clobKeeper.BatchCancelShortTermOrder(
+			ctx,
+			msg,
+		)
+
+		log.DebugLog(ctx, "Received new batch cancellation",
+			log.Tx, cometbftlog.NewLazySprintf("%X", tmhash.Sum(ctx.TxBytes())),
+			log.Error, err,
+		)
```
jayy04, line 152: nit:
```suggestion
		log.DebugLog(
		    ctx, 
		    "Received new batch cancellation",
			log.Tx, cometbftlog.NewLazySprintf("%X", tmhash.Sum(ctx.TxBytes())),
			log.Error, err,
		)
```

protocol/x/clob/keeper/orders.go
jayy04: then we can just call `k.MemClob.CancelOrder` directly? I guess not as future proof

jayy04: small nit: I kind of don't like this `+1` logic everywhere and feel like we should move it to within `validateGoodTilBlock`. 

haven't checked other references to `validateGoodTilBlock`, so will leave it to you to decide

jonfung-dydx: Hm we could call memclob directly. I'm ok with either.

Current approach
- `SendOffchainMessages` gets called multiple times for each 
- each cancel gets its own latency metric exposed
Calling memclob directly
- can coalesce offchain messages and call `SendOffchainMessages` once
- no latency metrics for each cancel

re: the +1 logic, i'm kind of for keeping it as it is since i don't want `validateGoodTilBlock` method to assume that it's being used in checktx. i'd prefer to keep it as a pure function of its parameters without assuming anything about usage

protocol/x/clob/keeper/orders.go
jonfung-dydx: hm i'm down to do the suggestion (1), but returning `[]error` seems a little clunky to me.

Reason being, usually when someone calls a function you expect to check one error to see if it failed or not.
I'm ok with returning just a BatchError, or using the [errors.join](https://pkg.go.dev/errors#Join) method, but it seems to just stack errors together separated by newlines. not sure how i feel about that.

protocol/x/clob/keeper/orders.go
jonfung-dydx: hmm same reasoning as above, i'd just like to reject the message asap. but this does take up some state reads, so this could warrant directly calling memclob cancel.

protocol/lib/log/constants.go
jonfung-dydx: this is the batch for tag values, i just haven't been bothered to move anything over to this constants file other than the callbacks

protocol/x/clob/keeper/orders.go
jonfung-dydx: Going to punt this. Details in ticket.
https://linear.app/dydx/issue/CT-649/modify-msgbatchcancelresponse-to-return-success-failure-ids

protocol/x/clob/ante/clob.go
```diff
@@ -44,6 +46,7 @@ func (cd ClobDecorator) AnteHandle(
 	simulate bool,
 	next sdk.AnteHandler,
 ) (sdk.Context, error) {
+	fmt.Println("clob ante handler")
```
jayy04, line 49: remove

protocol/x/clob/keeper/orders.go
jayy04: > re: the +1 logic, i'm kind of for keeping it as it is since i don't want validateGoodTilBlock method to assume that it's being used in checktx. i'd prefer to keep it as a pure function of its parameters without assuming anything about usage

I meant do something like 
``` 
func validateGTB(ctx, ...) {
    if ctx.IsCheckTx() {
        GTB = ctx.BlockHeight() + 1
    }
    GTB = ctx.BlockHeight()
}
```

slightly cleaner imo, but up to you

protocol/x/clob/keeper/orders.go
jayy04, line 81: should we move this and the GTB validation to a separate function (something like `PerformStatefulBatchCancelValidation`?

protocol/x/clob/keeper/orders.go
jayy04: I think this function can still return `([]uint32, []uint32)` here, the caller can decide what to do with it (e.g. returning `ErrBatchCancelAllFailed`

protocol/x/clob/e2e/batch_cancel_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// These tests are the same as the e2e tests for single order cancellations.
```
jayy04, line 13: can we add a few tests for batch cancelling multiple orders?

jayy04 (approved):

protocol/x/clob/e2e/batch_cancel_test.go
jonfung-dydx: sep pr

protocol/x/clob/keeper/orders.go
jonfung-dydx: done

PR #1135, opened 2024-03-04
feat: Proposal logic for Slinky

Eric-Warehime:
### Changelist
This PR introduces interfaces and implementations of those interfaces in the Prepare and Process Proposal handlers.

We introduce Slinky-specific proposal handlers via price update generator and market price decoders. Existing behavior of the application is preserved in order to allow the changes to be introduced to the repo and turned on in a later PR which will include the end to end flow and feature flagging.

### Test Plan
New changes include unit tests and are fully backwards compatible by using the default interfaces as existing impl.

### Author/Reviewer Checklist


protocol/app/prepare/prepare_proposal.go
```diff
@@ -71,8 +74,21 @@ func PrepareProposalHandler(
 			return &EmptyResponse, nil
 		}
 
+		var extCommitBzTx []byte
+		if len(req.Txs) >= slinkyabci.NumInjectedTxs {
+			extCommitBzTx = req.Txs[slinkyabci.OracleInfoIndex]
```
tqin7, line 79: is extended commit always at `OracleInfoIndex`?

protocol/app/prepare/prepare_proposal.go
```diff
@@ -71,8 +74,21 @@ func PrepareProposalHandler(
...
+		}
+
+		// get the update market prices tx
+		msg, err := priceUpdateGenerator.GetValidMarketPriceUpdates(ctx, extCommitBzTx)
```
tqin7, line 83: Maybe I'm missing something but this doesn't exclude the markets whose prices changed less than a threshold, right?

protocol/app/process/default_market_price_decoder.go
```diff
@@ -0,0 +1,42 @@
...
+
+// DecodeUpdateMarketPricesTx returns a new `UpdateMarketPricesTx` after validating the following:
+//   - decodes the given tx bytes
+//   - checks the num of msgs in the tx matches expectations
+//   - checks the msg is of expected type
+//
+// If error occurs during any of the checks, returns error.
+func (mpd *DefaultUpdateMarketPriceTxDecoder) DecodeUpdateMarketPricesTx(
+	ctx sdk.Context, txs [][]byte) (*UpdateMarketPricesTx, error) {
+	return DecodeUpdateMarketPricesTx(ctx, mpd.pk, mpd.txDecoder, txs[len(txs)+updateMarketPricesTxLenOffset])
```
Christopher-Li, line 35: can you explain the `updateMarketPricesTxLenOffset` constant?

protocol/app/process/market_price_decoder.go
```diff
@@ -0,0 +1,28 @@
+package process
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	pricestypes "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
+)
+
+// MarketPriceDecoder is an interface for decoding market price transactions, This interface is responsible
```
Christopher-Li, line 8: `MarketPriceDecoder` shouldnt this be `UpdateMarketPriceTxDecoder`?

protocol/app/process/default_market_price_decoder.go
Christopher-Li, line 35: Where is this inner `DecodeUpdateMarketPricesTx` defined?

protocol/app/process/slinky_market_price_decoder.go
```diff
@@ -0,0 +1,120 @@
...
+		// get the expected message from the injected vote-extensions
+		var err error
+		expectedMsg, err = mpd.agg.GetValidMarketPriceUpdates(ctx, txs[slinkyabci.OracleInfoIndex])
+		if err != nil {
+			return nil, getDecodingError(msgUpdateMarketPricesType, err)
+		}
+	}
+
+	// use the underlying decoder to get the UpdateMarketPricesTx
+	updateMarketPrices, err := mpd.decoder.DecodeUpdateMarketPricesTx(ctx, txs)
```
Christopher-Li, line 61: what is the purpose of this inner decoder? Is it so `SlinkyMarketPriceDecoder` can call `DefaultMarketPriceDecoder`?

protocol/app/process/default_market_price_decoder.go
tqin7, line 1: Why do we need to define this whole decoder struct? Maybe I'm missing something but it doesn't seem to be doing much

protocol/app/prepare/prepare_proposal.go
```diff
@@ -71,8 +74,21 @@ func PrepareProposalHandler(
 			return &EmptyResponse, nil
 		}
 
+		var extCommitBzTx []byte
+		if len(req.Txs) >= slinkyabci.NumInjectedTxs {
```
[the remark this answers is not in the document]
Eric-Warehime: Yeah the prepare proposal handler will inject the VEs as the first tx in the block. If there are no txs, it means there were no VEs injected. This is basically only going to happen if the protocol hasn't enabled VEs yet.

protocol/app/prepare/prepare_proposal.go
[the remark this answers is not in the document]
Eric-Warehime: It's not really determined in the dydx repo though--this is hard to tell here just because of the way I've had to separate out the PRs into manageable chunks. 

The [slinky prepare handler](https://github.com/skip-mev/slinky/blob/main/abci/proposals/proposals.go#L182) wraps the dydx handler. It injects the VE data into the `req.Txs` and then hands off execution to dydx so that we can construct the MarketPriceUpdates Tx with it. 

So the index at which the VEs are injected and the number of txs inject are both fixed in the Slinky impl. I can move the constants here, but changing them without also changing the upstream handler would just break. 

If it's important to be able to control this it can maybe be parametrized as inputs to the slinky prepare handler.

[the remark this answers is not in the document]
Eric-Warehime: Yes, the VEs commit info is always injected as the first Tx in the block. See comment above.

protocol/app/prepare/prepare_proposal.go
Eric-Warehime: My understanding is that this is handled in the Prices Keeper. 
https://github.com/dydxprotocol/v4-chain/pull/1139/files#diff-0da397be90196d47fd5225c12b6bfd6c7c4adc27b46e334f33776f510c0c1031R54

We only extend vote with "valid" market price updates. If a price update is not above the min_price threshold it would not be returned by that keeper method. Perhaps I'm mistaken though and this needs to be handled/validated outside the keeper?

protocol/app/process/default_market_price_decoder.go
Eric-Warehime: AFAIK we've just reused this in order to reference the MarketPricesTx which is injected as the last tx in the block.

protocol/app/process/slinky_market_price_decoder.go
Eric-Warehime: @nivasan1 can probably comment more on this as he implemented it.

The idea is twofold:
1. The existing behavior is preserved via just using `DefaultMarketPricesDecoder` so all of these changes are backwards compatible in the case where slinky is not yet enabled.
2. The new implementation wraps the default encoder because we still want all the same behavior. But additionally we want to do checks on the equality of VE data with the changes being made in the UpdateMarketPrices tx. 

So we first get the injected VE data, then get the UpdateMarketPricesTx via the default behavior, then verify that the two match.

protocol/app/process/market_price_decoder.go
nivasan1: That is correct, the name of the type in the docs should be `UpdateMarketPricesDecoder`

protocol/app/process/default_market_price_decoder.go
nivasan1: This is defined in `app/process/market_prices.go` this is used in the current dydx protocol to unmarshal price-update txs

protocol/app/process/slinky_market_price_decoder.go
nivasan1: That's correct. The `SlinkyUpdateMarketPriceDecoder`'s responsibility is to decode the price-update + determine whether or not the proposed prices followed exactly from the injected vote-extensions

protocol/app/process/default_market_price_decoder.go
nivasan1: The decoder struct is responsible for encapsulating the logical differences between dydx's ProcessProposal pre / post vote-extensions. Specifically
- The `DefaultMarketPriceDecoder` decodes price-updates from proposals as they are done in the pre-vote extensions protoocl
- The `SlinkyMarketPriceDecoder` decodes price-updates + verfies that they follow exactly from the injected vote-extensions

protocol/app/prepare/prepare_proposal.go
```diff
@@ -10,7 +10,10 @@ import (
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	"github.com/dydxprotocol/v4-chain/protocol/app/prepare/prices"
 	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+	pricetypes "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
+	slinkyabci "github.com/skip-mev/slinky/abci/types"
```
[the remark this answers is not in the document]
Eric-Warehime: Same way that comet and the sdk are maintained. Breaking changes will only occur along with a major version bump. Dependency updates should be looked at and tested to make sure regression aren't introduced. Security and bugfixes will be backported to all supported versions.

protocol/app/prepare/prepare_proposal.go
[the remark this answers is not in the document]
Christopher-Li: can we include this in a comment?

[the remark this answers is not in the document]
Christopher-Li: Could we also change this constant name? `NumInjectedTxs` doesn't really mean anything. Ideally this would be a function named `TxsContainsVEs` or the constant would be something like `MinTxRequiredForVEEnabled`

protocol/app/prepare/prepare_proposal.go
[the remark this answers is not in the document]
Christopher-Li: You're saying that storing `OracleInfoIndex = 0` in the dYdX codebase would required changes to Slinky protocol code? It seems to me like this constant of 0 is fixed because VE is always at index 0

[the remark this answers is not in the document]
Eric-Warehime: Updated in `protocol/app/constants/constants.go`.

It is fixed--there's no other place it can be. Changing this constant in either place would simply break the handlers.

protocol/app/process/default_market_price_decoder.go
Christopher-Li: If we stop supporting our own Price Daemon and remove VEs, would you then delete this decoder struct and remove `DefaultMarketPriceDecoder`?

Eric-Warehime: If I understand correctly I think yes.

The interface is just in place to support having the existing implementation and the new one. If we clean up the old code we don't really need the interface/struct anymore and it can just be the function defined in slinky market price decoder.

protocol/app/prepare/prepare_proposal.go
tqin7: makes sense!

PR #1138, opened 2024-03-04
[CT-646] stream offchain updates through stream manager

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -961,7 +961,9 @@ func New(
 	clobFlags := clobflags.GetClobFlagValuesFromOptions(appOpts)
 	logger.Info("Parsed CLOB flags", "Flags", clobFlags)
 
-	memClob := clobmodulememclob.NewMemClobPriceTimePriority(app.IndexerEventManager.Enabled())
+	memClob := clobmodulememclob.NewMemClobPriceTimePriority(
+		app.IndexerEventManager.Enabled() || app.GrpcStreamingManager.Enabled(),
+	)
 
 	app.ClobKeeper = clobmodulekeeper.NewKeeper(
 		appCodec,
@@ -1758,8 +1760,5 @@ func getGrpcStreamingManagerFromOptions(
 	logger log.Logger,
 ) (manager streamingtypes.GrpcStreamingManager) {
 	// TODO(CT-625): add command line flags for full node streaming.
-	if appFlags.NonValidatingFullNode {
-		return streaming.NewGrpcStreamingManager()
-	}
-	return streaming.NewNoopGrpcStreamingManager()
+	return streaming.NewGrpcStreamingManager()
 }
--- a/protocol/streaming/grpc/grpc_streaming_manager.go
+++ b/protocol/streaming/grpc/grpc_streaming_manager.go
@@ -24,7 +24,6 @@ type GrpcStreamingManagerImpl struct {
 type OrderbookSubscription struct {
 	clobPairIds []uint32
 	srv         clobtypes.Query_StreamOrderbookUpdatesServer
-	finished    chan bool
 }
 
 func NewGrpcStreamingManager() *GrpcStreamingManagerImpl {
@@ -43,21 +42,18 @@ func (sm *GrpcStreamingManagerImpl) Subscribe(
 	req clobtypes.StreamOrderbookUpdatesRequest,
 	srv clobtypes.Query_StreamOrderbookUpdatesServer,
 ) (
-	finished chan bool,
 	err error,
 ) {
 	clobPairIds := req.GetClobPairId()
 
 	// Perform some basic validation on the request.
 	if len(clobPairIds) == 0 {
-		return nil, clobtypes.ErrInvalidGrpcStreamingRequest
+		return clobtypes.ErrInvalidGrpcStreamingRequest
 	}
 
-	finished = make(chan bool)
 	subscription := &OrderbookSubscription{
 		clobPairIds: clobPairIds,
 		srv:         srv,
-		finished:    finished,
 	}
 
 	sm.Lock()
@@ -66,7 +62,7 @@ func (sm *GrpcStreamingManagerImpl) Subscribe(
 	sm.orderbookSubscriptions[sm.nextId] = subscription
 	sm.nextId++
 
-	return finished, nil
+	return nil
 }
 
 // SendOrderbookUpdates groups updates by their clob pair ids and
@@ -117,7 +113,6 @@ func (sm *GrpcStreamingManagerImpl) SendOrderbookUpdates(
 
 	// Clean up subscriptions that have been closed.
 	for _, id := range idsToRemove {
-		sm.orderbookSubscriptions[id].finished <- true
 		delete(sm.orderbookSubscriptions, id)
 	}
 }
--- a/protocol/streaming/grpc/grpc_streaming_manager.go
+++ b/protocol/streaming/grpc/grpc_streaming_manager.go
@@ -20,6 +20,7 @@ type GrpcStreamingManagerImpl struct {
 	nextId                 uint32
 }
 
+// OrderbookSubscription represents a active subscription to the orderbook updates stream.
 type OrderbookSubscription struct {
 	clobPairIds []uint32
 	srv         clobtypes.Query_StreamOrderbookUpdatesServer
@@ -37,6 +38,7 @@ func (sm *GrpcStreamingManagerImpl) Enabled() bool {
 }
 
 // Subscribe subscribes to the orderbook updates stream.
+// This function returns a channel that is used to signal termination when an error occurs.
 func (sm *GrpcStreamingManagerImpl) Subscribe(
 	req clobtypes.StreamOrderbookUpdatesRequest,
 	srv clobtypes.Query_StreamOrderbookUpdatesServer,
--- a/protocol/streaming/grpc/grpc_streaming_manager.go
+++ b/protocol/streaming/grpc/grpc_streaming_manager.go
@@ -17,7 +17,7 @@ type GrpcStreamingManagerImpl struct {
 
 	// orderbookSubscriptions maps subscription IDs to their respective orderbook subscriptions.
 	orderbookSubscriptions map[uint32]*OrderbookSubscription
-	nextId                 uint32
+	nextSubscriptionId     uint32
 }
 
 // OrderbookSubscription represents a active subscription to the orderbook updates stream.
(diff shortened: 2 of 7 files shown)


protocol/app/app.go
```diff
@@ -1758,8 +1760,5 @@ func getGrpcStreamingManagerFromOptions(
 	logger log.Logger,
 ) (manager streamingtypes.GrpcStreamingManager) {
 	// TODO(CT-625): add command line flags for full node streaming.
-	if appFlags.NonValidatingFullNode {
-		return streaming.NewGrpcStreamingManager()
-	}
-	return streaming.NewNoopGrpcStreamingManager()
+	return streaming.NewGrpcStreamingManager()
```
jayy04, line 1763: will add the if statement back in in the next PR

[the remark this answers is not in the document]
jonfung-dydx: Unfortunately the ec2 instance I am running this on does not support filesystem reads and writes. Can I paste the contents of my .pem file in the code?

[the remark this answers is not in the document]
jonfung-dydx: Unfortunately I have no money to use AWS secrets manager. What should I do here?

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+		clobPairIds: clobPairIds,
+		srv:         srv,
+		finished:    finished,
+	}
+
+	sm.Lock()
+	defer sm.Unlock()
+
+	sm.orderbookSubscriptions[sm.nextId] = subscription
+	sm.nextId++
```
jonfung-dydx, line 67: for my understanding, what is the point of the nextId variable if we are just using this as a unique key variable for a map? seems like we don't expose it anywhere here, where are we gonna expose subscription key?

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
 
 // GrpcStreamingManagerImpl is an implementation for managing gRPC streaming subscriptions.
 type GrpcStreamingManagerImpl struct {
+	sync.Mutex
+
+	// orderbookSubscriptions maps subscription IDs to their respective orderbook subscriptions.
+	orderbookSubscriptions map[uint32]*OrderbookSubscription
+	nextId                 uint32
```
jonfung-dydx, line 20: nit: `nextSubscriptionId`

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+		}
+		v1updates[clobPairId] = v1update
+	}
+
+	sm.Lock()
+	defer sm.Unlock()
+
+	// Send updates to subscribers.
+	idsToRemove := make([]uint32, 0)
+	for id, subscription := range sm.orderbookSubscriptions {
```
jonfung-dydx, line 102: we are not using id for anything?

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
 ) {
+	// Group updates by clob pair id.
+	updates := make(map[uint32]*clobtypes.OffchainUpdates)
+	for _, message := range offchainUpdates.Messages {
+		clobPairId := message.OrderId.ClobPairId
+		if _, ok := updates[clobPairId]; !ok {
+			updates[clobPairId] = clobtypes.NewOffchainUpdates()
+		}
+		updates[clobPairId].Messages = append(updates[clobPairId].Messages, message)
+	}
```
jonfung-dydx, line 85: Doesn't have to be now, but do we want to condense the messages somehow? We have this condensing function for indexer, but i don't think it works in this case. these events might be condensed differently.
```
func (om *OffchainUpdates) CondenseMessagesForReplay() {
```

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+					idsToRemove = append(idsToRemove, id)
+					break
+				}
+			}
+		}
+	}
+
+	// Clean up subscriptions that have been closed.
+	for _, id := range idsToRemove {
+		sm.orderbookSubscriptions[id].finished <- true
```
jonfung-dydx, line 120: close the channel if it's done and this is the only goroutine that sends stuff into this channel?

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+					},
+				); err != nil {
+					idsToRemove = append(idsToRemove, id)
+					break
+				}
+			}
+		}
+	}
+
+	// Clean up subscriptions that have been closed.
```
jonfung-dydx, line 118:
```suggestion
	// Clean up subscriptions that have been closed.
	// If a Send update has failed for any clob pair id, the whole subscription will be closed.
```

protocol/x/clob/types/errors.go
```diff
@@ -518,4 +518,15 @@ var (
 		10001,
 		"Subaccount cannot open more orders due to equity tier limit.",
 	)
+
+	ErrGrpcStreamingManagerNotEnabled = errorsmod.Register(
```
jonfung-dydx, line 522: nit: label this section of errors

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+		}
+		updates[clobPairId].Messages = append(updates[clobPairId].Messages, message)
+	}
+
+	// Unmarshal messages to v1 updates.
+	v1updates := make(map[uint32][]ocutypes.OffChainUpdateV1)
+	for clobPairId, update := range updates {
+		v1update, err := GetOffchainUpdatesV1(update)
+		if err != nil {
+			panic(err)
```
jonfung-dydx, line 92: i believe golang kills the whole process on a panic. Would this lead to the whole validator going bye bye? Do we want to have a recover?

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,125 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+					break
+				}
+			}
+		}
+	}
+
+	// Clean up subscriptions that have been closed.
+	for _, id := range idsToRemove {
+		sm.orderbookSubscriptions[id].finished <- true
+		delete(sm.orderbookSubscriptions, id)
```
jonfung-dydx, line 121: Are we sure that this delete will properly clean up the nested inner data structures? `grpc.ServerStream` might have a close function?

jonfung-dydx: Seems like there's no close function 🆗

protocol/streaming/grpc/grpc_streaming_manager.go
jonfung-dydx: edit: nvm, i see it for removal

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: I think this gets called before `SendOffchainMessages` is called? i.e. the message is already condensed

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: yeah I was trying to see what makes sense here - 

based on my understanding, MMs will run their own full nodes and these full nodes are only used for streaming information to their trading strategies. in this case, if something happens, probably not important to recover?

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: actually I think we can just use `ctx.Done()`? no need to have this extra channel

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: let me know your thoughts

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: yeah wasn't able to find a close function either 

not really sure if we need to do `ctx, cancelCtx := context.WithCancel(ctx)` to cancel the context, since the connection is likely already broken and the context is already done 🤔

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -9,30 +13,121 @@ var _ types.GrpcStreamingManager = (*GrpcStreamingManagerImpl)(nil)
...
+		orderbookSubscriptions: make(map[uint32]*OrderbookSubscription),
+	}
 }
 
 func (sm *GrpcStreamingManagerImpl) Enabled() bool {
 	return true
 }
 
 // Subscribe subscribes to the orderbook updates stream.
+// This function returns a channel that is used to signal termination when an error occurs.
```
jonfung-dydx, line 40: defunct

protocol/streaming/grpc/grpc_streaming_manager.go
jonfung-dydx: It gets called during Replay Operations but i feel it might not be used half the time due to this code:
```
	if replayUpdates != nil {
		offchainUpdates = replayUpdates
	}
```
I think it should be fine, unless you can think of any edge case where condensing messages drops anything vital for orderbook updates

protocol/streaming/grpc/grpc_streaming_manager.go
jonfung-dydx: https://github.com/grpc/grpc-go/blob/v1.60.x/stream.go#L262-L278
```
	// Possible context leak:
	// The cancel function for the child context we create will only be called
	// when RecvMsg returns a non-nil error, if the ClientConn is closed, or if
	// an error is generated by SendMsg.
```
I'm assuming an error will be generated by SendMsg in this case so the cancel function will be called.

jonfung-dydx (approved):

protocol/streaming/grpc/grpc_streaming_manager.go
jayy04: yeah should be fine to send for example (place + remove) for example

PR #1139, opened 2024-03-04
feat: VoteExtension Slinky logic

Eric-Warehime:
### Changelist
Implements the Slinky logic for vote extension handlers. The VerifyVoteExtension logic is not included in this PR since it will be entirely identical to the Slinky VerifyVoteHandler.

Behavior here is completely backward compatible since vote extension handlers are not hooked up to the app in this PR and additionally they won't be called until the protocol enables vote extensions.

### Test Plan
Unit tests included--perf tests to be done out of band. All implementation according to spec document.

### Author/Reviewer Checklist


protocol/app/vote_extensions/extend_vote_handler.go
```diff
@@ -0,0 +1,59 @@
...
+
+		// Update the market prices in the PricesKeeper, so that the GetValidMarketPriceUpdates
+		// function returns the latest available market prices.
+		updateMarketPricesMsg, ok := updatePricesTx.GetMsg().(*prices.MsgUpdateMarketPrices)
+		if !ok {
+			return nil, fmt.Errorf("expected %s, got %T", "MsgUpdateMarketPrices", updateMarketPricesMsg)
+		}
+
+		// Update the market prices in the PricesKeeper
+		err = e.PricesKeeper.UpdateMarketPrices(ctx, updateMarketPricesMsg.MarketPriceUpdates)
```
tqin7, line 51: Why are we updating market prices in `ExtendVote`?

protocol/x/prices/types/types.go
```diff
@@ -43,7 +45,15 @@ type PricesKeeper interface {
 		ctx sdk.Context,
 		linearInterpolateFunc func(v0 uint64, v1 uint64, ppm uint32) (uint64, error),
 	) error
+	GetValidMarketPriceUpdates(
```
Christopher-Li, line 48: Where is this function defined? I don't see it defined in this PR

protocol/x/prices/keeper/slinky_adapter.go
```diff
@@ -0,0 +1,73 @@
...
+	if len(split) != 2 {
+		return slinkytypes.CurrencyPair{}, fmt.Errorf("incorrectly formatted CurrencyPair: %s", marketPair)
+	}
+	cp := slinkytypes.CurrencyPair{
+		Base:  strings.ToUpper(split[0]),
+		Quote: strings.ToUpper(split[1]),
+	}
+
+	return cp, cp.ValidateBasic()
+}
```
Christopher-Li, line 73: in your [other PR](https://github.com/dydxprotocol/v4-chain/pull/1109/files) in (`protocol/daemons/slinky/client/market_pair_fetcher.go`), you import this function from Slinky, can we consolidate? If this function is fine to have here, lets just have stored in `v4-chain` rather than in Slinky

protocol/x/prices/keeper/slinky_adapter.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (k Keeper) GetPriceForCurrencyPair(ctx sdk.Context, cp slinkytypes.CurrencyPair) (oracletypes.QuotePrice, error) {
+	id, found := k.GetIDForCurrencyPair(ctx, cp)
+	if !found {
+		return oracletypes.QuotePrice{}, fmt.Errorf("currency pair %s not found", cp.String())
+	}
+	mp, err := k.GetMarketPrice(ctx, uint32(id))
+	if err != nil {
+		return oracletypes.QuotePrice{}, fmt.Errorf("currency pair %s not found", cp.String())
```
Christopher-Li, line 55: nit: This should have a different error, maybe "market price not found for currency pair %s"

protocol/app/vote_extensions/extend_vote_handler.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+
+	cometabci "github.com/cometbft/cometbft/abci/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/dydxprotocol/v4-chain/protocol/app/process"
+	prices "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
+)
+
+// ExtendVoteHandler is a wrapper around the Slinky ExtendVoteHandler. This wrapper is responsible for
```
Christopher-Li, line 13: nit: Link to the Slinky Extend Vote Handler

protocol/app/vote_extensions/oracle_client.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	oracleservicetypes "github.com/skip-mev/slinky/service/servers/oracle/types"
+	"google.golang.org/grpc"
+)
+
+// OracleClient is a wrapper around the default Slinky OracleClient interface. This object is responsible for requesting
```
Christopher-Li, line 14: nit: Link to Slink Oracle Client Interface

protocol/app/vote_extensions/oracle_client.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Start starts the OracleClient.
+func (o *OracleClient) Start(_ context.Context) error {
+	return nil
+}
+
+// Stop stops the OracleClient.
+func (o *OracleClient) Stop() error {
+	return nil
+}
```
Christopher-Li, line 36: shouldn't this contain more logic?

protocol/app/vote_extensions/oracle_client.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Prices is a wrapper around the Slinky OracleClient's Prices method.
```
Christopher-Li, line 38: Why is this a wrapper around. the Slinky OracleClient's Prices method when Slinky's OracleClient isn't even referenced?

protocol/app/vote_extensions/extend_vote_handler.go
Eric-Warehime: Because VEs are generated before deliver tx runs. 

When we tested the protocol changes initially we ran into the following scenario:
1. ExtendVote in block `N-1` runs and suggests a price change to price `New`.
2. Prepare/Process Proposal runs in block `N` and the price change is accepted into the block.
3. ExtendVote runs in block `N` and suggests to change the price to `New` since DeliverTx hasn't actually changed the x/prices state.
4. Block `N+1` runs and PrepareProposal creates an `UpdateMarketPrices` Tx which is rejected in ProcessProposal because of the `min_price_change` constraint being violated.

The chain in this case would perpetually propose an UpdateMarketPrices tx which would be rejected and it would never make progress.

The solution to this was to execute the changes to the PricesKeeper in `ExtendVote` before calling `GetValidMarketPriceUpdates`. This way all suggested price changes are based on the state the prices will be in after the current block is applied.

protocol/x/prices/types/types.go
Eric-Warehime: It is already included in the prices keeper https://github.com/dydxprotocol/v4-chain/blob/1aaa2ccdcd6757a31444673d0eb4aa0f0b674a93/protocol/x/prices/keeper/update_price.go#L31 

I'm just exposing it here so that it will show up in the generated mock.

protocol/app/vote_extensions/oracle_client.go
Eric-Warehime: Updated w/ a new name.

protocol/app/vote_extensions/oracle_client.go
Eric-Warehime: We moved the logic for actually querying the sidecar out of the VE handlers and into the daemon. The interface still works as the place in the VE handlers where we get the prices from--just the prices are coming from the keeper here instead of the sidecar.

protocol/app/vote_extensions/extend_vote_handler.go
Christopher-Li: IIuc, Since ExtendVote for block `N` occurs before block `N` is committed, the price updates proposed by block `N` may not be valid? 

I think I'm a little confused too, because I thought if in Block `N+1` received a consensus through vote extension which `New + Delta` where `Delta` is less than `min_price_change`, PrepareProposal would just ignore that price and would not create an `UpdateMarketPrices` Tx  with a price update

Eric-Warehime: > IIuc, Since ExtendVote for block N occurs before block N is committed, the price updates proposed by block N may not be valid?

Correct. The changes are valid as long as we first apply the changes in block N. 

> I think I'm a little confused too, because I thought if in Block N+1 received a consensus through vote extension which New + Delta where Delta is less than min_price_change, PrepareProposal would just ignore that price and would not create an UpdateMarketPrices Tx with a price update

No, the [Default](https://github.com/skip-mev/v4-chain/blob/9cf8a678503243efa2f98af0fd49291121c44b7f/protocol/app/prepare/prices/default_price_update_generator.go#L31) prepare handler which uses the existing logic calls the Prices keeper to generate the tx which means prices with less than min_price_change movement are ignored.


In the [Slinky](https://github.com/skip-mev/v4-chain/blob/9cf8a678503243efa2f98af0fd49291121c44b7f/protocol/app/prepare/prices/slinky_price_update_generator.go#L66) prepare handler we just aggregate over the data in the VEs so the responsibility for making sure that data is valid is in the ExtendVoteHandler.

protocol/app/vote_extensions/extend_vote_handler.go
Eric-Warehime: Added below

protocol/x/prices/keeper/slinky_adapter.go
Eric-Warehime: Ok, added in the most recent commit. Copie `lib/slinky/utils.go` which contains this function from the sidecar integration PR. Should be the single place this logic exists now.

Christopher-Li (approved):

protocol/app/vote_extensions/extend_vote_handler.go
tqin7: We have prices transaction as the last one in a block as things such as order matches are created based on old prices. If we update prices in ExtendVote, that will potentially make other transactions invalid

PR #1152, opened 2024-03-07
[CT-647] initialize new streams and send orderbook snapshot

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/streaming/grpc/grpc_streaming_manager.go
+++ b/protocol/streaming/grpc/grpc_streaming_manager.go
@@ -23,8 +23,8 @@ type GrpcStreamingManagerImpl struct {
 
 // OrderbookSubscription represents a active subscription to the orderbook updates stream.
 type OrderbookSubscription struct {
-	// Indicate whether the subscription has been initialized with a snapshot.
-	initialized bool
+	// Initialize the subscription with orderbook snapshots.
+	initialize sync.Once
 
 	// Clob pair ids to subscribe to.
 	clobPairIds []uint32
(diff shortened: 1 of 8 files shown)


protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -22,8 +23,14 @@ type GrpcStreamingManagerImpl struct {
 
 // OrderbookSubscription represents a active subscription to the orderbook updates stream.
 type OrderbookSubscription struct {
+	// Indicate whether the subscription has been initialized with a snapshot.
+	initialized bool
```
jonfung-dydx, line 27: If this works, it works, but I think `sync.Once` is a cleaner/more performant way to do this singleton/initialize pattern. If we do that we don't have to keep locking and unlocking on every subscription in `GetUninitializedClobPairIds()`.  We can just call the sync.once subscription for every subscription in the orderbook subscription array and be sure the code only runs once.

Maybe timebox exploring this approach? but i don't feel too strongly if this approach works since this is only run on full nodes (which I assume don't have to be as performant as validators) so feel free to ignore.

protocol/x/clob/keeper/keeper.go
```diff
@@ -219,3 +219,21 @@ func (k Keeper) InitMemStore(ctx sdk.Context) {
 func (k *Keeper) SetAnteHandler(anteHandler sdk.AnteHandler) {
 	k.antehandler = anteHandler
 }
+
+// InitializeNewGrpcStreams initializes new gRPC streams for all uninitialized clob pairs.
```
jonfung-dydx, line 223: by sending an orderbook snapshot

jonfung-dydx (approved):

PR #1154, opened 2024-03-07
[CT-631] choose random read replica

dydxwill:
### Changelist
choose random read replica

### Test Plan
tested in dev3
ran Lawrence's script polling `dydx1hhu24usgaxc9lfn879yzv2y5t64057gfwjmvz3` address

### Author/Reviewer Checklist


vincentwschau:
Did we consider just adding a route53 records to do this rather than doing it within the service? [Reference](https://repost.aws/knowledge-center/requests-rds-read-replicas)

Christopher-Li (commented):
instead of doing this this, I think we should hide the read replicas behind an LB

Christopher-Li:
actually nvm this should be fine and is less work

Christopher-Li (approved):

vincentwschau:
DNS / LB I think is still preferred, since it can be changed and applied w/o a release.

PR #1157, opened 2024-03-08
[OTE-209] Emit metrics gated through execution mode

affanv14:
### Changelist
- Emits metrics based list of acceptable exec modes which is passed as an argument

### Test Plan
Added a test which checks which whether the metric is emitted based on the mode pass through

### Author/Reviewer Checklist

### DIFF
--- a/protocol/lib/metrics/lib.go
+++ b/protocol/lib/metrics/lib.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	gometrics "github.com/hashicorp/go-metrics"
 )
 
@@ -12,6 +13,8 @@ import (
 // that supports float64 (i.e hashicorp go-metrics)
 
 type Label = gometrics.Label
+type TelemetryEmitWithLabelsFunc func(key []string, val float32, labels []gometrics.Label)
+type TelemetryEmityFunc func(key []string, val float32)
 
 // IncrCounterWithLabels provides a wrapper functionality for emitting a counter
 // metric with global labels (if any) along with the provided labels.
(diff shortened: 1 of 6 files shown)


protocol/lib/metrics/util.go
```diff
@@ -80,3 +80,18 @@ func GetCallbackMetricFromCtx(ctx sdk.Context) string {
 
 	return DeliverTx
 }
+
+func ContextuallySetGaugeWithLabels(
```
teddyding, line 84: There are other telemetry function that we use often e.g. `IncrCounterWithLabels`

Could we find a more generic pattern?
```
type TelemetryFunc func(keys []string, val float32, labels []gometrics.Label)

func ContextuallyEmitTelemetry(
	ctx types.Context,
	telemetryFunc TelemetryFunc,
	allowedModes []types.ExecMode,
 ......
) {
	contextExecMode := ctx.ExecMode()
	for _, mode := range allowedModes {
		if contextExecMode == mode {
			telemetryFunc(keys, val, labels)
			break // Exit once the correct mode is found and function is executed
		}
	}
}
```

protocol/lib/metrics/util.go
```diff
@@ -80,3 +80,18 @@ func GetCallbackMetricFromCtx(ctx sdk.Context) string {
...
+	ctx sdk.Context,
+	keys []string,
+	val float32,
+	labels []gometrics.Label,
+	allowedModes []sdk.ExecMode,
+) {
+	contextExecMode := ctx.ExecMode()
+	for _, mode := range allowedModes {
+		if contextExecMode == mode {
+			telemetry.SetGaugeWithLabels(keys, val, labels)
```
teddyding, line 94: Nit: add a break

teddyding, line 94: Can you check with @jonfung-dydx if we should be using functions [here](https://github.com/dydxprotocol/v4-chain/blob/454cb0c2a9d9fa58b8f38019b6e1e4569d95c939/protocol/lib/metrics/lib.go#L10)?

jonfung-dydx: can we move these functions over to the lib.go file? and use `SetGaugeWithLabels`?

also nit: `SetGaugeWithLabelsForExecutionMode` since all our logs area already contextualized

protocol/lib/metrics/util.go
affanv14: I see two common method signatures. Added one for each of them

protocol/lib/metrics/lib.go
```diff
@@ -90,3 +93,41 @@ func ModuleMeasureSinceWithLabels(
...
+	key []string,
+	val float32,
+	labels []gometrics.Label,
+) {
+	if isAllowedExecutionMode(ctx, allowedModes) {
+		telemtryFuncWithLabels(key, val, labels)
+	}
+}
+
+func EmitTelemetryForExecMode(
```
jonfung-dydx, line 123: Technically no need to do this since you can ...kwargs the labels parameter and then always use the `metrics.SetXXXXWithLabels` fn

protocol/lib/metrics/util.go
jonfung-dydx: yeah, all those places should be migrated over into one-key 😆 
https://www.notion.so/dydx/Metrics-Guidelines-f31f02e792bc4914abac2309df02e3b0?pvs=4#7e8b21909f404372940d062dde577e0b

affanv14: ah... got it!

jonfung-dydx (approved):

PR #1158, opened 2024-03-08
[TRA-81] Check isolated market constraints in UpdateSubaccount.

vincentwschau:
### Changelist
Adds logic to check whether updates passed into `UpdateSubaccount` are valid under the isolated markets constraints.

### Test Plan
Unit tests.

### Author/Reviewer Checklist


protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -292,6 +292,13 @@ func (k Keeper) UpdateSubaccounts(
 		perpIdToFundingIndex[perp.Params.Id] = perp.FundingIndex
 	}
 
+	// Check if the updates satisfy the isolated perpetual constraints.
+	success, successPerUpdate, err = k.checkIsolatedSubaccountConstraints(
```
teddyding, line 296: Shouldn't we put this in `internalCanUpdateSubaccounts`? I believe we maintain the constraint that `CanUpdateSubaccount` success guarantees `UpdateSubaccounts` success. This will no longer hold true if the input `updates` fails `checkIsolatedSubaccountConstraints`

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+)
+
+// checkIsolatedSubaccountConstaints will validate all `updates` to the relevant subaccounts against
```
teddyding, line 11: For my understanding, do we ever expect these constraints to fail during normal operations? If isolated subaccounts is abstracted away from the user, then upstream `x/clob` logic would be responsible for populating the updates correctly right? So this is more of an internal sanity check?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+		return types.ViolatesIsolatedSubaccountConstraints, nil
+	}
+
+	// There cannot be more than a single perpetual update if an update to an isolated perpetual
+	// exists in the slice of perpetual updates.
+	if hasIsolatedUpdate && len(settledUpdate.PerpetualUpdates) > 1 {
+		return types.ViolatesIsolatedSubaccountConstraints, nil
+	}
+
+	// Note  we can assume that if `hasIsolatedUpdate` is true, there is only a single perpetual
```
teddyding, line 133: Nit: extra space

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+		if !exists {
+			return types.UpdateCausedError, errorsmod.Wrap(
+				perptypes.ErrPerpetualDoesNotExist, lib.UintToString(perpetualUpdate.PerpetualId),
+			)
+		}
+
+		if marketType == perptypes.PerpetualMarketType_PERPETUAL_MARKET_TYPE_ISOLATED {
+			hasIsolatedUpdate = true
+			isolatedUpdatePerpetualId = perpetualUpdate.PerpetualId
+		}
```
teddyding, line 93: `break`?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+		marketType, exists := perpIdToMarketType[perpetualPosition.PerpetualId]
+		if !exists {
+			return types.UpdateCausedError, errorsmod.Wrap(
+				perptypes.ErrPerpetualDoesNotExist, lib.UintToString(perpetualPosition.PerpetualId),
+			)
+		}
+
+		if marketType == perptypes.PerpetualMarketType_PERPETUAL_MARKET_TYPE_ISOLATED {
+			isIsolatedSubaccount = true
+			isolatedPositionPerpetualId = perpetualPosition.PerpetualId
```
teddyding, line 111: `break` to avoid extra iterations

protocol/x/subaccounts/keeper/subaccount.go
shrenujb: +1 to this comment

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return success, successPerUpdate, nil
+}
+
+// Checks whether the perpetual updates to a settled subaccount violates constraints for isolated
+// perpetuals. This function assumes the settled subaccount is valid and does not violate the
+// the constraints.
+// The constraint being checked is:
+//   - a subaccount with a position in an isolated perpetual cannot have updates other perpetuals
```
shrenujb, line 64:
```suggestion
//   - a subaccount with a position in an isolated perpetual cannot have updates with other perpetuals
```

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+	var perpIdToMarketType = make(map[uint32]perptypes.PerpetualMarketType)
+
+	for _, perpetual := range perpetuals {
+		perpIdToMarketType[perpetual.GetId()] = perpetual.Params.MarketType
+	}
+
+	for i, u := range settledUpdates {
+		_, exists := idOfSettledUpdates[*u.SettledSubaccount.Id]
+
+		if exists {
```
shrenujb, line 41:
```suggestion
		if _, exists := idOfSettledUpdates[*u.SettledSubaccount.Id]; exists {
```

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+			return false, nil, types.ErrNonUniqueUpdatesSubaccount
+		}
```
shrenujb, line 43:
```suggestion
		
		if _, exists := idOfSettledUpdates[*u.SettledSubaccount.Id]; exists {
			return false, nil, types.ErrNonUniqueUpdatesSubaccount
		}
```

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+) (types.UpdateResult, error) {
+	// If there are no perpetual updates, then this update does not violate constraints for isolated
+	// markets.
+	if len(settledUpdate.PerpetualUpdates) == 0 {
+		return types.Success, nil
+	}
+
+	// Check if the updates contain an update to an isolated perpetual.
+	hasIsolatedUpdate := false
+	isolatedUpdatePerpetualId := uint32(0)
```
shrenujb, line 81: Is 0 a currently valid perpetual id? If so, can we set this to a value like -1?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,144 @@
...
+			return types.UpdateCausedError, errorsmod.Wrap(
+				perptypes.ErrPerpetualDoesNotExist, lib.UintToString(perpetualUpdate.PerpetualId),
+			)
+		}
+
+		if marketType == perptypes.PerpetualMarketType_PERPETUAL_MARKET_TYPE_ISOLATED {
+			hasIsolatedUpdate = true
+			isolatedUpdatePerpetualId = perpetualUpdate.PerpetualId
+		}
+	}
```
shrenujb, line 94: Do we need to add a constraint that if we see multiple isolated perpetual updates, that would be erroneous?

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: `CanUpdateSubaccounts` has a different constraint from `UpdateSubaccounts` which is why I decided to pull this function into `UpdateSubaccounts` and not `CanUpdateSubaccounts`.
`CanUpdateSubaccounts` does not require unique subaccounts in updates, but `UpdateSubaccounts` does.
There may not be a use of this specific behavior of `CanUpdateSubaccounts`, I can check, however I preferred changing `UpdateSubaccounts` rather than `CanUpdateSubaccounts` after seeing this difference between the 2 functions.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Upstream `x/clob` logic relies on  `UpdateSubaccounts` to determine if the updates to a subaccount for a match are valid.  This isn't an internal sanity check, this would be the function / logic that would ensure orders which result in a match that would lead to an invalid subaccount state (based on these new constraints from isolated markets) result in an error.
There is additional non-scoped work to add logic to `x/clob` to prevent users from even sending orders that would lead to matches that are invalid based on the isolated market constraints.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: perpetual ids are `uint32` so they can't be set to -1. 0 is a valid id, all unsigned integers are valid perpetual ids though.

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// A subaccount with perpetual positions in non-isolated perpetuals cannot have an update
+	// to an isolated perpetual.
+	if !isIsolatedSubaccount && hasPerpetualPositions && hasIsolatedUpdate {
+		return types.ViolatesIsolatedSubaccountConstraints, nil
+	}
+
+	// There cannot be more than a single perpetual update if an update to an isolated perpetual
+	// exists in the slice of perpetual updates.
+	if hasIsolatedUpdate && len(settledUpdate.PerpetualUpdates) > 1 {
```
shrenujb, line 130: Is it possible we have multiple perpetual update entries with the same isolated perpetual?

protocol/x/subaccounts/types/update.go
```diff
@@ -52,11 +52,12 @@ func GetErrorFromUpdateResults(
...
-	NewlyUndercollateralized:       "NewlyUndercollateralized",
-	StillUndercollateralized:       "StillUndercollateralized",
-	WithdrawalsAndTransfersBlocked: "WithdrawalsAndTransfersBlocked",
-	UpdateCausedError:              "UpdateCausedError",
+	Success:                               "Success",
+	NewlyUndercollateralized:              "NewlyUndercollateralized",
+	StillUndercollateralized:              "StillUndercollateralized",
+	WithdrawalsAndTransfersBlocked:        "WithdrawalsAndTransfersBlocked",
+	UpdateCausedError:                     "UpdateCausedError",
+	ViolatesIsolatedSubaccountConstraints: "ViolatesIsolatedSubaccountConstraints",
```
shrenujb, line 60: Is this how the error gets propagated to the client? Do we think we need to provide more info or would this be sufficient?

shrenujb (commented):
Looks good for the most part!

Can we also add an integration test for placing an orders for isolated markets if no other core changes are required?

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: To have multiple perpetual update entires for the same isolated perpetual, the length of the slice of perpetual updates would have to be > 1, given that each [PerpetualUpdate](https://github.com/dydxprotocol/v4-chain/blob/340801a3cc8013a8c6b61a0a3717b48a0fe70aeb/protocol/x/subaccounts/types/update.go#L89-L95) holds only the update for a single perpetual id.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: I also don't have strong opinons on this and would prefer an invalid id. I can use MAX_UINT32 as well, might be a better number since it's unlikely to be a valid perpetual id.

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: Yes, that's true that it can be added to the orderbook since `AddOrderToOrderbookCollatCheck` calls `CanUpdateSubaccounts`. However this is also a path where having a set of updates with non-unique subaccounts are run through `CanUpdateSubaccounts` but never through `UpdateSubaccounts` and just adding the above check to `CanUpdateSubaccounts` would break a working flow.

protocol/x/subaccounts/types/update.go
vincentwschau: There's several more layers of indirection before this error is sent to the client. I'm planning on looking at the error propagation once integration tests / more E2E testing is added.

vincentwschau:
>Can we also add an integration test for placing an orders for isolated markets if no other core changes are required?

Agree on adding more integration / E2E testing. Will add a TODO for this, however I don't want to include and increase scope of this PR.

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: Removed the unique subaccount check in `checkIsolatedSubaccountConstraints` and moved into `internalCanUpdateSubaccounts`. Specified in comments that the function does not consider constraints across multiple updates.

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -511,6 +517,7 @@ func checkPositionUpdatable(
 
 // internalCanUpdateSubaccounts will validate all `updates` to the relevant subaccounts.
 // The `updates` do not have to contain `Subaccounts` with unique `SubaccountIds`.
+// The `updates` do not have to contain `Subaccounts` with unique `SubaccountIds`.
```
shrenujb, line 520: nit: repeated line

shrenujb (approved):
LGTM

teddyding (approved):

PR #1161, opened 2024-03-11
[OTE-214] Calculate current OI and pass to `GetMarginRequirements` calculation

teddyding:
### Changelist
Calculate current OI and pass to GetMarginRequirements calculation 

### Test Plan
Unit tests

### Author/Reviewer Checklist


protocol/x/perpetuals/keeper/perpetual.go
```diff
@@ -1222,6 +1234,30 @@ func (k Keeper) ModifyFundingIndex(
 	return nil
 }
 
+func (k Keeper) ModifyOpenInterest(
```
teddyding, line 1237: Only used for test in this PR, but in a future will be used for actually updating open interest of a perp.

jayy04 (approved):

protocol/x/perpetuals/keeper/perpetual.go
jayy04, line 1237: Add a function comment?

PR #1165, opened 2024-03-12
[CT-515] Combine place, cancel, batch cancel rate limiters

jonfung-dydx:


protocol/x/clob/rate_limit/order_rate_limiter.go
```diff
@@ -193,12 +142,34 @@ func (r *cancelOrderRateLimiter) RateLimit(ctx sdk.Context, msg *types.MsgCancel
 	return err
 }
 
-func (r *cancelOrderRateLimiter) PruneRateLimits(ctx sdk.Context) {
+func (r *placeAndCancelOrderRateLimiter) RateLimitBatchCancelOrder(ctx sdk.Context, msg clobtypes.MsgBatchCancel) (err error) {
+	lib.AssertCheckTxMode(ctx)
+
+	// calcualate how mcuh each batch cancel should be weighted. Use 2 for now.
+	weight := uint32(2)
```
jonfung-dydx, line 149: TODO: check the ratio of how much work it takes to gossip/deserialize/sigverify messages compared to actually processing messages and then set this to be `X + YperId so like ceil(1+ 0.1 per ID)`

protocol/testing/genesis.sh
```diff
@@ -1404,14 +1404,10 @@ function edit_genesis() {
 	dasel put -t int -f "$GENESIS" '.app_state.clob.liquidations_config.fillable_price_config.spread_to_maintenance_margin_ratio_ppm' -v '1500000'  # 150%
 
 	# Block Rate Limit
-	# Max 50 short term orders per block
-	dasel put -t json -f "$GENESIS" '.app_state.clob.block_rate_limit_config.max_short_term_orders_per_n_blocks.[]' -v "{}"
-	dasel put -t int -f "$GENESIS" '.app_state.clob.block_rate_limit_config.max_short_term_orders_per_n_blocks.[0].limit' -v '200'
-	dasel put -t int -f "$GENESIS" '.app_state.clob.block_rate_limit_config.max_short_term_orders_per_n_blocks.[0].num_blocks' -v '1'
-	# Max 50 short term order cancellations per block
-	dasel put -t json -f "$GENESIS" '.app_state.clob.block_rate_limit_config.max_short_term_order_cancellations_per_n_blocks.[]' -v "{}"
```
jonfung-dydx, line 1412: don't know why the comment says 50 per block when it's 200 per block. Set to be 400 per block for the combined version for now.

proto/dydxprotocol/clob/block_rate_limit_config.proto
```diff
@@ -13,8 +13,10 @@ message BlockRateLimitConfiguration {
   // configurations.
   //
   // Specifying 0 values disables this rate limit.
+  // Deprecated in favor of `max_short_term_orders_and_cancels_per_n_blocks`
+  // for v5.x onwards.
   repeated MaxPerNBlocksRateLimit max_short_term_orders_per_n_blocks = 1
-      [ (gogoproto.nullable) = false ];
+      [ (gogoproto.nullable) = false, deprecated = true ];
```
jayy04, line 19: should we just remove this? and mark 1 and 3 as reserved?

jayy04: we haven't removed any fields in production, but I feel like this is probably the way to do it

protocol/x/clob/e2e/rate_limit_test.go
```diff
@@ -94,6 +95,30 @@ func TestRateLimitingOrders_RateLimitsAreEnforced(t *testing.T) {
 			firstMsg:  &CancelOrder_Alice_Num0_Id0_Clob1_GTB5,
 			secondMsg: &CancelOrder_Alice_Num1_Id0_Clob0_GTB20,
 		},
+		"Batch cancellations with same subaccounts": {
+			blockRateLimitConfig: clobtypes.BlockRateLimitConfiguration{
+				MaxShortTermOrdersAndCancelsPerNBlocks: []clobtypes.MaxPerNBlocksRateLimit{
+					{
+						NumBlocks: 2,
+						Limit:     3,
```
jayy04, line 103: for my understanding - why do we need to set limit to 3 here? doesn't batch cancel have a weight of 2? so we can equivalently set this to 2 per 2 blocks?

protocol/x/clob/e2e/rate_limit_test.go
```diff
@@ -170,6 +195,180 @@ func TestRateLimitingOrders_RateLimitsAreEnforced(t *testing.T) {
...
+			firstBatchSuccess: []bool{
+				true,
+				true,
+				true,
+			},
+			secondBatch: []sdktypes.Msg{
+				&PlaceOrder_Alice_Num0_Id0_Clob0_Buy5_Price10_GTB23, // 1-weight success @ 4
+				&CancelOrder_Alice_Num1_Id0_Clob0_GTB20,             // 1-weight success @ 5
+				&BatchCancel_Alice_Num0_Clob0_1_2_3_GTB20,           // 2-weight failure @ 7
+				&CancelOrder_Alice_Num0_Id0_Clob0_GTB23,             // 1-weight failure @ 8
```
jayy04, line 232: oh interesting - I didn't know we increment the counter for failed txns. I guess this has always be the behavior?

protocol/x/clob/e2e/rate_limit_test.go
```diff
@@ -513,6 +712,20 @@ func TestRateLimitingShortTermOrders_GuardedAgainstReplayAttacks(t *testing.T) {
...
+					{
+						NumBlocks: 1,
+						Limit:     3, // TODO FIX THIS ONCE WEIGHTS ARE DECIDED
+					},
+				},
+			},
+			replayLessGTB:    &BatchCancel_Alice_Num0_Clob0_1_2_3_GTB5,
+			replayGreaterGTB: &BatchCancel_Alice_Num0_Clob0_1_2_3_GTB27,
+			firstValidGTB:    &BatchCancel_Alice_Num0_Clob0_1_2_3_GTB20,
+			secondValidGTB:   &BatchCancel_Alice_Num0_Clob0_1_2_3_GTB20,
```
jayy04, line 727: these are the same transactions so would get rejected even without rate limits, right?

nbd but probably better to make one of them `Alice_Num1` like the cancellation tests above

protocol/x/clob/rate_limit/multi_block_rate_limiter.go
```diff
@@ -66,6 +67,10 @@ func NewMultiBlockRateLimiter[K comparable](context string, config []types.MaxPe
 }
 
 func (r *multiBlockRateLimiter[K]) RateLimit(ctx sdk.Context, key K) error {
+	return r.RateLimitIncrBy(ctx, key, 1)
+}
+
+func (r *multiBlockRateLimiter[K]) RateLimitIncrBy(ctx sdk.Context, key K, incrBy uint32) error {
```
jayy04, line 73: nice

protocol/x/clob/rate_limit/order_rate_limiter.go
```diff
@@ -193,12 +144,37 @@ func (r *cancelOrderRateLimiter) RateLimit(ctx sdk.Context, msg *types.MsgCancel
...
 
-func (r *cancelOrderRateLimiter) PruneRateLimits(ctx sdk.Context) {
+func (r *placeAndCancelOrderRateLimiter) RateLimitBatchCancelOrder(
+	ctx sdk.Context,
+	msg types.MsgBatchCancel,
+) (err error) {
+	lib.AssertCheckTxMode(ctx)
+
+	// calcualate how mcuh each batch cancel should be weighted. Use 2 for now.
+	weight := uint32(2)
```
jayy04, line 154: nit: declare constants at top of file

protocol/x/clob/rate_limit/order_rate_limiter.go
jayy04: maybe attach a ticket?

protocol/x/clob/rate_limit/order_rate_limiter.go
```diff
@@ -207,5 +183,6 @@ func (r *cancelOrderRateLimiter) PruneRateLimits(ctx sdk.Context) {
 	for key := range r.rateLimitedAccounts {
 		delete(r.rateLimitedAccounts, key)
 	}
-	r.checkStateShortTermRateLimiter.PruneRateLimits(ctx)
+	r.checkStateShortTermOrderPlaceCancelRateLimiter.PruneRateLimits(ctx)
+	r.checkStateStatefulOrderRateLimiter.PruneRateLimits(ctx)
```
jayy04, line 187: for my understanding - was the stateful order rate limit not getting pruned? or pruned somewhere else?

jayy04 (approved):

protocol/x/clob/rate_limit/order_rate_limiter.go
jonfung-dydx: we previously had:
place rate limiter
- prune short term place
- prune stateful place order
cancel rate limiter
- prune short term cancel

this combines them into both. Github just marked this diff as the one for cancel orders so it appears this was added.

protocol/x/clob/e2e/rate_limit_test.go
jonfung-dydx: I think you mean `BatchCancel_Alice_Num0_Clob1_1_2_3_GTB20` with CLOB1 not alice_num1 right? alice num 1 is the same subaccount owner so the rate limit is shared. I think the above test is incorrect. I will update it.

jonfung-dydx: also added some comments

protocol/x/clob/rate_limit/order_rate_limiter.go
jonfung-dydx: done

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -30,10 +31,41 @@ func perpetualsUpgrade(
 	}
 }
 
+func blockRateLimitConfigUpdate(
+	ctx sdk.Context,
+	clobKeeper clobtypes.ClobKeeper,
+) {
+	// Based off of https://docs.dydx.exchange/trading/rate_limits
+	blockRateLimitConfig := clobtypes.BlockRateLimitConfiguration{
```
roy-dydx, line 39: We should copy the old params from the existing state rather than setting them to constants.

We should initialize the new params to defaults rather than setting them to constants.

jayy04 (approved):

roy-dydx (approved):

PR #1170, opened 2024-03-13
[OTE-218] Add state validity test for perpetual market type

affanv14:
### Changelist
Adds an upgrade test which checks state validity for perpetual market type
Refactors testnet upgrade code so that it can be used across tests

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testing/containertest/testnet_test.go
+++ b/protocol/testing/containertest/testnet_test.go
@@ -295,46 +295,61 @@ func TestUpgrade(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestStateUpgradeForPerpetualMarketType(t *testing.T) {
+func TestStateUpgrade(t *testing.T) {
 	testnet, err := NewTestnetWithPreupgradeGenesis()
 	require.NoError(t, err, "failed to create testnet - is docker daemon running?")
 	err = testnet.Start()
 	require.NoError(t, err)
 	defer testnet.MustCleanUp()
 	node := testnet.Nodes["alice"]
-
 	nodeAddress := constants.AliceAccAddress.String()
-	queryClient := perpetuals.NewQueryClient
-	requestFunction := perpetuals.QueryClient.Perpetual
-	request := &perpetuals.QueryPerpetualRequest{
-		Id: 1,
-	}
 
-	resp, err := Query(
-		node,
-		queryClient,
-		requestFunction,
-		request,
-	)
-	require.NoError(t, err)
-	perpetualObject := &perpetuals.QueryPerpetualResponse{}
-	err = proto.UnmarshalText(resp.String(), perpetualObject)
-	require.NoError(t, err)
-	assert.Equal(t, perpetuals.PerpetualMarketType_PERPETUAL_MARKET_TYPE_UNSPECIFIED, perpetualObject.Perpetual.Params.MarketType)
+	preUpgradeChecks(node, t)
 
 	err = upgradeTestnet(nodeAddress, t, node)
 	require.NoError(t, err)
 
-	resp, err = Query(
+	postUpgradeChecks(node, t)
+}
+
+func preUpgradeChecks(node *Node, t *testing.T) {
+	preUpgradeCheckPerpetualMarketType(node, t)
+}
+
+func postUpgradeChecks(node *Node, t *testing.T) {
+	postUpgradecheckPerpetualMarketType(node, t)
+}
+
+func preUpgradeCheckPerpetualMarketType(node *Node, t *testing.T) {
+	perpetualsList := &perpetuals.QueryAllPerpetualsResponse{}
+	resp, err := Query(
 		node,
-		queryClient,
-		requestFunction,
-		request,
+		perpetuals.NewQueryClient,
+		perpetuals.QueryClient.AllPerpetuals,
+		&perpetuals.QueryAllPerpetualsRequest{},
 	)
 	require.NoError(t, err)
-	err = proto.UnmarshalText(resp.String(), perpetualObject)
+	err = proto.UnmarshalText(resp.String(), perpetualsList)
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


protocol/testing/containertest/testnet_test.go
```diff
@@ -291,7 +290,54 @@ func TestUpgrade(t *testing.T) {
 	require.NoError(t, err)
 	defer testnet.MustCleanUp()
 	node := testnet.Nodes["alice"]
+	nodeAddress := constants.AliceAccAddress.String()
+	err = upgradeTestnet(nodeAddress, t, node)
+	require.NoError(t, err)
+}
+
+func TestStateUpgradeForPerpetualMarketType(t *testing.T) {
```
shrenujb, line 298: I wouldn't make this test specific to perpetual market type but rather something that encompasses all state changes from one version to another.
I think it might be helpful to just call this TestStateUpgrade and move it into the uprade 5.0.0 folder if thats possible

protocol/testing/containertest/testnet_test.go
```diff
@@ -291,7 +290,54 @@ func TestUpgrade(t *testing.T) {
...
+	resp, err = Query(
+		node,
+		queryClient,
+		requestFunction,
+		request,
+	)
+	require.NoError(t, err)
+	err = proto.UnmarshalText(resp.String(), perpetualObject)
+	require.NoError(t, err)
+	assert.NotEqual(t, perpetuals.PerpetualMarketType_PERPETUAL_MARKET_TYPE_UNSPECIFIED, perpetualObject.Perpetual.Params.MarketType)
```
shrenujb, line 337: assert should check if the state was updated to CROSS since this is what the upgrade code does

protocol/testing/containertest/testnet_test.go
```diff
@@ -291,7 +290,54 @@ func TestUpgrade(t *testing.T) {
...
+	require.NoError(t, err, "failed to create testnet - is docker daemon running?")
+	err = testnet.Start()
+	require.NoError(t, err)
+	defer testnet.MustCleanUp()
+	node := testnet.Nodes["alice"]
+
+	nodeAddress := constants.AliceAccAddress.String()
+	queryClient := perpetuals.NewQueryClient
+	requestFunction := perpetuals.QueryClient.Perpetual
+	request := &perpetuals.QueryPerpetualRequest{
```
shrenujb, line 309: We should do this for all perpetuals and not just 1

protocol/testing/containertest/testnet_test.go
```diff
@@ -291,7 +290,54 @@ func TestUpgrade(t *testing.T) {
...
+	require.NoError(t, err)
+	perpetualObject := &perpetuals.QueryPerpetualResponse{}
+	err = proto.UnmarshalText(resp.String(), perpetualObject)
+	require.NoError(t, err)
+	assert.Equal(t, perpetuals.PerpetualMarketType_PERPETUAL_MARKET_TYPE_UNSPECIFIED, perpetualObject.Perpetual.Params.MarketType)
+
+	err = upgradeTestnet(nodeAddress, t, node)
+	require.NoError(t, err)
 
+	resp, err = Query(
```
shrenujb, line 328: Might be useful to put everything below in a separate helper function `postUpgradeChecks`

protocol/testing/containertest/testnet_test.go
affanv14: Good point! made the change so that its easier to add other state change tests. 

With regards to move it to upgrade 5.0.0 folder - isnt that for unit tests? all container tests are in the current folder

protocol/testing/containertest/testnet_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build all || container_test
```
teddyding, line 1: Did you mean to revert this?

protocol/testing/containertest/testnet_test.go
```diff
@@ -291,7 +290,69 @@ func TestUpgrade(t *testing.T) {
...
+	preUpgradeChecks(node, t)
+
+	err = upgradeTestnet(nodeAddress, t, node)
+	require.NoError(t, err)
+
+	postUpgradeChecks(node, t)
+}
+
+func preUpgradeChecks(node *Node, t *testing.T) {
+	preUpgradeCheckPerpetualMarketType(node, t)
```
teddyding, line 316: Add a comment here for others to add more tests
```
// Add test for your upgrade handler logic below
```

protocol/testing/containertest/testnet_test.go
shrenujb: I don't think it needs to be just for unit tests. It'll be nice to have this in the 5.0.0 folder because the next time someone is doing this for the next upgrade, they can just copy paste the 5.0.0 folder and have the entire basic framework for the upgrade changes

@teddyding to confirm if we are able to move this to the 5.0.0 folder

protocol/testing/containertest/testnet_test.go
affanv14: Good catch! didnt mean to remove this.

protocol/testing/containertest/testnet_test.go
teddyding: Yeah agree with @shrenujb - we don't strictly differentiate tests based on file paths, but instead by functionality. Most custom modules (e.g. `x/clob`) have both unit and e2e tests within their directories. So it makes sense to all testing related to `v5.0.0` under `protocol/app/upgrades/v5.0.0`. 

We can highlight the test type by naming the file something like `upgrade_container_test` under `protocol/app/upgrades/v5.0.0`

affanv14: Since the function depends significantly on container tests, moving the function over and importing references causes a circular dependency. Maybe creating another upgrade test folder within `protocol/testing` would be a better idea?

teddyding (approved):

PR #1172, opened 2024-03-13
[OTE-236] Evaluate subaccounts update end state with OIMF 

teddyding:
### Changelist
- Evaluate subaccounts update end state with OIMF, by temporarily modify perpetual state and reverting the state afterwards
- Set up existing tests with default perp OIs 

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testing/e2e/gov/wind_down_market_test.go
+++ b/protocol/testing/e2e/gov/wind_down_market_test.go
@@ -146,7 +146,7 @@ func TestWindDownMarketProposal(t *testing.T) {
 						genesisState.Params = constants.PerpetualsGenesisParams
 						genesisState.LiquidityTiers = constants.LiquidityTiers
 						genesisState.Perpetuals = []perptypes.Perpetual{
-							constants.BtcUsd_20PercentInitial_10PercentMaintenance,
+							constants.BtcUsd_20PercentInitial_10PercentMaintenance_OpenInterest1,
 						}
 					},
 				)


protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -609,6 +609,17 @@ func (k Keeper) internalCanUpdateSubaccounts(
 		}
 	}
 
+	var perpOpenInterestDelta *perptypes.OpenInterestDelta
+	// If update type is `Match`, calculate the open interest delta for the updated perpetual.
+	// Note: this assumes that `Match` can only happen for perpetuals.
```
teddyding, line 614: AFAIK this assumption will hold true for a very long time (we don't have plans to support spot), which is why I didn't include a TODO.

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -631,11 +642,52 @@ func (k Keeper) internalCanUpdateSubaccounts(
 			}
 		}
 
+		// Temporily apply open interest delta to perpetuals, so IMF is calculated based on open interest after the update.
+		// `perpOpenInterestDeltas` is only present for `Match` update type.
+		if perpOpenInterestDelta != nil {
+			if err := k.perpetualsKeeper.ModifyOpenInterest(
```
teddyding, line 648: Updating `perp.OpenInterest += delta` and then `perp.OpenInterest -= delta` technically modifies app hash as well. For `DeliverTx` this should be fine since this is called deterministically. For `CheckTx`, no need to keep each node `CheckState` the same

protocol/x/subaccounts/keeper/subaccount_test.go
```diff
@@ -3146,20 +3215,386 @@ func TestCanUpdateSubaccounts(t *testing.T) {
...
+					// 900_000 USDC (just enough to colalteralize 90 BTC at $50_000 and 20% IMF)
+					Quantums: dtypes.NewInt(900_000_000_000),
+				},
+			},
+			additionalTestSubaccounts: []types.Subaccount{
+				{
+					Id: &types.SubaccountId{
+						Owner:  "Bob",
+						Number: 0,
+					},
```
jayy04, line 3253: mega nit: just use constants.Bob_Num0

jayy04 (approved):

protocol/x/subaccounts/keeper/oimf.go
```diff
@@ -0,0 +1,89 @@
...
+	if len(settledUpdates[0].PerpetualUpdates) != 1 || len(settledUpdates[1].PerpetualUpdates) != 1 {
+		panic(
+			fmt.Sprintf(
+				types.ErrMatchUpdatesMustUpdateOnePerp,
+				settledUpdates,
+			),
+		)
+	}
+
+	if settledUpdates[0].PerpetualUpdates[0].PerpetualId != settledUpdates[1].PerpetualUpdates[0].PerpetualId {
```
vincentwschau, line 71: nit: The 2 perpetual updates should also have the same absolute value as their size right? If a subaccount is chaing by +1 BTC from a match, the other subaccount should be changing -1 BTC.

vincentwschau:
non-blocking nit, assume it may already to scoped to a different ticket. It would be good to update / add some E2E tests for this logic as well.

vincentwschau (approved):

PR #1175, opened 2024-03-14
[TRA-120] move cancel order msg server logic into keeper

tqin7:
### Changelist
move cancel stateful order logic from msg server to keeper

### Test Plan
existing tests are sufficient

### Author/Reviewer Checklist


protocol/x/clob/keeper/orders.go
```diff
@@ -241,6 +244,97 @@ func (k Keeper) PlaceShortTermOrder(
 	return orderSizeOptimisticallyFilledFromMatchingQuantums, orderStatus, err
 }
 
+// DeliverMsgCancelOrder delivers a MsgCancelOrder.
+func (k Keeper) DeliverMsgCancelOrder(
```
jayy04, line 248: nit: should we just leave this in `msg_server_cancel_order.go`? `orders.go` is kind of huge already

jayy04, line 248: nit: let's maybe call this `HandleMsgCancelOrder`? can also add a `lib.AssertDeliverTxMode(ctx)` or sometihng

protocol/x/clob/keeper/orders.go
```diff
@@ -241,6 +244,97 @@ func (k Keeper) PlaceShortTermOrder(
 	return orderSizeOptimisticallyFilledFromMatchingQuantums, orderStatus, err
 }
 
+// DeliverMsgCancelOrder delivers a MsgCancelOrder.
```
jayy04, line 247: Elaborate a little bit, something like this persists order cancellations on chain and sends out onchain indexer events

jayy04 (approved):

PR #1176, opened 2024-03-14
[OTE-221] Add query for PendingSendPacket

affanv14:
### Changelist
- Adds a query endpoint to check number of pending send packets

### Test Plan
- Added unit test to check query endpoint
- Added integration test for cobra to check if response threw an error
- ran a local testnet and queried for sanity

### Author/Reviewer Checklist


proto/dydxprotocol/ratelimit/query.proto
```diff
@@ -23,6 +24,12 @@ service Query {
     option (google.api.http).get =
         "/dydxprotocol/v4/ratelimit/capacity_by_denom";
   }
+  // Get all pending send packets
+  rpc GetAllPendingSendPackets(GetAllPendingSendPacketsRequest)
```
teddyding, line 28: Nit: conventionally, we name rpc methods without "Query" or "Get", but prefix the request object name with "Query". See [example](https://github.com/dydxprotocol/v4-chain/blob/d7333d4dbba60f20b6225d8ef788282775894e81/proto/dydxprotocol/perpetuals/query.proto#L20)

protocol/x/ratelimit/types/keys.go
```diff
@@ -1,6 +1,10 @@
 package types
 
-import fmt "fmt"
+import (
+	"bytes"
+	fmt "fmt"
```
teddyding, line 5:
```suggestion
	"fmt"
```

protocol/x/ratelimit/types/keys.go
```diff
@@ -27,3 +31,16 @@ const ()
 func GetPendingSendPacketKey(channelId string, sequenceNumber uint64) []byte {
 	return []byte(fmt.Sprintf("%s_%d", channelId, sequenceNumber))
 }
+
+func SplitPendingSendPacketKey(key []byte) (channelId string, sequenceNumber uint64) {
```
teddyding, line 35: Nit: can we add a unit test for this helper?

protocol/x/ratelimit/types/keys.go
```diff
@@ -27,3 +31,16 @@ const ()
...
+	parts := bytes.Split(key, []byte("_"))
+	if len(parts) != 2 {
+		panic(fmt.Sprintf("unexpected key format: %s", key))
```
teddyding, line 38: What does user see if we panic here? I assume nothing or some internal error? 

Let's:
- Log an error here so we could see this on server side
- Also return this error so the user can see this error as well (easier for troubleshooting)

The error string could also be more helpful (e.g. "unexpected PendingSendPacket key format...")

protocol/x/ratelimit/types/keys.go
teddyding: The unit test can test that: given `channelId` and `sequenceNumber` tuples, 
```
SplitPendingSendPacketKey(
  GetPendingSendPacketKey(channelId, sequenceNumber),
) == channelId, sequenceNumber
```

protocol/x/ratelimit/types/keys.go
```diff
@@ -27,3 +31,18 @@ const ()
 func GetPendingSendPacketKey(channelId string, sequenceNumber uint64) []byte {
 	return []byte(fmt.Sprintf("%s_%d", channelId, sequenceNumber))
 }
+
+func SplitPendingSendPacketKey(key []byte) (channelId string, sequenceNumber uint64, err error) {
+	err = nil
+	parts := bytes.Split(key, []byte("_"))
+	if len(parts) != 2 {
+		err = fmt.Errorf("unexpected PendingSendPacket key format: %s", key)
+		return
```
teddyding, line 40: nit: return value explicitly.

```
return "", 0, fmt.Errorf(...)
```

protocol/x/ratelimit/types/keys.go
```diff
@@ -27,3 +31,18 @@ const ()
...
+	}
+	channelId = string(parts[0])
+	// convert parts[1] to uint64 parts[1] is is a byte array with numeric characters of variable length
+
+	sequenceNumberInt, _ := strconv.Atoi(string(parts[1]))
+	sequenceNumber = uint64(sequenceNumberInt)
+	return
```
teddyding, line 47: nit: return value explicitly.



return channelId, sequenceNumber, nil

teddyding (approved):

PR #1178, opened 2024-03-14
[TRA-80] Update subaccount-to-subaccount transfers for collateral pools.

vincentwschau:
### Changelist
Added a new function to `x/subaccounts` keeper to transfer funds from subaccount to subaccount and send coins between collateral pools. Updated `x/sending` to use the new function instead of `UpdateSubaccounts`.

### Test Plan
Added new unit tests.

### Author/Reviewer Checklist


protocol/x/subaccounts/keeper/transfer_test.go
```diff
@@ -574,6 +574,599 @@ func TestWithdrawFundsFromSubaccountToAccount_DepositFundsFromAccountToSubaccoun
...
+			senderCollateralPoolBalance:    big.NewInt(500),
+			recipientCollateralPoolBalance: big.NewInt(600),
+			senderCollateralPoolAddr: authtypes.NewModuleAddress(
+				types.ModuleName + ":" + lib.UintToString(constants.PerpetualPosition_OneISOShort.PerpetualId),
+			),
+			recipientCollateralPoolAddr: types.ModuleAddress,
+			quantums:                    big.NewInt(500),
+			expectedErr:                 types.ErrFailedToUpdateSubaccounts,
+		},
+		"Send from non-isolated subaccount to isolated subaccount, collateral pool does not have enough balance": {
```
jayy04, line 947: for my understanding - collateral pools should never not have enough balance, right? should we panic instead of returning an error?

jayy04 (approved):

protocol/x/subaccounts/keeper/transfer_test.go
vincentwschau: Collateral pools could end up with not enough balance, in the same way the `subaccounts` module account could end up with not enough balance if withdrawal gating doesn't catch a negative TNC subaccount in time.
In terms of panic'ing, I think that could be fine, however it would diverge from the withdrawal flow, which returns an error.

shrenujb (commented):
Core logic looks sound to me! Could you please add integration tests for the new transfer functionality as well?

PR #1179, opened 2024-03-15
[TRA-116] add x/vault total shares store

tqin7:
### Changelist
Add `TotalShares` store (and getter and setter methods)

### Test Plan
added unit tests in this PR

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testutil/constants/vault.go
+++ b/protocol/testutil/constants/vault.go
@@ -1,0 +1,16 @@
+package constants
+
+import (
+	vaulttypes "github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+var (
+	Vault_Clob_0 = vaulttypes.VaultId{
+		Type:   vaulttypes.VaultType_VAULT_TYPE_CLOB,
+		Number: 0,
+	}
+	Vault_Clob_1 = vaulttypes.VaultId{
+		Type:   vaulttypes.VaultType_VAULT_TYPE_CLOB,
+		Number: 1,
+	}
+)
--- a/protocol/x/vault/keeper/shares.go
+++ b/protocol/x/vault/keeper/shares.go
@@ -1,0 +1,34 @@
+package keeper
+
+import (
+	"cosmossdk.io/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+// GetTotalShares gets TotalShares for a vault.
+func (k Keeper) GetTotalShares(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+) (val types.NumShares, exists bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.TotalSharesKeyPrefix))
+
+	b := store.Get(vaultId.ToStateKey())
+	if b == nil {
+		return val, false
+	}
+
+	k.cdc.MustUnmarshal(b, &val)
+	return val, true
+}
+
+// SetTotalShares sets TotalShares for a vault.
+func (k Keeper) SetTotalShares(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+	totalShares types.NumShares,
+) {
+	b := k.cdc.MustMarshal(&totalShares)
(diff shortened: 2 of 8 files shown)


protocol/testutil/constants/vault.go
```diff
@@ -0,0 +1,16 @@
+package constants
+
+import (
+	vaulttypes "github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+var (
+	Vault_Clob_0 = vaulttypes.VaultId{
+		Type:   vaulttypes.VaultType_VAULT_TYPE_CLOB,
+		Number: 0,
```
Christopher-Li, line 10: why is this field named `Number` instead of like `Id`?

protocol/x/vault/keeper/shares.go
```diff
@@ -0,0 +1,34 @@
...
+// GetTotalShares gets TotalShares for a vault.
+func (k Keeper) GetTotalShares(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+) (val types.NumShares, exists bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.TotalSharesKeyPrefix))
+
+	b := store.Get(vaultId.ToStateKey())
+	if b == nil {
+		return val, false
```
Christopher-Li, line 18: how is `val` being returned here without being initialized? Is `val` being in the return of this function also initialize it? Does that also mean `exists` is initialized?

protocol/x/vault/keeper/shares.go
```diff
@@ -0,0 +1,34 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+// GetTotalShares gets TotalShares for a vault.
+func (k Keeper) GetTotalShares(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+) (val types.NumShares, exists bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.TotalSharesKeyPrefix))
```
Christopher-Li, line 14: This is finding the existing keyvalue store that is associated with this keeper? Why do we use this instead of just storing this KV store directly in the keeper in `x/vault/keeper.go`?

protocol/x/vault/keeper/shares.go
```diff
@@ -0,0 +1,34 @@
...
+	if b == nil {
+		return val, false
+	}
+
+	k.cdc.MustUnmarshal(b, &val)
+	return val, true
+}
+
+// SetTotalShares sets TotalShares for a vault.
+func (k Keeper) SetTotalShares(
```
Christopher-Li, line 26: I wonder if instead of having `SetTotalShares`, we should have `incrementTotalShares`, `decrementTotalShares`, or `resetTotalShares`(setting total shares to 0). It seems like we will never set totalShares to an arbitrary number, so `SetTotalShares` would require additional arithmetic.

nit: Should we panic on `totalShares` being a negative number?

protocol/x/vault/keeper/shares.go
vincentwschau: Yes, that's correct. See [here](https://go.dev/tour/basics/7#:~:text=Go's%20return%20values%20may%20be,returns%20the%20named%20return%20values.) for more details.

protocol/x/vault/keeper/shares.go
tqin7: We usually create a separate store (within `KVStore`) for each "functionality"

protocol/x/vault/keeper/shares.go
tqin7: Yeah we can build these increment decrement on top of `Get` and `Set` for sure.

As for verifying whether `totalShares` is non-negative, perhaps we can add that later when building out `MsgDepositToVault`?

Christopher-Li (approved):

vincentwschau (approved):

protocol/x/vault/keeper/shares.go
tqin7: chatted with Vincent offline about this check on negative. will change this to return error if negative

PR #1182, opened 2024-03-15
perf: optimize `slinky_adapter` via in-memory cache [SKI-13]

nivasan1:
### Changelist
- Introduces a `CurrencyPairIDCache` to the `PricesKeeper`
- Reads from the `slinky_adapter` hit the cache first
- Caching semantics
    - Stores an in-memory cache between `currency-pair-id -> currency-pair`
    - `ticker -> ID`
    - The cache is updated dynamically, as new currency-pairs are queried + placed in vote-extensions
 
### Test Plan
- unit-tests, tested via petri

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/prices/keeper/currency_pair_id_cache.go
+++ b/protocol/x/prices/keeper/currency_pair_id_cache.go
@@ -1,0 +1,56 @@
+package keeper
+
+import "sync"
+
+// CurrencyPairIDCache handles the caching logic of currency-pairs to their corresponding IDs. This 
+// data-structure is thread-safe, allowing concurrent reads + synchronized writes.
+type CurrencyPairIDCache struct {
+	// ID -> CurrencyPair
+	idToCurrencyPair map[uint64]string
+	// CurrencyPair -> ID
+	currencyPairToID map[string]uint64
+	// lock
+	sync.RWMutex
+}
+
+// NewCurrencyPairIDCache creates a new CurrencyPairIDCache
+func NewCurrencyPairIDCache() *CurrencyPairIDCache {
+	return &CurrencyPairIDCache{
+		idToCurrencyPair: make(map[uint64]string),
+		currencyPairToID: make(map[string]uint64),
+	}
+}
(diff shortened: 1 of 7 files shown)


protocol/x/prices/keeper/market_param.go
```diff
@@ -48,6 +49,20 @@ func (k Keeper) ModifyMarketParam(
...
+	
+	// if the market pair has been changed, we need to update the in-memory market pair cache
+	if existingParam.Pair != updatedMarketParam.Pair {
+		// remove the old cache entry
+		k.currencyPairIDCache.Remove(uint64(existingParam.Id))
+
+		// add the new cache entry
+		cp, err := slinky.MarketPairToCurrencyPair(updatedMarketParam.Pair)
+		if err == nil {
+			k.currencyPairIDCache.AddCurrencyPair(uint64(updatedMarketParam.Id), cp.String())
```
Christopher-Li, line 61: didn't we decide on using the same format for currencyPair.string() and marketParam.Pair? Shouldn't these two always be equal

protocol/x/prices/keeper/currency_pair_id_cache.go
```diff
@@ -0,0 +1,44 @@
+package keeper
+
+// CurrencyPairIDCache handles the caching logic of currency-pairs to their corresponding IDs
+type CurrencyPairIDCache struct {
```
Christopher-Li, line 4: can you add a comment that this cache is lazily updated and does not necessarily contain all CPs?

protocol/x/prices/keeper/market_param.go
nivasan1: currency-pairs in slinky are marshalled as `BASE/QUOTE` whereas it looks like DYDX handles them as `BASE-QUOTE` hence the conversion. The same conversion is performed in the slinky-price-daemon as well

Christopher-Li (approved):

PR #1186, opened 2024-03-15
Emit metric to track withdrawal amounts

affanv14:
### Changelist
Emits metric only on finalize which keeps track of withdrawal amounts
Correct a slight typo in name

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new telemetry metrics for rate limit withdrawals, including insufficient withdrawal amount and withdrawal amount metrics.

### DIFF
--- a/protocol/x/ratelimit/keeper/keeper.go
+++ b/protocol/x/ratelimit/keeper/keeper.go
@@ -77,6 +77,15 @@ func (k Keeper) ProcessWithdrawal(
 			)
 		}
 
+		metrics.EmitTelemetryWithLabelsForExecMode(
+			ctx,
+			[]sdk.ExecMode{sdk.ExecModeFinalize},
+			metrics.IncrCounterWithLabels,
+			metrics.RateLimitWithdrawalAmount,
+			metrics.GetMetricValueFromBigInt(amount),
+			metrics.GetLabelForStringValue(metrics.RateLimitDenom, denom),
+		)
+
 		// Debit each capacity in the list by the amount of withdrawal.
 		newCapacityList[i] = dtypes.NewIntFromBigInt(
 			new(big.Int).Sub(
--- a/protocol/x/ratelimit/keeper/keeper.go
+++ b/protocol/x/ratelimit/keeper/keeper.go
@@ -77,15 +77,6 @@ func (k Keeper) ProcessWithdrawal(
 			)
 		}
 
-		metrics.EmitTelemetryWithLabelsForExecMode(
-			ctx,
-			[]sdk.ExecMode{sdk.ExecModeFinalize},
-			metrics.IncrCounterWithLabels,
-			metrics.RateLimitWithdrawalAmount,
-			metrics.GetMetricValueFromBigInt(amount),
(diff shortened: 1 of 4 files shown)


protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -77,6 +77,15 @@ func (k Keeper) ProcessWithdrawal(
 			)
 		}
 
+		metrics.EmitTelemetryWithLabelsForExecMode(
```
teddyding, line 80: Why put this in the for loop?

If there's no `CapacityList` for the denom then we won't execute this logic. Should probably place this at the start of this function?

teddyding: Additionally can we also add a counter `withdrawal_rate_limited` [here](https://github.com/dydxprotocol/v4-chain/blob/572244641e56a6446707f913a4f2c0ce7c6f9fa9/protocol/x/ratelimit/keeper/packet.go#L173) so we know how many times withdrawals are getting rate-limited. Thanks!

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -77,6 +77,15 @@ func (k Keeper) ProcessWithdrawal(
...
+			ctx,
+			[]sdk.ExecMode{sdk.ExecModeFinalize},
+			metrics.IncrCounterWithLabels,
```
teddyding, line 83: Instead of a `IncrCounter` here should we use `AddSample` instead? Thinking about how to easily graph this in Datadog.

We are interested in amount withdrawn with a certain time window. Similarly we have this DD [chart](https://app.datadoghq.com/dashboard/fbc-gkh-ywa/v4-clob?fromUser=true&fullscreen_end_ts=1711486952418&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=edit&fullscreen_start_ts=1710882152418&fullscreen_widget=3228782431089520&refresh_mode=sliding&view=spans&from_ts=1710882149774&to_ts=1711486949774&live=true) for amount liquidated in some window

protocol/x/ratelimit/keeper/keeper.go
```diff
@@ -77,6 +77,15 @@ func (k Keeper) ProcessWithdrawal(
...
+			metrics.RateLimitWithdrawalAmount,
+			metrics.GetMetricValueFromBigInt(amount),
```
teddyding, line 85: We should not emit `amount` as a label. For some context, search for "infinite cardinality" in Slack

protocol/x/ratelimit/keeper/keeper.go
affanv14: ah... this was an error on my part. ')' looked like '}' . Will fix

protocol/x/ratelimit/keeper/keeper.go
affanv14: amount isnt a label, its a value field

teddyding (approved):

affanv14:
https://github.com/Mergifyio backport release/protocol/v5.x

PR #1187, opened 2024-03-16
[TRA-102] add subaccountNumber field in assetPositions return result

shrenujb:
### Changelist
Add subAccountNumber as a field on AssetPositionResponseObject

### Test Plan
Unit tests

### Author/Reviewer Checklist


indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
```diff
@@ -52,8 +52,8 @@ const router: express.Router = express.Router();
 const controllerName: string = 'asset-positions-controller';
 
 @Route('assetPositions')
-class AddressesController extends Controller {
```
vincentwschau, line 55: This is a breaking change for the API, it should be created as a separate API and not replacing the current API.

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
```diff
@@ -159,11 +161,14 @@ router.get(
       subaccountNumber,
     }: AssetPositionRequest = matchedData(req) as AssetPositionRequest;
 
+    // The schema checks allow subaccountNumber to be a string, but we know it's a number here.
```
vincentwschau, line 164: Doesn't `subaccountNumber` already have the type `number`?

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
```diff
@@ -52,8 +52,8 @@ const router: express.Router = express.Router();
 const controllerName: string = 'asset-positions-controller';
 
 @Route('assetPositions')
-class AddressesController extends Controller {
-  @Get('/')
+class AssetPositionsController extends Controller {
+  @Get('/:address/:subaccountNumber')
```
vincentwschau, line 56: This should still be `/` because the parameters are query parameters and not path parameters.

vincentwschau (approved):

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
shrenujb: Yes, but that only applies to compile time checks. Someone can still pass in a string and the schema checks let that through

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
shrenujb: Those should still be the same?

I only add this because the other api will look like
`/:address/parentSubaccountNumber/:parentSubaccountNumber`

vincentwschau: If you look at the doc changes, this changes the `v4/assetPositions` documentation to have the URL be `fetch('https://dydx-testnet.imperator.co/v4/assetPositions/{address}/{subaccountNumber}?address=string&subaccountNumber=0',` which I don't think is correct as there are additional path parameters which was not added in this PR.

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
vincentwschau: yea, makes sense, I see that there's no explicit conversion to `number` only the `as` operator that tells typescript to treat the data as an `AssetPositionRequest`.

indexer/services/comlink/src/controllers/api/v4/asset-positions-controller.ts
shrenujb: This just allows you to add the args as path parameters. I think you still have the option to specify the args in the payload instead?

vincentwschau: `'https://dydx-testnet.imperator.co/v4/assetPositions/{address}/{subaccountNumber}?address=string&subaccountNumber=0'` is not a valid URL today. This means the documentation update is wrong right?
We don't want to change any URLs that work today, so I don't think adding the the args as path parameters is correct for the existing endpoint.

PR #1188, opened 2024-03-18
[TRA-126] Add E2E test for isolated subaccount transfers.

vincentwschau:
### Changelist
Add E2E test for transfers from / to isolated subaccounts.

Additional changes:
- cleaned up some tech debt I found where subaccounts with no perpetual positions should have `nil` and not an empty slice for test constants
- cleaned up some tech debt I found where perpetual positions should be initialized with a 0 funding index rather than the default value for the byte array for test constants

### Test Plan
Ran test.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/testutil/constants/subaccounts.go
+++ b/protocol/testutil/constants/subaccounts.go
@@ -12,7 +12,7 @@ var (
 		AssetPositions: []*satypes.AssetPosition{
 			&Usdc_Asset_10_000,
 		},
-		PerpetualPositions: []*satypes.PerpetualPosition{},
+		PerpetualPositions: nil,
 	}
 	Alice_Num0_100_000USD = satypes.Subaccount{
 		Id: &Alice_Num0,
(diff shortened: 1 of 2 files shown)


protocol/testutil/constants/subaccounts.go
```diff
@@ -80,8 +108,9 @@ var (
 		},
 		PerpetualPositions: []*satypes.PerpetualPosition{
 			{
-				PerpetualId: 0,
-				Quantums:    dtypes.NewInt(-10_000_000_000), // -100 BTC
+				PerpetualId:  0,
+				Quantums:     dtypes.NewInt(-10_000_000_000), // -100 BTC
+				FundingIndex: dtypes.NewInt(0),
```
shrenujb, line 113: Why were all of these changed? Doesn't seem related to this PR?

protocol/x/sending/e2e/isolated_subaccount_transfers_test.go
```diff
@@ -0,0 +1,394 @@
...
+			expectedCollateralPoolBalances: map[string]int64{
+				authtypes.NewModuleAddress(
+					satypes.ModuleName + ":" + lib.UintToString(constants.IsoUsd_IsolatedMarket.Params.Id),
+				).String(): 9_900_000_000, // $9,900 USDC
+				authtypes.NewModuleAddress(
+					satypes.ModuleName + ":" + lib.UintToString(constants.Iso2Usd_IsolatedMarket.Params.Id),
+				).String(): 10_100_000_000, // $10,100 USDC
+			},
+			expectedErr: "",
+		},
```
shrenujb, line 154: Another possible test case?
- collateral transfer from isolated subaccount to isolated subaccount with the same isolated perpetual, doesn't result in underlying transfer of collateral pool

protocol/testutil/constants/subaccounts.go
vincentwschau: I have it in the description:
>cleaned up some tech debt I found where perpetual positions should be initialized with a 0 funding index rather than the default value for the byte array for test constants

Mainly did this because I had to change it for the constants I used, which made the constants file inconsistent.

protocol/x/sending/e2e/isolated_subaccount_transfers_test.go
vincentwschau: Good idea, added.

shrenujb (approved):

PR #1193, opened 2024-03-18
feat: Enforce MarketParam.Pair uniqueness constraint

Eric-Warehime:
### Changelist
Enforces a uniqueness constraint on Create/Update operations for MarketParams.
This is required for Slinky since it uses Base/Quote as a compound primary key to reason about pairs being posted to chain as well as prices retrieved from the sidecar.

### Test Plan
Unit tests included. 

Mainnet data also shows no existing pairs with duplicates, so this should just prevent things in addition to avoiding this in governance votes in the interim.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/prices/keeper/market_param.go
+++ b/protocol/x/prices/keeper/market_param.go
@@ -44,6 +44,13 @@ func (k Keeper) ModifyMarketParam(
 			errorsmod.Wrapf(types.ErrMarketExponentCannotBeUpdated, lib.UintToString(updatedMarketParam.Id))
 	}
 
+	// Validate base/quote uniqueness
+	for _, market := range k.GetAllMarketParams(ctx) {
+		if market.Pair == updatedMarketParam.Pair && market.Id != updatedMarketParam.Id {
+			return types.MarketParam{}, errorsmod.Wrapf(types.ErrMarketParamPairAlreadyExists, updatedMarketParam.Pair)
+		}
+	}
+
 	// Store the modified market param.
 	marketParamStore := k.getMarketParamStore(ctx)
 	b := k.cdc.MustMarshal(&updatedMarketParam)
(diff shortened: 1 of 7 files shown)


protocol/x/prices/keeper/market_test.go
```diff
@@ -170,15 +170,29 @@ func TestCreateMarket_Errors(t *testing.T) {
...
 			).Error(),
 		},
 	}
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-			ctx, keeper, _, _, _, _ := keepertest.PricesKeepers(t)
+			ctx, keeper, _, _, _, mockTimeKeeper := keepertest.PricesKeepers(t)
 			ctx = ctx.WithTxBytes(constants.TestTxBytes)
 
+			mockTimeKeeper.On("Now").Return(constants.TimeT)
```
Christopher-Li, line 193: Why do we need a timekeeper now?

Eric-Warehime: The other test cases never succesfully create a market param. They all fail before hitting this line https://github.com/dydxprotocol/v4-chain/blob/ab14ab8b82c13c08afa1226f51a945ad3f00093c/protocol/x/prices/keeper/market.go#L64
The additional test case relies on a market param already existing so it has to be able to succeed once.

Christopher-Li (approved):

protocol/x/prices/keeper/market_param.go
```diff
@@ -44,6 +44,13 @@ func (k Keeper) ModifyMarketParam(
 			errorsmod.Wrapf(types.ErrMarketExponentCannotBeUpdated, lib.UintToString(updatedMarketParam.Id))
 	}
 
+	// Validate base/quote uniqueness
```
nivasan1, line 47: This doesn't seem necessary? Isn't there already a check that the updatedMarket must already exist in state + we've introduced a uniqueness constraint across all market-pairs in state?

nivasan1: Looks like this check is actually necessary, as it's possible for a `MarketParam` update to change the marketParam.Pair, and the updated Pair may conflict with one in state

Christopher-Li (changes requested):
great catch nikhil, i'm going to request changes for updates until we finalize

Christopher-Li (approved):

PR #1200, opened 2024-03-19
[TRA-139] Add logic to transfer collateral when open/close isolated perpetual position.

vincentwschau:
### Changelist
Update `UpdateSubaccounts` to transfer collateral between collateral pools when a subaccount opens/closes an isolated perpetual position.
Adds multiple helper functions to calculate / track the collateral transfers.

Note: This changes the assumption that `UpdateSubaccounts` only changes state in the `x/subaccount` module as it will now change state in the `x/bank` module as well.

### Test Plan
Unit tests. E2E tests are planned under this [ticket](https://linear.app/dydx/issue/TRA-99/add-integration-e2e-tests-for-placing-orders-in-isolated-markets-cross).

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/subaccounts/keeper/isolated_subaccount.go
+++ b/protocol/x/subaccounts/keeper/isolated_subaccount.go
@@ -2,16 +2,35 @@ package keeper
 
 import (
 	"math"
+	"math/big"
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	assettypes "github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
 	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
 	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+type positionStateTransition uint
+
+const (
+	opened positionStateTransition = iota
+	closed
+)
+
+// Represents a state transition for an isolated perpetual.
+type isolatedPerpetualStateTranisition struct {
+	perpetualId uint32
+	// TODO(DEC-715): Support non-USDC assets.
+	// USDC position size of the subaccount that has a state change for an isolated perpetual.
+	usdcQuantumsBeforeUpdate *big.Int
+	transition               positionStateTransition
+}
+
 // checkIsolatedSubaccountConstaints will validate all `updates` to the relevant subaccounts against
-// isolated subaccount constraints.
+// isolated subaccount constraints and computes whether each update leads to a state change
+// (open / close) to an isolated perpetual occurred due to the updates if they are valid.
 // This function checks each update in isolation, so if multiple updates for the same subaccount id
 // are passed in, they are not evaluated separately.
 // The input subaccounts must be settled.
@@ -20,6 +39,10 @@ import (
 // Returns a `successPerUpdates` value, which is a slice of `UpdateResult`.
 // These map to the updates and are used to indicate which of the updates
 // caused a failure, if any.
+// Returns a `isolatedPerpetualStateTransitions` value, which is a slice of
+// `isolatedPerpetualStateTransition`.
+// These map to the updates and are used to indicate which of the updates opened or closed an
+// isolated perpetual position.
 func (k Keeper) checkIsolatedSubaccountConstraints(
 	ctx sdk.Context,
 	settledUpdates []settledUpdate,
@@ -27,10 +50,12 @@ func (k Keeper) checkIsolatedSubaccountConstraints(
 ) (
 	success bool,
 	successPerUpdate []types.UpdateResult,
+	isolatedPerpetualStateTranisitions []*isolatedPerpetualStateTranisition,
 	err error,
 ) {
 	success = true
 	successPerUpdate = make([]types.UpdateResult, len(settledUpdates))
+	isolatedPerpetualStateTranisitions = make([]*isolatedPerpetualStateTranisition, len(settledUpdates))
 	var perpIdToMarketType = make(map[uint32]perptypes.PerpetualMarketType)
 
 	for _, perpetual := range perpetuals {
@@ -38,23 +63,25 @@ func (k Keeper) checkIsolatedSubaccountConstraints(
 	}
 
 	for i, u := range settledUpdates {
-		result, err := isValidIsolatedPerpetualUpdates(u, perpIdToMarketType)
+		result, stateTransition, err := processIsolatedPerpetualUpdates(u, perpIdToMarketType)
 		if err != nil {
-			return false, nil, err
+			return false, nil, nil, err
 		}
 		if result != types.Success {
 			success = false
 		}
 
 		successPerUpdate[i] = result
+		isolatedPerpetualStateTranisitions[i] = stateTransition
 	}
 
-	return success, successPerUpdate, nil
+	return success, successPerUpdate, isolatedPerpetualStateTranisitions, nil
 }
 
-// Checks whether the perpetual updates to a settled subaccount violates constraints for isolated
-// perpetuals. This function assumes the settled subaccount is valid and does not violate the
-// the constraints.
+// processIsolatedPerpetualUpdates checks whether the perpetual updates to a settled subaccount
+// violates constraints for isolated perpetuals and computes whether the perpetual updates result in
+// a state change (open / close) for an isolated perpetual position if the updates are valid.
+// This function assumes the settled subaccount is valid and does not violate the constraints.
 // The constraint being checked is:
 //   - a subaccount with a position in an isolated perpetual cannot have updates for other
 //     perpetuals
@@ -62,23 +89,27 @@ func (k Keeper) checkIsolatedSubaccountConstraints(
 //     perpetuals
 //   - a subaccount with no positions cannot be updated to have positions in multiple isolated
 //     perpetuals or a combination of isolated and non-isolated perpetuals
-func isValidIsolatedPerpetualUpdates(
+//
+// If there is a state change (open / close) from the perpetual updates, it is returned along with
+// the perpetual id of the isolated perpetual and the size of the USDC position in the subaccount.
+func processIsolatedPerpetualUpdates(
 	settledUpdate settledUpdate,
 	perpIdToMarketType map[uint32]perptypes.PerpetualMarketType,
-) (types.UpdateResult, error) {
+) (types.UpdateResult, *isolatedPerpetualStateTranisition, error) {
 	// If there are no perpetual updates, then this update does not violate constraints for isolated
 	// markets.
 	if len(settledUpdate.PerpetualUpdates) == 0 {
-		return types.Success, nil
+		return types.Success, nil, nil
 	}
 
 	// Check if the updates contain an update to an isolated perpetual.
 	hasIsolatedUpdate := false
 	isolatedUpdatePerpetualId := uint32(math.MaxUint32)
+	isolatedUpdate := (*types.PerpetualUpdate)(nil)
 	for _, perpetualUpdate := range settledUpdate.PerpetualUpdates {
 		marketType, exists := perpIdToMarketType[perpetualUpdate.PerpetualId]
 		if !exists {
-			return types.UpdateCausedError, errorsmod.Wrap(
+			return types.UpdateCausedError, nil, errorsmod.Wrap(
 				perptypes.ErrPerpetualDoesNotExist, lib.UintToString(perpetualUpdate.PerpetualId),
 			)
 		}
@@ -86,6 +117,10 @@ func isValidIsolatedPerpetualUpdates(
 		if marketType == perptypes.PerpetualMarketType_PERPETUAL_MARKET_TYPE_ISOLATED {
 			hasIsolatedUpdate = true
 			isolatedUpdatePerpetualId = perpetualUpdate.PerpetualId
+			isolatedUpdate = &types.PerpetualUpdate{
+				PerpetualId:      perpetualUpdate.PerpetualId,
+				BigQuantumsDelta: perpetualUpdate.GetBigQuantums(),
+			}
 			break
 		}
 	}
@@ -94,11 +129,12 @@ func isValidIsolatedPerpetualUpdates(
 	// Assumes the subaccount itself does not violate the isolated perpetual constraints.
 	isIsolatedSubaccount := false
 	isolatedPositionPerpetualId := uint32(math.MaxUint32)
+	isolatedPerpetualPosition := (*types.PerpetualPosition)(nil)
 	hasPerpetualPositions := len(settledUpdate.SettledSubaccount.PerpetualPositions) > 0
 	for _, perpetualPosition := range settledUpdate.SettledSubaccount.PerpetualPositions {
 		marketType, exists := perpIdToMarketType[perpetualPosition.PerpetualId]
 		if !exists {
-			return types.UpdateCausedError, errorsmod.Wrap(
+			return types.UpdateCausedError, nil, errorsmod.Wrap(
 				perptypes.ErrPerpetualDoesNotExist, lib.UintToString(perpetualPosition.PerpetualId),
 			)
 		}
@@ -106,6 +142,7 @@ func isValidIsolatedPerpetualUpdates(
 		if marketType == perptypes.PerpetualMarketType_PERPETUAL_MARKET_TYPE_ISOLATED {
 			isIsolatedSubaccount = true
 			isolatedPositionPerpetualId = perpetualPosition.PerpetualId
+			isolatedPerpetualPosition = perpetualPosition
 			break
 		}
 	}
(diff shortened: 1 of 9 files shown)


protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -2,16 +2,49 @@ package keeper
...
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	assettypes "github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
 	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
 	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+type positionStateTransition uint
```
teddyding, line 15: Can we add some documentation on what each value of `positionStateTransition` means? It was not immediately clear to me what it means for `isolatedPerpetualStateTransition.transition` to be `opened` or `closed` at first glance

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -526,6 +544,10 @@ func checkPositionUpdatable(
 // Returns a `successPerUpdates` value, which is a slice of `UpdateResult`.
 // These map to the updates and are used to indicate which of the updates
 // caused a failure, if any.
+// Returns a `isolatedPerpetualStateTransitions` value, which is a slice of
+// `isolatedPerpetualStateTransition`.
+// Thess map to the updates and are used to indicate which of the updates led to an isolated
```
teddyding, line 549: Typo?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -2,16 +2,49 @@ package keeper
...
+func (t positionStateTransition) String() string {
+	result, exists := positionStateTransitionStringMap[t]
+	if !exists {
+		return "UnexpectedStateTransitionError"
+	}
+
+	return result
+}
+
+// Represents a state transition for an isolated perpetual.
```
teddyding, line 36: Nit: `state transition for an isolated perpetual position`?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -2,16 +2,49 @@ package keeper
...
+type isolatedPerpetualStateTransition struct {
+	perpetualId uint32
+	// TODO(DEC-715): Support non-USDC assets.
+	// USDC position size of the subaccount that has a state change for an isolated perpetual.
+	usdcQuantumsBeforeUpdate *big.Int
+	transition               positionStateTransition
```
teddyding, line 42: Can we add some documentation wha `transition` means? 
Related, does `isolatedPerpetualStateTransition` only make sense when a position goes from `open` to `closed`, or a new position is `opened`? Does it not apply to position size change?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -2,16 +2,49 @@ package keeper
...
+
+	return result
+}
+
+// Represents a state transition for an isolated perpetual.
+type isolatedPerpetualStateTransition struct {
+	perpetualId uint32
+	// TODO(DEC-715): Support non-USDC assets.
+	// USDC position size of the subaccount that has a state change for an isolated perpetual.
+	usdcQuantumsBeforeUpdate *big.Int
```
teddyding, line 41: Conventionally we use `quoteQuantums` instead of `usdcQuantums` (the exact quote currency is abstracted away from subaccount logic)

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -20,72 +53,88 @@ import (
 // Returns a `successPerUpdates` value, which is a slice of `UpdateResult`.
 // These map to the updates and are used to indicate which of the updates
 // caused a failure, if any.
+// Returns a `isolatedPerpetualStateTransitions` value, which is a slice of
+// `isolatedPerpetualStateTransition`.
+// These map to the updates and are used to indicate which of the updates opened or closed an
+// isolated perpetual position.
 func (k Keeper) checkIsolatedSubaccountConstraints(
```
teddyding, line 60: nit: it seems weird that this function is named `checkIsolatedSubaccountConstraints` (focus: `check`) and it mainly calls `processIsolatedPerpetualUpdates`  (focus: `process`) for each `settledUpdate`. 

Should this instead be 
- inner function: `processSingleIsolatedPerpetualUpdate`
- outer function `processIsolatedPerpetualUpdates`

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -135,8 +186,148 @@ func isValidIsolatedPerpetualUpdates(
...
+			isolatedPerpetualPosition,
+			isolatedUpdate,
+		),
+		nil
+}
+
+// getIsolatedPerpetualStateTransition computes whether an isolated perpetual position will be
+// opened or closed for a subaccount given an isolated perpetual update for the subaccount.
+// Input subaccount account must be settled.
+func getIsolatedPerpetualStateTransition(
```
teddyding, line 204: For modularity, might make sense to have separate unit test for this function.

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -135,8 +186,148 @@ func isValidIsolatedPerpetualUpdates(
...
+	var fromModuleAddr sdk.AccAddress
+	var usdcQuantums *big.Int
+
+	// If an isolated perpetual position was opened in the subaccount, then move collateral equivalent
+	// to the USDC asset position size of the subaccount before the update from the
+	// cross-perpetual collateral pool to the isolated perpetual collateral pool.
+	if stateTransition.transition == opened {
+		toModuleAddr = isolatedCollateralPoolAddr
+		fromModuleAddr = types.ModuleAddress
+		usdcQuantums = stateTransition.usdcQuantumsBeforeUpdate
```
teddyding, line 275: `quoteQuantums`, same for remaining references

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -135,8 +186,148 @@ func isValidIsolatedPerpetualUpdates(
...
+		usdcQuantums,
+	)
+	if err != nil {
+		return err
+	}
+
+	if err = k.bankKeeper.SendCoins(
+		ctx,
+		fromModuleAddr,
+		toModuleAddr,
```
teddyding, line 326: For my understanding, when/where are module accounts for isolated market collateral pool initiated? Want to make sure we don't run into module account initialization issue again

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -306,6 +310,18 @@ func (k Keeper) UpdateSubaccounts(
 	// Apply the updates to asset positions.
 	UpdateAssetPositions(settledUpdates)
 
+	// Transfer collateral between collateral pools for any isolated perpetual positions that changed
+	// state due to an update.
+	for i, stateTransition := range isolatedPerpetualStateTransitions {
+		if err := k.transferCollateralForIsolatedPerpetual(
+			ctx,
+			settledUpdates[i].SettledSubaccount,
```
teddyding, line 318: We pass in `settledUpdates[i].SettledSubaccount` as `updatedSubaccounts` - is this because of logic above
```
	// Apply the updates to perpetual positions.
	UpdatePerpetualPositions(
		settledUpdates,
		perpIdToFundingIndex,
	)
	```
	
	If so, can we comment to make this clear?

File: protocol/x/subaccounts/keeper/isolated_subaccount.go [hunk 11]
```diff
@@ -135,8 +186,148 @@ func isValidIsolatedPerpetualUpdates(
...
+		return err
+	}
+	var toModuleAddr sdk.AccAddress
+	var fromModuleAddr sdk.AccAddress
+	var usdcQuantums *big.Int
+
+	// If an isolated perpetual position was opened in the subaccount, then move collateral equivalent
+	// to the USDC asset position size of the subaccount before the update from the
+	// cross-perpetual collateral pool to the isolated perpetual collateral pool.
+	if stateTransition.transition == opened {
```
teddyding, line 272: Related to this [question](https://github.com/dydxprotocol/v4-chain/pull/1200/files#r1531458231), is there no automatic collateral transfer involved in increasing a position? Or is there a separate transfer flow to increase collateral?

protocol/x/subaccounts/keeper/isolated_subaccount.go
shrenujb: The isolated market collateral pools are just normal accounts and not module accounts so we don't need to worry about the module account init

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Updated naming of the type as well.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Added comment.
>Related, does isolatedPerpetualStateTransition only make sense when a position goes from open to closed, or a new position is opened? Does it not apply to position size change?

That's correct, it only makes sense when a position goes from open to closed or when a new position is opened. All the collateral backing an existing position will already have been moved into the isolated perpetual's collateral pool (due to the logic in this PR) when it's size is changed so there's no need to do any collateral transfers when the position size changes.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Good point, I struggled with renaming, and your proposed function names sound good.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Added separate unit tests.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: They are initiated when funds are sent to them. We use `bank.SendCoins` and not any of the functions that require a module address to send funds.

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: Yes, that's correct. Added a comment.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: That is correct. Increasing / decreasing a position does not lead to automatic collateral transfer. To increase collateral for an isolated perpetual position, a user would have to transfer in additional collateral, the same way they would increase collateral for some set of cross-margined perpetual positions.

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -2,16 +2,19 @@ package keeper
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	assettypes "github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
 	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
 	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
-// checkIsolatedSubaccountConstaints will validate all `updates` to the relevant subaccounts against
-// isolated subaccount constraints.
+// processIsolatedSubaccountUpdates will validate all `updates` to the relevant subaccounts against
```
shrenujb, line 15: Why change the name to "process" instead?
We're still only checking constraints and not doing any writes here?

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -281,12 +284,13 @@ func (k Keeper) UpdateSubaccounts(
 	}
 
 	allPerps := k.perpetualsKeeper.GetAllPerpetuals(ctx)
-	success, successPerUpdate, err = k.internalCanUpdateSubaccounts(
+	success, successPerUpdate, isolatedPerpetualPositionStateTransitions, err := k.internalCanUpdateSubaccounts(
```
shrenujb, line 287: Do we really need to store the state transitions and get them from this func?
AFAIU, the 2 checks are
- opening the first position
- closing the last position

Can we just add this after the constraints check since those 2 are separate and avoid creating the whole state transition object?

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: There's computations done as well to return the state transitions, so "processing" is done, and it's not just checks. No strong opinions, but having the same verb across both functions (inner and outer) is something I prefer.

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: That was my initial thought as well, however it led to a lot of repeated logic for finding the perpetual updates / perpetual positions that are isolated so it was slightly cleaner to pass them through/back the functions that check constraints as both flows have very similar logic.
Also not strongly opinionated on this, as a separate function also de-couples the 2 operations (however right now they are still tightly coupled in concept).

shrenujb: My preference would be to decouple the 2 operations since and err towards the simpler logic.

How about just putting this in `UpdatePerpetualPositions`? You have all the information there and the collateral transfers are more closely tied to the actual perpetual update so it might make more sense here

vincentwschau: I still would need some data stored that gives enough information to do the collateral transfers, and wouldn't want the collateral transfer logic to be in the same state update function.
`UpdatePerpetualPositions` could be a decent candidate for the computation, however to calculate the data for the resulting collateral transfer the `AssetUpdates` need to be unsettled, so I'd prefer doing this in the "checking" / "processing" functions before any updates to state happen and separate the state change functions from the functions calculating what state changes are needed.

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -140,3 +138,193 @@ func isValidIsolatedPerpetualUpdates(
...
+	}
+	var toModuleAddr sdk.AccAddress
+	var fromModuleAddr sdk.AccAddress
+
+	// If an isolated perpetual position was opened in the subaccount, then move collateral from the
+	// cross-perpetual collateral pool to the isolated perpetual collateral pool.
+	if stateTransition.Transition == types.Opened {
+		toModuleAddr = isolatedCollateralPoolAddr
+		fromModuleAddr = types.ModuleAddress
+		// If the isolated perpetual position was closed, then move collateral from the isolated
```
shrenujb, line 263: nit: fix indent

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -140,3 +138,193 @@ func isValidIsolatedPerpetualUpdates(
...
+			types.ErrFailedToUpdateSubaccounts,
+			"Invalid state transition %v for isolated perpetual with id %d in subaccount with id %v",
+			stateTransition,
+			stateTransition.PerpetualId,
+			stateTransition.SubaccountId,
+		)
+	}
+
+	// If there are zero quantums to transfer, don't transfer collateral.
+	if stateTransition.QuoteQuantums.Sign() == 0 {
```
shrenujb, line 280: nit: maybe this should be at the top of the func?

protocol/x/subaccounts/keeper/isolated_subaccount.go
```diff
@@ -140,3 +138,193 @@ func isValidIsolatedPerpetualUpdates(
...
+			"Subaccount with id %v %s perpteual position with perpetual id %d with negative collateral %s to transfer",
+			stateTransition.SubaccountId,
+			stateTransition.Transition.String(),
+			stateTransition.PerpetualId,
+			stateTransition.QuoteQuantums.String(),
+		)
+	}
+
+	// Transfer collateral between collateral pools.
+	_, coinToTransfer, err := k.assetsKeeper.ConvertAssetToCoin(
```
shrenujb, line 298: informational q: what exactly are we converting for the collateral here?

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -306,6 +310,26 @@ func (k Keeper) UpdateSubaccounts(
...
+			return false, nil, err
+		}
+		if err := k.transferCollateralForIsolatedPerpetual(
+			ctx,
+			stateTransition,
+		); err != nil {
+			return false, nil, err
+		}
+	}
+
```
shrenujb, line 332: nit: Make this into a helper function

protocol/x/subaccounts/types/isolated_perpetual_position_state_transition_test.go
```diff
@@ -0,0 +1,36 @@
+package types_test
+
+import (
+	"testing"
+
+	"github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPositionStateTransitionString(t *testing.T) {
```
shrenujb, line 10: nit: Is this test necessary? Seems pretty trivial logic

shrenujb (commented):
Logic looks good. Just had a few nits and qs

Also, please add e2e tests for this case (or a TODO linear ticket) if this doesn't exist already

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: That's from `go fmt`, this indent is expected. I tried to fix it too and `go fmt` kept on changing it back.

protocol/x/subaccounts/keeper/isolated_subaccount.go
vincentwschau: Converts between `quantums` (units that `x/subaccounts` use to track balances) and `coins` (units that `x/bank` use to track balances).

protocol/x/subaccounts/types/isolated_perpetual_position_state_transition_test.go
vincentwschau: Not necessary, but could be useful if increase / decrease of position is a future state to be tracked.

vincentwschau:
E2E tests do have linear tickets already, noted in the PR already.

protocol/x/subaccounts/keeper/isolated_subaccount.go
shrenujb: weird

shrenujb (approved):

PR #1206, opened 2024-03-19
[TRA-134] construct simple vault orders

tqin7:
### Changelist
construct simple vault orders

### Test Plan
unit tests

### Author/Reviewer Checklist

### DIFF
--- a/protocol/lib/big_math.go
+++ b/protocol/lib/big_math.go
@@ -265,3 +265,27 @@ func warmCache() map[uint64]*big.Int {
 
 	return bigExponentValues
 }
+
+// BigRatRoundToNearestMultiple rounds `value` up/down to the nearest multiple of `base`.
+// Returns 0 if `base` is 0.
+func BigRatRoundToNearestMultiple(
+	value *big.Rat,
+	base uint32,
+	up bool,
+) uint64 {
+	if base == 0 {
+		return 0
+	}
+
+	quotient := new(big.Rat).Quo(
+		value,
+		new(big.Rat).SetUint64(uint64(base)),
+	)
+	quotientFloored := new(big.Int).Div(quotient.Num(), quotient.Denom())
+
+	if up && quotientFloored.Cmp(quotient.Num()) != 0 {
+		return (quotientFloored.Uint64() + 1) * uint64(base)
+	}
+
+	return quotientFloored.Uint64() * uint64(base)
+}
--- a/protocol/x/vault/keeper/orders.go
+++ b/protocol/x/vault/keeper/orders.go
@@ -1,8 +1,10 @@
 package keeper
 
 import (
+	"fmt"
 	"math/big"
 
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
@@ -45,24 +47,36 @@ func (k Keeper) RefreshAllVaultOrders(ctx sdk.Context) {
 func (k Keeper) GetVaultClobOrders(
 	ctx sdk.Context,
 	vaultId types.VaultId,
-) (orders []clobtypes.Order) {
+) (orders []*clobtypes.Order, err error) {
 	// Get clob pair, perpetual, market parameter, and market price that correspond to this vault.
 	clobPair, exists := k.clobKeeper.GetClobPair(ctx, clobtypes.ClobPairId(vaultId.Number))
 	if !exists {
-		return orders
+		return orders, errorsmod.Wrap(
+			types.ErrClobPairNotFound,
+			fmt.Sprintf("VaultId: %v", vaultId),
+		)
 	}
 	perpId := clobPair.Metadata.(*clobtypes.ClobPair_PerpetualClobMetadata).PerpetualClobMetadata.PerpetualId
 	perpetual, err := k.perpetualsKeeper.GetPerpetual(ctx, perpId)
 	if err != nil {
-		return orders
+		return orders, errorsmod.Wrap(
+			err,
+			fmt.Sprintf("VaultId: %v", vaultId),
+		)
 	}
 	marketParam, exists := k.pricesKeeper.GetMarketParam(ctx, perpetual.Params.MarketId)
 	if !exists {
-		return orders
+		return orders, errorsmod.Wrap(
+			types.ErrMarketParamNotFound,
+			fmt.Sprintf("VaultId: %v", vaultId),
+		)
 	}
 	marketPrice, err := k.pricesKeeper.GetMarketPrice(ctx, perpetual.Params.MarketId)
 	if err != nil {
-		return orders
+		return orders, errorsmod.Wrap(
+			err,
+			fmt.Sprintf("VaultId: %v", vaultId),
+		)
 	}
 
 	// Get vault (subaccount 0 of corresponding module account).
@@ -87,9 +101,10 @@ func (k Keeper) GetVaultClobOrders(
 		GoodTilBlockTime: uint32(ctx.BlockTime().Unix()) + ORDER_EXPIRATION_SECONDS,
 	}
 	// Construct one ask and one bid for each layer.
+	orders = make([]*clobtypes.Order, 2*NUM_LAYERS)
 	askSubticks := new(big.Rat).Set(subticks)
 	bidSubticks := new(big.Rat).Set(subticks)
-	for layer := uint8(1); layer <= NUM_LAYERS; layer++ {
+	for i := uint8(0); i < NUM_LAYERS; i++ {
 		// Calculate ask and bid subticks for this layer.
 		askSubticks = lib.BigRatMulPpm(askSubticks, lib.OneMillion+spreadPpm)
 		bidSubticks = lib.BigRatMulPpm(bidSubticks, lib.OneMillion-spreadPpm)
@@ -98,7 +113,7 @@ func (k Keeper) GetVaultClobOrders(
 		ask := clobtypes.Order{
 			OrderId: clobtypes.OrderId{
 				SubaccountId: vault,
-				ClientId:     k.GetVaultClobOrderClientId(ctx, clobtypes.Order_SIDE_SELL, uint8(layer)),
+				ClientId:     k.GetVaultClobOrderClientId(ctx, clobtypes.Order_SIDE_SELL, uint8(i+1)),
 				OrderFlags:   clobtypes.OrderIdFlags_LongTerm,
 				ClobPairId:   clobPair.Id,
 			},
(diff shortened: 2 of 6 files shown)


protocol/x/vault/keeper/orders.go
```diff
@@ -1,7 +1,30 @@
...
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+// TODO (TRA-118): store vault strategy constants in x/vault state.
+const (
```
tqin7, line 14: some of the constants here will be used when strategy is more fully implemented

protocol/x/vault/keeper/orders_test.go
```diff
@@ -1 +1,424 @@
...
+		constants.TestMarketPrices[1],
+	}
+	testPerps := []perptypes.Perpetual{
+		constants.BtcUsd_0DefaultFunding_0AtomicResolution,
+		constants.EthUsd_0DefaultFunding_9AtomicResolution,
+	}
+	// TODO (TRA-118): store vault strategy constants in x/vault state.
+	minBaseSpreadPpm := uint32(3_000)                   // 30bps
+	baseSpreadMinPriceChangePremiumPpm := uint32(1_500) // 15bps
+	orderExpirationSeconds := uint32(5)                 // 5 seconds
```
vincentwschau, line 47: nit: Use the constants declared in `x/vault`.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -1 +1,424 @@
...
+		constants.ClobPair_Eth,
+	}
+	testMarketParams := []pricestypes.MarketParam{
+		constants.TestMarketParams[0],
+		{
+			Id:                 constants.TestMarketParams[1].Id,
+			Pair:               constants.TestMarketParams[1].Pair,
+			Exponent:           constants.TestMarketParams[1].Exponent,
+			MinExchanges:       constants.TestMarketParams[1].MinExchanges,
+			MinPriceChangePpm:  4_200, // Set a high min price change to test spread calculation.
```
vincentwschau, line 32: nit: Add this to the test cases instead if it's a variable to test against.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -1 +1,424 @@
...
+	tests := map[string]struct {
+		/* --- Setup --- */
+		// Vault ID.
+		vaultId vaulttypes.VaultId
+
+		/* --- Expectations --- */
+		// Expected orders.
+		expectedOrders []clobtypes.Order
+	}{
+		"Get orders from Vault for Clob Pair 0": {
```
vincentwschau, line 112: nit: Non-blocking, the test cases are quite verbose and probably can be simplified to hold several parameters about each order rather than all the entire orders / constants could be declared for the orders.

protocol/x/vault/keeper/orders.go
```diff
@@ -10,3 +33,130 @@ import (
...
+			Quantums: clobPair.StepBaseQuantums, // TODO (TRA-144): Implement order size
+			Subticks: lib.BigRatRoundToNearestMultiple(
+				bidSubticks,
+				clobPair.SubticksPerTick,
+				false, // round down for bids
+			),
+			GoodTilOneof: goodTilBlockTime,
+		}
+
+		orders = append(orders, ask, bid)
```
vincentwschau, line 133: nit: Non-blocking, given you already know the number of layers, the slice could have been allocated with length 2*layers and this could be:
```
orders[2*i] = ask
orders[2*i+1] = bid
```

protocol/x/vault/keeper/orders.go
```diff
@@ -10,3 +33,130 @@ import (
...
+// - 2nd bit is `block height % 2`
+//   - block height bit alternates between 0 and 1 to ensure that client IDs
+//     are different in two consecutive blocks (otherwise, order placement would
+//     fail because the same order IDs are already marked for cancellation)
+//
+// - next 8 bits are `layer`
+func (k Keeper) GetVaultClobOrderClientId(
+	ctx sdk.Context,
+	side clobtypes.Order_Side,
+	layer uint8,
```
vincentwschau, line 151: nit: layer = 0 is invalid, given layer always starts at 1?

protocol/x/vault/keeper/orders_test.go
```diff
@@ -1 +1,424 @@
...
+		},
+		"Sell, Block Height Odd, Layer 0": {
+			side:             clobtypes.Order_SIDE_SELL, // 1<<31
+			blockHeight:      12345654321,               // 1<<30
+			layer:            0,                         // 0<<22
+			expectedClientId: 1<<31 | 1<<30 | 0<<22,
+		},
+		"Sell, Block Height Odd (negative), Layer 202": {
+			side:             clobtypes.Order_SIDE_SELL, // 1<<31
+			blockHeight:      -678987,                   // 1<<30
```
vincentwschau, line 399: nit: Weird test case given this is definitely an invalid state for the ctx.BlockHeight to be negative. Non-blocking though, maybe note that somewhere?

protocol/x/vault/keeper/orders.go
```diff
@@ -10,3 +33,130 @@ import (
...
+// Let n be number of layers, then the function returns orders at [a_1, b_1, a_2, b_2, ..., a_n, b_n]
+// where a_i and b_i are the ask price and bid price at i-th layer. To compute a_i and b_i:
+// - a_i = oraclePrice * (1 + spread)^i
+// - b_i = oraclePrice * (1 - spread)^i
+// TODO (TRA-144): Implement order size
+// TODO (TRA-114): Implement skew
+func (k Keeper) GetVaultClobOrders(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+) (orders []clobtypes.Order) {
```
vincentwschau, line 48: minor nit: non-blocking, do you expect the orders returned to be passed around into other functions / helpers alot? Could make sense to return a slice of pointers to order structs.

protocol/x/vault/keeper/orders_test.go
tqin7: In a later task, I'm going to remove the constants in `x/vault` and in this test file, initialize `x/vault` with these test params. will probably keep these test constants here (otherwise have to add them back in)

protocol/x/vault/keeper/orders.go
tqin7: will add this check!

protocol/x/vault/keeper/orders_test.go
tqin7: yeah just added this test case as `blockHeight` is an `int64`. wanted to make sure that a negative value doesn't break this function. will add a note on this

protocol/x/vault/keeper/orders.go
tqin7: Good catch! A slice of pointers will be more efficient

protocol/x/vault/keeper/orders_test.go
tqin7: yeah agreed. let me simplify these tests

protocol/lib/big_math.go
```diff
@@ -265,3 +265,27 @@ func warmCache() map[uint64]*big.Int {
...
+		return 0
+	}
+
+	quotient := new(big.Rat).Quo(
+		value,
+		new(big.Rat).SetUint64(uint64(base)),
+	)
+	quotientFloored := new(big.Int).Div(quotient.Num(), quotient.Denom())
+
+	if up && quotientFloored.Cmp(quotient.Num()) != 0 {
```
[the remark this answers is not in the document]
vincentwschau: `quotientFloored.Cmp(quotient.Num()) != 0` is only true when `quotient.Num() != quotientFloored` ([ref](https://pkg.go.dev/math/big#Int.Cmp)).
So this conditional is saying:
- if we want to round up (`up == true`) and if `quotient.Num() != quotientFloored` then add one to the floored quotient

protocol/lib/big_math.go
```diff
@@ -265,3 +265,27 @@ func warmCache() map[uint64]*big.Int {
 
 	return bigExponentValues
 }
+
+// BigRatRoundToNearestMultiple rounds `value` up/down to the nearest multiple of `base`.
+// Returns 0 if `base` is 0.
+func BigRatRoundToNearestMultiple(
+	value *big.Rat,
+	base uint32,
+	up bool,
```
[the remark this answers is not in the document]
tqin7: I think a boolean should suffice?

protocol/x/vault/keeper/orders.go
```diff
@@ -1,7 +1,32 @@
...
+	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+)
+
+// TODO (TRA-118): store vault strategy constants in x/vault state.
+const (
+	// Determines how many layers of orders a vault places.
+	// E.g. if num_levels=2, a vault places 2 asks and 2 bids.
+	NUM_LAYERS = uint8(2)
```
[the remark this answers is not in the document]
tqin7: oh it's just a made-up term. a vault places orders in terms of `layers` (layer 1, layer 2, ... where each layer gets further and further away from oracle price)

protocol/x/vault/keeper/orders.go
```diff
@@ -10,3 +35,144 @@ import (
...
+	// Get vault (subaccount 0 of corresponding module account).
+	vault := satypes.SubaccountId{
+		Owner:  vaultId.ToModuleAccountAddress(),
+		Number: 0,
+	}
+	// Calculate spread.
+	spreadPpm := lib.Max(
+		MIN_BASE_SPREAD_PPM,
+		BASE_SPREAD_MIN_PRICE_CHANGE_PREMIUM_PPM+marketParam.MinPriceChangePpm,
+	)
```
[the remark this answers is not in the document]
tqin7: [see research doc here](https://www.notion.so/dydx/Vault-Strategy-Description-Justification-7776169f15084a4187ca311bbb2f7af4?pvs=4#0dfb6403f449447a87b03d347efc7aa6)

protocol/x/vault/keeper/orders.go
```diff
@@ -10,3 +35,144 @@ import (
...
+				bidSubticks,
+				clobPair.SubticksPerTick,
+				false, // round down for bids
+			),
+			GoodTilOneof: goodTilBlockTime,
+		}
+
+		orders[2*i] = &ask
+		orders[2*i+1] = &bid
+	}
```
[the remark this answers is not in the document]
tqin7: don't think it's too much duplicate code? I feel having `ask` and `bid` explicitly here makes it very clear what each order is like and keeps the function simple

Christopher-Li (approved):

PR #1208, opened 2024-03-20
[TRA-125] Update withdrawal gating to be per collateral pool.

vincentwschau:
### Changelist
Update withdrawal gating logic to be done per collateral pool for negative TNC subaccounts. No changes to logic to block withdrawals after a chain halt.

### Test Plan
Unit tests updated. E2E tests update/change tracked in [TRA-165](https://linear.app/dydx/issue/TRA-165/update-add-e2e-test-for-per-collateral-pool-withdrawal-gating). State migration tracked in [TRA-167](https://linear.app/dydx/issue/TRA-167/add-state-migration-for-negative-subaccount-seen).

### Author/Reviewer Checklist


protocol/x/clob/keeper/process_operations.go
```diff
@@ -694,7 +694,15 @@ func (k Keeper) PersistMatchDeleveragingToState(
 			metrics.GetLabelForBoolValue(metrics.IsLong, position.GetIsLong()),
 			metrics.GetLabelForBoolValue(metrics.DeliverTx, true),
 		)
-		k.subaccountsKeeper.SetNegativeTncSubaccountSeenAtBlock(ctx, lib.MustConvertIntegerToUint32(ctx.BlockHeight()))
+		collateralPoolAddress, err := k.subaccountsKeeper.GetCollateralPoolFromPerpetualId(ctx, perpetualId)
+		if err != nil {
+			return err
+		}
```
jayy04, line 700: nit: can maybe just move this to `SetNegativeTncSubaccountSeenAtBlock`

jayy04: up to you though. no strong preference

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -3,28 +3,35 @@ package keeper
...
 func (k Keeper) GetNegativeTncSubaccountSeenAtBlock(
 	ctx sdk.Context,
+	collateralPoolAddr sdk.AccAddress,
 ) (uint32, bool) {
-	store := ctx.KVStore(k.storeKey)
-	return k.getNegativeTncSubaccountSeenAtBlock(store)
+	store := prefix.NewStore(
+		ctx.KVStore(k.storeKey),
+		[]byte(types.NegativeTncSubaccountForCollateralPoolSeenAtBlockKeyPrefix),
+	)
```
jayy04, line 22: do we need a state migration or no?

probably does need one I think

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -553,10 +560,17 @@ func (k Keeper) internalCanUpdateSubaccounts(
 
 	// Block all withdrawals and transfers if either of the following is true within the last
 	// `WITHDRAWAL_AND_TRANSFERS_BLOCKED_AFTER_NEGATIVE_TNC_SUBACCOUNT_SEEN_BLOCKS`:
-	// - There was a negative TNC subaccount seen.
+	// - There was a negative TNC subaccount seen for any of the collateral pools of subaccounts being updated
```
jayy04, line 563: for my understanding - we are blocking transfers both from and to the gated collateral pool, right? users also can't transfer to the pool

jayy04 (approved):

protocol/x/subaccounts/keeper/subaccount.go
vincentwschau: Yes, that's correct. Users can deposit into the gated collateral pool though still.

protocol/testutil/constants/clob_pair.go
```diff
@@ -127,4 +127,40 @@ var (
 		QuantumConversionExponent: -8,
 		Status:                    clobtypes.ClobPair_STATUS_PAUSED,
 	}
+	ClobPair_Iso = clobtypes.ClobPair{
```
jonfung-dydx, line 130: Nit: `ClobPairId_3_Iso`, same with below. `ClobPair_Iso2` has Id 4 so it gets a little confusing during tests, maybe ClobPairId_4_Iso

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -37,16 +44,20 @@ func (k Keeper) getNegativeTncSubaccountSeenAtBlock(
 }
 
 // SetNegativeTncSubaccountSeenAtBlock sets a block number in state where a negative TNC subaccount
-// was seen. This function will overwrite previous values at this key.
+// was seen for a specific collateral pool. This function will overwrite previous values at this key.
 // This function will panic if the old block height is greater than the new block height.
 func (k Keeper) SetNegativeTncSubaccountSeenAtBlock(
 	ctx sdk.Context,
+	collateralPoolAddr sdk.AccAddress,
```
jonfung-dydx, line 51: So collateral pools are tied to `sdk.AccAddress`, i.e Subaccounts module? shouldn't it be tied to an isolated market?

jonfung-dydx: Ok, talked to jay, seems like if we use the subaccounts module string that's the default cross margined iso pool.
Can we add this as a comment to make it more clear? also function above.

jonfung-dydx: Actually, i see [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/subaccounts/keeper/subaccount.go#L83) we are making an account address based on `"subaccount:<perpId>"`. Is there a reason we are using an sdk.Address`sdk.AccAddress` as the type of the key? Generating an acc address will hash the information so it's unreadable. I think collat pools are based off of perp ids purely. Would it make more sense to seperate the concept of an account address and use purely a `<prefix>:<perp id>` for this?

jonfung-dydx: seems like the only possible params here are <perp id> or <default cross margin perp id> so not sure if we should shoehorn this in an AccAddress

protocol/x/clob/e2e/withdrawal_gating_test.go
```diff
@@ -252,7 +252,10 @@ func TestWithdrawalGating_NegativeTncSubaccount_BlocksThenUnblocks(t *testing.T)
 					tApp.App.SubaccountsKeeper.GetSubaccount(ctx, *expectedSubaccount.Id),
 				)
 			}
-			negativeTncSubaccountSeenAtBlock, exists := tApp.App.SubaccountsKeeper.GetNegativeTncSubaccountSeenAtBlock(ctx)
+			negativeTncSubaccountSeenAtBlock, exists := tApp.App.SubaccountsKeeper.GetNegativeTncSubaccountSeenAtBlock(
+				ctx,
+				satypes.ModuleAddress,
```
jonfung-dydx, line 257: nit: can we use a hardcoded constant `DEFAULT_CROSS_MARGIN_KEY` instead of the module address? i don't think we would change the subaccount module name but it could in the future

protocol/x/clob/keeper/process_operations.go
jonfung-dydx: +1 to jays comment, related to my comment [here](https://github.com/dydxprotocol/v4-chain/pull/1208/files#r1534085785), I think i feel a little more strongly than jay haha

jonfung-dydx (approved):

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
vincentwschau: Discussed in slack, the actual place collateral is exiting / entering are sdk.AccAddress which can be derived from a perpetual Id. However, these accounts are associated with the `x/subaccounts` module so I don't think the addresses should be disassociated from the `x/subaccounts` module by removing the prefix.

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -951,3 +965,46 @@ func applyUpdatesToPositions[
...
+
+// getCollateralPoolAddresses gets a slice of collateral pool addresses for the subaccounts in the
+// slice of `settledUpdate`s passed in.
+// The slice will be de-duplicated and will contain unique collateral pool addresses.
+func (k Keeper) getCollateralPoolAddresses(
+	ctx sdk.Context,
+	settledUpdates []settledUpdate,
+) ([]sdk.AccAddress, error) {
+	collateralPoolAddressMap := make(map[string]bool)
+	collateralPoolAddresses := make([]sdk.AccAddress, 0)
```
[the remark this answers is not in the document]
vincentwschau: Why is this needed? There's no iteration on the map itself.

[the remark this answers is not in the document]
jonfung-dydx: yeah general helper function to extract keys of map without sorting, nit though

protocol/x/clob/e2e/withdrawal_gating_test.go
jonfung-dydx: just in case the subaccount module name changes, no need for a state migration. up to you

vincentwschau: Changed to a set of suffixes including a hard-coded constant for cross-perpetuals.

protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -80,7 +96,24 @@ func (k Keeper) SetNegativeTncSubaccountSeenAtBlock(
...
 	store.Set(
-		collateralPoolAddress.Bytes(),
+		[]byte(storeSuffix),
 		k.cdc.MustMarshal(&blockHeightValue),
 	)
+
+	return nil
+}
+
+func (k Keeper) getNegativeTncSubaccountStoreSuffx(
```
jonfung-dydx, line 106: suffix

protocol/x/subaccounts/keeper/subaccount.go
```diff
@@ -966,41 +966,49 @@ func applyUpdatesToPositions[
 	return result, nil
 }
 
-// getCollateralPoolAddresses gets a slice of collateral pool addresses for the subaccounts in the
-// slice of `settledUpdate`s passed in.
-// The slice will be de-duplicated and will contain unique collateral pool addresses.
-func (k Keeper) getCollateralPoolAddresses(
+// getNegativeTncSubaccountStoreSuffices gets a slice of negative tnc subaccount store suffices for
```
jonfung-dydx: might needa `s/suffices/suffixes` the whole file

protocol/x/subaccounts/types/keys.go
```diff
@@ -16,4 +16,7 @@ const (
 	// NegativeTncSubaccountForCollateralPoolSeenAtBlockKeyPrefix is the prefix for the store key that
 	// stores the last block a negative TNC subaccount was seen in state for a specific collateral pool.
 	NegativeTncSubaccountForCollateralPoolSeenAtBlockKeyPrefix = "NegSA:"
+	// Suffix for the store key to the last block a negative TNC subaccount was seen in state for the
+	// cross collateral pool.
+	CrossCollateralSuffix = "cross"
```
jonfung-dydx, line 21: i think the other suffixes have a colon, should we add that here?

jonfung-dydx: nvm, i see it is the suffix.

PR #1209, opened 2024-03-20
[CT-700] separate indexer and grpc streaming events

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/app.go
+++ b/protocol/app/app.go
@@ -994,9 +994,8 @@ func New(
 	clobFlags := clobflags.GetClobFlagValuesFromOptions(appOpts)
 	logger.Info("Parsed CLOB flags", "Flags", clobFlags)
 
-	memClob := clobmodulememclob.NewMemClobPriceTimePriority(
-		app.IndexerEventManager.Enabled() || app.GrpcStreamingManager.Enabled(),
-	)
+	memClob := clobmodulememclob.NewMemClobPriceTimePriority(app.IndexerEventManager.Enabled())
+	memClob.SetGenerateOrderbookUpdates(app.GrpcStreamingManager.Enabled())
 
 	app.ClobKeeper = clobmodulekeeper.NewKeeper(
 		appCodec,
--- a/protocol/x/clob/keeper/keeper.go
+++ b/protocol/x/clob/keeper/keeper.go
@@ -233,5 +233,13 @@ func (k Keeper) InitializeNewGrpcStreams(ctx sdk.Context) {
 		allUpdates.Append(update)
 	}
 
-	streamingManager.SendOrderbookUpdates(allUpdates, true)
+	k.SendOrderbookUpdates(allUpdates, true)
+}
+
+// SendOrderbookUpdates sends the offchain updates to the gRPC streaming manager.
+func (k Keeper) SendOrderbookUpdates(
+	offchainUpdates *types.OffchainUpdates,
+	snapshot bool,
+) {
+	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, snapshot)
 }
--- a/protocol/x/clob/keeper/keeper.go
+++ b/protocol/x/clob/keeper/keeper.go
@@ -241,5 +241,9 @@ func (k Keeper) SendOrderbookUpdates(
 	offchainUpdates *types.OffchainUpdates,
 	snapshot bool,
 ) {
+	if len(offchainUpdates.Messages) == 0 {
+		return
+	}
+
 	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, snapshot)
 }
--- a/protocol/x/clob/keeper/orders.go
+++ b/protocol/x/clob/keeper/orders.go
@@ -1310,8 +1310,6 @@ func (k Keeper) SendOffchainMessages(
 		}
 		k.GetIndexerEventManager().SendOffchainData(update)
 	}
-
-	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, false)
 }
 
 // getFillQuoteQuantums returns the total fillAmount price in quote quantums based on the maker subticks.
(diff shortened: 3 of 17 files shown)


protocol/x/clob/keeper/keeper.go
```diff
@@ -233,5 +233,13 @@ func (k Keeper) InitializeNewGrpcStreams(ctx sdk.Context) {
...
-	streamingManager.SendOrderbookUpdates(allUpdates, true)
+	k.SendOrderbookUpdates(allUpdates, true)
+}
+
+// SendOrderbookUpdates sends the offchain updates to the gRPC streaming manager.
+func (k Keeper) SendOrderbookUpdates(
+	offchainUpdates *types.OffchainUpdates,
+	snapshot bool,
+) {
+	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, snapshot)
```
jayy04, line 244: decouple indexer offchain events and grpc events

protocol/x/clob/keeper/orders.go
```diff
@@ -1310,8 +1310,6 @@ func (k Keeper) SendOffchainMessages(
 		}
 		k.GetIndexerEventManager().SendOffchainData(update)
 	}
-
-	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, false)
```
jayy04, line 1314: decouple grpc events and indexer events

protocol/x/clob/memclob/memclob.go
```diff
@@ -1216,6 +1216,18 @@ func (m *MemClobPriceTimePriority) PurgeInvalidMemclobState(
...
+	// to the grpc streams.
+	// Note that fully filled orders are removed from the orderbook in `RemoveOrderIfFilled` above.
+	allUpdates := types.NewOffchainUpdates()
+	for _, orderId := range filledOrderIds {
+		if m.openOrders.hasOrder(ctx, orderId) {
+			orderbookUpdate := m.GetOrderbookUpdatesForOrderUpdate(ctx, orderId)
+			allUpdates.Append(orderbookUpdate)
+		}
+	}
+	m.clobKeeper.SendOrderbookUpdates(allUpdates, false)
```
jonfung-dydx, line 1229: Would it make sense to move this from PurgeInvalidMemclobState to PCS in  abci.go? It seems that we depend on `filledOrderIds`, which is read off of `		processProposerMatchesEvents.OrderIdsFilledInLastBlock`. PCS is the main driver for using PPME and i don't think sending updates is related to purging memclob state?

protocol/x/clob/memclob/memclob.go
```diff
@@ -1510,6 +1522,10 @@ func (m *MemClobPriceTimePriority) mustAddOrderToOrderbook(
 	}
 
 	m.openOrders.mustAddOrderToOrderbook(ctx, newOrder, forceToFrontOfLevel)
+
+	// Send an orderbook update to grpc streams.
+	orderbookUpdate := m.GetOrderbookUpdatesForOrderPlacement(ctx, newOrder)
+	m.clobKeeper.SendOrderbookUpdates(orderbookUpdate, false)
```
jonfung-dydx, line 1528: i don't feel too strongly about this comment but would it make more sense to push these orderbook updates down a callstack level into the `func (m *memclobOpenOrders) mustRemoveOrder` memclobOpenOrders functions? Don't think we do this currently but in the future if we ever do direct order removal operations on the memclob open orders function orderbook updates won't be emitted.

jonfung-dydx: Same with the removal below

protocol/x/clob/memclob/memclob_grpc_streaming.go
```diff
@@ -86,3 +88,41 @@ func (m *MemClobPriceTimePriority) GetOffchainUpdatesForOrder(
 
 	return offchainUpdates
 }
+
+// GetOrderbookUpdatesForOrderRemoval returns a place order offchain message and
+func (m *MemClobPriceTimePriority) GetOrderbookUpdatesForOrderRemoval(
```
jonfung-dydx, line 93: In `MemclobPriceTimepriority` object we store a bool that indicates if we should ever ben emitting offchain events.
```
	// ---- Fields for determining if off-chain update messages should be generated ----
	generateOffchainUpdates bool
```
Can we do the same for orderbook updates? It would be nice to have all these no-op if this isn't a full node with streaming turned on.

jonfung-dydx (approved):

protocol/x/clob/memclob/memclob.go
jayy04: thought about this when I am implementing but `memclobOpenOrders` does not hold a reference to `memclob` and therefore does not hold to reference to clob keeper to send messages 😢

jayy04: let's keep it that way for now? memclob holding a reference to clob is already kind of ugly in my opinion so don't want to introduce another backwards pointer to the parent

protocol/app/app.go
```diff
@@ -994,9 +994,8 @@ func New(
 	clobFlags := clobflags.GetClobFlagValuesFromOptions(appOpts)
 	logger.Info("Parsed CLOB flags", "Flags", clobFlags)
 
-	memClob := clobmodulememclob.NewMemClobPriceTimePriority(
-		app.IndexerEventManager.Enabled() || app.GrpcStreamingManager.Enabled(),
-	)
+	memClob := clobmodulememclob.NewMemClobPriceTimePriority(app.IndexerEventManager.Enabled())
+	memClob.SetGenerateOrderbookUpdates(app.GrpcStreamingManager.Enabled())
```
jayy04, line 998: we should probably also get rid to the `app.IndexerEventManager.Enabled()` in the builder method and use the `Set...` method instead

jayy04: hard to add arguments to the constructor method since we have a ton of tests that need to be modified

PR #1213, opened 2024-03-20
[TRA-152] refresh all vault orders

tqin7:
### Changelist
refresh all vault orders

### Test Plan
added unit tests

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/vault/keeper/orders.go
+++ b/protocol/x/vault/keeper/orders.go
@@ -6,9 +6,11 @@ import (
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/dtypes"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/log"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
-	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
 )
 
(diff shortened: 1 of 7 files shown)


protocol/x/vault/keeper/orders.go
```diff
@@ -30,10 +32,79 @@ const (
...
+		k.cdc.MustUnmarshal(totalSharesIterator.Value(), &totalShares)
+
+		// Skip if TotalShares is non-positive.
+		if totalShares.NumShares.Cmp(dtypes.NewInt(0)) <= 0 {
+			continue
+		}
+
+		// Refresh orders depending on vault type.
+		switch vaultId.Type {
+		case types.VaultType_VAULT_TYPE_CLOB:
```
Christopher-Li, line 54: this is the only vault type we support currently right? 

nit: add comment that no other vault types exist

protocol/x/vault/keeper/orders.go
```diff
@@ -30,10 +32,79 @@ const (
 )
 
 // RefreshAllVaultOrders refreshes all orders for all vaults by
-// TODO(TRA-134)
 // 1. Cancelling all existing orders.
 // 2. Placing new orders.
 func (k Keeper) RefreshAllVaultOrders(ctx sdk.Context) {
```
Christopher-Li, line 37: will it be a final PR at the end that ties this to app.go?

Where is this called, would this be in CommitTx or in PrepareProposal?

protocol/x/vault/keeper/orders.go
```diff
@@ -30,10 +32,79 @@ const (
...
+	)
+	if err != nil {
+		log.ErrorLogWithError(ctx, "Failed to get vault clob orders to cancel", err, "vaultId", vaultId)
+		return err
+	}
+	for _, order := range ordersToCancel {
+		if _, exists := k.clobKeeper.GetLongTermOrderPlacement(ctx, order.OrderId); exists {
+			err := k.clobKeeper.HandleMsgCancelOrder(ctx, clobtypes.NewMsgCancelOrderStateful(
+				order.OrderId,
+				uint32(ctx.BlockTime().Unix())+ORDER_EXPIRATION_SECONDS,
```
Christopher-Li, line 78: why do we want to wait 5 seconds instead of just cancelling this now?

protocol/x/vault/keeper/orders.go
```diff
@@ -30,10 +32,79 @@ const (
...
+	}
+
+	// Place new CLOB orders.
+	ordersToPlace, err := k.GetVaultClobOrders(ctx, vaultId)
+	if err != nil {
+		log.ErrorLogWithError(ctx, "Failed to get vault clob orders to place", err, "vaultId", vaultId)
+		return err
+	}
+	for _, order := range ordersToPlace {
+		err := k.clobKeeper.HandleMsgPlaceOrder(ctx, clobtypes.NewMsgPlaceOrder(*order))
```
Christopher-Li, line 97: Should we make the optimization if the order already exists, do not remove it? Would save us two operations

protocol/x/vault/keeper/orders_test.go
```diff
@@ -23,6 +24,196 @@ const (
...
+
+			// Set total shares for each vault ID.
+			for i, vaultId := range tc.vaultIds {
+				err := tApp.App.VaultKeeper.SetTotalShares(ctx, vaultId, tc.totalShares[i])
+				require.NoError(t, err)
+			}
+
+			// Check that there's no stateful orders yet.
+			allStatefulOrders := tApp.App.ClobKeeper.GetAllStatefulOrders(ctx)
+			require.Len(t, allStatefulOrders, 0)
```
Christopher-Li, line 114: nit: Should we test that this is cancelling existing orders?

protocol/x/vault/keeper/orders.go
vincentwschau: nit: Add a default case and error there.

protocol/x/vault/keeper/orders.go
tqin7: It's already tied (see [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/vault/abci.go#L12)). At the end of every block, this will be called

protocol/x/vault/keeper/orders.go
tqin7: every block, we want to refresh these orders so that their sizes and prices are updated

protocol/x/vault/keeper/orders.go
tqin7: this cancels the order now (doesn't wait for 5 seconds). have to set this to be `currentBlockTime + 5` for this cancellation message to be valid

protocol/x/vault/keeper/orders_test.go
tqin7: Good call!

protocol/x/vault/keeper/orders.go
```diff
@@ -30,10 +32,82 @@ const (
...
+	ordersToCancel, err := k.GetVaultClobOrders(
+		ctx.WithBlockHeight(ctx.BlockHeight()-1),
+		vaultId,
+	)
+	if err != nil {
+		log.ErrorLogWithError(ctx, "Failed to get vault clob orders to cancel", err, "vaultId", vaultId)
+		return err
+	}
+	for _, order := range ordersToCancel {
+		if _, exists := k.clobKeeper.GetLongTermOrderPlacement(ctx, order.OrderId); exists {
```
sr33j, line 78: Is this method able to find orders that are partially filled?
For example, if the vault had half of its order filled, it would still get canceled in RefreshVaultClobOrders, correct?

protocol/x/vault/keeper/orders.go
tqin7: yep there can be. created a ticket [here](https://linear.app/dydx/issue/TRA-169/look-for-ways-to-minimize-order-placement-cancellation) to work on optimization later

sr33j (approved):

Christopher-Li (approved):

PR #1220, opened 2024-03-22
[TRA-135] Implement MsgDepositToVault

tqin7:
### Changelist
Implement `MsgDepositToVault`

### Test Plan
added unit tests / integration tests

### Author/Reviewer Checklist


protocol/x/vault/types/msg_deposit_to_vault_test.go
```diff
@@ -0,0 +1,64 @@
...
+			expectedErr: "Deposit amount is invalid",
+		},
+		"Failure: invalid subaccount owner": {
+			msg: types.MsgDepositToVault{
+				VaultId: &constants.Vault_Clob_0,
+				SubaccountId: &satypes.SubaccountId{
+					Owner:  "invalid-owner",
+					Number: 0,
+				},
+				QuoteQuantums: dtypes.NewInt(-1),
```
sr33j, line 48: nit: should you make QuoteQuantums a positive value here as to isolate which error we are going for?

tqin7: good catch. missed it

protocol/x/vault/keeper/msg_server_deposit_to_vault.go
```diff
@@ -11,5 +15,56 @@ func (k msgServer) DepositToVault(
...
+		msg.QuoteQuantums.BigInt(),
+	)
+	if err != nil {
+		return err
+	}
+
+	// Transfer from sender subaccount to vault.
+	// Note: Transfer should take place after minting shares for
+	// shares calculation to be correct.
+	err = k.subaccountsKeeper.TransferFundsFromSubaccountToSubaccount(
```
sr33j, line 47: nit: is there any situation in which this errors out?
if so, it could be mean that subaccount could hack us because they keep the money and get shares in the vault.
it doesn't really matter though, since we aren't exposing this to users.

protocol/x/vault/app_test.go
```diff
@@ -0,0 +1,300 @@
...
+						},
+					})
+				} else {
+					ctx = tApp.AdvanceToBlock(nextBlock, testapp.AdvanceToBlockOptions{})
+				}
+
+				// Check that total shares of the vault is as expected.
+				totalShares, exists := tApp.App.VaultKeeper.GetTotalShares(ctx, tc.vaultId)
+				require.True(t, exists)
+				require.Equal(t, tc.totalSharesHisotry[i], totalShares.NumShares.BigInt())
```
sr33j, line 296: nit: History is mispelled here

protocol/x/vault/keeper/msg_server_deposit_to_vault.go
tqin7: transaction is rolled back if it fails so there shouldn't be a scenario where a user gets shares but keeps the money

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,62 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+		// Don't mint shares if equity is non-positive.
+		if equity.Cmp(big.NewInt(0)) <= 0 {
+			return types.ErrNonPositiveEquity
+		}
+		// Mint `deposit * existing shares / vault equity` number of shares.
+		// For example:
+		// - a vault currently has 5000 shares and $4000 equity
+		// - each $1 is worth 5000 / 4000 = 1.25 shares
+		// - a deposit of $1000 should thus be given 1000 * 1.25 = 1250 shares
+		sharesToMint = sharesToMint.Mul(quantumsToDeposit, existingTotalShares)
```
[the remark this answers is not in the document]
tqin7: Yeah should be okay i think. just a matter of units

protocol/x/vault/types/msg_deposit_to_vault_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+
+	"github.com/dydxprotocol/v4-chain/protocol/dtypes"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgDepositToVault_ValidateBasic(t *testing.T) {
```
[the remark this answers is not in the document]
tqin7: yeah had that in my mind and then didn't add it in haha. added equity check in `app_test.go`

protocol/x/vault/keeper/shares.go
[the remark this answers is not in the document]
tqin7: let me update the comment here though

sr33j (approved):
nice!

protocol/x/vault/keeper/msg_server_deposit_to_vault.go
```diff
@@ -11,5 +15,56 @@ func (k msgServer) DepositToVault(
...
+		ctx,
+		*msg.SubaccountId,
+		*msg.VaultId.ToSubaccountId(),
+		0, // assetId
```
vincentwschau, line 51: nit: Use `constant.AssetUsdc.Id` to clearly denote this is the USDC asset.

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,62 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+		}
+		// Mint `deposit * existing shares / vault equity` number of shares.
+		// For example:
+		// - a vault currently has 5000 shares and 4000 equity (in quote quantums)
+		// - each quote quantum is worth 5000 / 4000 = 1.25 shares
+		// - a deposit of 1000 quote quantums should thus be given 1000 * 1.25 = 1250 shares
+		sharesToMint = sharesToMint.Mul(quantumsToDeposit, existingTotalShares)
+		sharesToMint = sharesToMint.Quo(sharesToMint, equity)
+	} else {
+		// Mint `quoteQuantums` number of shares.
```
vincentwschau, line 94: non-blocking: This code path executes if `existingTotalShares` is ever negative as well (don't think that could happen right now).

protocol/x/vault/keeper/msg_server.go
```diff
@@ -5,12 +5,12 @@ import (
 )
 
 type msgServer struct {
-	Keeper
+	Keeper types.VaultKeeper
```
Christopher-Li, line 8: was Keeper previously untyped?

protocol/x/vault/keeper/msg_server_deposit_to_vault_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/dydxprotocol/v4-chain/protocol/mocks"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	keepertest "github.com/dydxprotocol/v4-chain/protocol/testutil/keeper"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/keeper"
+	"github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgServerDepositToVault(t *testing.T) {
```
Christopher-Li, line 18: What exactly is this testing? it seems like you're just mocking a function to return an error, and checking that the error is returned when the function is called

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,62 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+			return types.ErrNonPositiveEquity
+		}
+		// Mint `deposit * existing shares / vault equity` number of shares.
+		// For example:
+		// - a vault currently has 5000 shares and 4000 equity (in quote quantums)
+		// - each quote quantum is worth 5000 / 4000 = 1.25 shares
+		// - a deposit of 1000 quote quantums should thus be given 1000 * 1.25 = 1250 shares
+		sharesToMint = sharesToMint.Mul(quantumsToDeposit, existingTotalShares)
+		sharesToMint = sharesToMint.Quo(sharesToMint, equity)
+	} else {
```
Christopher-Li, line 93: nit: It's usually cleaner to do the smaller block of logic first in an if statement. It would be slightly easier to follow for readers if we do the create vault case first

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,62 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+		// Get vault equity.
+		equity, err := k.GetVaultEquity(ctx, vaultId)
+		if err != nil {
+			return err
+		}
+		// Don't mint shares if equity is non-positive.
+		if equity.Cmp(big.NewInt(0)) <= 0 {
+			return types.ErrNonPositiveEquity
+		}
+		// Mint `deposit * existing shares / vault equity` number of shares.
```
Christopher-Li, line 86: nit: 
```suggestion
		// Mint `deposit equity* existing shares / existing vault equity` number of shares.
```

protocol/x/vault/app_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+// DepositorSetup represents the setup of a depositor.
+type DepositorSetup struct {
+	// Depositor.
+	depositor satypes.SubaccountId
+	// Initial balance of the depositor (in quote quantums).
+	depositorBalance *big.Int
+}
+
+func TestMsgDepositToVault(t *testing.T) {
```
Christopher-Li, line 44: The usual convention in this code base is that `file_name_test.go` is used to test `file_name.go`. Is this file supposed to test app.go code in vault?

protocol/x/vault/keeper/msg_server.go
vincentwschau: It was previously an embedded struct.

Question: Any reason we chose to stop embedding `Keeper` type into the struct?

tqin7: stopped embedding to be able to mock `x/vault`'s msgServer in unit tests

protocol/x/vault/keeper/msg_server_deposit_to_vault_test.go
tqin7: Yeah this is the pattern with existing `msgServer`s

protocol/x/vault/keeper/shares.go
tqin7: makes sense!

protocol/x/vault/keeper/shares.go
tqin7: yeah `SetTotalShares` prevents it from being negative

protocol/x/vault/keeper/msg_server_deposit_to_vault_test.go
tqin7: with your below comment on `app_test.go`, i'm going to remove mock testings here and move tests in `app_test.go` to this file

protocol/x/vault/app_test.go
tqin7: [see answer above](https://github.com/dydxprotocol/v4-chain/pull/1220#discussion_r1536562209)

protocol/x/vault/keeper/msg_server.go
tqin7: going to remove mock testings as they don't do much and revert back to embedded struct

protocol/x/vault/keeper/shares.go
tqin7: updated to `deposit (in quote quantums) * existing shares / vault equity (in quote quantums)`

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,72 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+		}
+		// Mint `deposit (in quote quantums) * existing shares / vault equity (in quote quantums)`
+		// number of shares.
+		// For example:
+		// - a vault currently has 5000 shares and 4000 equity (in quote quantums)
+		// - each quote quantum is worth 5000 / 4000 = 1.25 shares
+		// - a deposit of 1000 quote quantums should thus be given 1000 * 1.25 = 1250 shares
+		// For shares calculation to be accurate, scale everything by denominator and mint
+		// numerator number of shares.
+		// For example:
```
vincentwschau, line 99: How do you expect this to work with owner shares being calculated? Each time you "scale" shares, each owner's total number of shares changes, wouldn't this lead to # state changes being linearly scaling with number of depositors for a vault when minting (undesirable) vs constant number of state changes per deposit?

protocol/x/vault/keeper/shares.go
```diff
@@ -47,3 +56,72 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
...
+		// Don't mint shares if equity is non-positive.
+		if equity.Cmp(big.NewInt(0)) <= 0 {
+			return types.ErrNonPositiveEquity
+		}
+		// Mint `deposit (in quote quantums) * existing shares / vault equity (in quote quantums)`
+		// number of shares.
+		// For example:
+		// - a vault currently has 5000 shares and 4000 equity (in quote quantums)
+		// - each quote quantum is worth 5000 / 4000 = 1.25 shares
+		// - a deposit of 1000 quote quantums should thus be given 1000 * 1.25 = 1250 shares
```
vincentwschau, line 96: This is no longer accurate right? Due to the scaling logic the # of shares would be scaled by 4 (5000/4000 -> 5/4). So there would be 20,000 shares now in the vault and 5000 shares would be minted?

protocol/x/vault/keeper/shares.go
vincentwschau: If each deposit happened with uneven quote quantums / equity, the total number of shares would continue scaling up and the integer would get larger and larger exponentially (imagine I always deposit quote quantums that is co-prime with the equity) right?

vincentwschau: So each time `quote quantums * existing shares` / `equity` doesn't lead to a denominator of 1 we scale the total number of shares up?
E.g. equity is now 1001, with total shares 1001. Let's say equity now increases to 1002. If I deposit 1 quote quantum again, the math is:
(1 * (1001)) / 1002 = 1001/1002. Now we scale the # of shares again up by 1002?

tqin7: conclusion from offline discussion: store shares as rational numbers

protocol/x/vault/keeper/shares_test.go
```diff
@@ -77,98 +88,82 @@ func TestMintShares(t *testing.T) {
...
 		},
 		"Equity 1_000_000, TotalShares 1, Deposit 1": {
 			vaultId:           constants.Vault_Clob_1,
 			equity:            big.NewInt(1_000_000),
-			totalShares:       big.NewInt(1),
+			totalShares:       big.NewRat(1, 1),
 			quantumsToDeposit: big.NewInt(1),
-			// 1 * 1 / 1_000_000 = 1 / 1_000_000
+			// Should mint `1 / 1_000_000` shares.
 			// Should thus mint 1 share and scale existing shares by 1_000_000.
```
vincentwschau, line 137: nit: Remove this comment.

vincentwschau (approved):

PR #1226, opened 2024-03-22
[TRA-167] v5.0.0 state migration negative tnc subaccount store.

vincentwschau:
### Changelist
State migration for the negative tnc subaccount store.

### Test Plan
N/A

### Author/Reviewer Checklist


protocol/x/subaccounts/keeper/negative_tnc_subaccount.go
```diff
@@ -176,3 +176,24 @@ func (k Keeper) getLastBlockNegativeSubaccountSeen(
 	}
 	return lastBlockNegativeSubaccountSeen, negativeSubaccountExists, nil
 }
+
+// LegacyGetNegativeTncSubaccountSeenAtBlock gets the last block height a negative TNC subaccount was
+// seen in state and a boolean for whether it exists in state.
+// Note: This is the legacy implementation and meant to be used for the v5.0.0 state migration.
```
jonfung-dydx, line 182: let's explicitly mark the godoc deprecated?

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -98,11 +99,41 @@ func blockRateLimitConfigUpdate(
 	)
 }
 
+func negativeTncSubaccountSeenAtBlockUpgrade(
+	ctx sdk.Context,
+	perpetualsKeeper perptypes.PerpetualsKeeper,
+	subaccountsKeeper satypes.SubaccountsKeeper,
+) {
+	// Get block height stored by v4.x.x.
+	blockHeight, exists := subaccountsKeeper.LegacyGetNegativeTncSubaccountSeenAtBlock(ctx)
```
jonfung-dydx, line 108: qn: do we need to delete the old KV pair?

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -98,11 +99,41 @@ func blockRateLimitConfigUpdate(
...
+	}
+
+	// Migrate the value from the legacy store to the new store.
+	if err := subaccountsKeeper.SetNegativeTncSubaccountSeenAtBlock(
+		ctx,
+		perpetuals[0].Params.Id, // must be a cross-margined perpetual due to `perpetualsUpgrade`.
+		blockHeight,
+	); err != nil {
+		panic(fmt.Sprintf("failed to set negative tnc subaccount seen at block with value %d: %s", blockHeight, err))
+	}
```
jonfung-dydx, line 128: can we add some log statements to log out the pre and post upgrade data in the event of a success?

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -98,11 +99,41 @@ func blockRateLimitConfigUpdate(
...
+	// negative tnc subaccounts w/o perpetuals.
+	perpetuals := perpetualsKeeper.GetAllPerpetuals(ctx)
+	if len(perpetuals) == 0 {
+		return
+	}
+
+	// Migrate the value from the legacy store to the new store.
+	if err := subaccountsKeeper.SetNegativeTncSubaccountSeenAtBlock(
+		ctx,
+		perpetuals[0].Params.Id, // must be a cross-margined perpetual due to `perpetualsUpgrade`.
```
jonfung-dydx, line 124: for my understanding, we are just using perp 0 (btc) which should be cross margined? 
Might be nice to add a panic check to assert that perp 0 is cross margined but don't feel too strongly

jonfung-dydx (approved):

protocol/app/upgrades/v5.0.0/upgrade.go
vincentwschau: No need to explicitly delete.

protocol/app/upgrades/v5.0.0/upgrade.go
vincentwschau: >we are just using perp 0 (btc) which should be cross margined?
Not exactly. Since this upgrade contains both the "add market type" upgrade and the "change negative subaccount seen" upgrade, all perpetuals (as long as one exists) must be cross-margined. So perpetual 0 on any deployment of the chain must be cross-margined at this point.

PR #1229, opened 2024-03-23
[TRA-110] Implement GET /address/parentSubaccountNumber API

shrenujb:
### Changelist
Implement GET /address/parentSubaccountNumber API

### Test Plan
Added tests

### Author/Reviewer Checklist

### DIFF
--- a/indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
@@ -351,6 +351,7 @@ router.get(
   complianceAndGeoCheck,
   ExportResponseCodeStats({ controllerName }),
   async (req: express.Request, res: express.Response) => {
+    const start: number = Date.now();
     const {
       address,
     }: {
@@ -371,6 +372,11 @@ router.get(
         error,
         req,
         res,
+      );
+    } finally {
+      stats.timing(
+        `${config.SERVICE_NAME}.${controllerName}.get_addresses.timing`,
+        Date.now() - start,
       );
     }
   },
@@ -413,7 +419,7 @@ router.get(
       );
     } finally {
       stats.timing(
-        `${config.SERVICE_NAME}.${controllerName}.get_addresses.timing`,
+        `${config.SERVICE_NAME}.${controllerName}.get_subaccount.timing`,
         Date.now() - start,
       );
     }
@@ -460,7 +466,7 @@ router.get(
       );
     } finally {
       stats.timing(
-        `${config.SERVICE_NAME}.${controllerName}.get_addresses.timing`,
+        `${config.SERVICE_NAME}.${controllerName}.get_parentSubaccount.timing`,
         Date.now() - start,
       );
     }
--- a/indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
@@ -48,10 +48,10 @@ import {
   getTotalUnsettledFunding,
   handleControllerError,
   getPerpetualPositionsWithUpdatedFunding,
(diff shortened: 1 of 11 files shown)


indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
```diff
@@ -241,6 +241,107 @@ class AddressesController extends Controller {
...
+            subaccount.id,
+          ),
+          AssetTable.findAll(
+            {},
+            [],
+          ),
+          MarketTable.findAll(
+            {},
+            [],
+          ),
```
Christopher-Li, line 301: we don't need to pull this information for every subaccount, this only needs to be done once

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
```diff
@@ -241,6 +241,107 @@ class AddressesController extends Controller {
...
+          assets,
+          markets,
+          lastUpdatedFundingIndexMap,
+        ] = await Promise.all([
+          getOpenPerpetualPositionsForSubaccount(
+            subaccount.id,
+          ),
+          getAssetPositionsForSubaccount(
+            subaccount.id,
+          ),
```
Christopher-Li, line 293: Can we pull these values in bulk? I'm worried this will take too much time?

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
```diff
@@ -320,6 +421,53 @@ router.get(
...
+    const start: number = Date.now();
+    const {
+      address,
+      parentSubaccountNumber,
+    }: {
+      address: string,
+      parentSubaccountNumber: number,
+    } = matchedData(req) as ParentSubaccountRequest;
+
+    // Convert parentSubaccount to a number since
```
vincentwschau, line 441: nit: complete the comment.

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
shrenujb: I think this is a valid concern. There's a lot of ways to improve performance overall in all these APIs. My thinking right now is lets get these changes out first. If the performance is good enough, we don't need to tackle it immediately. If its not great, we can look into ways to improve perf

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
```diff
@@ -320,6 +420,53 @@ router.get(
...
+      return handleControllerError(
+        'AddressesController GET /:address/parentSubaccountNumber/:parentSubaccountNumber',
+        'Addresses subaccount error',
+        error,
+        req,
+        res,
+      );
+    } finally {
+      stats.timing(
+        `${config.SERVICE_NAME}.${controllerName}.get_addresses.timing`,
```
Christopher-Li, line 463: is this the correct metric?

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
Christopher-Li: that sounds good for now, we do need to monitor this, can you make sure the metric for this PR would be in v4-infrastructure and in our datadog for metrics?

vincentwschau (approved):

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
shrenujb: fixed

indexer/services/comlink/src/controllers/api/v4/addresses-controller.ts
shrenujb: should be available with the new commit

PR #1230, opened 2024-03-24
Store pruneable orders in a key-per-order format rather than key-per-height

roy-dydx:
### Changelist
- Store one order per key rather than a list of orders per height. This saves significant time in the serde and sorting of these orders.
- Add migration logic for upgrade handler

### Test Plan
- Modified existing tests and added test for migration

### Author/Reviewer Checklist


protocol/x/clob/abci_test.go
```diff
@@ -66,28 +62,6 @@ func assertFillAmountAndPruneState(
 		exists, _, _ := k.GetOrderFillAmount(ctx, orderId)
 		require.False(t, exists)
 	}
-
-	for blockHeight, orderIds := range expectedPruneableBlockHeights {
```
roy-dydx, line 70: Removing this chunk of code because `expectedPruneableBlockHeights` was never set, so this doesn't ever test anything.

protocol/x/clob/keeper/order_state.go
```diff
@@ -125,14 +128,32 @@ func (k Keeper) GetOrderFillAmount(
...
+	buf.Write(lib.Uint32ToKey(height))
+	buf.Write([]byte(":"))
+	return prefix.NewStore(ctx.KVStore(k.storeKey), buf.Bytes())
+}
+
+// AddOrdersForPruning creates or updates `orderIds` to state for potential future pruning from state.
 func (k Keeper) AddOrdersForPruning(ctx sdk.Context, orderIds []types.OrderId, prunableBlockHeight uint32) {
+	store := k.GetPruneableOrdersStore(ctx, prunableBlockHeight)
+	for _, orderId := range orderIds {
+		store.Set(
```
jonfung-dydx, line 143: We're still using the proto serialized version of orderid as the state key?

What is the difference between these two marshals?
`orderId.ToStateKey(),` `k.cdc.MustMarshal(&orderId),`

protocol/x/clob/keeper/order_state.go
```diff
@@ -125,14 +128,32 @@ func (k Keeper) GetOrderFillAmount(
...
+			orderId.ToStateKey(),
+			k.cdc.MustMarshal(&orderId),
+		)
+	}
+}
+
+// LegacyAddOrdersForPruning is the old key-per-height way of storing orders to prune.
+// DO NOT USE. Retained for testing purposes.
```
jonfung-dydx, line 151: DO NOT USE -> deprecated docstring

protocol/x/clob/keeper/order_state.go
```diff
@@ -221,14 +228,31 @@ func (k Keeper) PruneOrdersForBlockHeight(ctx sdk.Context, blockHeight uint32) (
 				)
 			}
 		}
+		potentiallyPrunableOrdersStore.Delete(it.Key())
```
jonfung-dydx, line 231: Do we need to prune the whole order `PO/12:` store after we prune all the K:V pairs in it? Or will the KV store automatically get gc'd once there are no keys?

protocol/x/clob/types/keys.go
```diff
@@ -45,9 +48,9 @@ const (
 	// OrderAmountFilledKeyPrefix is the prefix to retrieve the fill amount for an order.
 	OrderAmountFilledKeyPrefix = "Fill:"
 
-	// BlockHeightToPotentiallyPrunableOrdersPrefix is the prefix to retrieve a list of potentially prunable
-	// short term orders by block height.
-	BlockHeightToPotentiallyPrunableOrdersPrefix = "ExpHt:"
+	// LegacyBlockHeightToPotentiallyPrunableOrdersPrefix is the prefix to retrieve a list of potentially prunable
+	// short term orders by block height. Should not be used after migrating to key-per-order format.
```
jonfung-dydx, line 52: deprecate docstring

protocol/x/clob/keeper/order_state.go
roy-dydx: Yeah. I will change that for the whole codebase later. For this specific change it doesn't really matter since we are iterating over prefix rather than trying to query keys directly.

protocol/x/clob/keeper/order_state.go
roy-dydx: No. The prefix store is just a way of conveniently accessing keys with the same prefix. It's the same as querying the underlying store with the prefix prepended.

protocol/x/clob/keeper/order_state.go
roy-dydx: Right now those two marshals are the same, but it will be different later.

jonfung-dydx (approved):

jayy04 (approved):

roy-dydx:
[OTE-250]

PR #1237, opened 2024-03-25
[CT-708] Indexer track e2e latency

jonfung-dydx:
vulcan -> socks. Thread through the kafka headers.

Socks, emit the e2e latency.
Vulcan, emit the e2e latency.


indexer/services/vulcan/src/lib/on-message.ts
```diff
@@ -62,6 +62,18 @@ export async function onMessage(message: KafkaMessage): Promise<void> {
     },
   );
 
+  const originalMessageTimestamp = message.headers?.message_received_timestamp;
+  if (originalMessageTimestamp !== undefined) {
+    stats.timing(
+      `${config.SERVICE_NAME}.message_time_since_received`,
```
jonfung-dydx, line 68: better var name?

indexer/services/vulcan/src/lib/on-message.ts
```diff
@@ -86,7 +98,7 @@ export async function onMessage(message: KafkaMessage): Promise<void> {
     const handler: Handler = new (getHandler(update))!(
       getTransactionHashFromHeaders(message.headers),
     );
-    await handler.handleUpdate(update);
+    await handler.handleUpdate(update, message.headers ?? {});
```
dydxwill, line 102: can we also emit a stat here comparing to message_received_timestamp ?

indexer/services/socks/src/lib/message-forwarder.ts
```diff
@@ -125,12 +139,12 @@ export class MessageForwarder {
       return;
     }
 
-    const start: number = Date.now();
+    const startForwardMessage: number = Date.now();
     this.forwardMessage(messageToForward);
```
dydxwill, line 143: This function seems to send out websocket messages in batches for certain websocket topics. Can we pipe the initial timestamp all the way to the `sendMessage` function?

dydxwill (commented):
Can we also add unit tests?

indexer/services/vulcan/src/lib/on-message.ts
jonfung-dydx: we already do that [here](https://github.com/dydxprotocol/v4-chain/pull/1237/files#diff-fbcd2ef7f5fc44a4c9b34e080e07494f74a92010e4ef52fa449a1c6c98fe3cdcR65-R76) in the same code block right?

indexer/services/socks/src/lib/message-forwarder.ts
jonfung-dydx: Do you mean we should add a message_received_timestamp field to 
```
export interface OutgoingMessage {
  type: OutgoingMessageType;
  connection_id: string;
  message_id: number;
}
```?

File: indexer/services/vulcan/src/lib/on-message.ts (hunk 2 above)
Comment (dydxwill) [reply indexer/services/vulcan/src/lib/on-message.ts:102]: that's before handling the message though

File: indexer/services/socks/src/lib/message-forwarder.ts (hunk 3 above)
Comment (dydxwill) [reply indexer/services/socks/src/lib/message-forwarder.ts:143]: Nvm, the batches happen at periodic intervals so not necessary to pipe it through.

Does it make more sense to have the stat be `end - Number(message.headers?.message_received_timestamp)` (after calling `forwardMessage`) rather than `start - Number(message.headers?.message_received_timestamp)`?

Comment (jonfung-dydx) [reply indexer/services/socks/src/lib/message-forwarder.ts:143]: done

File: indexer/services/vulcan/__tests__/lib/on-message.test.ts [hunk 4]
```diff
@@ -64,7 +64,7 @@ describe('onMessage', () => {
     await onMessage(message);
 
     expect(handler).toHaveBeenCalledTimes(1);
-    expect(handleUpdateMock).toHaveBeenCalledWith(update);
+    expect(handleUpdateMock).toHaveBeenCalledWith(update, message.headers ?? {});
```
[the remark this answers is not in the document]
jonfung-dydx: no it is not the ?? operator sucks and promotes bad coding practices

dydxwill (approved):

PR #1238, opened 2024-03-26
[TRA-118] store x/vault parameters in state, add query, and update via governance

tqin7:
### Changelist
1. store x/vault parameters in state
2. query parameters
3. update parameters via governance

### Test Plan
unit tests

### Author/Reviewer Checklist


protocol/app/testdata/default_genesis_state.json
```diff
@@ -455,7 +455,16 @@
...
   "upgrade": {},
-  "vault": {},
+  "vault": {
+    "params": {
+      "layers": 2,
+      "spread_min_ppm": 3000,
+      "spread_buffer_ppm": 1500,
+      "skew_factor_ppm": 500000,
+      "order_size_ppm": 100000,
+      "order_expiration_seconds": 5
```
sr33j, line 465: i would recommend this making the `order_expiration_seconds` = 2, since this is more in line with a single block time.

protocol/x/vault/keeper/params_test.go
```diff
@@ -0,0 +1,45 @@
...
+	err := k.SetParams(ctx, newParams)
+	require.NoError(t, err)
+	require.Equal(t, newParams, k.GetParams(ctx))
+
+	// Set invalid params and get.
+	invalidParams := types.Params{
+		Layers:                 3,
+		SpreadMinPpm:           4_000,
+		SpreadBufferPpm:        2_000,
+		SkewFactorPpm:          1_000_000, // invalid
```
sr33j, line 38: talked about this in slack, just repeating here for clarity. I think we should allow skew greater than 1. I think the theoretical bound should be 1/order_size_pct, but don't know if we actively need to check this for v0.

tqin7: will remove check

protocol/app/testdata/default_genesis_state.json
tqin7: each block, we cancel pervious block's order and place new ones. but yeah having it as `2` makes sense

Issue #1239, opened 2024-03-26
Add pagination feature for transfers and trades


giorgionocera:
Given the large amount of data associated with trades and transfers provided by the [`/trades`](https://github.com/dydxprotocol/v4-chain/blob/82c2dec66f9ccab8929ff47ca9f757149329e6dc/indexer/services/comlink/src/controllers/api/v4/trades-controller.ts#L40) and [`/transfers`](https://github.com/dydxprotocol/v4-chain/blob/82c2dec66f9ccab8929ff47ca9f757149329e6dc/indexer/services/comlink/src/controllers/api/v4/transfers-controller.ts#L44) endpoints available through the [Indexer](https://github.com/dydxprotocol/v4-chain/tree/main/indexer), it may be useful to provide some pagination functionality to allow smarter access to the data.

PR #1244, opened 2024-03-26
Add support for pagination at table level.

giorgionocera:
### Changelist
This PR aims to solve #1239 issue.
Specifically, it modifies the low-level functions used by the API provided by the indexer to add the pagination feature.

### Test Plan
Tests are delivered with the features.
We have not tested the environment with a full live (development) environment, but have performed unit tests.
We encourage you to perform full (integration) testing before merging.

### Author/Reviewer Checklist
No breaking changes.

No breaking changes.

No changes in `PrepareProposal` OR `ProcessProposal`.

Probably should add feature:[support-pagination-in-transfers]

## Summary by CodeRabbit

- **New Features**
	- Introduced pagination support across various data retrieval functions, enabling users to fetch results with pagination metadata for better navigation.
	- Enhanced API responses to include pagination details, improving data handling and user interface interactions.

- **Refactor**
	- Modified data handling in controllers and services to destructure results for improved clarity and maintainability.

- **Tests**
	- Updated test cases to align with new data handling structures, ensuring consistency and reliability in automated testing.


indexer/packages/postgres/src/stores/fill-table.ts
```diff
@@ -156,11 +158,40 @@ export async function findAll(
...
+     */
+    const currentPage = Math.max(1, page);
+    const offset = (currentPage - 1) * limit;
+
+    /**
+     * We need to remove the sorting as it is not necessary in this case.
+     * Also a casting of the ts type is required since the infer of the type
+     * obtained from the count is not performed.
+     */
+    const count = await baseQuery.clone().clearOrder().count({ count: '*' }).first() as unknown as { count?: string };
```
vincentwschau, line 180: nit: Please add types to all declared variables.

vincentwschau, line 180: What's the sort order in this case then? Is there some default?

giorgionocera: Here we're just counting all the elements. No matter what order they are in, because the COUNT() function returns the number of lines that match a given criterion. For this reason we get the first element as it is the only one available.

dydxwill: Clients are expecting a reasonable default order per endpoint. That way, random records won't be skipped/repeated in subsequent pages. I think it makes sense to use ascending `eventId` order for trades/fills/transfers related endpoints, and `orderId` for the orders endpoint.

giorgionocera: I think there may be a misunderstanding here. The sort order is only removed for a query that asks for a count (a total number of rows), not for the list of items. 
Such a query is used to retrieve the total number of items, which we return to the client to say "there are x rows in total". This helps the user to know how many records have not yet been received and for which they could make requests.
Regarding the list of items, as you can see, the order is maintained because the `clone()` method _clones the current query chain, which is useful for reusing partial query snippets in other queries without mutating the original_.

I hope this helps to clarify things.

giorgionocera: Yeah, we "fixed" in 0d14f11d9d8046396db741056ad725881fca9699. 
Thank you!

dydxwill: I understand. I don't think the change is required in this file.

But can you pass in an orderBy param set to `eventId` ASC in fills-controller, trades-controller, and transfers-controller? And can you add unit tests to those controllers verifying the order? We also want pagination on the orders endpoint too.

indexer/packages/postgres/src/stores/fill-table.ts
```diff
@@ -156,11 +158,40 @@ export async function findAll(
...
+    const offset: number = (currentPage - 1) * limit;
+
+    /**
+     * We need to remove the sorting as it is not necessary in this case.
+     * Also a casting of the ts type is required since the infer of the type
+     * obtained from the count is not performed.
+     */
+    const count: { count?: string } = await baseQuery.clone().clearOrder().count({ count: '*' }).first() as unknown as { count?: string };
+
+    baseQuery = baseQuery.offset(offset).limit(limit);
```
vincentwschau, line 182: Question: How does this logic interact with additional items being added that are higher in the sort order than items returned?
E.g.
I want fills created after height X in descending order, page size of 10. I get page 2, which is the #10-#20 most recent fills. 10 new fills get added, and now I get page 3, would that be the same set of fills I got for page 2?

giorgionocera: To address your example: **yes, you are right**.

We use offset-based pagination to improve UX through parallel data retrieval, recognising the potential for data inconsistencies with new insertions. While this method provides fast access to any data segment, in certain scenarios it can lead to duplicate entries if data is added between two queries.

Conversely, cursor-based pagination offers better data consistency by using a pointer (cursor) to retrieve data sequentially, eliminating duplicates or missed entries due to data movement. However, it lacks the flexibility of parallel retrieval and direct access to specific data segments.

_The request for such a feature stems from the need to add pagination to retrieve a user's full trading history at the front-end._

**Given the frequency of data updates** relative to request receipt in our context, **we believe that offset-based pagination strikes the best balance for our current needs**, optimising the user experience while considering the trade-off in data consistency. In fact, in this scenario, if a user has a very long history, a sequential approach could actually degrade the user experience, as we would have to wait for each request to continue with the new one.

Anyway, give us more feedback, we are open to adapting our approach as requirements evolve.

indexer/packages/postgres/src/stores/fill-table.ts
giorgionocera: Got it 💪

In such a scenario, I think you might agree that adding the ordering feature is not strictly related to the pagination feature. So it might be interesting (and probably better) to add such a feature, **but via a new issue and pull request to keep the context**.

WRT the ability to add pagination on the orders endpoint too, that might be a good suggestion.

We'll work on it 👷

dydxwill: If that's the case, can you remove all the changes to the API in this PR? We don't want to be in a partial state of having pagination exposed but not having the correct order.

giorgionocera: Yeah. We will remove the "controller level" changes and leave only the "table level" changes.
Then we'll move to a new PR related to the same issue (since you want to keep the pagination and sorting related) to add both pagination and ordering at "controller level" and we can merge the two.

Do you agree?

dydxwill: Agreed, thank you!

indexer/services/roundtable/src/tasks/cancel-stale-orders.ts
```diff
@@ -22,7 +22,7 @@ export default async function runTask(): Promise<void> {
 
   const latestBlockHeight: number = parseInt(latestBlock.blockHeight, 10);
 
-  const staleOpenOrders: OrderFromDatabase[] = await OrderTable.findAll(
+  const { results: staleOpenOrders } = await OrderTable.findAll(
```
giorgionocera, line 25: Can you please add explicit type?

indexer/services/roundtable/__tests__/tasks/cancel-stale-orders.test.ts
```diff
@@ -145,7 +145,7 @@ describe('cancel-stale-orders', () => {
     expect(stats.gauge).toHaveBeenCalledWith('roundtable.num_stale_orders.count', 1);
     expect(stats.gauge).toHaveBeenCalledWith('roundtable.num_stale_orders_canceled.count', 1);
 
-    const ordersAfterTask: OrderFromDatabase[] = await OrderTable.findAll(
+    const { results: ordersAfterTask } = await OrderTable.findAll(
```
giorgionocera, line 148: Can you please add explicit type?

indexer/services/roundtable/__tests__/tasks/cancel-stale-orders.test.ts
```diff
@@ -83,7 +83,7 @@ describe('cancel-stale-orders', () => {
 
     await cancelStaleOrdersTask();
 
-    const ordersAfterTask: OrderFromDatabase[] = await OrderTable.findAll(
+    const { results: ordersAfterTask } = await OrderTable.findAll(
```
giorgionocera, line 86: Can you please add explicit type?

indexer/packages/postgres/__tests__/stores/order-table.test.ts
```diff
@@ -49,7 +49,7 @@ describe('Order store', () => {
   it('Successfully creates an Order with goodTilBlockTime', async () => {
     await OrderTable.create(defaultOrderGoodTilBlockTime);
 
-    const orders: OrderFromDatabase[] = await OrderTable.findAll({}, [], {});
+    const { results: orders } = await OrderTable.findAll({}, [], {});
```
giorgionocera, line 52: Can you please add explicit type?

indexer/packages/postgres/__tests__/stores/order-table.test.ts
```diff
@@ -67,7 +67,7 @@ describe('Order store', () => {
       }),
     ]);
 
-    const orders: OrderFromDatabase[] = await OrderTable.findAll({}, [], {
+    const { results: orders } = await OrderTable.findAll({}, [], {
```
giorgionocera, line 70: Can you please add explicit type?

indexer/packages/postgres/__tests__/stores/order-table.test.ts
```diff
@@ -106,7 +150,7 @@ describe('Order store', () => {
       }),
     ]);
 
-    const orders: OrderFromDatabase[] = await OrderTable.findAll(
+    const { results: orders } = await OrderTable.findAll(
```
giorgionocera, line 153: Can you please add explicit type?

indexer/packages/postgres/__tests__/stores/order-table.test.ts
```diff
@@ -202,7 +246,7 @@ describe('Order store', () => {
         OrderTable.create(defaultOrderGoodTilBlockTime),
       ]);
 
-      const orders: OrderFromDatabase[] = await OrderTable.findAll(
+      const { results: orders } = await OrderTable.findAll(
```
giorgionocera, line 249: Can you please add explicit type?

giorgionocera (commented):
The only thing that needs to be added is the explicit type. Could you please add it?
Thank you for your work! 🚀

giorgionocera (commented):
Thank you!

indexer/packages/postgres/src/stores/fill-table.ts
giorgionocera: As agreed, we have implemented the changes at table level. 
We're now moving to the controller level with a new PR!

indexer/packages/postgres/src/stores/fill-table.ts
```diff
@@ -49,10 +50,11 @@ export async function findAll(
     createdOnOrAfter,
     clientMetadata,
     fee,
+    page,
```
dydxwill, line 53: Ideally, we'd have a check that if page is set, then orderBy must contain eventId. Can you add a TODO for now?

dydxwill (approved):

indexer/packages/postgres/__tests__/stores/order-table.test.ts
dydxwill: +1.

Something like this: 
```
const response: { results: OrderFromDatabase[] } = await OrderTable.findAll({}, [], {});
    const { results: orders } = response;
```

indexer/packages/postgres/src/stores/fill-table.ts
giorgionocera: Done!

dydxwill:
https://github.com/Mergifyio%20backport%20release/indexer/v5.x

PR #1252, opened 2024-03-26
[CT-723] add block number + stage to grpc updates

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/app/app.go
```diff
@@ -1683,6 +1683,8 @@ func (app *App) PreBlocker(ctx sdk.Context, _ *abci.RequestFinalizeBlock) (*sdk.
 
 // BeginBlocker application updates every begin block
 func (app *App) BeginBlocker(ctx sdk.Context) (sdk.BeginBlock, error) {
+	ctx = ctx.WithExecMode(lib.ExecModeBeginBlock)
```
jonfung-dydx, line 1686: wanna do 
```
	ctx = log.AddPersistentTagsToLogger(ctx,
		log.Handler, XXXXXX,
		log.BlockHeight, ctx.BlockHeight(),
	)
```
while you are at it?

protocol/lib/context.go
```diff
@@ -11,6 +11,13 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/lib/log"
 )
 
+// Custom exec modes
+const (
+	ExecModeBeginBlock        = 100
```
jonfung-dydx, line 16: using ints for these values?

protocol/x/clob/keeper/keeper.go
```diff
@@ -233,17 +233,23 @@ func (k Keeper) InitializeNewGrpcStreams(ctx sdk.Context) {
...
+	ctx sdk.Context,
 	offchainUpdates *types.OffchainUpdates,
 	snapshot bool,
 ) {
 	if len(offchainUpdates.Messages) == 0 {
 		return
 	}
 
-	k.GetGrpcStreamingManager().SendOrderbookUpdates(offchainUpdates, snapshot)
+	k.GetGrpcStreamingManager().SendOrderbookUpdates(
```
jonfung-dydx, line 249: don't really feel strongly at all but why don't we pass in the ctx instead of the blockheight and exec mode

protocol/app/app.go
jonfung-dydx: use the strings not the ints though

jonfung-dydx (approved):

protocol/app/app.go
jayy04: you mean to move this from `x/clob`'s begin blocker to here?

protocol/lib/context.go
jayy04: these are meant to augment [these](https://github.com/cosmos/cosmos-sdk/blob/6e9528a2f5e71b38e4866c57b4e9efd82ee4ee12/types/context.go#L18-L31)

protocol/x/clob/keeper/keeper.go
jayy04: I think we can abstract away the context from the streaming manager. logic to get block height/exec mode doesn't really belong there.

let me know if you have a preference

protocol/app/app.go
jonfung-dydx: can just add some tags to the app-level app.go file

jayy04: going to punt on this for now so that I can merge. let's do this in a separate PR

PR #1253, opened 2024-03-26
[TRA-170] Track shares of owners in vaults

tqin7:
### Changelist
Track shares of owners in vaults

### Test Plan
Unit / integration tests

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/vault/keeper/shares.go
+++ b/protocol/x/vault/keeper/shares.go
@@ -62,6 +62,54 @@ func (k Keeper) getTotalSharesIterator(ctx sdk.Context) storetypes.Iterator {
 	return storetypes.KVStorePrefixIterator(store, []byte{})
 }
 
+// GetOwnerShares gets owner shares for an owner in a vault.
+func (k Keeper) GetOwnerShares(
+	ctx sdk.Context,
+	vaultId types.VaultId,
+	owner string,
+) (val types.NumShares, exists bool) {
+	store := k.getVaultOwnerSharesStore(ctx, vaultId)
+
+	b := store.Get([]byte(owner))
+	if b == nil {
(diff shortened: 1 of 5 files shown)


protocol/x/vault/keeper/shares.go
```diff
@@ -127,7 +175,37 @@ func (k Keeper) MintShares(
...
+			vaultId,
+			owner,
+			types.BigRatToNumShares(sharesToMint),
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Increase existing owner shares by sharesToMint.
+		existingOwnerShares, err := ownerShares.ToBigRat()
```
Christopher-Li, line 193: to confirm, there are no race conditions here because everything in abci is synchronous right?

Christopher-Li (approved):

protocol/x/vault/keeper/shares.go
```diff
@@ -127,7 +175,37 @@ func (k Keeper) MintShares(
...
+		if err != nil {
+			return err
+		}
+		err = k.SetOwnerShares(
+			ctx,
+			vaultId,
+			owner,
+			types.BigRatToNumShares(
+				existingOwnerShares.Add(existingOwnerShares, sharesToMint),
```
vincentwschau, line 202: Non-blocking: Since we always store integer denominators / numerators, both these integers can grow fairly quickly if an owner gets shares with denominators that are co-prime / has a large lowest common factor with the existing shares.
E.g. User has 1/1001 shares, deposits to get 1/1002 shares, now the denominator is 1001*1002.
Discussed this with @tqin7 offline, think this may not be an issue short-term but definitely something to think about long-term.

PR #1262, opened 2024-03-27
[TRA-114] implement full vault strategy

tqin7:
### Changelist
implement full vault strategy

### Test Plan
unit tests

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/vault/keeper/orders_test.go
+++ b/protocol/x/vault/keeper/orders_test.go
@@ -9,20 +9,13 @@ import (
 	"github.com/dydxprotocol/v4-chain/protocol/dtypes"
 	testapp "github.com/dydxprotocol/v4-chain/protocol/testutil/app"
 	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	assettypes "github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
 	clobtypes "github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 	perptypes "github.com/dydxprotocol/v4-chain/protocol/x/perpetuals/types"
 	pricestypes "github.com/dydxprotocol/v4-chain/protocol/x/prices/types"
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 	vaulttypes "github.com/dydxprotocol/v4-chain/protocol/x/vault/types"
 	"github.com/stretchr/testify/require"
-)
-
-// TODO (TRA-118): store vault strategy constants in x/vault state.
-const (
-	numLayers                          = uint8(2)
-	minBaseSpreadPpm                   = uint32(3_000) // 30 bps
-	baseSpreadMinPriceChangePremiumPpm = uint32(1_500) // 15 bps
-	orderExpirationSeconds             = uint32(5)     // 5 seconds
 )
 
 func TestRefreshAllVaultOrders(t *testing.T) {
@@ -79,7 +72,7 @@ func TestRefreshAllVaultOrders(t *testing.T) {
 								Id: vaultId.ToSubaccountId(),
 								AssetPositions: []*satypes.AssetPosition{
 									{
-										AssetId:  0,
+										AssetId:  assettypes.AssetUsdc.Id,
 										Quantums: dtypes.NewInt(1_000_000_000), // 1,000 USDC
 									},
 								},
@@ -188,7 +181,7 @@ func TestRefreshVaultClobOrders(t *testing.T) {
 								Id: tc.vaultId.ToSubaccountId(),
 								AssetPositions: []*satypes.AssetPosition{
 									{
-										AssetId:  0,
+										AssetId:  assettypes.AssetUsdc.Id,
 										Quantums: dtypes.NewInt(1_000_000_000), // 1,000 USDC
 									},
 								},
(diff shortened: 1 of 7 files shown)


protocol/x/vault/keeper/orders_test.go
```diff
@@ -249,79 +249,167 @@ func TestGetVaultClobOrders(t *testing.T) {
...
-				Id:                 constants.TestMarketParams[1].Id,
-				Pair:               constants.TestMarketParams[1].Pair,
-				Exponent:           constants.TestMarketParams[1].Exponent,
-				MinExchanges:       constants.TestMarketParams[1].MinExchanges,
-				MinPriceChangePpm:  4_200, // Set a high min price change to test spread calculation.
-				ExchangeConfigJson: constants.TestMarketParams[1].ExchangeConfigJson,
+			vaultParams: vaulttypes.Params{
+				Layers:                 3,       // 3 layers
+				SpreadMinPpm:           3_000,   // 30 bps
+				SpreadBufferPpm:        4_500,   // 15 bps
```
sr33j, line 313: nit: comment doesn't align with value here

protocol/x/vault/keeper/orders_test.go
```diff
@@ -249,79 +249,167 @@ func TestGetVaultClobOrders(t *testing.T) {
...
+				SkewFactorPpm:          700_000, // 0.7
+				OrderSizePpm:           50_000,  // 5%
+				OrderExpirationSeconds: 4,       // 4 seconds
 			},
-			marketPrice: constants.TestMarketPrices[1],
-			perpetual:   constants.EthUsd_20PercentInitial_10PercentMaintenance,
+			vaultId:            constants.Vault_Clob_1,
+			vaultAssetQuantums: big.NewInt(2_000_000_000), // 2,000 USDC
+			vaultInventory:     big.NewInt(-1_000),
```
sr33j, line 320: question: is inventory also in quantums?

protocol/x/vault/keeper/orders_test.go
```diff
@@ -249,79 +249,167 @@ func TestGetVaultClobOrders(t *testing.T) {
...
+				// spreadPpm = max(3_000, 4_500 + 50) = 4_550
+				// spread = 0.00455
+				// open_notional = -1_000 * 10^-9 * 3_000 * 10^6 = -3_000
+				// leverage = -3_000 / (2_000_000_000 - 3_000) = -3/1_999_997
+				// oracleSubticks = 3_000_000_000 * 10^(-6 - (-9) + (-9) - (-6)) = 3 * 10^9
+				// leverage_0 = leverage - 0 * 0.05 = -3/1_999_997
+				// skew_0 = 3 / 1_999_997 * 0.00455 * 0.7
+				// a_0 = 3 * 10^9 * (1 + skew_0) * (1 + 0.00455)^1 = 3_013_650_014.4
+				// round up to nearest multiple of subticks_per_tick=1_000.
+				3_013_651_000,
```
sr33j, line 345: something may be off here
if we have negative inventory, i would expect the quotes to be skewed up, but here they are centered around oracle price.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -249,79 +249,167 @@ func TestGetVaultClobOrders(t *testing.T) {
...
+				// b_2 = 3 * 10^9 * (1 + skew_2) / (1 + 0.00455)^3 ~= 2958477277.194
+				// round down to nearest multiple of subticks_per_tick=1_000.
+				2_958_477_000,
 			},
-			expectedOrderQuantums: []uint64{ // TODO (TRA-144): Implement order size
-				1000,
-				1000,
-				1000,
-				1000,
+			// order_size = 5% * 1999.997 / 3000 ~= 0.03333328333333334
```
sr33j, line 370: question: why is the equity slightly less than $2000 here?

protocol/x/vault/keeper/orders_test.go
```diff
@@ -249,79 +249,167 @@ func TestGetVaultClobOrders(t *testing.T) {
...
-				// a_2 = 3 * 10^9 * (1 + 0.0057)^2 = 3_034_297_470
-				// round up to nearest multiple of subticksPerTick=1000.
-				3_034_298_000,
-				// b_2 = 3 * 10^9 * (1 - 0.0057)^2 = 2_965_897_470
-				// round down to nearest multiple of subticksPerTick=1000.
-				2_965_897_000,
+				// spreadPpm = max(3_000, 4_500 + 50) = 4_550
+				// spread = 0.00455
+				// open_notional = -1_000 * 10^-9 * 3_000 * 10^6 = -3_000
+				// leverage = -3_000 / (2_000_000_000 - 3_000) = -3/1_999_997
```
sr33j, line 339: same question as below, why are we subtracting from the equity in the denominator?

protocol/x/vault/keeper/orders_test.go
sr33j: it might be the units of our inventory which i'm not understanding properly. 
if so can we make this test case have a much larger quantity.

protocol/x/vault/keeper/orders_test.go
tqin7: inventory is in base quantums

protocol/x/vault/keeper/orders_test.go
tqin7: because of the -1000 base quantums perpetual position. Equity takes into account both asset positions (USDC) and perpetual positions. Is that as expected?

protocol/x/vault/keeper/orders_test.go
tqin7: yeah i think the orders are skewed up a little bit. can increase to a larger quantity here to make that more visible

tqin7:
going to rename `order_size_ppm` to `order_size_fraction_ppm` in a separate PR

protocol/x/vault/keeper/orders_test.go
tqin7: see above

Christopher-Li (approved):

protocol/x/vault/keeper/orders_test.go
vincentwschau: Update naming? Or add comment noting how much of the base currency it is?

protocol/x/vault/keeper/orders_test.go
sr33j: im not sure I understand this. 
you could be short 1000 on ETH perp, but have even more equity since the current price is lower than your entry price.

protocol/x/vault/keeper/orders_test.go
sr33j: just making sure vaultAssetQuantums could be negative and the account could still have positive equity if the vaultInventory is valued higher.

note: the accounting is done such that we make USDC on every short and spend USDC for every long.

tqin7: Yeah asset position can be negative and subaccount can still have positive equity

sr33j (approved):
lgtm!

PR #1264, opened 2024-03-27
[TRA-115] decommission vaults at the beginning of a block

tqin7:
### Changelist
in beginblock, decommission a vault if its total shares is positive and equity is non-positive

### Test Plan
unit tests

### Author/Reviewer Checklist


protocol/x/vault/keeper/vault.go
```diff
@@ -0,0 +1,83 @@
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	return netCollateral, nil
+}
+
+// DecommissionVaults decommissions all vaults with positive shares and non-positive equity.
+func (k Keeper) DecommissionVaults(
```
Christopher-Li, line 32: nit: can we clarify this function name? It sounds like `DecomissionVaults` would decomission all of the vaults, maybe `DecommissionVaultsWithNegativeEquity

protocol/x/vault/keeper/vault.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		// Decommission vault if equity is non-positive.
+		if equity.Sign() <= 0 {
+			k.DecommissionVault(ctx, *vaultId)
+		}
+	}
+}
+
+// DecommissionVault decommissions a vault by deleting its tota shares and owner shares.
```
vincentwschau, line 67:
```suggestion
// DecommissionVault decommissions a vault by deleting its total shares and owner shares.
```

protocol/x/vault/keeper/vault.go
```diff
@@ -0,0 +1,83 @@
...
+	defer totalSharesIterator.Close()
+	for ; totalSharesIterator.Valid(); totalSharesIterator.Next() {
+		var totalShares types.NumShares
+		k.cdc.MustUnmarshal(totalSharesIterator.Value(), &totalShares)
+
+		// Skip if TotalShares is non-positive.
+		totalSharesRat, err := totalShares.ToBigRat()
+		if err != nil || totalSharesRat.Sign() <= 0 {
+			continue
+		}
```
Christopher-Li, line 46: is it positive for totalShares to be negative? How could that happen?

protocol/x/vault/keeper/vault.go
```diff
@@ -0,0 +1,83 @@
...
+
+		// Decommission vault if equity is non-positive.
+		if equity.Sign() <= 0 {
+			k.DecommissionVault(ctx, *vaultId)
+		}
+	}
+}
+
+// DecommissionVault decommissions a vault by deleting its tota shares and owner shares.
+func (k Keeper) DecommissionVault(
```
Christopher-Li, line 68: iiuc, if a user attempts to deposit in this vault before it's decommissioned, it will be recreated in the same block its decommissioned and a new vault will be created right?

protocol/x/vault/keeper/vault_test.go
```diff
@@ -0,0 +1,135 @@
...
+	require.Equal(t, true, exists)
+	require.Equal(t, shares, got)
+	got, exists = k.GetOwnerShares(ctx, vault0, constants.Alice_Num0.Owner)
+	require.Equal(t, true, exists)
+	require.Equal(t, shares, got)
+	// Check that total shares and owner shares are deleted for vault 1.
+	_, exists = k.GetTotalShares(ctx, vault1)
+	require.Equal(t, false, exists)
+	_, exists = k.GetOwnerShares(ctx, vault1, constants.Bob_Num0.Owner)
+	require.Equal(t, false, exists)
```
Christopher-Li, line 89: nit: shouldn't this be multiple separate test cases?

protocol/x/vault/keeper/vault_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func TestDecomissionVault(t *testing.T) {
+	tApp := testapp.NewTestAppBuilder(t).Build()
+	ctx := tApp.InitChain()
+	k := tApp.App.VaultKeeper
+
+	// Decomission a non-existent vault.
+	k.DecommissionVault(ctx, constants.Vault_Clob_0)
```
Christopher-Li, line 98: what are you testing with this?

Christopher-Li (approved):

protocol/x/vault/keeper/vault_test.go
```diff
@@ -0,0 +1,135 @@
+package keeper_test
```
vincentwschau, line 1: non-blocking nit: Both tests here can be broken up into separate test cases w/ descriptions which would be more helpful for someone running the test if anything fails.

protocol/x/vault/keeper/vault.go
tqin7: negative shouldn't happen, but just added this check here to be future-proof

protocol/x/vault/keeper/vault.go
tqin7: yep! correct

protocol/x/vault/keeper/vault_test.go
tqin7: making sure that the function doesn't panic on non-existent vaults

protocol/x/vault/keeper/vault.go
vincentwschau: Is that the case? If a user deposits to the vault before begin block, then it would have happened in the previous block. What re-creates it in the new block?

protocol/x/vault/keeper/vault.go
vincentwschau: Curious, will this have to be updated when withdrawal logic is added? It's possible for a vault to have 0 shares and 0 equity in that case right? Non-blocking as that isn't part of Phase 1.

tqin7: yeah 0 shares is possible. will think about that

protocol/x/vault/keeper/vault.go
tqin7: deposit tx is processed after begin block. so if a vault is ready to be decommissioned, begin block would decommission it and then when deposit tx is processed, it will be created again.

if a user deposits before begin block, then that deposit tx is processed in the previous block

PR #1273, opened 2024-03-28
[TRA-111] add vault query

tqin7:
### Changelist
add vault query

### Test Plan
unit tests

### Author/Reviewer Checklist


proto/dydxprotocol/vault/query.proto
```diff
@@ -22,3 +28,25 @@ message QueryParamsRequest {}
...
+}
+
+// QueryVaultResponse is a response type for the Vault RPC method.
+message QueryVaultResponse {
+  VaultId vault_id = 1 [ (gogoproto.nullable) = false ];;
+  dydxprotocol.subaccounts.SubaccountId subaccount_id = 2 [ (gogoproto.nullable) = false ];;
+  uint64 equity = 3;
+  uint64 inventory = 4;
+  uint64 total_shares = 5;
+  repeated OwnerShares all_owner_shares = 6 [ (gogoproto.nullable) = false ];;
```
vincentwschau, line 45: nit: Might be better to have a separate paginated query for this, considering there could be a lot of owners.

proto/dydxprotocol/vault/query.proto
```diff
@@ -22,3 +28,26 @@ message QueryParamsRequest {}
...
+
+// QueryVaultResponse is a response type for the Vault RPC method.
+message QueryVaultResponse {
+  VaultId vault_id = 1 [ (gogoproto.nullable) = false ];
+  dydxprotocol.subaccounts.SubaccountId subaccount_id = 2
+      [ (gogoproto.nullable) = false ];
+  uint64 equity = 3;
+  uint64 inventory = 4;
+  uint64 total_shares = 5;
+  repeated OwnerShares all_owner_shares = 6 [ (gogoproto.nullable) = false ];
```
vincentwschau, line 46: nit: Might be better to have a separate paginated query for this considering there may be many owners.

proto/dydxprotocol/vault/query.proto
tqin7: makes sense. created a ticket to add a separate query for owner shares. let me remove owner shares for now

proto/dydxprotocol/vault/query.proto
tqin7: makes sense. created a ticket to add a separate query for owner shares. let me remove owner shares for now

vincentwschau: Non-blocking, I see that the owner shares are stored as only integers, and not a struct with both the owner + the shares so it won't be as simple as using `query.Paginate` on the store.

tqin7: Yep will add a separate query for owner shares in phase 2

vincentwschau (approved):

PR #1282, opened 2024-03-29
[TRA-105] Add API for parent subaccount perpetual positions

shrenujb:
### Changelist
Add API to get perpetual positions by parent subaccount number

### Test Plan
Added tests

### Author/Reviewer Checklist


indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
```diff
@@ -150,6 +156,143 @@ class PerpetualPositionsController extends Controller {
...
+  ): Promise<PerpetualPositionResponse> {
+    // Get subaccountIds for all child subaccounts of the parent subaccount
+    // Create a record of subaccountId to subaccount number
+    const childIdtoSubaccountNumber: Record<string, number> = {};
+    getChildSubaccountNums(parentSubaccountNumber).forEach(
+      (subaccountNum: number) => {
+        childIdtoSubaccountNumber[SubaccountTable.uuid(address, subaccountNum)] = subaccountNum;
+      },
+    );
+    const childSubaccountIds: string[] = Object.keys(childIdtoSubaccountNumber);
```
Christopher-Li, line 177: nit: I think this should be in a separate function since we need it for every one of these parent subaccount number endpoints

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
```diff
@@ -150,6 +156,143 @@ class PerpetualPositionsController extends Controller {
...
+      markets,
+    ]: [
+      PerpetualPositionFromDatabase[],
+      MarketFromDatabase[],
+    ] = await Promise.all([
+      PerpetualPositionTable.findAll(
+        {
+          subaccountId: childSubaccountIds,
+          status,
+          limit,
```
Christopher-Li, line 190: do we want to remove limit here?

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
```diff
@@ -150,6 +156,143 @@ class PerpetualPositionsController extends Controller {
...
+        ),
+        BlockTable.getLatest(),
+      ]);
+
+      if (subaccounts.length === 0 || latestBlock === undefined) {
+        throw new NotFoundError(
+          `Found OPEN perpetual positions but no subaccounts with address ${address} ` +
+            `and parent Subaccount number ${parentSubaccountNumber}`,
+        );
+      }
```
Christopher-Li, line 237: I don't think this is possible, because we have a foreign key on the positions table for the subaccount to exist

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
shrenujb: There does exist a helper function to getChildSubaccountIds already but I also need to create the map above for this controller specifically, so this line here does not make a difference

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
shrenujb: This existed in the listPositions API and was brought over as a result. Should this be removed in the original API as well and if so, why?

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
shrenujb: I agree it shouldn't be possible. Also brought over from the original API. Should I remove it from there too?

indexer/services/comlink/src/controllers/api/v4/perpetual-positions-controller.ts
vincentwschau: We should keep the limit, this should be the 100 open perpetual positions on the parent-subaccount so it would include positions on child subaccounts.

vincentwschau (approved):

PR #1301, opened 2024-04-01
Remove custom ping message from socks

dydxwill:
### Changelist
- Remove custom ping message from socks. It'll be considered a no-op now. In the future, after updating documentation, we should consider this an invalid message.
- Add handling for default ping messages.
- After experimenting in testnet, I saw that we continue to receive a bunch of messages after hitting the rate limits & calling `ws.close` for the old custom ping messages. So, explicitly remove event listening on messages when we are rate limited for invalid messages.

### Test Plan
Unit tested.

### Author/Reviewer Checklist


indexer/services/socks/src/websocket/index.ts
```diff
@@ -396,7 +382,6 @@ export class Index {
 
       // Delete subscription data.
       this.subscriptions.remove(connectionId);
-      this.pingHandler.handleDisconnect(connectionId);
```
vincentwschau, line 399: nit: Can you also remove the `onMessage` listener in `disconnect`, before the `terminate` call? If there is no heartbeat, we directly call `disconnect` rather than `ws.close` so we'd want to ensure there's no message listener on these connections as well.

dydxwill: I don't think that's necessary. Terminate should destroy the websocket. https://stackoverflow.com/questions/74081816/when-is-it-preferable-to-use-ws-terminate-vs-ws-close

dydxwill: Nvm added

vincentwschau (approved):

PR #1308, opened 2024-04-01
[SKI-23] Add v5.0.0 enable vote extensions upgrade handler

Eric-Warehime:
### Changelist
Adds enable vote extension upgrade handler and test.

### Test Plan
Added container test.

### Author/Reviewer Checklist


protocol/app/upgrades/v5.0.0/upgrade_container_test.go
```diff
@@ -120,6 +137,20 @@ func preUpgradeCheckPerpetualMarketType(node *containertest.Node, t *testing.T)
...
+	resp, err := containertest.Query(
+		node,
+		consensus.NewQueryClient,
+		consensus.QueryClient.Params,
+		&consensus.QueryParamsRequest{},
+	)
+	require.NoError(t, err)
+	err = proto.UnmarshalText(resp.String(), consensusParams)
+	require.NoError(t, err)
+	assert.True(t, consensusParams.Params.Abci.VoteExtensionsEnableHeight > int64(0))
```
nivasan1, line 151: Shld this not be checking that the `VoteExtensionEnableHeight == height_of_upgrade + 1`

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -204,6 +206,39 @@ func initializeOIMFCaps(
...
+	ctx.Logger().Info(
+		fmt.Sprintf(
+			"Successfully set VoteExtensionsEnableHeight to %d\n",
+			veEnableHeight,
+		),
+	)
+}
+
 // Initialize open interest for perpetuals
 func initializePerpOpenInterest(
```
nivasan1, line 243: Is this the upgrade-handler entry-point? I'd prefer that we explicitly say this is upgrading consensus-params

nivasan1 (approved):
lgtm

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -204,6 +206,39 @@ func initializeOIMFCaps(
...
+	currentParams, err := keeper.Params(ctx, &consensustypes.QueryParamsRequest{})
+	if err != nil {
+		panic(fmt.Sprintf("failed to retrieve existing consensus params in VE upgrade handler: %s", err))
+	}
+	if currentParams.Params.Abci.VoteExtensionsEnableHeight != 0 {
+		panic(fmt.Sprintf(
+			"unable to update VE Enable Height since its current value of %d is already non-zero",
+			currentParams.Params.Abci.VoteExtensionsEnableHeight))
+	}
+	veEnableHeight := ctx.BlockHeight() + 1
```
davidterpay, line 222: I would prefer to have the enabled height be buffered with a few heights in between. Given that comet and cosmos have delayed updates for things like val-sets, it may also be the the case the consensus params are delayed in how quickly they are propagated.

Something like `ctx.BlockHeight() + 4/5`

protocol/app/upgrades/v5.0.0/upgrade_container_test.go
Christopher-Li: + 1

protocol/app/upgrades/v5.0.0/upgrade.go
Christopher-Li: Could you elaborate on this, what is the difference between using currentHeight + 1 versus currentHeight + 5

protocol/app/upgrades/v5.0.0/upgrade_container_test.go
Eric-Warehime: It was not clear to me how to get the height of the upgrade via these container test objects. Can see if it's doable though.

Eric-Warehime: Updated.

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -204,6 +206,39 @@ func initializeOIMFCaps(
...
+	currentParams, err := keeper.Params(ctx, &consensustypes.QueryParamsRequest{})
+	if err != nil {
+		panic(fmt.Sprintf("failed to retrieve existing consensus params in VE upgrade handler: %s", err))
+	}
+	if currentParams.Params.Abci.VoteExtensionsEnableHeight != 0 {
+		panic(fmt.Sprintf(
+			"unable to update VE Enable Height since its current value of %d is already non-zero",
+			currentParams.Params.Abci.VoteExtensionsEnableHeight))
+	}
+	veEnableHeight := ctx.BlockHeight() + 4
```
Christopher-Li, line 222: I think + 1 should actually be fine according to this documentation:
https://docs.cosmos.network/main/build/architecture/adr-064-abci-2.0#upgrading

Eric-Warehime: Yeah, I think david wants to get a little bit of separation between when the other upgrades are applied and when the VEs start hitting the chain. I don't really think it makes a different one way or another personally.

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -204,6 +206,39 @@ func initializeOIMFCaps(
...
+	currentParams.Params.Abci.VoteExtensionsEnableHeight = veEnableHeight
```
teddyding, line 223: Nit: inline and declare constant in `v5.0.0/constants.go` for readability

`currentParams.Params.Abci.VoteExtensionsEnableHeight = ctx.BlockHeight() + VoteExtensionEnableHeightDelta`

teddyding, line 223: Nit: inline and define constant in `v5.0.0/constants.go`, e.g.:

```
currentParams.Params.Abci.VoteExtensionsEnableHeight = ctx.BlockHeight() + VEEnableHeightDelta
```

protocol/app/upgrades/v5.0.0/upgrade.go
```diff
@@ -204,6 +206,39 @@ func initializeOIMFCaps(
...
+		Validator: currentParams.Params.Validator,
+		Abci:      currentParams.Params.Abci,
+	})
+	if err != nil {
+		panic(fmt.Sprintf("failed to update consensus params with VE Enable Height %d: %s", veEnableHeight, err))
+	}
+	ctx.Logger().Info(
+		fmt.Sprintf(
+			"Successfully set VoteExtensionsEnableHeight to %d\n",
+			veEnableHeight,
```
teddyding, line 237: Nit: just log out the entire `currentParams.Params` so you don't need to define `veEnableHeight`?

protocol/app/upgrades/v5.0.0/upgrade_container_test.go
```diff
@@ -120,6 +137,20 @@ func preUpgradeCheckPerpetualMarketType(node *containertest.Node, t *testing.T)
...
+	resp, err := containertest.Query(
+		node,
+		consensus.NewQueryClient,
+		consensus.QueryClient.Params,
+		&consensus.QueryParamsRequest{},
+	)
+	require.NoError(t, err)
+	err = proto.UnmarshalText(resp.String(), consensusParams)
+	require.NoError(t, err)
+	assert.Equal(t, int64(14), consensusParams.Params.Abci.VoteExtensionsEnableHeight)
```
teddyding, line 151: Magic number - can we add documentation

protocol/app/upgrades/v5.0.0/upgrade.go
teddyding: Agree either could work. For my own understanding, did we remove price daemon in `v5` or is it available as a backup? If the former, does this mean no price update will happen for `deltaBlock`?

Eric-Warehime: If you use cosmovisor and your node restarts on the upgrade height it should be running with Prepare/Process Proposal handlers which expect price updates to come from vote extensions. If vote extensions are not yet enabled in the consensus parameters, it will generate an empty `MsgUpdateMarketPrices` and the ProcessProposal handler will only accept an empty message as well.

So to answer your question, yes every block we delay VEs for will have no price updates in it.

Christopher-Li (approved):

Christopher-Li:
https://github.com/Mergifyio backport release/protocol/v5.x

PR #1327, opened 2024-04-03
Remove unnecessary message serialization in clob AnteHandle

roy-dydx:
### Changelist
Especially due to message serialization, this is more expensive than it should be. If msg is necessary, it can be serialized in individual logs.

### Test Plan
N/A

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/ante/clob.go
+++ b/protocol/x/clob/ante/clob.go
@@ -63,20 +63,8 @@ func (cd ClobDecorator) AnteHandle(
 	msgs := tx.GetMsgs()
 	var msg = msgs[0]
 
-	// Set request-level logging tags
-	ctx = log.AddPersistentTagsToLogger(ctx,
-		log.Module, log.Clob,
-		log.Callback, lib.TxMode(ctx),
-		log.BlockHeight, ctx.BlockHeight()+1,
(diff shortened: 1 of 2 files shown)


protocol/x/clob/ante/clob.go
```diff
@@ -63,20 +63,8 @@ func (cd ClobDecorator) AnteHandle(
 	msgs := tx.GetMsgs()
 	var msg = msgs[0]
 
-	// Set request-level logging tags
-	ctx = log.AddPersistentTagsToLogger(ctx,
-		log.Module, log.Clob,
-		log.Callback, lib.TxMode(ctx),
```
roy-dydx, line 69: @jonfung-dydx Lmk if there are any specific logs that need Msg. Don't want to serialize it for every CheckTx.

jonfung-dydx: leave the Module=Clob tag

roy-dydx: Leaving that out for now to avoid the hundreds or thousands of allocs per second.

roy-dydx:
@Mergifyio backport release/protocol/v5.x

PR #1333, opened 2024-04-04
[TRA-213] return error if shares to mint is rounded down to zero

tqin7:
### Changelist
return error if shares to mint is rounded down to zero

### Test Plan
- unit tests
- localnet

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/vault/keeper/shares.go
+++ b/protocol/x/vault/keeper/shares.go
@@ -143,6 +143,11 @@ func (k Keeper) MintShares(
 		sharesToMint = new(big.Int).Set(quantumsToDeposit)
 		sharesToMint = sharesToMint.Mul(sharesToMint, existingTotalShares)
 		sharesToMint = sharesToMint.Quo(sharesToMint, equity)
+
+		// Return error if `sharesToMint` is rounded down to 0.
+		if sharesToMint.Sign() == 0 {
+			return types.ErrZeroSharesToMint
+		}
 	}
 
 	// Increase TotalShares of the vault.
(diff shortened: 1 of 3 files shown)


protocol/x/vault/keeper/shares.go
```diff
@@ -143,6 +143,11 @@ func (k Keeper) MintShares(
 		sharesToMint = new(big.Int).Set(quantumsToDeposit)
 		sharesToMint = sharesToMint.Mul(sharesToMint, existingTotalShares)
 		sharesToMint = sharesToMint.Quo(sharesToMint, equity)
+
+		// Return error if `sharesToMint` is rounded down to 0.
+		if sharesToMint.Sign() == 0 {
```
shrenujb, line 148: Does this mean that at some point, there will be a minimum amount needed to be deposited into the vault in order to get owner shares?

Christopher-Li (approved):

protocol/x/vault/keeper/shares.go
tqin7: just looking to prevent the scenario where a user deposits some amount but the shares they get are rounded down to 0

shrenujb: Makes sense but in that case you would return an error saying you can't deposit this amount due to a rounding error. Would it make sense to add the minimum deposit amount required to the error?

vincentwschau: This does mean that there will be a minimum amount to be deposited per vault, however it won't be a hard limit but something based on the equity / shares ratio of the vault.

shrenujb: > This does mean that there will be a minimum amount to be deposited per vault,

We should probably specify what this minimum amount is in the error message and give the user more clues rather than "Cannot mint 0 shares"?

tqin7:
@Mergifyio backport release/protocol/v5.x

PR #1385, opened 2024-04-11
Stream DeliverTx Order Fills

jonfung-dydx:


protocol/app/app.go
```diff
@@ -343,6 +344,9 @@ type App struct {
 	// Slinky
 	oraclePrometheusServer *promserver.PrometheusServer
 	oracleMetrics          servicemetrics.Metrics
+
+	// Grpc Streaming Test Client
+	GrpcStreamingTestClient *streamingclient.GrpcClient
```
jayy04, line 349: we shouldn't commit this to main or your feature branch. this should be for testing only

PR #1419, opened 2024-04-26
[OTE-255]: Fix RemoveDisallowMsgs: tx parse error

Christopher-Li:
### Changelist
Fix RemoveDisallowMsgs: tx parse error

### Test Plan
Unit tests

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/prepare/other_msgs.go
+++ b/protocol/app/prepare/other_msgs.go
@@ -9,8 +9,11 @@ import (
 	"github.com/cosmos/gogoproto/proto"
 	gometrics "github.com/hashicorp/go-metrics"
 
+	"github.com/dydxprotocol/v4-chain/protocol/app/constants"
 	"github.com/dydxprotocol/v4-chain/protocol/lib/ante"
 	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+	compression "github.com/skip-mev/slinky/abci/strategies/codec"
+	"github.com/skip-mev/slinky/abci/ve"
 )
 
 // GetGroupMsgOther returns two separate slices of byte txs given a single slice of byte txs and max bytes.
@@ -36,7 +39,12 @@ func GetGroupMsgOther(availableTxs [][]byte, maxBytes uint64) ([][]byte, [][]byt
 }
 
 // RemoveDisallowMsgs removes any txs that contain a disallowed msg.
-func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) [][]byte {
+func RemoveDisallowMsgs(
+	ctx sdk.Context,
+	decoder sdk.TxDecoder,
+	extCommitCodec compression.ExtendedCommitCodec,
+	txs [][]byte,
+) [][]byte {
 	defer telemetry.ModuleMeasureSince(
 		ModuleName,
 		time.Now(),
@@ -46,6 +54,16 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
+		// If tx is index 0 and VE enabled, this ts is a VE tx and should be decoded with the extCommitCodec.
+		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
+			_, err := extCommitCodec.Decode(txBytes)
+			if err != nil {
+				ctx.Logger().Error(fmt.Sprintf("RemoveDisallowMsgs: failed to decode VE tx: %v", err))
+			}
+			// VE tx should never be removed
+			continue
+		}
+
 		// Decode tx so we can read msgs.
 		tx, err := decoder(txBytes)
 		if err != nil {
@@ -54,7 +72,7 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 		}
 
 		// For each msg in tx, check if it is disallowed.
-		containsDisllowMsg := false
+		containsDisallowMsg := false
 		for _, msg := range tx.GetMsgs() {
 			if ante.IsDisallowExternalSubmitMsg(msg) {
 				telemetry.IncrCounterWithLabels(
@@ -62,13 +80,13 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 					1,
 					[]gometrics.Label{metrics.GetLabelForStringValue(metrics.Detail, proto.MessageName(msg))},
 				)
-				containsDisllowMsg = true
+				containsDisallowMsg = true
 				break // break out of loop over msgs.
 			}
 		}
 
 		// If tx contains disallowed msg, skip it.
-		if containsDisllowMsg {
+		if containsDisallowMsg {
 			ctx.Logger().Error(
 				fmt.Sprintf("RemoveDisallowMsgs: skipping tx with disallowed msg. Size: %d", len(txBytes)))
 			continue // continue to next tx.
--- a/protocol/app/prepare/other_msgs.go
+++ b/protocol/app/prepare/other_msgs.go
@@ -54,13 +54,15 @@ func RemoveDisallowMsgs(
 
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
-		// If tx is index 0 and VE enabled, this ts is a VE tx and should be decoded with the extCommitCodec.
+		// If tx is index 0 and VE enabled, this tx is a VE tx and should be decoded with the extCommitCodec.
 		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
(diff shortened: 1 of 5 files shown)


protocol/app/prepare/prepare_proposal.go
```diff
@@ -75,7 +77,7 @@ func PrepareProposalHandler(
 		}
 
 		// Grab the injected VEs from the previous block.
-		// If VEs are not enabled, no tx will have been injected.
+		// If VEs are not enabled,
```
teddyding, line 80: Nit: comment is not complete

protocol/app/prepare/other_msgs.go
```diff
@@ -46,6 +54,16 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
+		// If tx is index 0 and VE enabled, this ts is a VE tx and should be decoded with the extCommitCodec.
+		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
```
nivasan1, line 58: Should we be removing any other transactions that decode into `extCommitInfo`? for instance if a proposer attempts to include an extended commit tx later in the block as well?

protocol/app/prepare/prepare_proposal.go
```diff
@@ -75,8 +77,9 @@ func PrepareProposalHandler(
 		}
 
 		// Grab the injected VEs from the previous block.
-		// If VEs are not enabled, no tx will have been injected.
 		var extCommitBzTx []byte
+		// Sanity check to ensure that there is at least 1 tx. This should never return false unless
+		// before VE are enabled, there are no tx in the block.
```
ttl33, line 82: I am confused around "there are no tx in the block". Shouldn't regular user submitted txs be part of req.Txs?

protocol/app/prepare/other_msgs.go
ttl33, line 58: can there be a situation where the VE is enabled but the first tx is not oracle VE info? for example, what happens when there's no oracle update because validators didn't add VEs?

protocol/app/prepare/other_msgs.go
```diff
@@ -46,6 +54,16 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
+		// If tx is index 0 and VE enabled, this ts is a VE tx and should be decoded with the extCommitCodec.
```
ttl33, line 57:
```suggestion
		// If tx is index 0 and VE enabled, this tx is a VE tx and should be decoded with the extCommitCodec.
```

protocol/app/prepare/other_msgs.go
```diff
@@ -46,6 +54,16 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
...
+		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
+			_, err := extCommitCodec.Decode(txBytes)
+			if err != nil {
+				ctx.Logger().Error(fmt.Sprintf("RemoveDisallowMsgs: failed to decode VE tx: %v", err))
+			}
+			// VE tx should never be removed
```
ttl33, line 63:
```suggestion
			// VE tx should not be part of `OtherMsgs`. It's already used by price update info.
```

protocol/app/prepare/other_msgs.go
nivasan1: The expected behavior of the dydx PrepareProposal handlers is that the first tx of a block (after vote-extensions are enabled) will always unmarshal to an `extendedCommitInfo`, if validators broadcast empty vote-extensions in the previous block (assume ves were enabled 2 blocks ago), the first tx will contain an extendedCommitInfo containing the empty vote-extensions.

protocol/app/prepare/other_msgs.go
```diff
@@ -36,7 +39,12 @@ func GetGroupMsgOther(availableTxs [][]byte, maxBytes uint64) ([][]byte, [][]byt
 }
 
 // RemoveDisallowMsgs removes any txs that contain a disallowed msg.
-func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) [][]byte {
+func RemoveDisallowMsgs(
+	ctx sdk.Context,
+	decoder sdk.TxDecoder,
+	extCommitCodec compression.ExtendedCommitCodec,
+	txs [][]byte,
```
ttl33, line 46: nit/alternative idea: when VE is enabled, the txs passed into this function should have already filtered out the VE txs. For example, use `txs[1:][]` to pass this function. 

the reason why i'd recommend this is so that this function remains true to its original intention of just filtering out DisallowMsgs. it seems like we are piggy backing this extra func

ttl33: this is optional and can be done later when we have more time

Christopher-Li: The only reason I'm not doing this is because after the v5.0.0 upgrade, there is [an initial 4 blocks](https://github.com/dydxprotocol/v4-chain/blob/8454dc9dbbfe571f824a6df7b82ef07616ed508f/protocol/app/upgrades/v5.0.0/upgrade.go#L223) where VE are not enabled. During those 4 blocks, we still want to pass in all transactions. I found this logic to be simplier

protocol/app/prepare/other_msgs.go
Christopher-Li: They will be removed because the decoder on line 70 will be unable to decode it and the tx will not be included in the final list of tx

nivasan1: That makes sense, we should be error logging in those scenarios anyway

protocol/app/prepare/other_msgs.go
```diff
@@ -46,6 +54,18 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
...
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
+		// If tx is index 0 and VE enabled, this tx is a VE tx and should be decoded with the extCommitCodec.
+		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
+			_, err := extCommitCodec.Decode(txBytes)
+			if err != nil {
+				ctx.Logger().Error(fmt.Sprintf("RemoveDisallowMsgs: failed to decode VE tx: %v", err))
+			}
+			// VE tx is excluded from `OtherMsgs` (`continue` basically skips this tx from being
+			// included in `OtherMsgs`. This is because the VE tx info is already used by price
```
ttl33, line 64:
```suggestion
			// included in `OtherMsgs`). This is because the VE tx info is already used by price
```

protocol/app/prepare/other_msgs.go
```diff
@@ -46,6 +54,18 @@ func RemoveDisallowMsgs(ctx sdk.Context, decoder sdk.TxDecoder, txs [][]byte) []
 
 	var filteredTxs [][]byte
 	for i, txBytes := range txs {
+		// If tx is index 0 and VE enabled, this tx is a VE tx and should be decoded with the extCommitCodec.
+		if i == constants.OracleInfoIndex && ve.VoteExtensionsEnabled(ctx) {
+			_, err := extCommitCodec.Decode(txBytes)
```
ttl33, line 59: This feels like a waste of work (ie decoding). In the final state, the protocol will expect the first txbytes to be VE related. 

Can we file a ticket to remove/skip this? I think we should either pass in `txs[1:]` as an input after the upgrade.

ttl33 (approved):

protocol/app/prepare/prepare_proposal.go
Christopher-Li: what I meant is that if there are no user submitted txs to be a part of req.Txs, then the block would be empty right?

protocol/app/prepare/other_msgs.go
Christopher-Li: yup working towards that now

Christopher-Li: take a look a neweest implementation

nivasan1 (approved):
lgtm

Christopher-Li:
https://github.com/Mergifyio backport release/protocol/v5.x

protocol/app/prepare/prepare_proposal.go
ttl33: I see. I think we should file a ticket and add TODOs to remove any "do this before VE is enabled" logic once v5.0 ships.

What do you think?

Christopher-Li: I can file a ticket to just clean it up, I think adding the TODOs at this point would be more work

Christopher-Li: https://linear.app/dydx/issue/SKI-36/cleanup-all-code-that-only-triggers-if-ve-is-not-enabled

PR #1448, opened 2024-04-30
Orderbook Fills emission

jonfung-dydx:
Emit orderbook fills through grpc stream.

Tested via
https://github.com/dydxprotocol/v4-chain/pull/1471/commits


protocol/x/clob/abci.go
```diff
@@ -174,45 +174,6 @@ func PrepareCheckState(
 		offchainUpdates,
 	)
 
-	// For orders that are filled in the last block, send an orderbook update to the grpc streams.
```
jonfung-dydx, line 177: Axed this whole section to send out order updates.
Reason being:
- PCS check state is set to deliver state, so no additional updates are needed.
- since fills are directly emitted immediately on orderbook update, matched orders don't need to be updated

protocol/x/clob/keeper/order_state.go
```diff
@@ -276,6 +276,19 @@ func (k Keeper) RemoveOrderFillAmount(ctx sdk.Context, orderId types.OrderId) {
 		[]byte(types.OrderAmountFilledKeyPrefix),
 	)
 	memStore.Delete(orderId.ToStateKey())
+
+	// If grpc stream is on, zero out the fill amount.
```
jonfung-dydx, line 280: Zero'ing out fill amounts when fill amounts are removed. I added this to the innermost statewrite/remove method. I could have added it to each of the 3-4 callsites but this might be cleaner. Let me know if the other approach is preferred.

protocol/x/clob/keeper/order_state.go
```diff
@@ -286,27 +299,5 @@ func (k Keeper) PruneStateFillAmountsForShortTermOrders(
...
-					ctx,
-					orderId,
-					0, // Total filled quantums is zero because it's been pruned from state.
-				); success {
-					allUpdates.AddUpdateMessage(orderId, message)
-				}
-			}
-		}
-		k.SendOrderbookUpdates(ctx, allUpdates, false)
-	}
```
jonfung-dydx, line 311: No need since this zero'ing out fills is punted to the innermost state remove method

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -141,6 +141,63 @@ func (sm *GrpcStreamingManagerImpl) SendOrderbookUpdates(
 	}
 }
 
+// SendOrderbookFillUpdates groups fills by their clob pair ids and
+// sends messages to the subscribers.
+func (sm *GrpcStreamingManagerImpl) SendOrderbookFillUpdates(
+	ctx sdk.Context,
+	orderbookFills []clobtypes.StreamOrderbookFill,
+) {
+	// Group fills by clob pair id.
```
jayy04, line 150: nit: add a latency metric for this

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -141,6 +141,63 @@ func (sm *GrpcStreamingManagerImpl) SendOrderbookUpdates(
...
+				streamUpdatesForSubscription = append(streamUpdatesForSubscription, update...)
+			}
+		}
+
+		if len(streamUpdatesForSubscription) > 0 {
+			if err := subscription.srv.Send(
+				&clobtypes.StreamOrderbookUpdatesResponse{
+					Updates:     streamUpdatesForSubscription,
+					BlockHeight: lib.MustConvertIntegerToUint32(ctx.BlockHeight()),
+					ExecMode:    uint32(ctx.ExecMode()),
```
jayy04, line 186: nit: let's pass this in explicitly similar to the other function (in case we move this to a separate go routine, we don't have to worry about ctx changing under the hood

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -141,6 +141,63 @@ func (sm *GrpcStreamingManagerImpl) SendOrderbookUpdates(
 	}
 }
 
+// SendOrderbookFillUpdates groups fills by their clob pair ids and
+// sends messages to the subscribers.
+func (sm *GrpcStreamingManagerImpl) SendOrderbookFillUpdates(
```
jayy04, line 146: similar looking functions - worth to create helper?

nbd - will let you make the call

protocol/x/clob/keeper/process_operations.go
```diff
@@ -19,6 +19,25 @@ import (
...
+func fetchOrdersInvolvedInOpQueue(
+	operations []types.InternalOperation,
+) (orderIdSet map[types.OrderId]struct{}) {
+	orderIdSet = make(map[types.OrderId]struct{})
+	for _, operation := range operations {
+		if shortTermOrderPlacement := operation.GetShortTermOrderPlacement(); shortTermOrderPlacement != nil {
+			orderId := shortTermOrderPlacement.GetOrder().OrderId
+			orderIdSet[orderId] = struct{}{}
+		}
+		if clobMatch := operation.GetMatch(); clobMatch != nil {
```
jonfung-dydx, line 33: I added all the order ids involved in a match, but technically there were no errors in `dev` without this statement. 
I think technically we can remove this since I think it's an invariant that every short term order Id that shows up in the any opqueue matches must appear before as a short term order placement. Additionally, there would be no short term order placements that show up without being involved in a match.

What do you think?

protocol/x/clob/keeper/process_operations.go
```diff
@@ -38,6 +57,26 @@ func (k Keeper) ProcessProposerOperations(
 		return errorsmod.Wrapf(types.ErrInvalidMsgProposedOperations, "Error: %+v", err)
 	}
 
+	// If grpc streams are on, send absolute fill amounts from local + proposed opqueue to the grpc stream.
+	// This must be sent out to account for checkState being discarded and deliverState being used.
+	if streamingManager := k.GetGrpcStreamingManager(); streamingManager.Enabled() {
+		localValidatorOperationsQueue, _ := k.MemClob.GetOperationsToReplay(ctx)
```
jonfung-dydx, line 63: Venn diagram. Update both proposed and local.

protocol/x/clob/keeper/process_operations.go
```diff
@@ -518,6 +559,26 @@ func (k Keeper) PersistMatchOrdersToState(
...
+					MatchOrders: matchOrders,
+				},
+			},
+			&takerOrder,
+			makerOrders,
+		)
+		streamingManager.SendOrderbookFillUpdates(
+			ctx,
+			[]types.StreamOrderbookFill{
+				streamOrderbookFill,
```
jonfung-dydx, line 577: The only location in the code in which we emit fills through the grpc stream.

protocol/x/clob/memclob/memclob.go
```diff
@@ -386,7 +386,20 @@ func (m *MemClobPriceTimePriority) mustUpdateMemclobStateWithMatches(
 	}
 
 	// Add the new matches to the operations queue.
-	m.operationsToPropose.MustAddMatchToOperationsQueue(takerOrder, makerFillWithOrders)
+	internalOperation := m.operationsToPropose.MustAddMatchToOperationsQueue(takerOrder, makerFillWithOrders)
+	// If orderbook updates are on, send an orderbook update with the fill to grpc streams.
+	if m.generateOrderbookUpdates {
```
jonfung-dydx, line 391: Optimistic fill emission for any matches that occur on this validator

protocol/x/clob/memclob/memclob.go
```diff
@@ -876,6 +889,20 @@ func (m *MemClobPriceTimePriority) matchOrder(
...
+		if !order.IsLiquidation() {
+			normalOrder := order.MustGetOrder()
+			updates := m.GetOrderbookUpdatesForOrderUpdate(ctx, normalOrder.OrderId)
+			allUpdates.Append(updates)
+		}
+		for _, fill := range newMakerFills {
+			updates := m.GetOrderbookUpdatesForOrderUpdate(ctx, fill.MakerOrderId)
+			allUpdates.Append(updates)
+		}
+		m.clobKeeper.SendOrderbookUpdates(ctx, allUpdates, false)
```
jonfung-dydx, line 905: Removing this didn't trigger any errors, but i think we still should have this -- It makes sense that we need to reset fill amounts if state was discarded.

2 qns
- do we need to reset anything else other than the order ids in the fills?
- do we need to reset orderbook placements for your side of streaming?

protocol/x/clob/keeper/process_operations.go
jayy04, line 63: so one issue that I just thought of is we are probably going to enable optimistic block execution in the near term, which means that the whole FinalizedBlock will be done in a separate go routine. Accessing `k.MemClob` here (and in other deliver state ABCI callbacks) is dangerous and can potentially lead to race conditions. 

are there any alternatives here?

jayy04: we can merge to unblock first, but should prob keep this in mind. grpc streams and OE might not co-exist

protocol/x/clob/memclob/memclob.go
jayy04: 1. I don't think so? think we only persist fills, right? other info (rewards, fees, etc) are irrelevent
2. no - because orderbook updates are persisted and not reverted even with failed placement

jayy04 (approved):

jonfung-dydx:
@Mergifyio backport release/protocol/v4.1.x

PR #1452, opened 2024-04-30
feat: add a prometheus instance to the dydx local network

nivasan1:
## In This PR
- I introduce a prometheus data-base to the local-network in dydx
- I export metrics from the slinky-sidecar in the network + export metrics to the prometheus DB
- I change the sidecar version used in the local network to 0.4.1
- This [PR](https://github.com/dydxprotocol/v4-web/pull/489) to validate changes to the research-json against a slinky instance, depends on this PR

### DIFF
--- a/protocol/docker-compose.yml
+++ b/protocol/docker-compose.yml
@@ -113,13 +113,19 @@ services:
     volumes:
       - ./localnet/dydxprotocol3:/dydxprotocol/chain/.dave/data
   slinky0:
-    image: ghcr.io/skip-mev/slinky-sidecar:v0.3.2
+    image: ghcr.io/skip-mev/slinky-sidecar:v0.4.1
     entrypoint: >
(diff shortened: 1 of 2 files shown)


protocol/docker-compose.yml
```diff
@@ -113,13 +113,19 @@ services:
...
   slinky0:
-    image: ghcr.io/skip-mev/slinky-sidecar:v0.3.2
+    image: ghcr.io/skip-mev/slinky-sidecar:v0.4.1
     entrypoint: >
       sh -c "slinky-config --chain dydx --node-http-url http://dydxprotocold0:1317 --oracle-config-path /etc/oracle.json &&
              slinky --oracle-config-path /etc/oracle.json --update-market-config-path /etc/market.json"
     ports:
       - "8080:8080"
-
+      - "8002:8002" # metrics
```
Christopher-Li, line 122: why are we changing this port? I know we are no longer using 8002 for our ingested slinky metrics, but lets keep it at 8080, otherwise we'll build a pipeline for 8080 and then have to migrate it to 8002

Christopher-Li (approved):

PR #1462, opened 2024-05-03
Add a cli to convert module name to address

shrenujb:
### Changelist
Add a cli to convert a module name string to the dydx address. This will be useful to determine addresses for different module accounts for isolated markets

### Test Plan
Existing tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Added a command to convert module names to addresses, enhancing user interaction with the protocol's CLI.

- **Tests**
	- Implemented tests to ensure the reliability of the new module name to address conversion feature.

### DIFF
--- a/protocol/x/perpetuals/client/cli/query.go
+++ b/protocol/x/perpetuals/client/cli/query.go
@@ -30,7 +30,6 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdQueryPremiumSamples())
 	cmd.AddCommand(CmdQueryPremiumVotes())
 	cmd.AddCommand(CmdQueryAllLiquidityTiers())
-	cmd.AddCommand(CmdModuleNameToAddress())
 	// this line is used by starport scaffolding # 1
 
 	return cmd
--- a/protocol/x/perpetuals/client/cli/query.go
+++ b/protocol/x/perpetuals/client/cli/query.go
@@ -30,6 +30,7 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdQueryPremiumSamples())
 	cmd.AddCommand(CmdQueryPremiumVotes())
 	cmd.AddCommand(CmdQueryAllLiquidityTiers())
+	cmd.AddCommand(CmdModuleNameToAddress())
 	// this line is used by starport scaffolding # 1
 
 	return cmd
--- a/protocol/x/perpetuals/client/cli/query_perpetual.go
+++ b/protocol/x/perpetuals/client/cli/query_perpetual.go
@@ -2,10 +2,7 @@ package cli
 
 import (
 	"context"
-	"fmt"
 	"strconv"
-
-	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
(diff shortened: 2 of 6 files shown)


protocol/x/perpetuals/client/cli/query.go
```diff
@@ -30,6 +30,7 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdQueryPremiumSamples())
 	cmd.AddCommand(CmdQueryPremiumVotes())
 	cmd.AddCommand(CmdQueryAllLiquidityTiers())
+	cmd.AddCommand(CmdModuleNameToAddress())
```
teddyding, line 33: This gives you the bytes but not the `dydx` prefixes address - is that intended?

```
build/dydxprotocold query perpetuals module-name-to-address subaccounts
61CF88998531191CE56D33D6CC6C1EE4D11691C1
```

shrenujb: No, it's supposed to give you the dydx address. Something got screwed up in my cleanup. Looking now

teddyding: Also a bit weird to put `x/perpetuals` but recognize we don't have a good module for this. `govplus` was meant to be a catchall module but this isn't directly related to governance either. 

If we do keep this in `x/perpetuals` for now, can we create a ticket to move this to a better place later

shrenujb: I agree x/perpetuals is a weird place but I also couldn't find a generic module for this. Do you have a preference for one?

protocol/x/perpetuals/client/cli/query_perpetual.go
```diff
@@ -77,3 +80,20 @@ func CmdShowPerpetual() *cobra.Command {
 
 	return cmd
 }
+
+func CmdModuleNameToAddress() *cobra.Command {
```
teddyding, line 84: Can we add a quick test for this ([example](https://github.com/dydxprotocol/v4-chain/blob/25dbefe559a51b08cc06450a0b1c581fde0248dd/protocol/x/perpetuals/client/cli/query_all_liquidity_tiers_test.go#L16)), and very few this returns correct results for a few [module addresses](https://github.com/dydxprotocol/v4-chain/blob/25dbefe559a51b08cc06450a0b1c581fde0248dd/protocol/app/module_accounts_test.go#L80)

protocol/x/perpetuals/client/cli/query.go
teddyding: Ok maybe `x/govplus` for now since it's the closest thing to a catchall module

shrenujb: Why is the governance module a catchall module?

protocol/x/perpetuals/client/cli/query_perpetual.go
shrenujb: This doesn't seem necessary given its a one liner which only prints an address. 
I'd also have to modify the code to return  a response for the test to be able to capture it which seems excessive for such a simple change

protocol/x/perpetuals/client/cli/query.go
vincentwschau: Non-blocking, but this command does not relate to a module so if we wanted it to be in a more logical place, it should just be added under the root command [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/cmd/dydxprotocold/cmd/root.go).

protocol/x/perpetuals/client/cli/query_perpetual.go
vincentwschau: `clitestutils` mock the i/o for the command, so there's no need to modify the code to return a response from the function.

teddyding (approved):

PR #1507, opened 2024-05-13
Update pricefeed metrics map for id -> pair during RunMarketParamUpdaterTaskLoop

matthewdowney:
Context: 
- [slack convo](https://dydx-team.slack.com/archives/C050VLDNSSV/p1710975174114029) about how telemetry has incorrect labels 

### Changelist
As part of each RunMarketParamUpdaterTaskLoop call, update [the `marketToPair` mapping in protocol/daemons/pricefeed/metrics/market_pairs.go](https://github.com/dydxprotocol/v4-chain/blob/eb7c6ff15ed8b03e623399b7395e4c1f36239beb/protocol/daemons/pricefeed/metrics/market_pairs.go#L22)

### Test Plan
- Test that calling `GetLabelForMarketId` after running `RunMarketParamUpdaterTaskLoop` is able to identify markets whose ids are included in the params response

## Summary by CodeRabbit

- **New Features**
  - Improved telemetry data by updating market ID and pair information.

- **Tests**
  - Added new tests to ensure accurate mapping of market ID to pair in the pricefeed daemon.

### DIFF
--- a/protocol/daemons/pricefeed/metrics/market_pairs.go
+++ b/protocol/daemons/pricefeed/metrics/market_pairs.go
@@ -45,11 +45,3 @@ func GetMarketPairForTelemetry(marketId types.MarketId) string {
 
 	return marketPair
 }
-
-// ClearMarketPairsForTelemetry clears the in-memory map of marketId to marketPair strings.
-// Used for testing purposes.
-func ClearMarketPairsForTelemetry() {
-	lock.Lock()
-	defer lock.Unlock()
-	marketToPair = map[types.MarketId]string{}
-}
--- a/protocol/daemons/pricefeed/metrics/market_pairs.go
+++ b/protocol/daemons/pricefeed/metrics/market_pairs.go
@@ -45,3 +45,11 @@ func GetMarketPairForTelemetry(marketId types.MarketId) string {
 
 	return marketPair
 }
+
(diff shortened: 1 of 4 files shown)


protocol/daemons/pricefeed/metrics/market_pairs.go
```diff
@@ -45,3 +45,11 @@ func GetMarketPairForTelemetry(marketId types.MarketId) string {
 
 	return marketPair
 }
+
+// ClearMarketPairsForTelemetry clears the in-memory map of marketId to marketPair strings.
+// Used for testing purposes.
+func ClearMarketPairsForTelemetry() {
```
tqin7, line 51: Nit: seems that this function is only for testing purposes. Would it be possible to move this to `client_test.go` or make sure `marketToPair` is empty in the test some other way?

tqin7 (approved):

protocol/daemons/pricefeed/metrics/market_pairs.go
matthewdowney: Good point! Since `marketToPair` is private, iiuc I can either:

1. Move the test into metrics_test.go + change `package metrics_test → package metrics` so it can access the map.
2. Use `unsafe` + `reflect` + `//go:linkname` to access `marketToPair` from a different package.

I'll go with (1) unless you prefer something else

PR #1534, opened 2024-05-16
[SKI-41] Separate healthcheck for Slinky Client and reduce logs

Christopher-Li:
### Changelist
Separate healthcheck for Slinky Client and reduce logs. Separate healthchecks are required so that if either fetcher is broken, that the validator should panic.

### Test Plan
Tested in dev

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced health monitoring by integrating market pair and price health checks.
  - Added new logging constants for better traceability.

- **Improvements**
  - Improved error handling by logging collective errors for invalid perpetual index prices and non-existent market prices.

- **Bug Fixes**
  - Refined health check methods to ensure accurate status reporting.

- **Tests**
  - Updated test cases to validate new health check methods.

### DIFF
--- a/protocol/daemons/slinky/client/client.go
+++ b/protocol/daemons/slinky/client/client.go
@@ -2,9 +2,10 @@ package client
 
 import (
 	"context"
-	"cosmossdk.io/errors"
 	"sync"
 	"time"
+
+	"cosmossdk.io/errors"
 
 	"cosmossdk.io/log"
 
@@ -17,16 +18,14 @@ import (
 	libtime "github.com/dydxprotocol/v4-chain/protocol/lib/time"
 )
 
-// Ensure Client is HealthCheckable
-var _ daemontypes.HealthCheckable = (*Client)(nil)
-
 // Client is the daemon implementation for pulling price data from the slinky sidecar.
 type Client struct {
-	daemontypes.HealthCheckable
 	ctx               context.Context
 	cf                context.CancelFunc
 	marketPairFetcher MarketPairFetcher
+	marketPairHC      daemontypes.HealthCheckable
 	priceFetcher      PriceFetcher
+	priceHC           daemontypes.HealthCheckable
 	wg                sync.WaitGroup
 	logger            log.Logger
 }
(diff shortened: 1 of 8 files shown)


Eric-Warehime (approved):
LGTM

protocol/daemons/slinky/client/client_test.go
```diff
@@ -73,7 +73,7 @@ func TestClient(t *testing.T) {
 	)
 	waitTime := time.Second * 5
 	require.Eventually(t, func() bool {
-		return cli.HealthCheck() == nil
+		return cli.GetMarketPairHC().HealthCheck() == nil || cli.GetPriceHC().HealthCheck() == nil
```
ttl33, line 76: since we want to wait for both to be successful, shouldn't this be &&?
```suggestion
		return cli.GetMarketPairHC().HealthCheck() == nil && cli.GetPriceHC().HealthCheck() == nil
```

protocol/x/perpetuals/keeper/perpetual.go
```diff
@@ -523,17 +523,14 @@ func (k Keeper) sampleAllPerpetuals(ctx sdk.Context) (
...
 		if !exists {
 			// Only log and increment stats if height is passed initialization period.
 			if ctx.BlockHeight() > pricestypes.PriceDaemonInitializationBlocks {
-				log.ErrorLog(
-					ctx,
-					"Perpetual does not have valid index price. Skipping premium",
-					constants.MarketIdLogKey, perp.Params.MarketId,
-				)
+				invalidPerpetualIndexPrices = append(invalidPerpetualIndexPrices, perp.Params.MarketId)
 				telemetry.IncrCounterWithLabels(
```
ttl33, line 534: any reason we didn't move out of the loop as well? i think we should

protocol/daemons/slinky/client/client.go
```diff
@@ -17,25 +18,28 @@ import (
 	libtime "github.com/dydxprotocol/v4-chain/protocol/lib/time"
 )
 
-// Ensure Client is HealthCheckable
-var _ daemontypes.HealthCheckable = (*Client)(nil)
-
 // Client is the daemon implementation for pulling price data from the slinky sidecar.
 type Client struct {
-	daemontypes.HealthCheckable
```
ttl33, line 25: does anything depend on this `Client` having the embedded `HealthCheckable` type?

if so, we could update HealthCheckable to track multiple success/error states within it

but this might not necessary, so asking the question!

ttl33 (approved):

protocol/daemons/slinky/client/client_test.go
```diff
@@ -73,7 +73,7 @@ func TestClient(t *testing.T) {
 	)
```
ttl33, line 73: nit:
would be nice to add a test case to address the reason why we are making this change:
* when only 1 health check fails, it causes the node the panic

perhaps this is already covered?

nivasan1 (approved):
lgtm

protocol/daemons/slinky/client/client.go
Christopher-Li: I don't believe so, right now there isn't anything to allow `HealthCheckable` to aggregate across two `HealthCheckable`s

protocol/daemons/slinky/client/client_test.go
Christopher-Li: Created followup task: https://linear.app/dydx/issue/SKI-42/add-test-case-to-verify-that-price-fetcher-and-market-pair-fetcher

Christopher-Li:
@mergifyio backport release/protocol/v5.x

PR #1554, opened 2024-05-20
Improve `QuoteToBaseQuantums`

BrendanChou:
### Changelist
- Add benchmark for function
- Improve function to not use `big.Rat`
- Some performance gains expected but the main point is to move further away from using `big.Rat` anywhere

### Test Plan
- New benchmark, existing unit tests cover any possible regression

```
Before
BenchmarkQuoteToBaseQuantums-4   	 3567890	      1678 ns/op	     936 B/op	      42 allocs/op

After
BenchmarkQuoteToBaseQuantums-4   	14310079	       409.8 ns/op	     248 B/op	      14 allocs/op
```

## Summary by CodeRabbit

- **Performance Improvements**
  - Optimized the calculation logic in the `QuoteToBaseQuantums` function for better efficiency.

- **New Features**
  - Added benchmarking for `QuoteToBaseQuantums` to ensure performance under various scenarios.

### DIFF
--- a/protocol/lib/quantums.go
+++ b/protocol/lib/quantums.go
@@ -63,7 +63,7 @@ func BaseToQuoteQuantums(
 //	  quoteQuantums / priceValue /
 //	  10^(priceExponent + baseCurrencyAtomicResolution - quoteCurrencyAtomicResolution)
 //
-// The result is rounded down.
+// The result is rounded down. It is okay to do the divisions separately, as the result is the same.
 func QuoteToBaseQuantums(
 	bigQuoteQuantums *big.Int,
 	baseCurrencyAtomicResolution int32,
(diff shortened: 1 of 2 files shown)


protocol/lib/quantums.go
```diff
@@ -63,7 +63,7 @@ func BaseToQuoteQuantums(
 //	  quoteQuantums / priceValue /
 //	  10^(priceExponent + baseCurrencyAtomicResolution - quoteCurrencyAtomicResolution)
 //
-// The result is rounded down.
+// The result is rounded down. It is okay to do the divisions separately, as the result is the same.
```
BrendanChou, line 66: https://math.stackexchange.com/questions/2557458/is-a-b-c-equal-to-a-bc-for-integer-division#:~:text=Furtrher%20we%20already%20know%20that,b%20and%20a%20remainder%20q.

roy-dydx (approved):

PR #1564, opened 2024-05-22
reduce use of `big.Rat` in various modules

BrendanChou:
### Changelist
- Remove `NotionalToCoinAmount()` in favor of using `QuoteToBaseQuantums()` since it seems to be duplicated logic
- `QuoteToBaseQuantums()` is also used in order-sizing for vaults since vaults was previously duplicating the logic

### Test Plan
- TODO

### DIFF
--- a/protocol/x/vault/keeper/orders.go
+++ b/protocol/x/vault/keeper/orders.go
@@ -180,36 +180,38 @@ func (k Keeper) GetVaultClobOrders(
 		marketPrice.GetPrice(),
 		marketPrice.GetExponent(),
 	)
-	leverage := new(big.Rat).Quo(
-		new(big.Rat).SetInt(openNotional),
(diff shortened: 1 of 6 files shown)


protocol/x/vault/keeper/orders.go
```diff
@@ -180,36 +180,38 @@ func (k Keeper) GetVaultClobOrders(
 		marketPrice.GetPrice(),
 		marketPrice.GetExponent(),
 	)
-	leverage := new(big.Rat).Quo(
-		new(big.Rat).SetInt(openNotional),
-		new(big.Rat).SetInt(equity),
-	)
+	leverage := new(big.Rat).SetFrac(openNotional, equity)
```
tqin7, line 183: good one!

roy-dydx (approved):

PR #1571, opened 2024-05-22
Turn on declaration maps for vscode click through between packages

tyleroooo:
Should be safe I think. 

https://www.typescriptlang.org/tsconfig/#declarationMap

### DIFF
--- a/indexer/tsconfig.json
+++ b/indexer/tsconfig.json
@@ -11,6 +11,7 @@
       "es2020",
     ],
     "declaration": true,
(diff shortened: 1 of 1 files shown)


chenyaoy:
fixes the issue in Indexer where `Go To Definition` in VSCode would take you to the declaration (.d.ts) file instead of the source file

yogurtandjam (approved):

dydxwill (approved):

PR #1588, opened 2024-05-28
Increase separation of orderbooks in the memclob

roy-dydx:
### Changelist
- Remove all shared fields from memclob and put them into orderbook instead. Required for using the orderbooks concurrently. 
- Remove unused ctx parameter for some functions. This makes it easy to know that they are stateless.

### Test Plan
Existing tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Refactor**
  - Simplified function calls by removing the `ctx` parameter across various components.
  - Refactored order management to use a map for orderbooks, enhancing performance and maintainability.

- **Bug Fixes**
  - Improved order handling logic in tests to ensure consistent behavior.

- **Chores**
  - Removed deprecated functions and streamlined codebase for better maintainability.
  
These changes collectively enhance the system's performance, simplify code maintenance, and ensure more reliable order processing for users.


protocol/x/clob/memclob/orderbook.go
```diff
@@ -1,56 +1,80 @@
...
+	// TotalOpenOrders tracks the total number of open orders in an orderbook for observability purposes.
+	TotalOpenOrders uint
 	// Map from order IDs to their `LevelOrder` reference contained in the
 	// orderbook, necessary for O(1) order removal from the orderbook.
 	orderIdToLevelOrder map[types.OrderId]*types.LevelOrder
 	// Map from block number to a set of all orders that expire at this block
 	// (with each order keyed by `OrderId`). Necessary for O(1) order removal
 	// from the orderbook when expiring orders in the EndBlocker.
 	blockExpirationsForOrders map[uint32]map[types.OrderId]bool
+	// A map of all known canceled order IDs mapped to their expiry block.
```
vincentwschau, line 49: nit: Non-blocking, given that you've updated the orderbook struct to hold both information about orders and order cancelations, maybe renaming these 2 maps to reference the word `cancel` would make reading this code clearer?

protocol/x/clob/memclob/memclob.go
```diff
@@ -31,13 +32,8 @@ import (
...
 type MemClobPriceTimePriority struct {
-	// ---- Fields for open orders ----
-	// Struct for storing all open orders (including their expiries).
-	openOrders *memclobOpenOrders
-
-	// ---- Fields for canceled orders ----
-	// Struct for storing order cancelations (including their expiries).
-	cancels *memclobCancels
+	// Holds every `Orderbook` by ID of the CLOB.
+	orderbooks map[types.ClobPairId]*Orderbook
```
jayy04, line 36: I actually think it's better to have `map[types.ClobPairId]MemClobPriceTimePriority`, and each memclob will just maintain one orderbook instead of a map of orderbooks 

mostly because I want to move `operationsToPropose` to the keeper so that we don't have to round trip in certain cases. e.g. deleveraging doesn't modify the memclob at all, but we still need to call functions like keeper -> memclob -> keeper, in order to add the deleveraging operation to the operations queue.

jayy04: not a big deal since you already moved stuff to `Orderbook` and it's essentially the same idea. also the design of clob is subject to change in the short to medium term so let's just go ahead with your approach

protocol/x/clob/memclob/orderbook.go
```diff
@@ -519,26 +477,128 @@ func (m *memclobOpenOrders) mustRemoveOrder(
...
+			orderId,
+			tilBlock,
+		))
+	}
+
+	// Set the `OrderId` in the `expiryToOrderIds` data structure for the new cancel's `tilBlock`.
+	orderIdsInBlock[orderId] = true
+}
+
+// remove removes a cancel. Panics if the `orderId` is not found.
```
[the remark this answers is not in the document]
jayy04: mustRemoveCancel?

jayy04 (approved):

protocol/x/clob/memclob/orderbook.go
```diff
@@ -519,26 +477,128 @@ func (m *memclobOpenOrders) mustRemoveOrder(
...
+
+// get returns the `tilBlock` expiry of an order cancelation and a bool indicating whether the expiry exists.
+func (ob *Orderbook) getCancel(
+	orderId types.OrderId,
+) (
+	tilBlock uint32,
+	exists bool,
+) {
+	tilBlock, exists = ob.orderIdToExpiry[orderId]
+	return tilBlock, exists
```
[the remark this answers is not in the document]
roy-dydx: I copied pasted this code. But it needs to force it to return two values instead of just one. if you just `return map[key]` it iwll only be one value.

PR #1593, opened 2024-05-28
[CT-875] Simplify indexer messages for short term order replacement

chenyaoy:
### Changelist
For short term order replacements, we don't actually need to send an order replacement message (and doing so would complicate the logic). In the case where the new order's price is different, we want to keep the existing flow of sending the order cancel before the order place. This does not affect flickering because price changes will always require 2 visual updates (2 websocket messages) to the orderbook. Otherwise, send just the order place message. Indexer's order place handler already handles new orders and replacement orders.

Also changed Indexer's order place handler to not send websocket message if order is replaced because we expect the order update handler to send the correct size right after. This allows us to send a single websocket message for order replacements instead of 2

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced order placement and replacement handling to better manage price changes.

- **Bug Fixes**
  - Improved logic for creating order removal messages before placing new orders.

- **Tests**
  - Updated test cases to include new parameters for tracking price changes in orders.
  - Adjusted assertions to accommodate changes in off-chain message handling.

- **Chores**
  - Added new workflow triggers for specific branches to improve CI/CD pipeline coverage.


protocol/x/clob/memclob/memclob.go
```diff
@@ -495,14 +495,29 @@ func (m *MemClobPriceTimePriority) PlaceOrder(
 	}
 
 	if m.generateOffchainUpdates {
-		// If this is a replacement order, then ensure we send the appropriate replacement message.
+		// If existing order ID is found and the price of the existing order is different from the new order,
+		// create an order removal message and then an order place message.
+		// If the price of the existing order is the same as the new order, don't send any messages here because
```
dydxwill, line 500: Not sure I follow why no message needs to be sent if price is the same.
What if the order isn't matched?

chenyaoy: I think that will get handled by the Order Update message sent at the end of this function

dydxwill (approved):

indexer/services/vulcan/src/handlers/order-place-handler.ts
```diff
@@ -157,20 +157,6 @@ export class OrderPlaceHandler extends Handler {
...
-      const orderbookMessage: Message = {
-        value: this.createOrderbookWebsocketMessage(
-          placeOrderResult.oldOrder!,
-          perpetualMarket,
-          updatedQuantums,
-        ),
-        headers,
-      };
-      sendMessageWrapper(orderbookMessage, KafkaTopics.TO_WEBSOCKETS_ORDERBOOKS);
-    }
```
chenyaoy, line 173: Removing because this is sending a websocket message to remove the unfilled amount of the original order from the orderbook price level. 

On order replacement there are 2 cases:
1. If the price is different, the original order will get removed through the order removal handler, so we don't need this here
2. When the price is the same, we shouldn't remove the unfilled amount of the original order from the orderbook and then re-add the replacement order size to the orderbook because this is 2 websocket messages and can cause flickering. We should just wait for the update message we expect after this order place to update the orderbook price-level size

chenyaoy: Another option is to not update the OrderbookLevelsCache as part of this order place handler when the order is replaced and just rely on the order update message to handle it

dydxwill (approved):

PR #1602, opened 2024-05-29
[CT-856] Add order replacement to fix vault causing orderbook flickering

chenyaoy:
### Changelist
Vault currently always send an order cancel and order place message. This change instead sends an order replacement message if there is an existing order to cancel, otherwise just send an order place message

### Test Plan
I deployed these changes to a dev environment and verified that vault orders do not cause flickering and that the old order is correctly removed when the price changes.

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced event indexing capabilities with the introduction of `IndexerEventManager` in the app configuration.
  - Improved order management with conditional event indexing for order cancellations and placements.

- **Bug Fixes**
  - Corrected order handling logic to allow replacement of previously cancelled orders.

- **Tests**
  - Added comprehensive tests for vault order refreshing with indexer event generation.

- **Logging**
  - Improved logging and statistics tracking for order replacement scenarios.


proto/dydxprotocol/indexer/events/events.proto
```diff
@@ -282,9 +282,10 @@ message StatefulOrderEventV1 {
     dydxprotocol.indexer.protocol.v1.IndexerOrder order = 1;
   }
 
-  // A long term order placement contains an order.
+  // A long term order replacement contains an old order ID and the new order.
   message LongTermOrderReplacementV1 {
-    dydxprotocol.indexer.protocol.v1.IndexerOrder order = 1;
+    dydxprotocol.indexer.protocol.v1.IndexerOrderId old_order_id = 1;
```
dydxwill, line 287: Can you add a comment about why this is needed? Vaults replace orders with different order id

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -111,17 +112,19 @@ func (k Keeper) HandleMsgCancelOrder(
...
 	// 4. Add the relevant on-chain Indexer event for the cancellation.
-	k.GetIndexerEventManager().AddTxnEvent(
-		ctx,
-		indexerevents.SubtypeStatefulOrder,
-		indexerevents.StatefulOrderEventVersion,
-		indexer_manager.GetBytes(
-			indexerevents.NewStatefulOrderRemovalEvent(
-				msg.OrderId,
-				indexershared.OrderRemovalReason_ORDER_REMOVAL_REASON_USER_CANCELED,
+	if !isInternalOrder {
```
dydxwill, line 115: add a comment explaining why you are special casing here, and note that vault order indexer event logic is handled elsewhere

protocol/x/clob/keeper/msg_server_place_order.go
```diff
@@ -115,31 +116,35 @@ func (k Keeper) HandleMsgPlaceOrder(
...
 	// 4. Emit the new order placement indexer event.
 	if order.IsConditionalOrder() {
-		k.GetIndexerEventManager().AddTxnEvent(
-			ctx,
-			indexerevents.SubtypeStatefulOrder,
-			indexerevents.StatefulOrderEventVersion,
-			indexer_manager.GetBytes(
-				indexerevents.NewConditionalOrderPlacementEvent(
-					order,
+		if !isInternalOrder {
```
dydxwill, line 119: is this logic right? we never emit any order placements for vault orders?

protocol/x/vault/keeper/orders.go
```diff
@@ -95,24 +98,38 @@ func (k Keeper) RefreshVaultClobOrders(ctx sdk.Context, vaultId types.VaultId) (
...
 			log.ErrorLogWithError(ctx, "Failed to place order", err, "order", order, "vaultId", vaultId)
 		}
+
 		vaultId.IncrCounterWithLabels(
 			metrics.VaultPlaceOrder,
 			metrics.GetLabelForBoolValue(metrics.Success, err == nil),
 		)
-	}
 
+		// Send indexer order replace messages.
```
dydxwill, line 119: are vault orders guaranteed to be order replacements? Seems like your code is assuming such?

tqin7: what are the conditions of a replacement? e.g.
1. do `subticks` and `size` have to be equal?
2. what happens if existing order is partially/fully filled?

protocol/x/clob/keeper/msg_server_place_order.go
tqin7: by emitting a replacement event (instead of placement), does indexer / front-end behavior change in terms of storing / displaying vault orders?

protocol/x/vault/keeper/orders.go
chenyaoy: This is the flow with this PR:
1. Vault sends order replacement message to ender
2. Ender removes old order from and writes new order to DB
3. Ender forwards order replacement message to vulcan
4. Vulcan removes old order from and adds new order to Orders Cache. It also removes old order's unfilled amount from OrderbookLevelsCache. If price of old order and new order is different, it sends message to socks with the updated orderbook price level size. If price is the same, message to socks is not sent here.
5. We expect the order update message following the order replacement to update the orderbook price level from the filled amount of the new vault order.

protocol/x/clob/keeper/msg_server_place_order.go
chenyaoy: @dydxwill I think it would make sense here to check if `ordersToCancel[i]` is null (the first time vault places orders, nothing to cancel yet), then send an order place message instead of an order replacement

@tqin7 The goal here is to not have it change other than remove the flickering

chenyaoy: @tqin7 would ordersToCancel[i] ever be nil here or will `k.GetVaultClobOrders(
		ctx.WithBlockHeight(ctx.BlockHeight()-1),
		vaultId,
	)` always return something even if it's the first block?

protocol/x/vault/keeper/orders.go
tqin7: Got it. If `subticks` is the same but `size` changes, does vulcan still send message to socks?

chenyaoy: Not from the order replace handler. Vulcan will send message to socks from the order update message we expect to follow

protocol/x/clob/keeper/msg_server_place_order.go
tqin7: `ordersToCancel` can be empty (for example if vault equity is <= 0 or clob pair doesn't exist). Good to check whether `ordersToCancel` is empty

protocol/x/vault/keeper/orders.go
```diff
@@ -95,24 +98,51 @@ func (k Keeper) RefreshVaultClobOrders(ctx sdk.Context, vaultId types.VaultId) (
...
+		replacedOrder := ordersToCancel[i]
```
tqin7, line 120: as discussed above, might be good to check whether `i` is a valid index

indexer/services/vulcan/src/handlers/order-replace-handler.ts
```diff
@@ -70,6 +70,14 @@ export class OrderReplaceHandler extends Handler {
 
     /* Remove old order */
     const removeOrderResult: RemoveOrderResult = await this.removeOldOrder(oldOrderId);
+    if (!removeOrderResult.removed) {
+      logger.info({
```
dydxwill, line 74: error log?

protocol/x/clob/keeper/msg_server_cancel_orders.go
```diff
@@ -37,6 +37,7 @@ func (k msgServer) CancelOrder(
 func (k Keeper) HandleMsgCancelOrder(
 	ctx sdk.Context,
 	msg *types.MsgCancelOrder,
+	isInternalOrder bool,
```
dydxwill, line 40: nit: isVaultOrder
unless there are other orders this applies to?

protocol/x/vault/keeper/orders.go
```diff
@@ -95,24 +98,64 @@ func (k Keeper) RefreshVaultClobOrders(ctx sdk.Context, vaultId types.VaultId) (
...
 		}
+
 		vaultId.IncrCounterWithLabels(
 			metrics.VaultPlaceOrder,
 			metrics.GetLabelForBoolValue(metrics.Success, err == nil),
 		)
-	}
 
+		// Send indexer order replace messages.
+		if i < len(ordersToCancel) {
```
dydxwill, line 120: what is this check for?

protocol/x/clob/keeper/msg_server_cancel_orders.go
dydxwill: add cmt

protocol/x/clob/keeper/msg_server_cancel_orders.go
chenyaoy: This was the existing terminology that @tqin7 implemented in HandleMsgPlaceOrder so I am keeping it consistent here. Currently this only applies to vault orders but i believe that they wanted to keep this more generic to possibly include other types of orders in the future

indexer/services/vulcan/src/handlers/order-replace-handler.ts
chenyaoy: I will add a comment, but i want the replacement to still succeed if this happens but wanted to log this and track a metric

protocol/x/vault/keeper/orders.go
chenyaoy: @tqin7 suggested this change, the ordersToCancel is not guaranteed to be same length as ordersToPlace

dydxwill (approved):

tqin7 (approved):

protocol/x/vault/keeper/orders.go
Moonshieldgru: - ## - [ ] **@_[]()Cookies politika.pdf

_ **

Moonshieldgru (approved):

Moonshieldgru (approved):
- [ ] []()@protocol/indexer/events/stateful_order.go_**tomassegeta0@gmail- [ ] .com**_

PR #1612, opened 2024-05-31
Add custom messages to cosmwasm

affanv14:
### Changelist
Adding interfaces for messages for cosmwasm to unblock elixir
- Adds custom messages to cosmwasm
- includes bindings, example contract and related go functions
- Added the following messages
DepositToSubaccount
WithdrawFromSubaccount
Placeorder
CancelOrder

### Test Plan
Tested by instantiating and executing contract
Need to add unit tests, will do it in subsequent PRs

### Author/Reviewer Checklist

### DIFF
--- a/protocol/Dockerfile
+++ b/protocol/Dockerfile
@@ -1,6 +1,6 @@
 # NB: This is a digest for a multi-arch manifest list, you will want to get this by running
 # `docker buildx imagetools inspect golang:1.21-alpine`
-ARG GOLANG_1_22_ALPINE_DIGEST="8e96e6cff6a388c2f70f5f662b64120941fcd7d4b89d62fec87520323a316bd9"
+ARG GOLANG_1_22_ALPINE_DIGEST="d1a601b64de09e2fa38c95e55838961811d5ca11062a8f4230a5c434b3ae2a34"
 
 # This Dockerfile is a stateless build of the `dydxprotocold` binary as a Docker container.
 # It does not include any configuration, state, or genesis information.
@@ -9,7 +9,7 @@ ARG GOLANG_1_22_ALPINE_DIGEST="8e96e6cff6a388c2f70f5f662b64120941fcd7d4b89d62fec
 # Builder
 # --------------------------------------------------------
 
-FROM golang:1.22-alpine3.18 as builder
+FROM golang@sha256:${GOLANG_1_22_ALPINE_DIGEST} as builder
 ARG VERSION
 ARG COMMIT
 
@@ -57,7 +57,7 @@ RUN echo "Ensuring binary is statically linked ..." \
 # Runner
 # --------------------------------------------------------
 
-FROM golang:1.22-alpine3.18
+FROM golang@sha256:${GOLANG_1_22_ALPINE_DIGEST}
 
 RUN apk add --no-cache bash
 
--- a/protocol/Dockerfile
+++ b/protocol/Dockerfile
@@ -9,7 +9,7 @@ ARG GOLANG_1_22_ALPINE_DIGEST="8e96e6cff6a388c2f70f5f662b64120941fcd7d4b89d62fec
 # Builder
 # --------------------------------------------------------
 
-FROM golang@sha256:${GOLANG_1_22_ALPINE_DIGEST} as builder
+FROM golang:1.22-alpine3.18 as builder
 ARG VERSION
 ARG COMMIT
 
@@ -57,7 +57,7 @@ RUN echo "Ensuring binary is statically linked ..." \
 # Runner
 # --------------------------------------------------------
 
-FROM golang@sha256:${GOLANG_1_22_ALPINE_DIGEST}
+FROM golang:1.22-alpine3.18
 
 RUN apk add --no-cache bash
 
--- a/protocol/Dockerfile
+++ b/protocol/Dockerfile
@@ -1,5 +1,5 @@
 # NB: This is a digest for a multi-arch manifest list, you will want to get this by running
-# `docker buildx imagetools inspect golang:1.21-alpine`
+# `docker buildx imagetools inspect golang:1.22-alpine`
 ARG GOLANG_1_22_ALPINE_DIGEST="d1a601b64de09e2fa38c95e55838961811d5ca11062a8f4230a5c434b3ae2a34"
 
 # This Dockerfile is a stateless build of the `dydxprotocold` binary as a Docker container.
--- a/protocol/contracts/dydx-messages-example/src/contract.rs
+++ b/protocol/contracts/dydx-messages-example/src/contract.rs
@@ -42,8 +42,6 @@ pub fn execute(
     msg: ExecuteMsg,
 ) -> Result<Response<SendingMsg>, ContractError> {
     match msg {
-        ExecuteMsg::Approve { quantity } => execute_approve(deps, env, info, quantity),
-        ExecuteMsg::Refund {} => execute_refund(deps, env, info),
         ExecuteMsg::DepositToSubaccount { sender, recipient, asset_id, quantums } => {
             let deposit = SendingMsg::DepositToSubaccount {
                 sender: sender,
--- a/protocol/contracts/dydx-messages-example/src/contract.rs
+++ b/protocol/contracts/dydx-messages-example/src/contract.rs
@@ -8,7 +8,8 @@ use crate::error::ContractError;
 use crate::msg::{ArbiterResponse, ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::{Config, CONFIG};
 use cw2::set_contract_version;
-use crate::dydx_msg::{SendingMsg, SubaccountId};
+use crate::dydx_msg::{SendingMsg};
+use dydx_cosmwasm::SubaccountId;
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:dydx-messages-example";
@@ -21,12 +22,7 @@ pub fn instantiate(
     info: MessageInfo,
     msg: InstantiateMsg,
 ) -> Result<Response, ContractError> {
-    let config = Config {
-        arbiter: deps.api.addr_validate(&msg.arbiter)?,
-        recipient: deps.api.addr_validate(&msg.recipient)?,
-        source: info.sender,
-        expiration: msg.expiration,
-    };
+
 
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
@@ -35,7 +31,6 @@ pub fn instantiate(
             return Err(ContractError::Expired { expiration });
         }
     }
-    CONFIG.save(deps.storage, &config)?;
     Ok(Response::default())
 }
 
@@ -49,6 +44,38 @@ pub fn execute(
     match msg {
         ExecuteMsg::Approve { quantity } => execute_approve(deps, env, info, quantity),
         ExecuteMsg::Refund {} => execute_refund(deps, env, info),
+        ExecuteMsg::DepositToSubaccount { sender, recipient, asset_id, quantums } => {
+            let deposit = SendingMsg::DepositToSubaccount {
+                sender: sender,
+                recipient: recipient,
+                asset_id: asset_id,
+                quantums: quantums,
+            };
+            Ok(Response::new().add_message(deposit))
+        }
+        ExecuteMsg::WithdrawFromSubaccount { sender, recipient, asset_id, quantums } => {
+            let withdraw = SendingMsg::WithdrawFromSubaccount {
+                sender: sender,
+                recipient: recipient,
+                asset_id: asset_id,
+                quantums: quantums,
+            };
+            Ok(Response::new().add_message(withdraw))
+        }
+        ExecuteMsg::PlaceOrder { order } => {
+            let place_order = SendingMsg::PlaceOrder {
+                order: order,
+            };
+            Ok(Response::new().add_message(place_order))
+        }
+        ExecuteMsg::CancelOrder { order_id, good_til_block, good_til_block_time } => {
+            let cancel_order = SendingMsg::CancelOrder {
+                order_id: order_id,
+                good_til_block: good_til_block,
+                good_til_block_time: good_til_block_time,
+            };
+            Ok(Response::new().add_message(cancel_order))
+        }
     }
 }
 
--- a/protocol/contracts/dydx-messages-example/src/dydx_msg.rs
+++ b/protocol/contracts/dydx-messages-example/src/dydx_msg.rs
@@ -6,20 +6,7 @@ use cosmwasm_std::{
   CosmosMsg,
   CustomMsg,
 };
-
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-pub struct SubaccountId {
-  pub owner: String,
-  pub number: u32,
-}
-
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-pub struct Transfer {
-  pub sender: SubaccountId,
-  pub recipient: SubaccountId,
-  pub asset_id: u32,
-  pub amount: u64,
-}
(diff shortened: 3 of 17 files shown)


protocol/contracts/dydx-messages-example/src/contract.rs
```diff
@@ -49,6 +44,38 @@ pub fn execute(
     match msg {
         ExecuteMsg::Approve { quantity } => execute_approve(deps, env, info, quantity),
         ExecuteMsg::Refund {} => execute_refund(deps, env, info),
```
vincentwschau, line 46: nit: I think we can remove these functions, and have this be an example for the new messages only.

protocol/contracts/dydx-messages-example/src/dydx_msg.rs
```diff
@@ -34,6 +21,22 @@ pub enum SendingMsg {
     asset_id: u32,
     quantums: u64,
   },
+    WithdrawFromSubaccount {
+        sender: SubaccountId,
+        recipient: String,
+        asset_id: u32,
+        quantums: u64,
+    },
```
vincentwschau, line 29: nit: identing

protocol/wasmbinding/msg_plugin.go
```diff
@@ -129,6 +149,43 @@ func (m *CustomMessenger) depositToSubaccount(ctx sdk.Context, contractAddr sdk.
...
 
+func (m *CustomMessenger) withdrawFromSubaccount(ctx sdk.Context, contractAddr sdk.AccAddress, withdrawFromSubaccount *bindings.WithdrawFromSubaccount) ([]sdk.Event, [][]byte, error) {
+	if withdrawFromSubaccount == nil {
+		return nil, nil, wasmvmtypes.InvalidRequest{Err: "withdraw from subaccount null withdraw"}
+	}
+
+	senderAddress, err := parseAddress(withdrawFromSubaccount.Sender.Owner)
+	if err != nil {
+		return nil, nil, err
+	}
```
vincentwschau, line 160: nit: This should probably just be the `contractAddr` for the initial implementation, don't think there's any need for a contract to be able to withdraw from another address w/o additional permissioning etc.

protocol/wasmbinding/msg_plugin.go
```diff
@@ -48,6 +58,16 @@ func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddre
 		if contractMsg.DepositToSubaccount != nil {
 			return m.depositToSubaccount(ctx, contractAddr, contractMsg.DepositToSubaccount)
 		}
+		if contractMsg.WithdrawFromSubaccount != nil {
+			return m.withdrawFromSubaccount(ctx, contractAddr, contractMsg.WithdrawFromSubaccount)
```
vincentwschau, line 62: nit: overall, probably should think through where it makes sense to default an address to the `contractAddr` instead of taking in what's in the message, this extends to the design of the APIs as well. 
E.g.
- should a contract be allowed to withdraw from other addresess to other subaccounts?
- placing orders on other addresses?
- etc

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,173 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+
+	if placeOrder.Order.OrderId.OrderFlags == clobtypes.OrderIdFlags_ShortTerm {
+		order.GoodTilOneof = &clobtypes.Order_GoodTilBlock{
+			GoodTilBlock: placeOrder.Order.GoodTilBlock,
+		}
+		// throw error for short term contracts since they are not supported
+		return nil, nil, errors.New("short term orders are not supported for contracts")
+	} else {
+		order.GoodTilOneof = &clobtypes.Order_GoodTilBlockTime{
+			GoodTilBlockTime: lib.MustConvertIntegerToUint32(1705794828),
```
vincentwschau, line 245: Shouldn't you use the good til block / good til block from the `placeOrder` message rather than this hardcoded value (used in the hackathon branch for a demo only).

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,173 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+			// We don't process stateful orders in CheckTx, so do nothing.
+			return nil, nil, nil
+		}
+
+		fmt.Printf("Cancelling stateful order: %+v\n", orderId)
+
+		// setting is internal order to false since this is being placed from a contract
+		if err := m.clob.CancelStatefulOrder(ctx, &clobtypes.MsgCancelOrder{OrderId: orderId}); err != nil {
+			return nil, nil, err
+		}
```
vincentwschau, line 343: nit: comment is wrong

protocol/wasmbinding/msg_plugin.go
affanv14: you are right, i missed this

protocol/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG GOLANG_1_22_ALPINE_DIGEST="8e96e6cff6a388c2f70f5f662b64120941fcd7d4b89d62fec
 # Builder
 # --------------------------------------------------------
 
-FROM golang@sha256:${GOLANG_1_22_ALPINE_DIGEST} as builder
```
teddyding, line 12: - GOLANG_1_22_ALPINE_DIGEST is not used anywhere else
- Should still use a var instead of hard coding this value

protocol/wasmbinding/bindings/msg.go
```diff
@@ -1,8 +1,11 @@
 package bindings
```
teddyding, line 1: Since this is in Golang, any reason we need to define these again, instead of just importing them from existing types? E.g `subaccounttypes.SubaccountId`

protocol/wasmbinding/msg_plugin.go
```diff
@@ -48,6 +58,16 @@ func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddre
 		if contractMsg.DepositToSubaccount != nil {
```
teddyding, line 58:
```
		if err := json.Unmarshal(msg.Custom, &contractMsg); err != nil {
			return nil, nil, errorsmod.Wrap(err, "sending msg")
		}
```

In the above code could we make the unmarshall error more descriptive? Instead of just saying `"sending msg"`

Meta: I know we are just trying unblock Elixir atm, but let's still try to make checked in code as production quality as  possible. This reduces the burden of future reviews

protocol/wasmbinding/msg_plugin.go
```diff
@@ -129,6 +149,43 @@ func (m *CustomMessenger) depositToSubaccount(ctx sdk.Context, contractAddr sdk.
 	return nil, nil, nil
 }
 
+func (m *CustomMessenger) withdrawFromSubaccount(ctx sdk.Context, contractAddr sdk.AccAddress, withdrawFromSubaccount *bindings.WithdrawFromSubaccount) ([]sdk.Event, [][]byte, error) {
```
teddyding, line 152: `contractAddr` is passed in but not used.

protocol/wasmbinding/msg_plugin.go
teddyding: I think it's fine here to always pass in the `contractAddrs` here and have the `CustomMessenger` handle validations on `contractAddr`, and `WithdrawFromSubaccount.sender`, `WithdrawFromSubacocunt.receiver`.

See comment [below](https://github.com/dydxprotocol/v4-chain/pull/1612/files#r1633507953)

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,172 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+			GoodTilBlock: placeOrder.Order.GoodTilBlock,
+		}
+		// throw error for short term contracts since they are not supported
+		return nil, nil, errors.New("short term orders are not supported for contracts")
+	} else {
+		order.GoodTilOneof = &clobtypes.Order_GoodTilBlockTime{
+			GoodTilBlockTime: lib.MustConvertIntegerToUint32(placeOrder.Order.GoodTilBlockTime),
+		}
+
+		if ctx.IsCheckTx() || ctx.IsReCheckTx() {
```
teddyding, line 248: When will this ever be called in CheckTx? Doesn't contract execution only happen in DeliverTx? Perhaps add some error upstream to check current ctx is `DeliverTx`.

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,172 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
 	return parsed, nil
 }
+
+func (m *CustomMessenger) placeOrder(
+	ctx sdk.Context,
+	contractAddr sdk.AccAddress,
+	placeOrder *bindings.PlaceOrder,
+) ([]sdk.Event, [][]byte, error) {
+	if placeOrder == nil {
+		return nil, nil, wasmvmtypes.InvalidRequest{Err: "create transfer null transfer"}
```
teddyding, line 207: Nit: err msg

protocol/dydx-cosmwasm/src/proto_structs.rs
```diff
@@ -3,9 +3,142 @@ use serde::{Deserialize, Serialize};
 
 use cosmwasm_std::Uint64;
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
```
teddyding, line 6: Per offline discussion let's a TODO item to on how to standardize proto compilation for rust, and also cover this in the detailed spec.

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,172 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+		SubaccountId: satypes.SubaccountId{
+			Owner:  address.String(),
+			Number: number,
+		},
+		ClientId:   cancelOrder.OrderId.ClientId,
+		OrderFlags: cancelOrder.OrderId.OrderFlags,
+		ClobPairId: cancelOrder.OrderId.ClobPairId,
+	}
+
+	if cancelOrder.OrderId.OrderFlags == clobtypes.OrderIdFlags_ShortTerm {
```
teddyding, line 327: Why do we need the logic (from line 327 to 367)? Shouldn't just call some clob function as entry point?

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,172 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+			GoodTilBlockTime: lib.MustConvertIntegerToUint32(placeOrder.Order.GoodTilBlockTime),
+		}
+
+		if ctx.IsCheckTx() || ctx.IsReCheckTx() {
+			// We don't process stateful orders in CheckTx, so do nothing.
+			return nil, nil, nil
+		}
+
+		fmt.Printf("Placing stateful order: %+v\n", order)
+		processProposerMatchesEvents := m.clob.GetProcessProposerMatchesEvents(ctx)
```
teddyding, line 254: Why do we need this?

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +197,172 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+		OrderFlags: cancelOrder.OrderId.OrderFlags,
+		ClobPairId: cancelOrder.OrderId.ClobPairId,
+	}
+
+	if cancelOrder.OrderId.OrderFlags == clobtypes.OrderIdFlags_ShortTerm {
+		// Send error if trying to cancel a short term order. Since short term orders
+		// are not supported for contracts
+		return nil, nil, errors.New("short term orders  are not supported for contracts")
+
+	} else {
```
teddyding, line 332: Why do we need the logic (from line 333 to 367) besides just calling the entrypoint `CancelStatefulOrder`?

protocol/Dockerfile
affanv14: Good point.

protocol/wasmbinding/msg_plugin.go
affanv14: This seems to be called in CheckTx actually - I added this so that it does not

protocol/wasmbinding/bindings/msg.go
```diff
@@ -1,8 +1,11 @@
...
-	CreateTransfer      *CreateTransfer      `json:"create_transfer,omitempty"`
-	DepositToSubaccount *DepositToSubaccount `json:"deposit_to_subaccount,omitempty"`
+	CreateTransfer         *CreateTransfer         `json:"create_transfer,omitempty"`
+	DepositToSubaccount    *DepositToSubaccount    `json:"deposit_to_subaccount,omitempty"`
+	WithdrawFromSubaccount *WithdrawFromSubaccount `json:"withdraw_from_subaccount,omitempty"`
+	PlaceOrder             *PlaceOrder             `json:"place_order,omitempty"`
+	CancelOrder            *CancelOrder            `json:"cancel_order,omitempty"`
 }
 
 type SubaccountId struct {
```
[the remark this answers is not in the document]
affanv14: Just kept what was there from hackathon. I updated it all

protocol/wasmbinding/msg_plugin.go
affanv14: I agree it should just be for 3. Should be easy enough to address

protocol/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # NB: This is a digest for a multi-arch manifest list, you will want to get this by running
 # `docker buildx imagetools inspect golang:1.21-alpine`
```
teddyding, line 2: Update comment

protocol/dydx-cosmwasm/src/msg.rs
```diff
@@ -0,0 +1,110 @@
...
+use serde::{Deserialize, Serialize};
+use serde_repr::*;
+use cosmwasm_std::{
+  CosmosMsg,
+  CustomMsg,
+};
+
+use crate::SubaccountId;
+
+// TODO(affan): handle issue with `GoodTilOneof` in `PlaceOrder` and `CancelOrder` not serializing correctly
```
teddyding, line 11: Nit: For all TODO, link to a Linear ticket as well

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +106,33 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+
+func (m *CustomMessenger) placeOrder(
+	ctx sdk.Context,
+	contractAddr sdk.AccAddress,
+	placeOrder *clobtypes.MsgPlaceOrder,
+) ([]sdk.Event, [][]byte, error) {
+	if placeOrder == nil {
+		return nil, nil, wasmvmtypes.InvalidRequest{Err: "Invalid place order request: No order data provided"}
+	}
+	if ctx.IsCheckTx() || ctx.IsReCheckTx() {
```
teddyding, line 118: Instead here, we should just panic if CheckTx at the top of `DispatchMsg`? Assuming it's true that Wasm execution doesn't do anything during CheckTx

protocol/wasmbinding/msg_plugin.go
```diff
@@ -140,3 +106,33 @@ func parseAddress(addr string) (sdk.AccAddress, error) {
...
+
+func (m *CustomMessenger) cancelOrder(
+	ctx sdk.Context,
+	contractAddr sdk.AccAddress,
+	cancelOrder *clobtypes.MsgCancelOrder,
+) ([]sdk.Event, [][]byte, error) {
+	if cancelOrder == nil {
+		return nil, nil, wasmvmtypes.InvalidRequest{Err: "Invalid cancel order request: No order data provided"}
+	}
+	if ctx.IsCheckTx() || ctx.IsReCheckTx() {
```
teddyding, line 133: Same here as above

PR #1628, opened 2024-06-05
schedule fork upgrade at snapshot height

teddyding:


protocol/app/upgrades.go
```diff
@@ -16,7 +16,18 @@ var (
 	Upgrades = []upgrades.Upgrade{
 		v5_0_0.Upgrade,
 	}
-	Forks = []upgrades.Fork{}
+	Forks = []upgrades.Fork{
+		{
+			UpgradeHeight: 17272001,
```
teddyding, line 21: Rule of thumb: this height should match the height in `data/upgrade-info.json`

protocol/app/upgrades.go
```diff
@@ -16,7 +16,18 @@ var (
 	Upgrades = []upgrades.Upgrade{
 		v5_0_0.Upgrade,
```
teddyding, line 17: This object is required

PR #1633, opened 2024-06-06
GRPC Streaming Batching

jonfung-dydx:
Two flags added for maximum buffer size and flush interval ms
batch updates + gouroutine to emit them

TODO: exit all subscriptions upon loop overlap

## Summary by CodeRabbit

- **New Features**
  - Introduced a new gRPC streaming manager with enhanced buffering and periodic flushing capabilities.
  - Added methods for sending snapshots and stopping the streaming manager.

- **Bug Fixes**
  - Removed unnecessary parameters from various functions to streamline orderbook updates.

- **Refactor**
  - Renamed and restructured fields in orderbook update responses for consistency.
  - Updated gRPC streaming manager to handle updates via a cache and periodic flushes.

- **Documentation**
  - Improved comments and documentation for better clarity on handling updates and snapshots.

- **Tests**
  - Added tests for new gRPC streaming flags and validation logic.


protocol/app/flags/flags.go
```diff
@@ -50,8 +55,11 @@ const (
 	DefaultNonValidatingFullNode = false
 	DefaultDdErrorTrackingFormat = false
 
-	DefaultGrpcStreamingEnabled = false
-	DefaultVEOracleEnabled      = true
+	DefaultGrpcStreamingEnabled         = false
+	DefaultGrpcStreamingFlushIntervalMs = 50
+	DefaultGrpcStreamingMaxBufferSize   = 10000
```
jayy04, line 60: is 10k reasonable here? should we go higher in case of spikes in traffic?

jayy04: i guess if we are doing 10ms, 10k should be enough

protocol/streaming/grpc/grpc_streaming_manager.go
```diff
@@ -25,6 +25,15 @@ type GrpcStreamingManagerImpl struct {
 	// orderbookSubscriptions maps subscription IDs to their respective orderbook subscriptions.
 	orderbookSubscriptions map[uint32]*OrderbookSubscription
 	nextSubscriptionId     uint32
+
+	// grpc stream will batch and flush out messages every 10 ms.
+	ticker *time.Ticker
+	done   chan bool
+	// map of clob pair id to stream updates.
+	streamUpdateCache map[uint32][]clobtypes.StreamUpdate
```
jayy04, line 33: not related to this PR, but in memclob we should probably only generate updates for clob pair ids that are relevant. e.g. those with at lease one subscriber

protocol/app/flags/flags.go
jonfung-dydx: i'll use these for now and then tune later based on metrics

jayy04 (approved):

jonfung-dydx:
@mergifyio backport v5.x

PR #1662, opened 2024-06-10
[CT-913] Fix issue where pnl wasn't being calculated for some subaccounts

dydxwill:
### Changelist
Fix issue where pnl wasn't being calculated for some subaccounts

### Test Plan
Tested in env

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced functionality to provide additional information about the latest processed block time and the count of records.

- **Refactor**
  - Updated methods to return both `maxBlockTime` and `count` for better data insights.
  - Improved logging to include more detailed information about subaccount processing.

- **Tests**
  - Updated test cases to reflect changes in method returns.
  - Added a new test case for subaccount limit handling.

### DIFF
--- a/indexer/packages/postgres/src/stores/pnl-ticks-table.ts
+++ b/indexer/packages/postgres/src/stores/pnl-ticks-table.ts
@@ -212,17 +212,24 @@ function convertPnlTicksFromDatabaseToPnlTicksCreateObject(
   return _.omit(pnlTicksFromDatabase, PnlTicksColumns.id);
 }
 
-export async function findLatestProcessedBlocktime(): Promise<string> {
+export async function findLatestProcessedBlocktimeAndCount(): Promise<{
+  maxBlockTime: string,
+  count: number,
+}> {
   const result: {
-    rows: [{ max: string }]
+    rows: [{ max: string, count: number }]
   } = await knexReadReplica.getConnection().raw(
     `
-    SELECT MAX("blockTime")
+    SELECT MAX("blockTime") as max, COUNT(*) as count
     FROM "pnl_ticks"
     `
     ,
-  ) as unknown as { rows: [{ max: string }] };
-  return result.rows[0].max || ZERO_TIME_ISO_8601;
+  ) as unknown as { rows: [{ max: string, count: number }] };
+
+  return {
+    maxBlockTime: result.rows[0].max || ZERO_TIME_ISO_8601,
+    count: Number(result.rows[0].count) || 0,
+  };
 }
 
 export async function findMostRecentPnlTickForEachAccount(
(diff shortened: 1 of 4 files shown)


indexer/services/roundtable/__tests__/tasks/create-pnl-ticks.test.ts
```diff
@@ -305,7 +312,7 @@ describe('create-pnl-ticks', () => {
       expect(pnlTicks.length).toEqual(1);
       expect(logger.info).toHaveBeenCalledWith(
         expect.objectContaining({
-          message: 'Skipping run because update interval has not been reached',
+          message: 'Skipping run because update interval has not been reached and all subaccounts have been processed',
```
Christopher-Li, line 315: nit: create separate test to check the count condition

indexer/packages/postgres/src/stores/pnl-ticks-table.ts
```diff
@@ -212,17 +212,24 @@ function convertPnlTicksFromDatabaseToPnlTicksCreateObject(
   return _.omit(pnlTicksFromDatabase, PnlTicksColumns.id);
 }
 
-export async function findLatestProcessedBlocktime(): Promise<string> {
+export async function findLatestProcessedBlocktimeAndCount(): Promise<{
+  maxBlockTime: string,
+  count: number,
```
Christopher-Li, line 217: im a little confused, why are we counting the number or rows in `pnl_ticks`? How does that relate to "all subaccounts have been processed"

indexer/services/roundtable/__tests__/tasks/create-pnl-ticks.test.ts
dydxwill: done

indexer/packages/postgres/src/stores/pnl-ticks-table.ts
dydxwill: This function is used to determine whether or not to run the pnl task.

Previously, we were looking at the last block time in the pnl table to determine whether or not to run the pnl task. This worked, but then we imposed a limit on the number of subaccounts/run to compute pnl for. What happened was we started to skip computing pnl on certain subaccounts because we were hitting the limit/run. So, this PR changes this check to see if the last run computed pnl for LIMIT # of subaccounts, and if so, proceed with the task & check the pnl timestamps of the subaccounts individually.

dydxwill: https://github.com/dydxprotocol/v4-chain/blob/c26b4593a8ff5f03c4e205def779562ec5a4ecc0/indexer/services/roundtable/src/helpers/pnl-ticks-helper.ts#L82

Christopher-Li (approved):

dydxwill:
@mergifyio backport release/indexer/v5.x

PR #1668, opened 2024-06-11
[performance]: Don't use branched context for OI calculation if not needed

BrendanChou:
### Changelist
- `CacheContext()` uses a significant amount of time on the profiler. Don't use it if not needed (i.e. the OI delta is zero)

## Summary by CodeRabbit

- **Refactor**
  - Improved logic for calculating new OIMF after an open interest increase to ensure more accurate updates.
  - Enhanced handling of open interest delta scenarios for better stability and reliability.


BrendanChou:
Removing as this might be breaking due to gas calculation costs

PR #1684, opened 2024-06-12
Set all ints in documentation to ints rather than numbers

Christopher-Li:
### Changelist
Set all ints in documentation to ints rather than numbers

### Test Plan
Look at documentation

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated API documentation to standardize the representation of `subaccountNumber` as an integer instead of a floating-point number.
  
- **New Features**
  - Added validation for several fields including `pageSize`, `totalResults`, `offset`, and `subaccountNumber` in various response objects.

- **Bug Fixes**
  - Updated Swagger JSON schema to ensure `subaccountNumber` is correctly typed as an integer (`int32`) instead of a floating-point number (`double`).

### DIFF
--- a/indexer/services/comlink/src/types.ts
+++ b/indexer/services/comlink/src/types.ts
@@ -43,8 +43,17 @@ export enum RequestMethod {
 
 /* ------- Pagination ------- */
 export interface PaginationResponse {
+  /**
+   * @isInt
+   */
   pageSize?: number,
+  /**
+   * @isInt
+   */
   totalResults?: number,
+  /**
+   * @isInt
+   */
   offset?: number,
 }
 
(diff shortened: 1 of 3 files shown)


indexer/services/comlink/src/types.ts
```diff
@@ -57,6 +66,9 @@ export interface AddressResponse {
 
 export interface SubaccountResponseObject {
   address: string,
+  /**
+   * @isInt
+   */
```
shrenujb, line 71: Wow this is so ugly. Is this just a workaround or do we need to do this for any integer in TS?

Christopher-Li: yeah we need this for any integer in the documentation we're using

shrenujb (approved):

vincentwschau (approved):

PR #1707, opened 2024-06-17
GRPC streaming move lock above send orderbook updates

jonfung-dydx:
We were iterating over the orderbook subscriptions without the lock.


vincentwschau (approved):
nit: TODO for tests?

jayy04 (approved):

PR #1708, opened 2024-06-17
Fix CI for cosmwasm integrated V4-chain

affanv14:
### Changelist
- Update dockerfile to support x86 arch(CI uses x86)
- update genesis file
- Fix broken tests by including new cosmwasm messages
- Add wasm module to sim tests
- Fix broken container upgrade tests by adding store key for wasm
- One of the simulation tests is still broken, been hard to replicate - I'll dive into this in a later PR

### Test Plan

### Author/Reviewer Checklist


protocol/app/module_accounts_test.go
```diff
@@ -34,6 +35,7 @@ func TestModuleAccountsToAddresses(t *testing.T) {
 		vestmoduletypes.CommunityTreasuryAccountName: "dydx15ztc7xy42tn2ukkc0qjthkucw9ac63pgp70urn",
 		vestmoduletypes.CommunityVesterAccountName:   "dydx1wxje320an3karyc6mjw4zghs300dmrjkwn7xtk",
 		icatypes.ModuleName:                          "dydx1vlthgax23ca9syk7xgaz347xmf4nunefw3cnv8",
+		wasmtypes.ModuleName:                         "dydx1xds4f0m87ajl3a6az6s2enhxrd0wta48eljwng",
```
teddyding, line 38: Usually there's no need to explicitly define a module address as a module account, which could lead to unexpected behavior when user send funds to it. Any specific reason here?

protocol/app/module_accounts_test.go
```diff
@@ -72,6 +74,7 @@ func TestMaccPerms(t *testing.T) {
 		"rewards_vester":         nil,
 		"community_treasury":     nil,
 		"community_vester":       nil,
+		"wasm":                   {"burner"},
```
teddyding, line 77: Why does `wasm` need to have burner permission? Can we document?

protocol/app/msgs/all_msgs.go
```diff
@@ -342,6 +342,65 @@ var (
 		"/ibc.applications.interchain_accounts.controller.v1.MsgUpdateParamsResponse":              {},
 		"/ibc.applications.interchain_accounts.host.v1.MsgUpdateParams":                            {},
 		"/ibc.applications.interchain_accounts.host.v1.MsgUpdateParamsResponse":                    {},
+
+		// wasm messages
+		"/cosmwasm.wasm.v1.AcceptedMessageKeysFilter":                  {},
```
teddyding, line 347: Can you add a TODO ticket to audit if all messages are needed? We want to support the bare minimal necessary

protocol/app/module_accounts_test.go
teddyding: Usually there's no need to explicitly define a module address as a module/

If we do want to add it as a module account, we should explicitly initialize the module account in the upgrade handler to avoid unexpected panic behavior when user send funds to it (relevant [outage](https://www.notion.so/dydx/Module-accounts-not-correctly-initialized-9cbb83d403704f8da8e14617fb499f6f?pvs=4))

protocol/app/upgrades/v5.0.0/constants.go
```diff
@@ -18,6 +19,7 @@ var Upgrade = upgrades.Upgrade{
 	StoreUpgrades: store.StoreUpgrades{
 		Added: []string{
 			vaulttypes.StoreKey,
+			wasmtypes.StoreKey,
```
teddyding, line 22: This should not be in v5.0.0 upgrade which has already happened. We don't yet know which version will CosmWasm be released in, could you create a dummy upgrade version (`v_upgrade_cosmwasm`)?

protocol/app/msgs/all_msgs.go
affanv14: Will do

protocol/app/upgrades/v5.0.0/constants.go
affanv14: I was going to put it in v6 after merging with main. I agree this is a better approach though

protocol/app/module_accounts_test.go
affanv14: I saw that Buner permission was already given to cosmwasm in the feature branch, this is just adding a test to it. I'll sync with you on module accounts

protocol/scripts/genesis/sample_pregenesis.json
```diff
@@ -1834,9 +1828,21 @@
...
+    },
+    "wasm": {
+      "codes": [],
+      "contracts": [],
+      "params": {
+        "code_upload_access": {
+          "addresses": [],
+          "permission": "Everybody"
+        },
+        "instantiate_default_permission": "Everybody"
```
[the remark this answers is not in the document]
affanv14: Agreed this is not the end state - was going to add it as part of gov task

protocol/app/upgrades/v5.0.0/constants.go
affanv14: Actually, for now - it may be better to just skip upgrade tests(let container tests fail) than adding the extra code since we arent doing anything for container tests yet. I can add this in later when we are closer to merging

affanv14: changed my mind, added upgrade tests but duplicated a bit of code. Will remove duplicated code on actual upgrade

protocol/app/module_accounts_test.go
```diff
@@ -72,6 +74,9 @@ func TestMaccPerms(t *testing.T) {
 		"rewards_vester":         nil,
 		"community_treasury":     nil,
 		"community_vester":       nil,
+		// required for burn msg in wasmvm
+		// https://github.com/CosmWasm/wasmvm/blob/1c3fdc2a4402e527617ec72fe53f114b24899a01/types/msg.go#L116-L121
+		"wasm": {"burner"},
```
teddyding, line 79: What does this mean exactly? Can any contract dispatch `BurnMsg`, and what are the restrictions? Also this comment should not be in the test file but in the actual permissions

protocol/app/upgrades/v5.0.0/constants.go
teddyding: Again, we should not be modifying anything related to `v5.0.0` upgrade. This needs to be in `v.cosmawm.0`

protocol/app/msgs/normal_msgs.go
```diff
@@ -211,6 +212,65 @@ var (
 
 		// ica
 		"/ibc.applications.interchain_accounts.v1.InterchainAccount": nil,
+
+		// wasm
+		"/cosmwasm.wasm.v1.AcceptedMessageKeysFilter":                  nil,
```
teddyding, line 217: As commented [here](https://www.notion.so/dydx/CosmWasm-Implementation-Spec-aa96d7ae25c64ccaadc1f61827c69e73?pvs=4#0fb0e1cbe36e4506ba25342385a9dffc), we should audit these messages, only support necessary ones and move others to unsupported [messages](https://github.com/dydxprotocol/v4-chain/blob/b0ca05073a250992478a39c3614f081cbdf79d46/protocol/lib/ante/unsupported_msgs.go#L11). Could you add a TODO item for this?

affanv14: There is a todo item in linear https://linear.app/dydx/issue/OTE-461/audit-messages-allowed-by-cosmwasm

affanv14: I'll add a comment for TODO

protocol/Dockerfile
```diff
@@ -26,12 +27,15 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
 
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.5.0/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep 2687afbdae1bc6c7c8b05ae20dfb8ffc7ddc5b4e056697d0f37853dfe294e913
-
-# Copy the library you want to the final location that will be found by the linker flag `-lwasmvm_muslc`
-RUN cp /lib/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.a
-
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
+  wget -O /lib/libwasmvm_muslc.a https://github.com/CosmWasm/wasmvm/releases/download/${WASMVM_VERSION}/libwasmvm_muslc.aarch64.a; \
```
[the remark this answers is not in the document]
affanv14: So that this runs in both x86 and aarch64 architecture. CI fails without x86 support

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
```diff
@@ -0,0 +1,164 @@
...
+
+		// Try to get the account in state.
+		acc := ak.GetAccount(ctx, addr)
+		if acc != nil {
+			// Account has been initialized.
+			macc, isModuleAccount := acc.(sdk.ModuleAccountI)
+			if isModuleAccount {
+				// Module account was correctly initialized. Skipping
+				ctx.Logger().Info(fmt.Sprintf(
+					"module account %+v was correctly initialized. No-op",
```
teddyding, line 47:
```suggestion
					"module account %+v has already been initialized. No-op",
```

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
```diff
@@ -0,0 +1,164 @@
...
+		maccI := (ak.NewAccount(ctx, newModuleAccount)).(sdk.ModuleAccountI) // this set the account number
+		ak.SetModuleAccount(ctx, maccI)
+		ctx.Logger().Info(fmt.Sprintf(
+			"Successfully initialized module account in state: %+v",
+			newModuleAccount,
+		))
+	}
+}
+
+func removeStatefulFOKOrders(ctx sdk.Context, k clobtypes.ClobKeeper) {
```
teddyding, line 90: Why is this needed?

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
```diff
@@ -0,0 +1,164 @@
...
+							clobtypes.OrderRemoval_REMOVAL_REASON_CONDITIONAL_FOK_COULD_NOT_BE_FULLY_FILLED,
+						),
+					),
+				),
+			)
+		}
+	}
+}
+
+func initRevShareModuleState(
```
teddyding, line 115: Why is this needed? This upgrade handler should handle CosmWasm migrations only

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
```diff
@@ -0,0 +1,164 @@
...
+	priceKeeper pricetypes.PricesKeeper,
+	ak authkeeper.AccountKeeper,
+) upgradetypes.UpgradeHandler {
+	return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		sdkCtx := lib.UnwrapSDKContext(ctx, "app/upgrades")
+		sdkCtx.Logger().Info(fmt.Sprintf("Running %s Upgrade...", UpgradeName))
+
+		InitializeModuleAccs(sdkCtx, ak)
+		// Remove all stateful FOK orders from state.
+		removeStatefulFOKOrders(sdkCtx, clobKeeper)
```
teddyding, line 155: Remove anything that's not necessary for CosmWasm upgrade

affanv14: We're upgrading from 5-> cosmwasm upgrade. Some code there is for new modules and migrations that need to be applied for 5->6. I was planning to remove this later when we are merging to main

protocol/app/upgrades/v5.0.0/constants.go
affanv14: Missed this. Good catch!

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
teddyding: Given v6[ code complete](https://www.notion.so/dydx/Code-Complete-Tested-13d0fc7c14134f1dae2a2e5b269ca89d?pvs=4) is tomorrow, let's plan on upgrading from v6. I think this requires setting up the pre-upgrade with v6 binary which doesn't exist yet. For now to reduce friction let's comment out the upgrade container test for CosmWasm, and set it up later when we are ready to merge into `main`.

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
affanv14: discussed in https://github.com/dydxprotocol/v4-chain/pull/1708#discussion_r1670834232

protocol/app/upgrades/v.cosmwasm.0/upgrade.go
affanv14: discussed in https://github.com/dydxprotocol/v4-chain/pull/1708#discussion_r1670834232

protocol/app/module_accounts_test.go
affanv14: Updated the comment. Went through the code and IIUC any contract can dispatch BurnMsg for itself https://github.com/CosmWasm/wasmd/blob/v0.51.0/x/wasm/keeper/handler_plugin.go#L225-L244

teddyding: Ok thanks. It looks like contract can only burn coins that belong to itself
`// Important if a contract controls significant token supply that must be retired.` what do you mean by this? Seems we can leave out this comment

protocol/app/msgs/all_msgs.go
teddyding: Can you comment inline `TODO(linear-ticket)`

protocol/app/msgs/normal_msgs.go
teddyding: Note clarified the linear ticket a bit more to disambiguate from another audit.

teddyding (approved):

protocol/app/module_accounts_test.go
affanv14: Ah... i added that to just say that if a contract has a lot of tokens and is retired, those tokens need to be burned

PR #1730, opened 2024-06-19
add margin.Risk type

BrendanChou:
### Changelist
- Add new `protocol/lib/metric` package to be used across several modules
- Modify `AssetsKeeper` to return `NetCollateral` and `MarginRequirements` in the same function

### Test Plan
Existing tests cover the current logic. AssetsKeeper tests were combined/modified for the combined function

### DIFF
--- a/protocol/lib/margin/risk.go
+++ b/protocol/lib/margin/risk.go
@@ -1,0 +1,26 @@
+package margin
+
+import (
+	"math/big"
+)
+
+type Risk struct {
+	MMR *big.Int // Maintenance Margin Requirement
+	IMR *big.Int // Initial Margin Requirement
+	NC  *big.Int // Net Collateral
+}
+
+// AddInPlace adds the values of b to a (in-place).
+func (a *Risk) AddInPlace(b Risk) {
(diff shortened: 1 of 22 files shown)


jayy04 (approved):

jayy04 (approved):

protocol/x/assets/lib/lib.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"math/big"
+
+	"github.com/dydxprotocol/v4-chain/protocol/lib/margin"
+	"github.com/dydxprotocol/v4-chain/protocol/x/assets/types"
+)
+
+// GetNetCollateralAndMarginRequirements returns the net collateral, initial margin, and maintenance margin
+// that a given position (quantums) for a given assetId contributes to an account.
+func GetNetCollateralAndMarginRequirements(
```
mananbordia, line 12: Naming for this function seems a bit odd.
IMR and MMR are always returned as zero which makes no sense.

mananbordia: cc: @jayy04

PR #1735, opened 2024-06-19
Send order remove subaccount message on order replace

chenyaoy:
### Changelist
Send a subaccount websocket message on order replace for the cancelled order so that the FE displays the correct information

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/indexer/services/vulcan/src/handlers/order-replace-handler.ts
+++ b/indexer/services/vulcan/src/handlers/order-replace-handler.ts
@@ -94,6 +94,17 @@ export class OrderReplaceHandler extends Handler {
       }
     }
 
+    /* We don't want to fail if old order is not found (new order should still be placed),
+    so log and track metric */
+    if (!removeOrderResult.removed) {
+      logger.info({
+        at: 'OrderReplaceHandler#handle',
+        message: 'Old order not found in cache',
+        oldOrderId,
+      });
+      stats.increment(`${config.SERVICE_NAME}.replace_order_handler.old_order_not_found_in_cache`, 1);
+    }
+
     /* Place new order */
     const order: IndexerOrder = orderReplace.order!;
     const placementStatus: OrderPlaceV1_OrderPlacementStatus = orderReplace.placementStatus;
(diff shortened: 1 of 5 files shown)


indexer/services/vulcan/src/handlers/order-replace-handler.ts
```diff
@@ -69,7 +85,25 @@ export class OrderReplaceHandler extends Handler {
...
+
+    /* We don't want to fail if old order is not found (new order should still be placed),
+    so log and track metric */
+    if (!removeOrderResult.removed) {
+      logger.info({
+        at: 'OrderReplaceHandler#handle',
+        message: 'Old order not found in cache',
+        oldOrderId,
+      });
+      stats.increment(`${config.SERVICE_NAME}.replace_order_handler.old_order_not_found_in_cache`, 1);
```
dydxwill, line 105: should we return here?

indexer/services/vulcan/src/handlers/order-replace-handler.ts
```diff
@@ -200,9 +274,114 @@ export class OrderReplaceHandler extends Handler {
...
+  protected async cancelOrderInPostgres(
+    oldOrderId: IndexerOrderId,
+  ): Promise<OrderFromDatabase | undefined> {
+    return OrderTable.update({
+      id: OrderTable.orderIdToUuid(oldOrderId),
+      status: OrderStatus.CANCELED,
+    });
+  }
+
+  protected createSubaccountWebsocketMessageFromRemoveOrderResult(
```
dydxwill, line 323: is this used in order-remove-handler as well? can we consolidate? feel free to tackle in a separate PR

dydxwill (approved):

indexer/services/vulcan/src/handlers/order-replace-handler.ts
chenyaoy: No, i still want to place the new order if the old order wasn't found

indexer/services/vulcan/src/handlers/order-replace-handler.ts
chenyaoy: I made some slight changes to it because the order replace message doesn't contain an order removal reason, so i hardcode it to `OrderRemovalReason.ORDER_REMOVAL_REASON_USER_CANCELED` in the order replace handler, which matches the old behavior of canceling and placing a new order

PR #1763, opened 2024-06-24
Add helper methods to compare risks

jayy04:
### Changelist
- Add helper methods to compare risks for future safety heap PRs. Note that this only compares the TNC/MR ratio of the two risk objects, and does not include other constraints (e.g. margin requirements can not increase when subaccount is already under-collateralized) that are used in state transition validations. 

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to compare risks between two accounts, improving accuracy by handling rounding errors.

- **Tests**
  - Added tests to ensure the new risk comparison method works correctly with different scenarios.

### DIFF
--- a/protocol/lib/margin/risk.go
+++ b/protocol/lib/margin/risk.go
@@ -53,6 +53,19 @@ func (a *Risk) IsLiquidatable() bool {
 	return a.MMR.Sign() > 0 && a.MMR.Cmp(a.NC) > 0
 }
 
+// Cmp compares the risks of two accounts.
+// Returns -1 if a is less risky than b, 0 if they are equally risky, and 1 if a is more risky than b.
+
+// Note that here we are effectively checking that
+// `a.NetCollateral / a.MaintenanceMargin >= b.NetCollateral / b.MaintenanceMargin`.
+// However, to avoid rounding errors, we factor this as
+// `a.NetCollateral * b.MaintenanceMargin >= b.NetCollateral * a.MaintenanceMargin`.
+func (a *Risk) Cmp(b Risk) int {
+	ANcMultBMmr := new(big.Int).Mul(a.NC, b.MMR)
+	BNcMultAMmr := new(big.Int).Mul(b.NC, a.MMR)
+	return BNcMultAMmr.Cmp(ANcMultBMmr)
+}
+
 func mustExist(i *big.Int) *big.Int {
 	if i == nil {
 		return new(big.Int)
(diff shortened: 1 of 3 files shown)


BrendanChou (approved):

protocol/lib/margin/risk_test.go
```diff
@@ -220,3 +220,93 @@ func TestRisk_IsLiquidatable(t *testing.T) {
 		})
 	}
 }
+
+func TestRisk_Cmp(t *testing.T) {
```
BrendanChou, line 224: need tests with zero MMR

protocol/lib/margin/risk.go
```diff
@@ -53,6 +53,19 @@ func (a *Risk) IsLiquidatable() bool {
...
 }
 
+// Cmp compares the risks of two accounts.
+// Returns -1 if a is less risky than b, 0 if they are equally risky, and 1 if a is more risky than b.
+
+// Note that here we are effectively checking that
+// `a.NetCollateral / a.MaintenanceMargin >= b.NetCollateral / b.MaintenanceMargin`.
+// However, to avoid rounding errors, we factor this as
+// `a.NetCollateral * b.MaintenanceMargin >= b.NetCollateral * a.MaintenanceMargin`.
+func (a *Risk) Cmp(b Risk) int {
```
BrendanChou, line 63: - consider if `Cmp` is the clearest name as opposed to `CmpRisk` or `CmpSafety` (would be the inverse)
- also consider if we need strict ordering at all (?) like if 2/1 should `Cmp==0` to 4/2

BrendanChou (changes requested):

protocol/lib/margin/risk_test.go
jayy04: Added!

protocol/lib/margin/risk.go
jayy04: Think `Cmp` is fine if we want to make it a method on `Risk`. CmpRisk / CmpSafety sounds like a library function that takes in two `Risk` objects. Let me know if you have a preference here. 

> also consider if we need strict ordering at all (?) like if 2/1 should Cmp==0 to 4/2
Added strict ordering. Tie break using NC if a.MMR = b.MMR = 0, otherwise, tie break using MMR. Added corresponding unit tests 🙏

BrendanChou (approved):

PR #1776, opened 2024-06-25
GPRC Streaming change default flag options, break upon connection order

jonfung-dydx:
- change max channel buffer size to be 2000
- change max batch (buffer) size to be 2000
- Instead of immediately removing the subscription in the subscription's goroutine thread, simply end the goroutine and let the main thread clean up the subscription

## Summary by CodeRabbit

- **Improvements**
  - Adjusted default values for gRPC streaming parameters to optimize resource usage and performance.
  - Enhanced code comments for better clarity in the gRPC streaming manager's subscription logic.

- **Bug Fixes**
  - Corrected metric increment logic in the gRPC streaming manager for accurate performance tracking.


jayy04 (approved):

jonfung-dydx:
@mergify.io backport protocol/v6.x

PR #1810, opened 2024-06-28
[TRA-444] Emit a metric to record market mapper revenue per market

shrenujb:
### Changelist
Emit a metric to record market mapper revenue per market

### Test Plan
Look at datadog

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
    - Added metric tracking for fees transferred to the market mapper. 

- **Chores**
    - Removed outdated TODO comment related to monitoring.

### DIFF
--- a/protocol/x/subaccounts/keeper/transfer.go
+++ b/protocol/x/subaccounts/keeper/transfer.go
@@ -3,6 +3,8 @@ package keeper
 import (
 	"math/big"
 
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+
 	"github.com/dydxprotocol/v4-chain/protocol/lib/log"
 
 	errorsmod "cosmossdk.io/errors"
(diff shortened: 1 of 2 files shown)


protocol/x/subaccounts/keeper/transfer.go
```diff
@@ -262,8 +264,17 @@ func (k Keeper) DistributeFees(
...
 
+	// Emit a metric for the amount of fees transferred to the market mapper
+	labels := []metrics.Label{
+		metrics.GetLabelForIntValue(metrics.MarketId, int(perpetual.Params.MarketId)),
+	}
+	metrics.AddSampleWithLabels(
+		metrics.MarketMapperRevenue,
+		metrics.GetMetricValueFromBigInt(marketMapperShare),
+		labels...,
+	)
```
vincentwschau, line 275: What do you plan on doing with this metric? It's emitted per block, but gets sampled, does that mean we get some approx. sum over the blocks or approx. average?

shrenujb: Hmm i hope these metrics don't get sampled. At least the IncrCounter and AddSample ones.
How do you know these are sampled?

shrenujb: I see we use the same metric to count liquidations and deleveraging: https://github.com/dydxprotocol/v4-chain/blob/dfd4f6798265101bbd52a17c92da3d3a1a87652c/protocol/x/clob/keeper/deleveraging.go#L111

shrenujb: Asked around and these metrics are not sampled by default

vincentwschau: `AddSample` does do summarization of metrics, but the summary does track the `sum` which is what we need. Can you update the metric name to indicate this, e.g. similar to the deleverage metric with the `distribution` suffix?

vincentwschau (approved):

PR #1836, opened 2024-07-02
Add upgrade handler for v5.1.0

vincentwschau:
### Changelist
Add upgrade handler for v5.1.0.

### Test Plan
Updated container test, also tested upgrading from v5.0.5 in staging.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/app/upgrades.go
+++ b/protocol/app/upgrades.go
@@ -3,7 +3,7 @@ package app
 import (
 	"fmt"
 
-	v5_0_0 "github.com/dydxprotocol/v4-chain/protocol/app/upgrades/v5.0.0"
+	v5_1_0 "github.com/dydxprotocol/v4-chain/protocol/app/upgrades/v5.1.0"
 
 	upgradetypes "cosmossdk.io/x/upgrade/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -14,7 +14,7 @@ var (
 	// `Upgrades` defines the upgrade handlers and store loaders for the application.
 	// New upgrades should be added to this slice after they are implemented.
 	Upgrades = []upgrades.Upgrade{
-		v5_0_0.Upgrade,
+		v5_1_0.Upgrade,
 	}
 	Forks = []upgrades.Fork{}
 )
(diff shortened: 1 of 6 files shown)


protocol/app/upgrades/v5.1.0/upgrade_container_test.go
```diff
@@ -0,0 +1,187 @@
+//go:build all || container_test
```
vincentwschau, line 1: Nothing in state to test after the upgrade, just an empty test to run the upgrade.

protocol/app/upgrades/v5.1.0/upgrade_container_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+func preUpgradeChecks(node *containertest.Node, t *testing.T) {
+	// Add test for your upgrade handler logic below
+}
+
+func postUpgradeChecks(node *containertest.Node, t *testing.T) {
+	// Add test for your upgrade handler logic below
+}
+
+func placeOrders(node *containertest.Node, t *testing.T) {
```
vincentwschau, line 52: Copied from previous upgrade test to sanity check network is running.

protocol/app/upgrades/v5.0.0/upgrade_container_test.go
```diff
@@ -1,362 +0,0 @@
...
-	"github.com/stretchr/testify/require"
-)
-
-const (
-	AliceBobBTCQuantums = 1_000_000
-	CarlDaveBTCQuantums = 2_000_000
-	CarlDaveETHQuantums = 4_000_000
-)
-
-func TestStateUpgrade(t *testing.T) {
```
teddyding, line 28: For my understanding, why do we need to delete this?

teddyding (approved):

teddyding:
For documentation purpose, should we frontport the v5.1.0 upgrade handler to `main` branch? So that the upgrade path is always clear.

protocol/app/upgrades/v5.0.0/upgrade_container_test.go
vincentwschau: It runs using the same container test setup I think, and will break since we're setting it up for a 5.0.5 to 5.1.0 upgrade and not a 4.0.0 to 5.0.0 upgrade.

vincentwschau:
We can front-port it, I'll set that up afterwards.

vincentwschau:
@Mergifyio backport main

PR #1846, opened 2024-07-03
feat(oe): Deprecate `marketToCreatedAt` on prices keeper

teddyding:
### Changelist
Deprecate `marketToCreatedAt`, an in-memory object used only to adjust log levels. This is motivated by OE integration which requires:
- Access to in-memory object on keepers to be thread safe
- Write to in-memory objects on keepers to occur after canonical block is finalized
Also fix a bug that causes excessive logging. 

Since `marketToCreatedAt` doesn't provide much value to protocol, deprecating it is the easiest option. 

### Test Plan
Covered by existing tests. 

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved market creation and availability logic by removing outdated time-based checks.
  - Enhanced logging for non-existent market prices.

- **Refactor**
  - Simplified internal data structures and removed unnecessary imports to optimize performance.

- **Tests**
  - Updated test cases to reflect changes in market availability logic, removing obsolete tests.

### DIFF
--- a/protocol/x/prices/keeper/market.go
+++ b/protocol/x/prices/keeper/market.go
@@ -86,33 +86,12 @@ func (k Keeper) CreateMarket(
 		),
 	)
 
-	k.marketToCreatedAt[marketParam.Id] = k.timeProvider.Now()
 	metrics.SetMarketPairForTelemetry(marketParam.Id, marketParam.Pair)
 
 	// create a new market rev share
 	k.RevShareKeeper.CreateNewMarketRevShare(ctx, marketParam.Id)
 
 	return marketParam, nil
-}
-
-// IsRecentlyAvailable returns true if the market was recently made available to the pricefeed daemon. A market is
-// considered recently available either if it was recently created, or if the pricefeed daemon was recently started. If
-// an index price does not exist for a recently available market, the protocol does not consider this an error
-// condition, as it is expected that the pricefeed daemon will eventually provide a price for the market within a
-// few seconds.
-func (k Keeper) IsRecentlyAvailable(ctx sdk.Context, marketId uint32) bool {
-	createdAt, ok := k.marketToCreatedAt[marketId]
-
-	if !ok {
-		return false
-	}
-
-	// The comparison condition considers both market age and price daemon warmup time because a market can be
-	// created before or after the daemon starts. We use block height as a proxy for daemon warmup time because
-	// the price daemon is started when the gRPC service comes up, which typically occurs just before the first
-	// block is processed.
-	return k.timeProvider.Now().Sub(createdAt) < types.MarketIsRecentDuration ||
-		ctx.BlockHeight() < types.PriceDaemonInitializationBlocks
 }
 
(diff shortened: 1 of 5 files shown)


protocol/x/prices/keeper/update_price.go
```diff
@@ -74,22 +74,9 @@ func (k Keeper) GetValidMarketPriceUpdates(
...
-			// Conditionally log missing index prices at least 20s after genesis/restart/market creation. We expect that
-			// there will be a delay in populating index prices after network genesis or a network restart, or when a
-			// market is created, it takes the daemon some time to warm up.
-			if !k.IsRecentlyAvailable(ctx, marketId) {
-				nonExistentMarkets = append(nonExistentMarkets, marketId)
-			}
+			nonExistentMarkets = append(nonExistentMarkets, marketId)
 			continue
 		}
-		if len(nonExistentMarkets) > 0 {
```
teddyding, line 85: This code block should not be within this for loop, which just results in logging out this error in every iteration afterwards, even if just once market is missing index price (see example error [log](https://app.datadoghq.com/logs?query=environment%3A%28staging%20OR%20mainnet%20OR%20testnet%29%20%22Index%20price%20for%20markets%20does%20not%20exist%22%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1719439365000&to_ts=1720044165000&live=false)). Moved logging to after the loop.

protocol/x/prices/keeper/update_price.go
```diff
@@ -134,6 +121,13 @@ func (k Keeper) GetValidMarketPriceUpdates(
 		}
 	}
 
+	if len(nonExistentMarkets) > 0 {
+		ctx.Logger().Warn(
```
teddyding, line 125: Prices are managed by Slinky sidecar and this log rarely has material signal, so use `Warn` instead

protocol/x/prices/keeper/market.go
```diff
@@ -95,26 +94,6 @@ func (k Keeper) CreateMarket(
...
-
-	if !ok {
-		return false
-	}
-
-	// The comparison condition considers both market age and price daemon warmup time because a market can be
-	// created before or after the daemon starts. We use block height as a proxy for daemon warmup time because
-	// the price daemon is started when the gRPC service comes up, which typically occurs just before the first
-	// block is processed.
-	return k.timeProvider.Now().Sub(createdAt) < types.MarketIsRecentDuration ||
```
BrendanChou, line 114: remove any newly unused constants like `types.MarketIsRecentDuration` from `params.go`

BrendanChou (approved):

ttl33 (approved):

protocol/x/prices/keeper/update_price.go
teddyding: Removed `MarketIsRecentDuration`. `PriceDaemonInitializationBlocks` is still used in some other legacy logging logic, which we can deprecate in a future PR.

PR #1849, opened 2024-07-04
Ignore short-term clob messages with non-zero tx timeout.

vincentwschau:
### Changelist
Short-term orders with non-zero transaction timeouts are invalid as `GoodTilBlock` already ensures they are not processed. Ignore such place order transactions in the ante handler during `CheckTx`.

### Test Plan
Unit tests.

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Implemented a hotfix to ignore short-term clob messages in transactions with a non-zero timeout height, enhancing transaction reliability.

### DIFF
--- a/protocol/x/clob/ante/clob.go
+++ b/protocol/x/clob/ante/clob.go
@@ -115,7 +115,7 @@ func (cd ClobDecorator) AnteHandle(
 
 			// HOTFIX: Ignore any short-term place orders in a transaction with a timeout height.
 			if timeoutHeight > 0 && ctx.IsCheckTx() {
-				log.InfoLog(
+				log.WarnLog(
 					ctx,
 					"Ignored short-term place order with non-zero timeout height",
 					timeoutHeightLogKey,
--- a/protocol/x/clob/ante/clob.go
+++ b/protocol/x/clob/ante/clob.go
@@ -12,6 +12,10 @@ import (
 	satypes "github.com/dydxprotocol/v4-chain/protocol/x/subaccounts/types"
 )
 
+var (
+	timeoutHeightLogKey = "TimeoutHeight"
+)
+
 // ClobDecorator is an AnteDecorator which is responsible for:
 //   - adding short term order placements and cancelations to the in-memory orderbook (`CheckTx` only).
 //   - adding stateful order placements and cancelations to state (`CheckTx` and `RecheckTx` only).
@@ -68,6 +72,9 @@ func (cd ClobDecorator) AnteHandle(
 		)
 	}
 
+	// HOTFIX: Ignore any short-term clob messages in a transaction with a timeout height.
+	timeoutHeight := GetTimeoutHeight(tx)
+
 	msgs := tx.GetMsgs()
 	var msg = msgs[0]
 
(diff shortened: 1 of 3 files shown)


protocol/x/clob/ante/clob.go
```diff
@@ -130,6 +159,17 @@ func (cd ClobDecorator) AnteHandle(
 			return next(ctx, tx, simulate)
 		}
 
+		// HOTFIX: Ignore any short-term clob messages in a transaction with a timeout height.
```
teddyding, line 162: For my understanding? Why do we need hotfix for `MsgBatchCancel` and `MsgCancel`? They will never be attached in the operations queue, so do not go through the AnteHandler during `PrepareProposal`.

protocol/x/clob/ante/clob.go
```diff
@@ -106,6 +124,17 @@ func (cd ClobDecorator) AnteHandle(
 				return next(ctx, tx, simulate)
 			}
 
+			// HOTFIX: Ignore any short-term clob messages in a transaction with a timeout height.
+			if timeoutHeight > 0 {
```
teddyding, line 128: For my understanding why can't we just return error here for ST placement in CheckTx? 

```
			if timeoutHeight > 0 && ctx.IsCheckTx(){
				// return error
			}
			```

File: protocol/x/clob/ante/clob.go (hunk 1 above)
Comment (vincentwschau) [reply protocol/x/clob/ante/clob.go:162]: True, that can be removed.

File: protocol/x/clob/ante/clob.go (hunk 2 above)
Comment (vincentwschau) [reply protocol/x/clob/ante/clob.go]: Returned an error instead.

Comment (vincentwschau) [reply protocol/x/clob/ante/clob.go]: Original reason for not adding an error, I'm wary of adding errors in a rolling hotfix but this should be fine since it just rejects during checktx.

File: protocol/x/clob/ante/clob.go [hunk 3]
```diff
@@ -68,6 +72,9 @@ func (cd ClobDecorator) AnteHandle(
 		)
 	}
 
+	// HOTFIX: Ignore any short-term clob messages in a transaction with a timeout height.
+	timeoutHeight := GetTimeoutHeight(tx)
```
teddyding, line 76: Nit: move to inline below

protocol/x/clob/ante/clob.go
```diff
@@ -106,6 +113,20 @@ func (cd ClobDecorator) AnteHandle(
 				return next(ctx, tx, simulate)
 			}
 
+			// HOTFIX: Ignore any short-term place orders in a transaction with a timeout height.
+			if timeoutHeight > 0 && ctx.IsCheckTx() {
+				log.InfoLog(
+					ctx,
+					"Ignored short-term place order with non-zero timeout height",
```
teddyding, line 120: Nit: instead of "Ignore", "Reject"

teddyding (approved):

protocol/x/clob/ante/clob.go
```diff
@@ -81,6 +88,17 @@ func (cd ClobDecorator) AnteHandle(
 				return next(ctx, tx, simulate)
 			}
 
+			// HOTFIX: Ignore any short-term clob messages in a transaction with a timeout height.
+			if timeoutHeight > 0 {
+				log.InfoLog(
```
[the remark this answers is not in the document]
vincentwschau: Our new logger doesn't have `Warn`, doesn't seem like it's too important to change to `Warn` either.

PR #1858, opened 2024-07-05
Log order id and good til block of rejected st orders.

vincentwschau:
### Changelist
Add additional logging when rejecting ST orders to have better visibility as to who is sending the order and the difference between `TimeoutHeight` and `GoodTilBlock`.

### Test Plan
N/A

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/ante/clob.go
+++ b/protocol/x/clob/ante/clob.go
@@ -14,6 +14,8 @@ import (
 
 var (
 	timeoutHeightLogKey = "TimeoutHeight"
+	orderIdLogKey       = "OrderId"
+	goodTilBlockLogKey  = "GoodTilBlock"
 )
 
 // ClobDecorator is an AnteDecorator which is responsible for:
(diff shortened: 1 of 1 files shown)


teddyding (approved):

protocol/x/clob/ante/clob.go
```diff
@@ -118,6 +120,10 @@ func (cd ClobDecorator) AnteHandle(
 					"Rejected short-term place order with non-zero timeout height < goodTilBlock",
```
ttl33, line 120: just to double check, does `ctx` have the address?

ttl33 (approved):

protocol/x/clob/ante/clob.go
vincentwschau: No I don't think so.

teddyding: Are you asking if `ctx` carries the transaction sender address? I believe not.

vincentwschau:
@Mergifyio backport release/protocol/v5.1.x

vincentwschau:
@Mergifyio backport main

PR #1875, opened 2024-07-09
Store stateful order expirations individually in state

roy-dydx:
### Changelist
- Store these individually to prevent lots of serde when receiving many orders with the same GTT

### Test Plan
- Modified and added tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved stateful order management by renaming and restructuring functions.
  - Enhanced efficiency in removing expired stateful orders.

- **Tests**
  - Updated test cases to align with the new function names and signatures.

- **Chores**
  - Cleaned up deprecated function calls and unused imports.

### DIFF
--- a/protocol/x/clob/keeper/process_proposer_matches_events.go
+++ b/protocol/x/clob/keeper/process_proposer_matches_events.go
@@ -5,6 +5,7 @@ import (
 	"math"
 
 	"cosmossdk.io/store/prefix"
+	storetypes "cosmossdk.io/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
@@ -71,14 +72,17 @@ func (k Keeper) InitializeProcessProposerMatchesEvents(
 
 // ResetAllDeliveredOrderIds resets the lists of delivered order ids. This should be reset every block.
 func (k Keeper) ResetAllDeliveredOrderIds(ctx sdk.Context) {
-	k.ResetUnorderedOrderIds(ctx, types.DeliveredCancelKeyPrefix)
+	memStore := ctx.KVStore(k.memKey)
+	k.ResetUnorderedOrderIds(ctx, memStore, types.DeliveredCancelKeyPrefix)
 	k.ResetOrderedOrderIds(
 		ctx,
+		memStore,
 		types.OrderedDeliveredLongTermOrderKeyPrefix,
 		types.OrderedDeliveredLongTermOrderIndexKey,
 	)
 	k.ResetOrderedOrderIds(
 		ctx,
+		memStore,
 		types.OrderedDeliveredConditionalOrderKeyPrefix,
 		types.OrderedDeliveredConditionalOrderIndexKey,
 	)
@@ -89,6 +93,7 @@ func (k Keeper) ResetAllDeliveredOrderIds(ctx sdk.Context) {
 func (k Keeper) AddDeliveredLongTermOrderId(ctx sdk.Context, orderId types.OrderId) {
 	k.AppendOrderedOrderId(
 		ctx,
+		ctx.KVStore(k.memKey),
 		types.OrderedDeliveredLongTermOrderKeyPrefix,
 		types.OrderedDeliveredLongTermOrderIndexKey,
 		orderId,
@@ -100,6 +105,7 @@ func (k Keeper) AddDeliveredLongTermOrderId(ctx sdk.Context, orderId types.Order
 func (k Keeper) AddDeliveredConditionalOrderId(ctx sdk.Context, orderId types.OrderId) {
 	k.AppendOrderedOrderId(
 		ctx,
+		ctx.KVStore(k.memKey),
 		types.OrderedDeliveredConditionalOrderKeyPrefix,
 		types.OrderedDeliveredConditionalOrderIndexKey,
 		orderId,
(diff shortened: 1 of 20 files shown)


protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -175,23 +175,67 @@ func (k Keeper) DeleteLongTermOrderPlacement(
...
+// returns the removed order ids as a slice.
+func (k Keeper) RemoveExpiredStatefulOrders(ctx sdk.Context, blockTime time.Time) (
+	expiredOrderIds []types.OrderId,
+) {
+	expiredOrderIds = make([]types.OrderId, 0)
+	store := ctx.KVStore(k.storeKey)
+	start := []byte(fmt.Sprintf(types.StatefulOrdersExpirationsKeyPrefix, sdk.FormatTimeString(time.Time{})))
+	end := []byte(fmt.Sprintf(types.StatefulOrdersExpirationsKeyPrefix, sdk.FormatTimeString(blockTime)))
+	// Increment the byte representation of `/` at the end of the prefix so that blockTime is included in the iteration
+	end[len(end)-1] += 1
```
jayy04, line 230: can this overflow? not sure what formatted time string this is using

there is a util function in sdk store types, which might be better
```
// InclusiveEndBytes returns the []byte that would end a
// range query such that the input would be included
func InclusiveEndBytes(inclusiveBytes []byte) []byte {
	return append(inclusiveBytes, byte(0x00))
}
```

jayy04 (approved):

protocol/x/clob/keeper/stateful_order_state.go
roy-dydx: Since the key prefix is `Exp/%s:`, if you do that, you'd get `Exp/{time}:\00` as the exclusive end, which would not cover all the keys that the orders are stored it. For example `"Exp/{time}:{order_id}" would not get iterated over.

The idea is to increment the byte representation of `:` by 1 so that all order ids at the ending time would be included.

If you can think of better ways of doing this, I'm open to it.

protocol/x/clob/keeper/process_proposer_matches_events.go
```diff
@@ -109,77 +115,77 @@ func (k Keeper) AddDeliveredConditionalOrderId(ctx sdk.Context, orderId types.Or
...
 	}
-	memStore.Delete([]byte(indexKey))
+	store.Delete([]byte(indexKey))
 }
 
-func (k Keeper) ResetUnorderedOrderIds(ctx sdk.Context, keyPrefix string) {
-	memStore := ctx.KVStore(k.memKey)
-	store := prefix.NewStore(memStore, []byte(keyPrefix))
-	it := store.Iterator(nil, nil)
+func (k Keeper) ResetUnorderedOrderIds(ctx sdk.Context, store storetypes.KVStore, keyPrefix string) {
```
teddyding, line 153: For my understanding what's the motivation behind changing this interface?

protocol/x/clob/keeper/stateful_order_state_test.go
```diff
@@ -1216,12 +795,6 @@ func TestRemoveExpiredStatefulOrdersTimeSlices(t *testing.T) {
 			},
```
teddyding, line 795: Can we add a test where multiple orders under the inclusive end blocktime are all removed?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -175,23 +175,67 @@ func (k Keeper) DeleteLongTermOrderPlacement(
...
+// RemoveExpiredStatefulOrders removes the stateful order id expirations up to `blockTime` and
+// returns the removed order ids as a slice.
+func (k Keeper) RemoveExpiredStatefulOrders(ctx sdk.Context, blockTime time.Time) (
+	expiredOrderIds []types.OrderId,
+) {
+	expiredOrderIds = make([]types.OrderId, 0)
+	store := ctx.KVStore(k.storeKey)
+	start := []byte(fmt.Sprintf(types.StatefulOrdersExpirationsKeyPrefix, sdk.FormatTimeString(time.Time{})))
+	end := []byte(fmt.Sprintf(types.StatefulOrdersExpirationsKeyPrefix, sdk.FormatTimeString(blockTime)))
+	// Increment the byte representation of `/` at the end of the prefix so that blockTime is included in the iteration
```
teddyding, line 229:
```suggestion
	// Increment the byte representation of `:` at the end of the prefix so that blockTime is included in the iteration
```
Do you mean this, given the key prefix format is `Exp/%s:`?

protocol/x/clob/keeper/stateful_order_state.go
teddyding: Thinking out loud - could we define another `StatefulOrdersExpirationsKeyPrefixExclusiveEnd = "Exp/%s;"`, and do `end := []byte(fmt.Sprintf(types.StatefulOrdersExpirationsKeyPrefixExclusiveEnd, sdk.FormatTimeString(blockTime)))`

protocol/x/clob/keeper/process_proposer_matches_events.go
roy-dydx: I'm using it with persisted KVStore in this PR.

protocol/x/clob/keeper/stateful_order_state.go
roy-dydx: Yeah that's pretty much the same thing though.

jayy04: there is also another function that we can use called `PrefixEndBytes`

```
// PrefixEndBytes returns the []byte that would end a
// range query for all []byte with a certain prefix
// Deals with last byte of prefix being FF without overflowing
func PrefixEndBytes(prefix []byte) []byte {
	if len(prefix) == 0 {
		return nil
	}

	end := make([]byte, len(prefix))
	copy(end, prefix)

	for {
		if end[len(end)-1] != byte(255) {
			end[len(end)-1]++
			break
		}

		end = end[:len(end)-1]

		if len(end) == 0 {
			end = nil
			break
		}
	}

	return end
}
```

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -175,23 +175,67 @@ func (k Keeper) DeleteLongTermOrderPlacement(
 	)
 }
 
-// GetStatefulOrdersTimeSlice gets a slice of stateful order IDs that expire at `goodTilBlockTime`,
-// sorted by order ID.
-func (k Keeper) GetStatefulOrdersTimeSlice(ctx sdk.Context, goodTilBlockTime time.Time) (
+// GetStatefulOrderIdExpirations gets a slice stateful order IDs that expire at `goodTilBlockTime`,
```
vincentwschau, line 178:
```suggestion
// GetStatefulOrderIdExpirations gets a slice of stateful order IDs that expire at `goodTilBlockTime`,
```

protocol/x/clob/keeper/stateful_order_state.go
jayy04: which does what you are trying to do, and handles overflow

protocol/x/clob/keeper/stateful_order_state_test.go
roy-dydx: Modified the existing test to cover this

PR #1886, opened 2024-07-10
[TRA-169] keep existing vault orders if price and size don't need updates

tqin7:
### Changelist
this behavior is achieved by introducing a new store:
```
most recent client IDs: vaultId -> client IDs of most recently placed orders by a vault
```
- this is necessary because if we don't place/cancel each block, we no longer can deterministically calculate the IDs
of the most recently placed orders (previously, we just calculate this with last block's height)

also set order expiration to be 60s into the future as a vault wants its order to stay as long as possible until an update is needed.

### Test Plan
unit / integration tests
localnet testing

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Added functionality to store and retrieve the most recent client IDs for vaults.
  - Updated order expiration parameter to 60 seconds.
- **Bug Fixes**
  - Improved order management logic for vaults.
- **Tests**
  - Added comprehensive tests for new client ID and order management features.
- **Refactor**
  - Refactored initialization logic and import statements for better efficiency.


protocol/app/app.go
```diff
@@ -1288,16 +1288,14 @@ func New(
...
-		// processed before any vault order expirations (handled by clob)
-		vaultmoduletypes.ModuleName,
 		clobmoduletypes.ModuleName,
 		sendingmoduletypes.ModuleName,
 		vestmoduletypes.ModuleName,
 		rewardsmoduletypes.ModuleName,
 		epochsmoduletypes.ModuleName,
 		govplusmoduletypes.ModuleName,
 		delaymsgmoduletypes.ModuleName,
+		vaultmoduletypes.ModuleName,
```
tqin7, line 1298: previously, vault endblocker was moved to be in front of clob to cancel orders before they expire to avoid flickering (as order expiration was set to be a very small value). With this PR, moving vault endblocker back as order expiration is significantly increased

protocol/x/vault/keeper/orders.go
```diff
@@ -72,78 +73,46 @@ func (k Keeper) RefreshAllVaultOrders(ctx sdk.Context) {
 }
 
 // RefreshVaultClobOrders refreshes orders of a CLOB vault.
+// Note: Order IDs are deterministically constructed based on layer and side. An order ID has its
+// last bit flipped only upon order replacement.
 func (k Keeper) RefreshVaultClobOrders(ctx sdk.Context, vaultId types.VaultId) (err error) {
```
tqin7, line 78: main logic change is in this function

protocol/app/app.go
vincentwschau: Why does the endblocker for the vault module need to be moved back?

tqin7: if vault endblocker is before clob's:
let's say an order is due for expiration, vault endblocker will run and see that the order still exists (thus don't do anything if price / size don't need updates), then clob endblocker will remove expired order. Then there's no order anymore until the next block's endblocker

protocol/x/vault/keeper/orders_test.go
```diff
@@ -287,34 +284,69 @@ func TestRefreshVaultClobOrders(t *testing.T) {
...
+			ctx = tApp.AdvanceToBlock(
+				uint32(tApp.GetBlockHeight())+1,
+				testapp.AdvanceToBlockOptions{
+					BlockTime: ctx.BlockTime().Add(
+						time.Second * time.Duration(params.OrderExpirationSeconds+1),
+					),
+				},
+			)
+			newOrders = tApp.App.ClobKeeper.GetAllStatefulOrders(ctx)
+			require.Len(t, newOrders, int(params.Layers*2))
```
chenyaoy, line 349: Check that client IDs are different from before?

chenyaoy:
What happens when existing vault orders are filled or partially filled? Does the new order quantums change from the old order's quantums and trigger a replacement order?

protocol/app/app.go
vincentwschau: Note this behavior in a comment.

tqin7:
> What happens when existing vault orders are filled or partially filled? Does the new order quantums change from the old order's quantums and trigger a replacement order?

Yep. Replacement happens when price / size changes

protocol/x/vault/keeper/orders.go
```diff
@@ -72,78 +73,46 @@ func (k Keeper) RefreshAllVaultOrders(ctx sdk.Context) {
...
-				ctx,
-				indexerevents.SubtypeStatefulOrder,
-				indexerevents.StatefulOrderEventVersion,
-				indexer_manager.GetBytes(
-					indexerevents.NewLongTermOrderReplacementEvent(
-						*replacedOrderId,
-						*order,
-					),
-				),
-			)
```
vincentwschau, line 144: note: For expired orders, this breaks the order flickering logic since orders will have the expire + place events sent for them.

protocol/x/vault/keeper/orders_test.go
vincentwschau: Or check that the expiration timestamps on the orders are updated.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -287,34 +284,69 @@ func TestRefreshVaultClobOrders(t *testing.T) {
...
+			// Advance to next block with no price updates / order matches and vault should not refresh its orders.
+			msgSender.Clear()
+			ctx = tApp.AdvanceToBlock(uint32(tApp.GetBlockHeight())+1, testapp.AdvanceToBlockOptions{})
+			require.Equal(
+				t,
+				initialOrders,
+				tApp.App.ClobKeeper.GetAllStatefulOrders(ctx),
+			)
+
+			// Advance to next block with price updates and vault should replace its old orders with new ones.
```
vincentwschau, line 309: Update test with case for filled orders / change in quantums.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -915,83 +947,29 @@ func TestGetVaultClobOrderIds(t *testing.T) {
...
-func TestGetVaultClobOrderClientId(t *testing.T) {
-	tests := map[string]struct {
-		/* --- Setup --- */
-		// side.
-		side clobtypes.Order_Side
-		// block height.
-		blockHeight int64
-		// layer.
-		layer uint8
+func TestGetSetMostRecentClientIds(t *testing.T) {
```
vincentwschau, line 950: nit: Prefer testcase organization for this test rather than commented cases.

protocol/x/vault/keeper/orders.go
```diff
@@ -72,78 +73,46 @@ func (k Keeper) RefreshAllVaultOrders(ctx sdk.Context) {
...
+				// Flip last bit of old client ID to get new client ID to make sure they are different
+				// as order placement fails if the same order ID is already marked for cancellation.
+				orderToPlace.OrderId.ClientId = oldClientId ^ 1
+				err = k.ReplaceVaultClobOrder(ctx, vaultId, oldOrderId, orderToPlace)
+			}
+		}
+		if err != nil {
+			log.ErrorLogWithError(ctx, "Failed to place/replace vault clob order", err, "vaultId", vaultId)
 		}
+		clientIds[i] = orderToPlace.OrderId.ClientId
```
vincentwschau, line 112: nit: Make it explicit that if the order is not replaced `oldClientId` is what' stored into the `clientIds`.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -287,33 +293,115 @@ func TestRefreshVaultClobOrders(t *testing.T) {
...
-				require.Len(t, allStatefulOrders, int(params.Layers*2))
-				// Check that the orders are as expected.
-				expectedOrders, err := tApp.App.VaultKeeper.GetVaultClobOrders(ctx, tc.vaultId)
-				require.NoError(t, err)
-				for i := uint32(0); i < params.Layers*2; i++ {
-					require.Equal(t, *expectedOrders[i], allStatefulOrders[i])
-				}
+			}
+
+			// Vault should place its initial orders.
```
vincentwschau, line 305: nit: Add a check after each block that `mostRecentClientIds` of the vault is the same as all the stateful orders.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -287,33 +293,115 @@ func TestRefreshVaultClobOrders(t *testing.T) {
...
+			require.Len(t, newOrders, int(params.Layers*2))
+			for _, newOrder := range newOrders {
+				require.Equal(
+					t,
+					uint32(ctx.BlockTime().Unix())+params.OrderExpirationSeconds,
+					newOrder.GetGoodTilBlockTime(),
+				)
+			}
+
+			// Advance to next block where vault should replace its orders to update their sizes.
```
vincentwschau, line 372: non-blocking: Add TODO for filled stateful vault order testing if possible.

protocol/x/vault/keeper/orders_test.go
```diff
@@ -287,33 +293,115 @@ func TestRefreshVaultClobOrders(t *testing.T) {
...
-				for i := uint32(0); i < params.Layers*2; i++ {
-					require.Equal(t, *expectedOrders[i], allStatefulOrders[i])
-				}
+			}
+
+			// Vault should place its initial orders.
+			initialOrders := tApp.App.ClobKeeper.GetAllStatefulOrders(ctx)
+			require.Len(t, initialOrders, int(params.Layers*2))
+
+			// Advance to next block with no price updates / order matches and vault should not refresh its orders.
```
vincentwschau, line 309: nit: Advance a few more blocks while checking that orders do not get refreshed.

vincentwschau (approved):

protocol/x/vault/keeper/orders.go
tqin7: correct. seems okay to me as expiration occurs much less frequently than before. 

however, one thing we can do might be to replace order (send replacement event) [here](https://github.com/dydxprotocol/v4-chain/blob/tq/tra-169/protocol/x/vault/keeper/orders.go#L99) (i.e. replace order if order gets filled or expires)

PR #1941, opened 2024-07-18
OTE-558 Add scaffolding for module

jerryfan01234:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced the `AccountPlus` module, enhancing account management with new functionalities.
  - Added the `InitializeAccount` function for initializing account states.
  - Enhanced data integrity by ensuring certain fields in the `AccountState` and `GenesisState` are non-nullable.

- **Bug Fixes**
  - Improved error handling in account state retrieval processes.

- **Tests**
  - Implemented extensive test cases for the `InitializeAccount` function, ensuring robust validation of account management.
  
- **Chores**
  - Established groundwork for future development with placeholder functions and tests related to timestamp nonces.


protocol/x/accountplus/types/keys.go
```diff
@@ -2,8 +2,9 @@ package types
 
 // Module name and store keys.
 const (
-	// ModuleName defines the module name.
-	ModuleName = "accountplus"
+	// The Account module uses "acc" as its module name.
+	// KVStore keys cannot have other keys as prefixes so we prepend "dydx" to "accountplus"
+	ModuleName = "dydxaccountplus"
```
Christopher-Li, line 7: are there any issues with the name being dydxaccountplus, and everything else with this module being named accountplus?

protocol/x/accountplus/keeper/keeper.go
```diff
@@ -62,6 +63,40 @@ func (k Keeper) SetGenesisState(ctx sdk.Context, data types.GenesisState) error
...
+
+	if _, found := k.getAccountState(store, address); found {
+		return types.AccountState{}, errors.New(
+			"Cannot initialize AccountState for address with existing AccountState, address: " + address.String(),
+		)
+	}
+
+	initialAccountState := types.AccountState{
+		Address:               address.String(),
+		TimestampNonceDetails: DeepCopyTimestampNonceDetails(InitialTimestampNonceDetails),
```
teddyding, line 77: Let's make the [field definition](https://github.com/dydxprotocol/v4-chain/blob/e86691c003285eb0d00babeadfb8d7828fedc150/proto/dydxprotocol/accountplus/accountplus.proto#L9) not nullable ([eg](https://github.com/dydxprotocol/v4-chain/blob/e86691c003285eb0d00babeadfb8d7828fedc150/proto/dydxprotocol/clob/tx.proto#L78)) so this is never a pointer.

protocol/x/accountplus/testutils/testutils.go
```diff
@@ -0,0 +1,52 @@
+package testutils
+
+import (
+	"github.com/dydxprotocol/v4-chain/protocol/x/accountplus/types"
+	"github.com/google/go-cmp/cmp"
+)
+
+func CompareTimestampNonceDetails(actualDetails, expectedDetails *types.TimestampNonceDetails) bool {
```
teddyding, line 8: Why do we need this special compare function? And do we still need it if it's no longer a pointer?

protocol/x/accountplus/keeper/keeper.go
Christopher-Li: If the field is not nullable, will go automatically copy by value?

protocol/x/accountplus/testutils/testutils.go
jerryfan01234: Not too familiar with Go but my understanding is that TimestampNonceDetails is a struct which contains a slice so comparison is not directly supported. Moreover, slice comparison in Go is by reference. 

Many unit tests will come down to comparing TimestampNonceDetails so I thought it would be helpful to create a helper function (same applies for AccountState).

As for if theres are not pointers, i think the above issues are still relevant.

protocol/x/accountplus/keeper/keeper.go
```diff
@@ -62,6 +63,40 @@ func (k Keeper) SetGenesisState(ctx sdk.Context, data types.GenesisState) error
 	return nil
 }
 
+func (k Keeper) InitializeAccount(ctx sdk.Context, address sdk.AccAddress) (types.AccountState, error) {
```
teddyding, line 66: This function is not used yet right? When will it be used? Trying to understand why we need this function at all (initializing to empty account state)

protocol/x/accountplus/keeper/keeper.go
jerryfan01234: I believe structs are copied by value but if the struct has a slice field, the copy's slice will reference the same slice.

protocol/x/accountplus/keeper/keeper.go
jerryfan01234: Sorry I actually wanted to remove this function for this CR but the commit history got kinda messed up. An upcoming PR will implement business logic in the validator and will use this function. Feel free to review this then.

teddyding: When would we want to initialize empty account state? Wouldn't it always be non-empty when it's first created (i.e. one nonce?)

protocol/x/accountplus/keeper/keeper.go
teddyding: The nullable comment is not just for deep copy, but more for proto conventions (since dealing with pointers is annoying)

teddyding: I'm also wondering why we need to keep this `InitialTimestampNonceDetails` at all. As asked above, when would we ever want to write an empty account state to [state](https://github.com/dydxprotocol/v4-chain/pull/1941/files#r1687176199)?

protocol/x/accountplus/keeper/keeper.go
jerryfan01234: My thought for initializing account was that as users submit tx, if they did not previously have an AccountState (corresponding to accountplus), a new account with default substructs (eg. timestamp nonce {[], 0} will be created. 

Each subcomponent of accountplusmodule, such as timestamp nonce, will be responsible for updating the default state when it becomes relevant. In ts-nonce case, it will be when the user first uses a ts-nonce.

protocol/x/accountplus/testutils/testutils.go
teddyding: > TimestampNonceDetails is a struct which contains a slice so comparison is not directly supported.

IIRC `require.Equal` uses `reflect.DeepEqual` so it should be supported. Otherwise we would've needed to write a lot of recursive compare utilities

protocol/x/accountplus/keeper/keeper.go
teddyding: > My thought for initializing account was that as users submit tx, if they did not previously have an AccountState (corresponding to accountplus), a new account with default substructs (eg. timestamp nonce {[], 0} will be created.

This seems pretty wasteful. In general state storage is a scare resource in blockchain, so we prefer to not add a node vs. having an empty node. Similarly, there was a project to remove empty subaccount state when their no longer have any outstanding balance/positions.

Also it seems natural that any transaction that's no using this special timestamp nonce should no result in any state change in `x/accountplus` state

teddyding:
Could you add label `feature:timestamp_nonce` to this and future PRs

teddyding (approved):

PR #1946, opened 2024-07-18
add v5.2 upgrade handler and upgrade container test

tqin7:
### Changelist
add v5.2 upgrade handler and upgrade container test

### Test Plan
container test (done)
manual upgrade test on staging (planned)

### Author/Reviewer Checklist


protocol/app/upgrades/v5.2.0/upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+	ctx sdk.Context,
+	vaultKeeper vaultkeeper.Keeper,
+) {
+	allVaults := vaultKeeper.GetAllVaults(ctx)
+	for _, vault := range allVaults {
+		vaultKeeper.AddVaultToAddressStore(ctx, *vault.VaultId)
+	}
+}
+
+// setVaultOrderExpirationTo60Seconds sets vault module param `OrderExpirationSeconds` to 60.
```
vincentwschau, line 47: nit: `Sixty` instead of `60`?

protocol/app/upgrades/v5.2.0/upgrade_container_test.go
```diff
@@ -0,0 +1,165 @@
...
+	require.NoError(t, err)
+}
+
+/*** Preupgrade Check ***/
+func preUpgradeChecks(node *containertest.Node, t *testing.T) {
+	// Add test for your upgrade handler logic below
+}
+
+/*** Postupgrade Check ***/
+func postUpgradeChecks(node *containertest.Node, t *testing.T) {
```
vincentwschau, line 98: nit: Check that vault addresses were added to the vault state holding addresses.

vincentwschau:
The container test is failing.

protocol/app/upgrades/v5.2.0/upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+	configurator module.Configurator,
+	clobKeeper clobtypes.ClobKeeper,
+	vaultKeeper vaultkeeper.Keeper,
+) upgradetypes.UpgradeHandler {
+	return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		sdkCtx := lib.UnwrapSDKContext(ctx, "app/upgrades")
+		sdkCtx.Logger().Info(fmt.Sprintf("Running %s Upgrade...", UpgradeName))
+
+		// Migrate x/clob.
+		clobKeeper.UnsafeMigrateOrderExpirationState(sdkCtx)
```
vincentwschau, line 26: Is there any preupgrade state needed to ensure the logic in this function is executed?

protocol/app/upgrades/v5.2.0/upgrade_container_test.go
tqin7: i don't know if that's accessible through `node` / container tests. Instead, I opted to check whether every vault belongs to the best fee tier.

tqin7:
> The container test is failing.

oh that was the 5.1 upgrade test. removed

protocol/app/upgrades/v5.2.0/upgrade.go
tqin7: vaults quoting should be enough. however, let me add explicit logic to place long term orders

protocol/app/upgrades/v5.2.0/upgrade_container_test.go
vincentwschau: non-blocking nit: Check that they aren't part of the best fee tier pre-upgrade?

protocol/Makefile
```diff
@@ -233,11 +233,13 @@ test-unit-and-integration:
...
-	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' ./...
+test-container: # run ./app/upgrades/v5.2.0 first to avoid resource conflict
+	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' ./app/upgrades/v5.2.0
+	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' $$(go list ./... | grep -v './app/upgrades/v5.2.0')
 
-test-container-accept:
-	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' ./... -args -accept
+test-container-accept: # run ./app/upgrades/v5.2.0 first to avoid resource conflict
+	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' ./app/upgrades/v5.2.0 -args -accept
+	@SKIP_DISABLED=true VERSION=$(VERSION) go test -mod=readonly -tags='container_test $(build_tags)' $$(go list ./... | grep -v './app/upgrades/v5.2.0') -args -accept
```
vincentwschau, line 242: non-blocking: What was the issue here? This change means that any future upgrade testing would require updating these command as well, so wondering if there was a better solution that specifically specifying the file to run.

vincentwschau (approved):

protocol/Makefile
tqin7: without this change, `app/upgrades/v5.2.0` and `testing/container` are being run in parallel, leading to resource conflict (attempting to use the same port). This is more of an issue for 5.2 because 5.2's upgrade test takes long as it needs to wait until block 50

in `main`, don't think this change is needed as we won't run 5.2 upgrade test

tqin7:
@mergifyio backport main

PR #2000, opened 2024-07-31
[OTE-600] Emit metrics for geoblock endpoint

affanv14:
### Changelist
Emit metric to track changes in state for geoblock endpoint 

### Test Plan
Added unit test

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced compliance monitoring by tracking changes in compliance status and incrementing related statistics.
  
- **Tests**
	- Added a new test case to ensure correct statistics are captured during geoblock compliance status updates.

### DIFF
--- a/indexer/services/comlink/__tests__/controllers/api/v4/compliance-v2-controller.test.ts
+++ b/indexer/services/comlink/__tests__/controllers/api/v4/compliance-v2-controller.test.ts
@@ -414,6 +414,11 @@ describe('ComplianceV2Controller', () => {
         reason: ComplianceReason.US_GEO,
       }));
 
+      expect(stats.increment).toHaveBeenCalledWith(`${config.SERVICE_NAME}.compliance-v2-controller.geo_block.compliance_status_changed.count`,
+        {
+          newStatus: ComplianceStatus.BLOCKED,
+        });
+
       expect(response.body.status).toEqual(ComplianceStatus.BLOCKED);
       expect(response.body.reason).toEqual(ComplianceReason.US_GEO);
       expect(response.body.updatedAt).toBeDefined();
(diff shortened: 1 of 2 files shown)


indexer/services/comlink/__tests__/controllers/api/v4/compliance-v2-controller.test.ts
```diff
@@ -633,5 +634,30 @@ describe('ComplianceV2Controller', () => {
       expect(response.body.reason).toEqual(ComplianceReason.US_GEO);
       expect(response.body.updatedAt).toBeDefined();
     });
+
+    it('Stats being sent for the geoblock compliance', async () => {
```
Christopher-Li, line 638: nit: just add this expectation on all of the other tests, instead of having a separate test for this. If you're looking for something especially clean, follow [this example](https://github.com/dydxprotocol/v4-chain/blob/9d56ab8d3dc8e37a69612713048ddb9bdc5bd2d4/indexer/services/ender/__tests__/handlers/order-fills/liquidation-handler.test.ts#L1204)

indexer/services/comlink/src/controllers/api/v4/compliance-v2-controller.ts
```diff
@@ -276,6 +276,17 @@ router.post(
         complianceStatus,
         updatedAt,
       );
+      if (complianceStatus[0] !== complianceStatusFromDatabase &&
```
Christopher-Li, line 279: when would `complianceStatus[0] !== complianceStatusFromDatabase` ever be equal?

I also don't think you can access `complianceStatus[0]`, because `complianceStatus` might have length 0

affanv14: Good catch!

affanv14: There are cases where it can be equal - like first_strike -> first_strike  or compliant->compliant

indexer/services/comlink/src/controllers/api/v4/compliance-v2-controller.ts
```diff
@@ -276,6 +276,19 @@ router.post(
         complianceStatus,
         updatedAt,
       );
+      if (complianceStatus.length === 0 ||
+        complianceStatus[0] !== complianceStatusFromDatabase) {
+        if (complianceStatusFromDatabase !== undefined &&
+          complianceStatusFromDatabase.status !== ComplianceStatus.COMPLIANT
```
dydxwill, line 282: when would it ever change from non-compliant to compliant?

affanv14: we're tracking states which are not COMPLIANT... there is a state where address has not status and becomes compliant which doesnt need to be tracked

dydxwill (approved):

affanv14:
@mergify backport https://github.com/dydxprotocol/v4-chain/tree/release/indexer/v5.x

affanv14:
https://github.com/Mergifyio backport release/indexer/v5.x

PR #2068, opened 2024-08-09
Full node streaming batch size reset to 2000, properly zero out cache on flush

jonfung-dydx:
- properly zero out the cache
- emit metrics on # streamUpdates
- reset max batch size default to 2000


jayy04 (approved):

protocol/streaming/full_node_streaming_manager.go
```diff
@@ -501,8 +503,8 @@ func (sm *FullNodeStreamingManagerImpl) FlushStreamUpdatesWithLock() {
 		}
 	}
 
-	clear(sm.streamUpdateCache)
-	clear(sm.streamUpdateSubscriptionCache)
+	sm.streamUpdateCache = nil
```
dydxwill, line 506: will this work? vs. setting to empty array

jonfung-dydx: yep, tested

PR #2074, opened 2024-08-12
[DO NOT MERGE] Merge `feature/cosmwasm` into `main` 

teddyding:
### Changelist

### Test Plan

### Author/Reviewer Checklist


chipshort (commented):
Not a full review, just some comments I had while looking at the code

protocol/app/ante.go
```diff
@@ -396,6 +404,9 @@ func (h *lockingAnteHandler) otherMsgAnteHandle(ctx sdk.Context, tx sdk.Tx, simu
 		defer h.globalLock.Unlock()
 	}
 
+	if ctx, err = h.executeCosmWasm.AnteHandle(ctx, tx, simulate, noOpAnteHandle); err != nil {
```
pinosu, line 407: you should also add wasmd ante handlers :

- `CountTXDecorator`  is useful because it provides position data to the contract via [Env](https://github.com/CosmWasm/wasmd/blob/f478697fce67983546f5729b04f5389161310a8b/x/wasm/types/types.go#L271) . The motivation is to add a unique ID to the contract environment to distinguish between multiple executions within the same block but in different tx versus a reentrant call within the same msg or tx.
- `LimitSimulationGasDecorator`  limits the maximum gas available in simulations, preventing the "infinite gas" issues.
So to avoid well known issues and keep the full compatibility both ante handlers should be provided.

protocol/app/module_accounts.go
```diff
@@ -35,6 +36,9 @@ var (
 		govtypes.ModuleName:            {authtypes.Burner},
 		ibctransfertypes.ModuleName:    {authtypes.Minter, authtypes.Burner},
 		icatypes.ModuleName:            nil,
+		// https://github.com/CosmWasm/wasmvm/blob/1c3fdc2a4402e527617ec72fe53f114b24899a01/types/msg.go#L116-L121
+		// required to burn given coins from the contract's account.
+		wasmtypes.ModuleName: {authtypes.Burner},
```
pinosu, line 41: 👍

pinosu (commented):
Added some comments

protocol/dydx-cosmwasm/src/query.rs
```diff
@@ -0,0 +1,48 @@
...
+#[serde(rename_all = "snake_case")]
+pub struct DydxQueryWrapper {
+    pub route: DydxRoute,
+    pub query_data: DydxQuery,
+}
+
+// implement custom query
+impl CustomQuery for DydxQueryWrapper {}
+
+/// SeiQuery is defines available query datas
```
uint, line 19:
```suggestion
/// SeiQuery defines available query datas
```

protocol/dydx-cosmwasm/src/serializable_int.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct SerializableInt {
+    i: BigInt,
+}
+
+impl SerializableInt {
+    pub fn new(i: BigInt) -> Self {
+        Self { i }
+    }
+    pub fn to_big_int(&self) -> &BigInt {
```
uint, line 16:
```suggestion
    pub fn as_big_int(&self) -> &BigInt {
```
This is more idiomatic for reference conversions. See [example](https://doc.rust-lang.org/std/string/struct.String.html#method.as_str)

protocol/contracts/dydx-messages-example/src/contract.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::error::ContractError;
+use crate::msg::{ArbiterResponse, ExecuteMsg, InstantiateMsg};
+use crate::state::{Config, CONFIG};
+use cw2::set_contract_version;
+use dydx_cosmwasm::DydxMsg;
+use dydx_cosmwasm::{DydxQuerier, DydxQueryWrapper, SubaccountId};
+use dydx_cosmwasm::DydxQuery;
+
+// version info for migration info
```
uint, line 16:
```suggestion
// version info for migration
```

protocol/contracts/dydx-messages-example/src/contract.rs
```diff
@@ -0,0 +1,54 @@
...
+) -> Result<Response<DydxMsg>, ContractError> {
+    match msg {
+        ExecuteMsg::DydxMsg(dydx_msg) => {
+            Ok(Response::new().add_message(dydx_msg))
+        }
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps<DydxQueryWrapper>, _env: Env, msg: DydxQuery) -> StdResult<Binary> {
```
uint, line 46: Is this contract meant as an example for contract devs to use as a reference? I worry accepting the `DydxQuery` type from the bindings lib would be confusing to them. It's more typical for a contract to define its own `QueryMsg`.

uint (commented):
I reviewed the Rust bits and they look fine save for nitpicks.

PR #2082, opened 2024-08-13
Enforce uniqueness of timestamp nonce

jerryfan01234:
### Changelist
Add check for uniqueness of ts nonce

### Test Plan
1. existing and new unit test passed
2. ran chain locally and performed following test
```
get current time
submit transaction with seq=above time
wait for block to be committed and reset cache by restarting container
submit prev transaction  
```
Second tx failed with error code 32 as intended (https://github.com/cosmos/cosmos-sdk/blob/main/types/errors/errors.go)

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Implemented a uniqueness check for timestamp nonces during updates, ensuring no duplicates are added.

- **Bug Fixes**
	- Enhanced rejection logic for non-unique timestamp nonces, preventing erroneous updates.

- **Tests**
	- Improved clarity and specificity of existing test cases.
	- Added tests to validate the rejection of duplicate timestamp nonces.
	- Introduced a new test for verifying slice membership with the `SliceContains` function.

### DIFF
--- a/protocol/x/accountplus/keeper/timestampnonce.go
+++ b/protocol/x/accountplus/keeper/timestampnonce.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
 	"github.com/dydxprotocol/v4-chain/protocol/x/accountplus/types"
 )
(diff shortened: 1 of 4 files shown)


protocol/x/accountplus/keeper/timestampnonce.go
```diff
@@ -115,3 +120,12 @@ func isLargerThanSmallestValue(value uint64, values []uint64) (bool, int) {
 		return false, minIndex
 	}
 }
+
+func contains[T comparable](list []T, value T) bool {
```
teddyding, line 124: Make function public so we can write separate unit tests.

Also since if we made this generalized, might as well move to [here](https://github.com/dydxprotocol/v4-chain/blob/18b7707a932cf9b06552a88178770e717e688965/protocol/lib/collections.go#L9)

roy-dydx (approved):

jerryfan01234:
@Mergify backport release/protocol/v6.x

PR #2093, opened 2024-08-15
Add new liquidity tiers to genesis

chenyaoy:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new and redefined liquidity tiers, including "Small-Cap" and "Isolated," enhancing the granularity of liquidity management.
	- Added multiple new liquidity tiers like "Mid-Cap" and "FX" with specific parameters for better trading strategy support.

- **Bug Fixes**
	- Improved the handling of liquidity tier parameters, ensuring better consistency and accuracy in the management of financial metrics.

### DIFF
--- a/protocol/testing/genesis.sh
+++ b/protocol/testing/genesis.sh
@@ -196,9 +196,9 @@ function edit_genesis() {
 	dasel put -t json -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[]' -v "{}"
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].id' -v '6'
 	dasel put -t string -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].name' -v 'FX'
-	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].initial_margin_ppm' -v '1000' # 0.1%
+	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].initial_margin_ppm' -v '10000' # 1%
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].maintenance_fraction_ppm' -v '500000' # 50% of IM
-	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].impact_notional' -v '2500000000' # 2_500 USDC (2.5 USDC / 0.1%)
+	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].impact_notional' -v '2500000000' # 2_500 USDC (25 USDC / 1%)
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].open_interest_lower_cap' -v '500000000000' # 500k USDC
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[6].open_interest_upper_cap' -v '1000000000000' # 1 million USDC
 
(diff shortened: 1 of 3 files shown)


protocol/testing/genesis.sh
```diff
@@ -137,18 +137,16 @@ function edit_genesis() {
 	dasel put -t string -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].name' -v 'Large-Cap'
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].initial_margin_ppm' -v '50000' # 5%
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].maintenance_fraction_ppm' -v '600000' # 60% of IM
-	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].base_position_notional' -v '1000000000000' # 1_000_000 USDC
```
teddyding, line 140: Why did we remove this?

protocol/testing/genesis.sh
```diff
@@ -137,18 +137,16 @@ function edit_genesis() {
...
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].impact_notional' -v '10000000000' # 10_000 USDC (500 USDC / 5%)
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].open_interest_lower_cap' -v '0' # OIMF doesn't apply to Large-Cap
 	dasel put -t int -f "$GENESIS" '.app_state.perpetuals.liquidity_tiers.[0].open_interest_upper_cap' -v '0' # OIMF doesn't apply to Large-Cap
 
-	# Liquidity Tier: Mid-Cap
+	# Liquidity Tier: Small-Cap
```
teddyding, line 144: We seem to follow conventions that the liquidity tier list goes from Largest cap -> long tail. Can we keep Mid-cap at `1` instead of `5` below?

chenyaoy: The params for `1` aren't changing but wanted to rename to `small-cap` and create a new `mid-cap` tier so that markets currently on tier 1 aren't affected. But your point makes sense, we'd just have to put the following messages in a single proposal:
1. Rename liquidity tier 1 to mid-cap and change params, add small-cap as tier 5
2. Update all perpetuals with liquidity tier 1 to liquidity tier 5

@yujin512 what do you think?

yujin512: Agreed the ordering according to IMF makes sense! But if we follow the logic thoroughly, the ordering would be

- FX
- Large Cap
- Mid cap (new)
- Isolated
- Small cap (previously Mid cap)
- Long-tail
- Safety.

which means updating all perpetuals to be in the correct tier. Also wondering if it would be worth reordering every time we want to add new tiers, which will likely be in the direction of increasing leverage and therefore shifting the rest down.

teddyding: Ok agree that we should avoid changing the existing perpetual objects for simplicity/reduce room for error.

For my understanding, if a mainnet market is currently `Mid-cap`, do we want them to be `Mid-cap` or `Small-cap` after the change? Is it not the former?

chenyaoy: Current Mid-cap markets will now be renamed to Small-cap, with the params as before. A new Mid-cap liquidity tier will be created with a different set of param values (e.g. 20x leverage instead of 10x of the old mid-cap tier). See here: https://www.notion.so/dydx/Research-liquidity-recommendations-d3b96d1fde374158841490b93c32de48?pvs=4#bf9047f7f06e487687d879a0f290e60c

yujin512: Currently mid-cap markets should have the same param values as now. Only renaming because the new liquidity tier (ex: SOL) makes sense to be between large cap and small cap!

yujin512 (approved):
lgtm

teddyding (approved):

PR #2124, opened 2024-08-21
skip liquidation task loop if last committed block height is the same

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling in test cases for more precise validation of expected errors.
  
- **New Features**
	- Improved efficiency of the liquidation daemon task loop by preventing unnecessary processing when no new blocks are committed.
	- Added state management capabilities to track the last processed block height within the task loop.

### DIFF
--- a/protocol/daemons/liquidation/client/sub_task_runner.go
+++ b/protocol/daemons/liquidation/client/sub_task_runner.go
@@ -26,7 +26,9 @@ type SubTaskRunner interface {
 	) error
 }
 
-type SubTaskRunnerImpl struct{}
+type SubTaskRunnerImpl struct {
+	lastLoopBlockHeight uint32
+}
 
 // Ensure SubTaskRunnerImpl implements the SubTaskRunner interface.
 var _ SubTaskRunner = (*SubTaskRunnerImpl)(nil)
@@ -48,6 +50,14 @@ func (s *SubTaskRunnerImpl) RunLiquidationDaemonTaskLoop(
 	lastCommittedBlockHeight, err := daemonClient.GetPreviousBlockInfo(ctx)
 	if err != nil {
 		return err
+	}
+
+	if lastCommittedBlockHeight == s.lastLoopBlockHeight {
+		daemonClient.logger.Info(
+			"Skipping liquidation daemon task loop as no new block has been committed",
+			"blockHeight", lastCommittedBlockHeight,
+		)
+		return nil
 	}
 
 	// 1. Fetch all information needed to calculate total net collateral and margin requirements.
--- a/protocol/daemons/liquidation/client/sub_task_runner.go
+++ b/protocol/daemons/liquidation/client/sub_task_runner.go
@@ -60,6 +60,8 @@ func (s *SubTaskRunnerImpl) RunLiquidationDaemonTaskLoop(
 		return nil
 	}
 
+	s.lastLoopBlockHeight = lastCommittedBlockHeight
+
 	// 1. Fetch all information needed to calculate total net collateral and margin requirements.
 	subaccounts,
 		perpInfos,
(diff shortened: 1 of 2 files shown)


protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -50,6 +52,16 @@ func (s *SubTaskRunnerImpl) RunLiquidationDaemonTaskLoop(
...
 
+	if lastCommittedBlockHeight == s.lastLoopBlockHeight {
+		daemonClient.logger.Info(
+			"Skipping liquidation daemon task loop as no new block has been committed",
+			"blockHeight", lastCommittedBlockHeight,
+		)
+		return nil
+	}
+
+	s.lastLoopBlockHeight = lastCommittedBlockHeight
```
teddyding, line 63: Playing devil's advocate: are we sure this is thread-safe? Could two liquidation daemon loop overlap?

jayy04: no daemon loops don't overlap since we are not spawning new go routines 
```
for {
		select {
		case <-ticker.C:
			if err := s.RunLiquidationDaemonTaskLoop(
				ctx,
				client,
				flags.Liquidation,
			); err != nil {
				// TODO(DEC-947): Move daemon shutdown to application.
				client.logger.Error("Liquidations daemon returned error", "error", err)
				client.ReportFailure(err)
			} else {
				client.ReportSuccess()
			}
		case <-stop:
			return
		}
	}
```

protocol/daemons/liquidation/client/sub_task_runner.go
```diff
@@ -50,6 +52,16 @@ func (s *SubTaskRunnerImpl) RunLiquidationDaemonTaskLoop(
 		return err
 	}
 
+	if lastCommittedBlockHeight == s.lastLoopBlockHeight {
```
teddyding, line 55: IIUC if application restarts, `lastLoopBlockHeight` is by default `0` and we always continue below logic. If so could we document?

protocol/daemons/liquidation/client/sub_task_runner.go
teddyding, line 63: Nit: I wonder if here's the best place to assign `s.lastLoopBlockHeight = lastCommittedBlockHeight` since we haven't sent any current block info to application yet. Wdyt think about deferring the statement `s.lastLoopBlockHeight = lastCommittedBlockHeight`?

jayy04: good question. thought about it a little bit more 

putting it here strictly guarantees that loop doesn't run for the same block height, regardless of errors. If we defer the statement to the end (e.g. after `SendLiquidatableSubaccountIds`), theoretically it's possible that `SendLiquidatableSubaccountIds` only partially succeeds when first few requests goes through and later requests fail. In this case, we would be running into the same issue. I think might be slightly better here for correctness?

teddyding: `defer`ing guarantees that the `s.lastLoopBlockHeight = lastCommittedBlockHeight` is always run before return, even if `SendLiquidatableSubaccountIds` fails right? Trying to understand when we'd send `SendLiquidatableSubaccountIds`again for the same block. 

No strong preference though, the current implementation also works. Up to you. Approved to unblock

teddyding (approved):

protocol/daemons/liquidation/client/sub_task_runner.go
jayy04: ohh you mean `defer` and update block height regardless of error? I misunderstood

yeah I think it's basically the same - let's keep it as is to unstuck the testnet first

jayy04:
@Mergifyio backport release/protocol/v5.2.x

jayy04:
@Mergifyio backport release/protocol/v6.x

PR #2131, opened 2024-08-21
Update compliance roundtable batch size and add stale compliance metrics

jerryfan01234:
### Changelist
Changed default compliance query batch size to match elliptic rate limit
Added metrics to track number of addresses with stale compliance update

### Test Plan
Flag has already been used successfully in imperator mainnet with the default value in PR
Metrics pass unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new metrics to track active and inactive addresses with stale compliance data, enhancing compliance monitoring capabilities.
	- Improved the logic for managing address queries based on remaining query limits.

- **Configuration Changes**
	- Updated the compliance provider query batch size from 100 to 15, enforcing stricter rate limiting for compliance data queries.

- **Bug Fixes**
	- Enhanced the tracking and processing of compliance data, ensuring more accurate metrics and better control flow in the update process.


indexer/services/roundtable/src/tasks/update-compliance-data.ts
```diff
@@ -165,6 +176,12 @@ export default async function runTask(
       undefined,
       { provider: complianceProvider.provider },
     );
+    stats.gauge(
+      `${config.SERVICE_NAME}.${taskName}.num_inactive_addresses_with_stale_compliance`,
+      inactiveAddressesWithStaleCompliance,
+      undefined,
```
Christopher-Li, line 182: what is this undefined?

jerryfan01234: it is the sample rate. For the other count metrics, we used undefined so i kept it the same for this count metric.

Christopher-Li: Let's use `STATS_NO_SAMPLING` this constant

indexer/services/roundtable/src/config.ts
```diff
@@ -177,7 +177,8 @@ export const configSchema = {
   MAX_COMPLIANCE_DATA_AGE_SECONDS: parseInteger({ default: 2_630_000 }), // 1 month
   MAX_ACTIVE_COMPLIANCE_DATA_AGE_SECONDS: parseInteger({ default: 86_400 }), // 1 day
   MAX_COMPLIANCE_DATA_QUERY_PER_LOOP: parseInteger({ default: 100 }),
-  COMPLIANCE_PROVIDER_QUERY_BATCH_SIZE: parseInteger({ default: 100 }),
+  // v2/wallet/synchronous rate limit is 15/s https://developers.elliptic.co/docs/configuration
+  COMPLIANCE_PROVIDER_QUERY_BATCH_SIZE: parseInteger({ default: 15 }),
```
vincentwschau, line 181: Will this be enough to ensure all the addresses in the table are checked? Should the frequency of the loop be increased if not?

jerryfan01234: The total number of addresses attempted per roundtable runs is unaffected. Currently ~60 requests are 200 because we still have ~40 invalid accounts. At the rate of ~60 per 5min, we should be catching up by ~15k a day, which is what i have observed. 
If the number of active accounts skyrockets or the number of invalid addresses increases to near 100, we will start to accumulate stale rows. The new metrics (will add alarming soon) will catch if this happens.

Christopher-Li (approved):

jerryfan01234:
@Mergifyio backport release/indexer/v6.x

PR #2151, opened 2024-08-23
[TRA-530] When changing market pair name, validate that new pair exists in the market map

hwray:
### Changelist
* When changing market pair name, validate that new pair exists in the market map

### Test Plan
* Updated tests in `market_param_test.go`:
   * Changed "valid" tests to enforce that new pair ticker exists in MarketMap
   * Added new "error" test to ensure the correct error is fired upon failure

### Author/Reviewer Checklist
- [] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [] If the PR has breaking postgres changes to the indexer add the `indexer-postgres-breaking` label.
- [] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for modifying market parameters, ensuring only valid market parameters are processed.
	- Improved error handling when updating market parameters related to tickers.

- **Bug Fixes**
	- Expanded test cases for error handling in market parameter modifications to cover scenarios where the ticker is absent in the MarketMap.

- **Tests**
	- Updated test cases to ensure robustness in the modification and error handling of market parameters, including new scenarios for invalid pairs.

### DIFF
--- a/protocol/x/prices/keeper/market_param.go
+++ b/protocol/x/prices/keeper/market_param.go
@@ -50,10 +50,7 @@ func (k Keeper) ModifyMarketParam(
 		}
 	}
 	if _, err := k.MarketMapKeeper.GetMarket(ctx, updatedMarketParam.Pair); err != nil {
-		return types.MarketParam{}, errorsmod.Wrapf(
-			types.ErrTickerNotFoundInMarketMap,
-			updatedMarketParam.Pair,
-		)
+		return types.MarketParam{}, errorsmod.Wrapf(types.ErrTickerNotFoundInMarketMap, updatedMarketParam.Pair)
 	}
 
 	// Store the modified market param.
--- a/protocol/x/prices/keeper/market_param.go
+++ b/protocol/x/prices/keeper/market_param.go
@@ -49,8 +49,11 @@ func (k Keeper) ModifyMarketParam(
 			return types.MarketParam{}, errorsmod.Wrapf(types.ErrMarketParamPairAlreadyExists, updatedMarketParam.Pair)
 		}
 	}
-	if _, err := k.GetMarketPrice(ctx, updatedMarketParam.Id); err != nil {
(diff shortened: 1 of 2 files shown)


protocol/x/prices/keeper/market_param.go
```diff
@@ -50,6 +50,15 @@ func (k Keeper) ModifyMarketParam(
 		}
 	}
 
+	// Validate that modified market param has a corresponding ticker in MarketMap
+	cp, err := slinky.MarketPairToCurrencyPair(updatedMarketParam.Pair)
+	if err != nil {
```
vincentwschau, line 55: nit: Is this error case already covered by a separate test / un-reachable due to validation done before hand on the pair?

hwray: Looks like the only caller of `ModifyMarketParam` is `UpdateMarketParam` in the `prices` keeper. That function doesn't perform this check, and none of its callers do either AFAICT (thought it starts getting tough to trace through the RPC calls). 

I also ran a quick search on `MarketPairToCurrencyPair` func and didn't see references to it in related / surrounding code paths. 

Just noticed we actually check the same condition on line 73 of this file: https://github.com/dydxprotocol/v4-chain/pull/2151/files#diff-b5fe4e18251360d63261212057a8fd1fb2520da6bbc27eae56c709c08a6bf0abR73

Which suggests this check is necessary but probably can be de-duped with the existing one

vincentwschau (approved):

chenyaoy (approved):

PR #2169, opened 2024-08-28
add share unlocking mechanism to megavault

tqin7:
### Changelist
- add msg and keeper method on shares unlocking
- delay a MsgUnlockShares when locking shares

### Test Plan
unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for unlocking shares, allowing users to manage their shares based on transaction block height.
	- Added support for delayed message processing to enhance message handling capabilities within the vault module.
	- Expanded internal message processing to recognize the new `MsgUnlockShares` type.
	- Implemented comprehensive testing for share locking and unlocking functionalities, improving reliability.

- **Bug Fixes**
	- Enhanced validation checks during the unlocking process to ensure integrity.


protocol/x/vault/keeper/shares.go
```diff
@@ -2,11 +2,13 @@ package keeper
```
vincentwschau, line 2: Is it possible to use the E2E testing framework to add an E2E test of the flow of:
- lock shares til block X
- advance to block X
- check that shares are unlocked

protocol/x/vault/keeper/msg_server_unlock_shares_test.go
```diff
@@ -0,0 +1,92 @@
...
+					},
+				},
+			}
+			if tc.msg.OwnerAddress != "" {
+				err := k.SetOwnerShareUnlocks(ctx, tc.msg.OwnerAddress, ownerShareUnlocks)
+				require.NoError(t, err)
+			}
+
+			res, err := ms.UnlockShares(ctx, tc.msg)
+			newLockedShares, exists := k.GetOwnerShareUnlocks(ctx, tc.msg.OwnerAddress)
```
vincentwschau, line 78: nit: Check `newLockedShares` is what's expected as well.

protocol/x/vault/keeper/shares.go
tqin7: yeah this will be possible once there's a tx that calls `LockShares`. I was thinking once PML is integrated, we can add this e2e test. (before we have a tx that calls `LockShares`, e2e won't work)

tqin7: added a task [here](https://linear.app/dydx/issue/TRA-585/add-e2e-test-on-shares-locking-once-pml-is-integrated)

vincentwschau: Are you sure there's a need for a tx that calls `LockShares`? Isn't there an instance of the `VaultsKeeper` that's accessible on the test app in the E2E test framework?

tqin7: we can call `VaultsKeeper.LockShares` keeper method but i don't think the state changes from that call persists once you advance to next block (as it's not part of a tx)

vincentwschau (approved):

protocol/x/vault/keeper/shares.go
vincentwschau: Good point, possibly an E2E test with a delayed message w/ unlock shares could be useful (just to check authorities/everything else works) but not necessary.

PR #2174, opened 2024-08-29
[OTE-729] Affiliates update fills table

jerryfan01234:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new property, `affiliateEarnedRevShare`, across various data structures and API responses to enhance tracking of affiliate revenue sharing.
	- Updated API documentation and schemas to include `affiliateEarnedRevShare`, providing users with detailed revenue-sharing information.

- **Bug Fixes**
	- Improved test coverage by incorporating `affiliateEarnedRevShare` in relevant test cases, ensuring accurate validation of functionality.


Christopher-Li (commented):
Three things:
1. I think this affects websocket documentation. When fills occur, I believe the fill object is forwarded in order-handler and/or liquidation-handler to kafka for websockets. If thats the case, could you update websocket-documentation.md?
2. I understand we're adding this column and index to the `fills` table which is our largest table. Could you test this in testnet? Basically just run `\timing` in postgres then run the generated commands to add a column and add an index and see how long it takes? I'm a little concerned with adding an index in this way, because we usually add an index like [this](https://github.com/dydxprotocol/v4-chain/blob/main/indexer/packages/postgres/src/db/migrations/migration_files/20240717142808_pnl_ticks_subaccount_id_index.ts#L13). `CONCURRENTLY` matters so that the table isn't locked and so we dont cause an outage while adding the index.
3. Is there a corresponding ticket to this PR?

jerryfan01234:
> Three things:
> 
> 1. I think this affects websocket documentation. When fills occur, I believe the fill object is forwarded in order-handler and/or liquidation-handler to kafka for websockets. If thats the case, could you update websocket-documentation.md?
> 2. I understand we're adding this column and index to the `fills` table which is our largest table. Could you test this in testnet? Basically just run `\timing` in postgres then run the generated commands to add a column and add an index and see how long it takes? I'm a little concerned with adding an index in this way, because we usually add an index like [this](https://github.com/dydxprotocol/v4-chain/blob/main/indexer/packages/postgres/src/db/migrations/migration_files/20240717142808_pnl_ticks_subaccount_id_index.ts#L13). `CONCURRENTLY` matters so that the table isn't locked and so we dont cause an outage while adding the index.
> 3. Is there a corresponding ticket to this PR?

1. Will check websocket and update documentation as needed
2. Good point, I will test.
3. Changed PR title to include ticket

Christopher-Li (approved):

PR #2233, opened 2024-09-10
[OTE-757] add update affiliate info roundtable

jerryfan01234:
### Changelist
Add roundtable task to update affiliate info
Minor modifications to dbs via migrations
Minor style changes to total volume roundtable

### Test Plan
local tests passed

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new asynchronous function for updating affiliate information based on a specified time window.
	- Created a comprehensive test suite for the `update-affiliate-info` task to validate various update scenarios.
	- Enhanced the functionality of the wallet total volume update process.
	- Added configuration options for controlling affiliate information updates.
	- Implemented a new task to periodically update affiliate information based on configuration settings.
	- Enhanced the testing framework for the affiliate information store, improving robustness and coverage.

- **Bug Fixes**
	- Improved the persistent caching mechanism for affiliate information updates.

- **Documentation**
	- Enhanced test coverage for affiliate information updates and wallet total volume functionalities.

- **Chores**
	- Refactored test cases for better organization and clarity, including the addition of new helper functions.


indexer/packages/postgres/src/db/migrations/migration_files/20240913142157_change_affiliate_info_decimal_precision_and_add_total_referred_volume.ts
```diff
@@ -1,22 +0,0 @@
...
-
-export async function down(knex: Knex): Promise<void> {
-  return knex.schema.alterTable('affiliate_info', (table) => {
-    table.decimal('affiliateEarnings').alter();
-    table.decimal('totalReferredFees').alter();
-    table.decimal('referredNetProtocolEarnings').alter();
-
-    table.dropColumn('referredTotalVolume');
-  });
-}
```
affanv14, line 22: Would deleting a migration cause issues with the database here?

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,245 @@
...
+      referredMakerTrades: 0,
+      referredTakerTrades: 1,
+      totalReferredFees: '1000',
+      totalReferredUsers: 2,
+      referredNetProtocolEarnings: '500',
+      firstReferralBlockHeight: '1',
+      referredTotalVolume: '1',
+    };
+    expect(updatedInfo).toEqual(expectedAffiliateInfo);
+    const lastUpdateTime2 = await getAffiliateInfoUpdateTime();
```
affanv14, line 112: nit: typing

affanv14: ANd below as well

indexer/packages/postgres/src/db/migrations/migration_files/20240913142157_change_affiliate_info_decimal_precision_and_add_total_referred_volume.ts
jerryfan01234: The newly added migration and the deleted one are almost the same, with compatible up/downs. For better comptibility, let me modify the file in place instead of adding/deleting.

indexer/packages/postgres/src/db/migrations/migration_files/20240913142157_change_affiliate_info_decimal_precision_and_add_total_referred_volume.ts
```diff
@@ -1,21 +1,22 @@
...
 }
 
 export async function down(knex: Knex): Promise<void> {
   return knex.schema.alterTable('affiliate_info', (table) => {
-    table.decimal('affiliateEarnings').alter();
-    table.decimal('totalReferredFees').alter();
-    table.decimal('referredNetProtocolEarnings').alter();
+    table.decimal('affiliateEarnings').notNullable().defaultTo(0).alter();
+    table.decimal('totalReferredFees').notNullable().defaultTo(0).alter();
+    table.decimal('referredNetProtocolEarnings').notNullable().defaultTo(0).alter();
```
Christopher-Li, line 19: Lets try not to make this a habit, it's better to create a new migration to delete the rows above and change them to notNullable and default to 0. Especially since the cutoff for the release is today

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  });
+
+  afterEach(async () => {
+    await dbHelpers.clearData();
+  });
+
+  it('Successfully updates affiliate info and persistent cache multiple times', async () => {
+    const startDt = DateTime.utc();
+
+    // Set persistent cache affiliateInfoUpdateTIme so task does not use backfill windows
```
Christopher-Li, line 40:
```suggestion
    // Set persistent cache affiliateInfoUpdateTime so task does not use backfill windows
```

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,245 @@
...
+      value: startDt.toISO(),
+    });
+
+    // First task run: add refereal w/o any fills
+    // defaultWallet2 will be affiliate and defaultWallet will be referee
+    await AffiliateReferredUsersTable.create({
+      affiliateAddress: testConstants.defaultWallet2.address,
+      refereeAddress: testConstants.defaultWallet.address,
+      referredAtBlock: '1',
+    });
```
Christopher-Li, line 52: await Promise.all

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,245 @@
...
+      refereeAddress: testConstants.defaultWallet3.address,
+      referredAtBlock: '2',
+    });
+
+    await affiliateInfoUpdateTask();
+
+    updatedInfo = await AffiliateInfoTable.findById(
+      testConstants.defaultWallet2.address,
+    );
+    expectedAffiliateInfo = {
```
Christopher-Li, line 100: better to use a second variable, and keep updatedInfo and expectedAffiliateInfo as `const`

indexer/services/roundtable/__tests__/tasks/update-wallet-total-volume.test.ts
```diff
@@ -213,9 +189,11 @@ describe('update-wallet-total-volume', () => {
 });
 
 async function getTotalVolumeUpdateTime(): Promise<DateTime | undefined> {
-  const persistentCache = await PersistentCacheTable.findById('totalVolumeUpdateTime');
-  const lastUpdateTime1 = persistentCache?.value
+  const persistentCache = await PersistentCacheTable.findById(
+    PersistentCacheKeys.TOTAL_VOLUME_UPDATE_TIME,
+  );
+  const lastUpdateTime = persistentCache?.value
```
Christopher-Li, line 195: nit: typing

indexer/packages/postgres/src/stores/wallet-table.ts
```diff
@@ -123,7 +124,7 @@ export async function updateTotalVolume(
   windowEndTs: string,
 ) : Promise<void> {
 
-  await knexReadReplica.getConnection().raw(
+  await knexPrimary.raw(
```
Christopher-Li, line 127: after reviewing some of this, I'm being more convinced that we should fix the nits in the previous PRs for three reasons:
- As I'm reviewing a PR to `add update affiliate info into roundtable`, there are a bunch of unrelated changes that are confusing as a reader
- Reviewing this PR now requires context from multiple previous PRs, and is impossible to track.

I would prefer that a change like this is completely separated from affiliate info roundtable. This is real business logic changes that if caused a bug would be difficult to track when reviewing PR history. Ie why would adding updates to affiliate info roundtable change the behavior of updateTotalVolume?

indexer/services/roundtable/src/tasks/update-wallet-total-volume.ts
```diff
@@ -1,50 +1,48 @@
...
+    if (windowEndTime > lastUpdateTime.plus({ days: 1 })) {
+      windowEndTime = lastUpdateTime.plus({ days: 1 });
     }
 
-    await WalletTable.updateTotalVolume(lastUpdateTime.toISO(), currentTime.toISO());
-    await PersistentCacheTable.upsert({
-      key: persistentCacheKey,
-      value: currentTime.toISO(),
-    });
+    await WalletTable.updateTotalVolume(lastUpdateTime.toISO(), windowEndTime.toISO());
```
Christopher-Li, line 37: same as above, I would never expect this change to exist in `add update affiliate info roundtable`, please move into  a separate PR

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+export default async function runTask(): Promise<void> {
+  try {
+    const start = Date.now();
+    const persistentCacheEntry: PersistentCacheFromDatabase | undefined = await PersistentCacheTable
+      .findById(PersistentCacheKeys.AFFILIATE_INFO_UPDATE_TIME);
+
+    if (!persistentCacheEntry) {
+      logger.info({
+        at: 'update-affiliate-info#runTask',
+        message: `No previous ${PersistentCacheKeys.AFFILIATE_INFO_UPDATE_TIME} found in persistent cache table. Will use default value: ${defaultLastUpdateTime}`,
```
Christopher-Li, line 23: this is different from `update-wallet-total-volume`, because the first instance of fills is 10/26/23, but since this task is updating affiliate data for affiliate addresses, you can just use 9/19/23, because there will never be any affiliate data before today

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+import { DateTime } from 'luxon';
+
+import config from '../config';
+
+const defaultLastUpdateTime: string = '2023-10-26T00:00:00Z';
+
+/**
+ * Update the affiliate info for all affiliate addresses.
+ */
+export default async function runTask(): Promise<void> {
```
Christopher-Li, line 14: don't you have to update index.ts so this task is run?

indexer/packages/postgres/src/stores/affiliate-info-table.ts
```diff
@@ -94,6 +97,165 @@ export async function findById(
     .returning('*');
 }
 
+export async function updateInfo(
```
Christopher-Li, line 100: can you add a jsdoc explaining what this function does? Also I know you merged the logic to update persistentcache in this function just like in updateWalletVolume, but it might be better to separate it. When I read updateInfo, I don't expect that the persistent cache is being updated.

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+      ? persistentCacheEntry.value
+      : defaultLastUpdateTime);
+    let windowEndTime = DateTime.utc();
+
+    // During backfilling, we process one day at a time to reduce roundtable runtime.
+    if (windowEndTime > lastUpdateTime.plus({ days: 1 })) {
+      windowEndTime = lastUpdateTime.plus({ days: 1 });
+    }
+
+    await AffiliateInfoTable.updateInfo(lastUpdateTime.toISO(), windowEndTime.toISO());
```
Christopher-Li, line 37: Wrap these events in a transaction. There is a race condition:

1. current task reads persistent cache
2. another task overwrites persistent cache
3. current task updates affiliateInfoTable and persistent cache not knowing that the `lastUpdateTime` is stale

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    stats.timing(
```
Christopher-Li, line 39: nit: already tracked here: https://github.com/dydxprotocol/v4-chain/blob/main/indexer/services/roundtable/src/helpers/loops-helper.ts#L176

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+      });
+    }
+
+    const lastUpdateTime: DateTime = DateTime.fromISO(persistentCacheEntry
+      ? persistentCacheEntry.value
+      : defaultLastUpdateTime);
+    let windowEndTime = DateTime.utc();
+
+    // During backfilling, we process one day at a time to reduce roundtable runtime.
+    if (windowEndTime > lastUpdateTime.plus({ days: 1 })) {
```
Christopher-Li, line 33: this is a pretty expensive operation, I'm concerned that if you processed a day of data it could cause issues, I would recommend starting at 1 hour

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,50 @@
...
+      logger.info({
+        at: 'update-affiliate-info#runTask',
+        message: `No previous ${PersistentCacheKeys.AFFILIATE_INFO_UPDATE_TIME} found in persistent cache table. Will use default value: ${defaultLastUpdateTime}`,
+      });
+    }
+
+    const lastUpdateTime: DateTime = DateTime.fromISO(persistentCacheEntry
+      ? persistentCacheEntry.value
+      : defaultLastUpdateTime);
+    let windowEndTime = DateTime.utc();
```
Christopher-Li, line 30: Upper bound should not be at DateTime.utc(), it should be set at the timestamp of the latest block height, because if the indexer is stalled, we can only update affiliate info up to the latest block time. See [here](https://github.com/dydxprotocol/v4-chain/blob/main/indexer/services/roundtable/src/tasks/aggregate-trading-rewards.ts#L153) as an example, you should also change update-wallet-total-volume to match the logic

indexer/packages/postgres/src/stores/affiliate-info-table.ts
```diff
@@ -94,6 +97,165 @@ export async function findById(
...
+-- Step 5: Upsert new affiliateInfoLastUpdateTime to persistent_cache table
+INSERT INTO persistent_cache (key, value)
+VALUES ('${PersistentCacheKeys.AFFILIATE_INFO_UPDATE_TIME}', '${windowEndTs}')
+ON CONFLICT (key) 
+DO UPDATE SET value = EXCLUDED.value;
+
+COMMIT;
+    `,
+  );
+}
```
Christopher-Li, line 258: Do you have any idea of the runtime of this SQL statement? I understand that half of the tables do not exist, but you should be able to test the speed of the fills table statements and joins to ensure that the runtime is acceptable. At the very least, every field you are filtering on should have an index, because these are big tables

indexer/packages/postgres/src/db/migrations/migration_files/20240913142157_change_affiliate_info_decimal_precision_and_add_total_referred_volume.ts
Christopher-Li: deleting, and modifying migrations are both breaking changes, its always preferred to add a new migration

indexer/packages/postgres/src/db/migrations/migration_files/20240913142157_change_affiliate_info_decimal_precision_and_add_total_referred_volume.ts
jerryfan01234: Ok let me just make a new migration this time around as well then

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: fixed

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: changed

indexer/packages/postgres/src/stores/affiliate-info-table.ts
jerryfan01234: This was tested during wallet total update on imperator db. For a day it is <1s.

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: Fixed

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: This was tested during wallet total update on imperator db. For a day it is <1s.

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: Removed

indexer/services/roundtable/src/tasks/update-wallet-total-volume.ts
jerryfan01234: same as above

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: Good point. To resolve this race condition, I could modify the query in AffiliateInfoTable.updateInfo to:
1. get cache time and lock row for update
2. get block time
3. compare block height to cache time
4. current logic for updating and setting new cache time

indexer/packages/postgres/src/stores/wallet-table.ts
jerryfan01234: But im ok with just addressing everything in one PR as well

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
Christopher-Li: just using a transaction is sufficient. Getting block time doesnt even need to be included in the race conditions

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,63 @@
...
+export default async function runTask(): Promise<void> {
+  const latestBlock: BlockFromDatabase = await BlockTable.getLatest();
+  if (latestBlock.time === null) {
+    throw Error('Failed to get latest block time');
+  }
+
+  // Wrap getting cache, updating info, and setting cache in one transaction with row locking to
+  // prevent race condition on persistent cache rows between read and write.
+  const txId: number = await Transaction.start();
+  await Transaction.setIsolationLevel(txId, IsolationLevel.REPEATABLE_READ);
```
jerryfan01234, line 28: This is the minimum isolation level to ensure that the row we read from persistent cache cannot be modified during the transaction. The side effect is that it also applies to the update query as well, which would lock all the rows in subaccounts belonging to affiliates. 

The entire roundtable should be fairly quick, but I still dont feel great about locking subaccounts. Thoughts?

An alternative solution is to explicity set the locking level in the query via
SELECT ... FOR UPDATE
but this would require more code change

indexer/packages/postgres/__tests__/stores/affiliate-info-table.test.ts
```diff
@@ -80,7 +260,7 @@ describe('Affiliate info store', () => {
     });
 
     it('Successfully filters by address', async () => {
-      const infos: AffiliateInfoFromDatabase[] = await AffiliateInfoTable
+      const infos: AffiliateInfoFromDatabase[] | undefined = await AffiliateInfoTable
```
Christopher-Li, line 263: why can this be undefined now?

indexer/packages/postgres/__tests__/stores/affiliate-info-table.test.ts
```diff
@@ -175,3 +355,66 @@ describe('Affiliate info store', () => {
...
+    ...defaultFill,
+    liquidity: Liquidity.MAKER,
+    subaccountId: defaultOrder.subaccountId,
+    createdAt: referenceDt.minus({ minutes: 2 }).toISO(),
+    eventId: defaultTendermintEventId4,
+    price: '1',
+    size: '1',
+    fee: '1000',
+    affiliateRevShare: '500',
+  });
```
Christopher-Li, line 417: await promise.all for the above fills

indexer/packages/postgres/__tests__/stores/wallet-table.test.ts
```diff
@@ -90,48 +90,68 @@ describe('Wallet store', () => {
...
-  it('Successfully updates totalVolumeUpdateTime in persistent cache', async () => {
-    const leftBound = DateTime.utc().minus({ hours: 1 });
-    const rightBound = DateTime.utc();
-    await WalletTable.updateTotalVolume(leftBound.toISO(), rightBound.toISO());
-
-    const persistentCache = await PersistentCacheTable.findById('totalVolumeUpdateTime');
-    const lastUpdateTime = persistentCache?.value
-      ? DateTime.fromISO(persistentCache.value)
-      : undefined;
+  describe('Wallet .updateTotalVolume()', () => {
```
Christopher-Li, line 93: nit: since top level already denotes that this is for wallet, you can just do 
```suggestion
  describe('updateTotalVolume', () => {
```

indexer/packages/postgres/__tests__/stores/wallet-table.test.ts
```diff
@@ -90,48 +90,68 @@ describe('Wallet store', () => {
...
+      // Updates persistent cache value
+      leftBound = referenceDt.minus({ hours: 1 });
+      rightBound = referenceDt;
+
+      await WalletTable.updateTotalVolume(leftBound.toISO(), rightBound.toISO());
+
+      persistentCache = await PersistentCacheTable.findById(
+        PersistentCacheKeys.TOTAL_VOLUME_UPDATE_TIME,
+      );
+      lastUpdateTime = persistentCache?.value;
```
Christopher-Li, line 149: The preference in general is to use const and keep objects immutable. Constantly overwriting them is an antipattern and can be confusing. When possible please use `const` instead of `let` even in tests

indexer/packages/postgres/src/db/migrations/migration_files/20240919142157_change_affiliate_info_default_value.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
```
Christopher-Li, line 3: 🔥 thanks Jerry!

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,276 @@
...
+      size: '1',
+      fee: '1000',
+      affiliateRevShare: '500',
+    });
+
+    // Create block at current time
+    await BlockTable.create({
+      blockHeight: '3',
+      time: DateTime.utc().toISO(),
+    });
```
Christopher-Li, line 177: await Promise.all

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    // Simulate backfill
+    let backfillTime: DateTime | undefined = await getAffiliateInfoUpdateTime();
```
Christopher-Li, line 180: same feedback as above with the let

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
Christopher-Li: I thought about it more, I think we actually dont need the transaciton. Roundtable has logic to prevent mutliple of the same roundtable task from running at once and theres only one task editting this data right now. If we have two, then we'll have to worry about race conditions

jerryfan01234: In that case, I will keep it in a transaction to enable rolling back in case any of the operations fail, but I will not set any isolation level.

indexer/packages/postgres/__tests__/stores/affiliate-info-table.test.ts
jerryfan01234: probably missed a merge conflict from previous PR

indexer/services/roundtable/__tests__/tasks/update-affiliate-info.test.ts
jerryfan01234: But with this one i actually want to update the value as part of the while loop control.

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,60 @@
...
+      ? persistentCacheEntry.value
+      : defaultLastUpdateTime);
+
+    let windowEndTime = DateTime.fromISO(latestBlock.time);
+    // During backfilling, we process one day at a time to reduce roundtable runtime.
+    if (windowEndTime > windowStartTime.plus({ days: 1 })) {
+      windowEndTime = windowStartTime.plus({ days: 1 });
+    }
+
+    await AffiliateInfoTable.updateInfo(windowStartTime.toISO(), windowEndTime.toISO(), { txId });
```
Christopher-Li, line 45: nit: let's add a log to print out window start and end time

indexer/packages/postgres/src/stores/affiliate-info-table.ts
```diff
@@ -94,6 +97,186 @@ export async function findById(
...
+ * @function updateInfo
+ * @param {string} windowStartTs - The exclusive start timestamp for filtering fills.
+ * @param {string} windowEndTs - The inclusive end timestamp for filtering fill.
+ * @param {Options} [options={ txId: undefined }] - Optional transaction ID or additional options.
+ * @returns {Promise<void>}
+ */
+export async function updateInfo(
+  windowStartTs: string, // exclusive
+  windowEndTs: string, // inclusive
+  options: Options = { txId: undefined },
```
Christopher-Li, line 116: I think you should just accept txId directly here instead of using options. Options have a [readReplica](https://github.com/dydxprotocol/v4-chain/blob/main/indexer/packages/postgres/src/types/utility-types.ts#L20) field that if populated, one would assume you would use the readReplica. However for this type of operation, we would never accept readReplica = true, because it is a write operation

indexer/packages/postgres/__tests__/stores/wallet-table.test.ts
Christopher-Li: bump on this

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
jerryfan01234: sure

indexer/services/roundtable/src/tasks/update-affiliate-info.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export default async function runTask(): Promise<void> {
+  // Wrap getting cache, updating info, and setting cache in one transaction so that persistent
+  // cache and affilitate info table are in sync.
+  const txId: number = await Transaction.start();
+  try {
+    const latestBlock: BlockFromDatabase = await BlockTable.getLatest();
+    if (latestBlock.time === null) {
+      throw Error('Failed to get latest block time');
+    }
```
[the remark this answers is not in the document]
jerryfan01234: I have a serarate branch with this chagne already made, will PR later today

Christopher-Li (approved):

affanv14 (approved):

PR #2249, opened 2024-09-12
add check for invalid skew factor

sr33j:
### Changelist
What is the change:
- Ensure that the product of the skew factor and half spread parameters is less than 2

Reason for change:
- It is dangerous for the skew to offset the vault's distance from oracle after a single fill, since one could take advantage of the vault AMM buy buying then selling for an immediate profit
- Assuming the general form of skewing is `skew ~ skew_factor*leverage*halfspread`, Then ensure  `skew_factor*leverage*halfspread < vault spread `
- Assume the leverage = order_size_pct after a single fill: `skew_factor*order_size_pct*halfspread <  2*halfspread`
- Therefore, `skew_factor*order_size_pct < 2`

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/vault/types/params.go
+++ b/protocol/x/vault/types/params.go
@@ -41,6 +41,10 @@ func (p QuotingParams) Validate() error {
 	if p.ActivationThresholdQuoteQuantums.Sign() < 0 {
 		return ErrInvalidActivationThresholdQuoteQuantums
 	}
+	// Skew factor times order_size_pct must be less than 2 to avoid skewing over the spread
+	if uint64(p.SkewFactorPpm)*uint64(p.OrderSizePctPpm) >= 2_000_000*1_000_000 {
+		return ErrInvalidSkewFactor
+	}
 
 	return nil
 }
--- a/protocol/x/vault/types/params_test.go
+++ b/protocol/x/vault/types/params_test.go
@@ -80,6 +80,30 @@ func TestValidateQuotingParams(t *testing.T) {
 			},
 			expectedErr: types.ErrInvalidActivationThresholdQuoteQuantums,
 		},
+		"Failure - SkewFactorPpm is high. Product of SkewFactorPpm and OrderSizePctPpm is above threshold.": {
+			params: types.QuotingParams{
+				Layers:                           2,
+				SpreadMinPpm:                     3_000,
+				SpreadBufferPpm:                  1_500,
+				SkewFactorPpm:                    20_000_000,
+				OrderSizePctPpm:                  100_000,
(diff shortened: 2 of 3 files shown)


protocol/x/vault/types/params_test.go
```diff
@@ -80,6 +80,30 @@ func TestValidateQuotingParams(t *testing.T) {
 			},
 			expectedErr: types.ErrInvalidActivationThresholdQuoteQuantums,
 		},
+		"Failure - SkewFactorPpm is high. Product of SkewFactorPpm and OrderSizePctPpm is above threshold.": {
```
vincentwschau, line 83: nit: Add a test where skew and order size are max uint32 to make sure no overflow errors.

vincentwschau (approved):

protocol/x/vault/types/params_test.go
```diff
@@ -80,6 +81,42 @@ func TestValidateQuotingParams(t *testing.T) {
...
+				SpreadMinPpm:                     3_000,
+				SpreadBufferPpm:                  1_500,
+				SkewFactorPpm:                    2_000_000,
+				OrderSizePctPpm:                  1_000_000,
+				OrderExpirationSeconds:           5,
+				ActivationThresholdQuoteQuantums: dtypes.NewInt(1),
+			},
+			expectedErr: types.ErrInvalidSkewFactor,
+		},
+		"Failure - Both OrderSizePctPpm and SkewFactorPpm are MaxUint8. Product of SkewFactorPpm and OrderSizePctPpm is above threshold.": {
```
vincentwschau, line 108: nit:
```suggestion
		"Failure - Both OrderSizePctPpm and SkewFactorPpm are MaxUint32. Product of SkewFactorPpm and OrderSizePctPpm is above threshold.": {
```

protocol/x/vault/types/params.go
```diff
@@ -42,6 +42,10 @@ func (p QuotingParams) Validate() error {
 	if p.ActivationThresholdQuoteQuantums.Sign() < 0 {
 		return ErrInvalidActivationThresholdQuoteQuantums
 	}
+	// Skew factor times order_size_pct must be less than 2 to avoid skewing over the spread
+	if uint64(p.SkewFactorPpm)*uint64(p.OrderSizePctPpm) >= 2_000_000*1_000_000 {
```
tqin7, line 46: it might be better to use `big.Int` here to do multiplication + comparison. using uint64 won't overflow here with these two being uint32s, but in case we want to be future proof

tqin7 (approved):

tqin7 (approved):

PR #2253, opened 2024-09-13
[OTE-456] FNS x OE: stage `FinalizeBlock` events and emit in `Precommit`

teddyding:
### Changelist
Store order fill stream events in a transient store during block processing
Do the following in Precommit, after consensus agrees on a block.
- Send onchain order fills
- Send onchain subacount updates
- Send OrderbookUpdates to synchronize optimistic fills in our local ops queue.

### Test Plan
(TODO) FNS integration testing on dev

### Author/Reviewer Checklist

## Summary by CodeRabbit

Here are the updated release notes based on the provided changes:

- **New Features**
  - Added functionality to stage and retrieve events related to order fills and subaccount updates during the block finalization process.
  - Introduced methods to handle batch streaming of updates after a block is finalized, ensuring synchronization between on-chain state and off-chain operations.

- **Improvements**
  - Enhanced the streaming manager component to better manage and respond to events related to block finalization.
  - Streamlined the handling of finalized updates for order fills and subaccounts, improving performance and maintainability.

- **Chores**
  - Added new constants and metrics for better monitoring and management of streaming events.


protocol/streaming/full_node_streaming_manager.go
```diff
@@ -703,6 +793,86 @@ func (sm *FullNodeStreamingManagerImpl) GetSubaccountSnapshotsForInitStreams(
...
+	ctx sdk.Context,
+	orderBookUpdatesToSyncLocalOpsQueue *clobtypes.OffchainUpdates,
+	perpetualIdToClobPairId map[uint32][]clobtypes.ClobPairId,
+) {
+	// Flush all pending updates, since we want the onchain updates to arrive in a batch.
+	sm.FlushStreamUpdates()
+
+	finalizedFills, finalizedSubaccountUpdates := sm.getStagedEventsFromFinalizeBlock(ctx)
+
+	// TODO(CT-1190): Stream below in a single batch.
```
teddyding, line 809: Will do this in a follow-up PR

proto/dydxprotocol/clob/query.proto
```diff
@@ -198,6 +198,15 @@ message StreamUpdate {
   }
 }
 
+// StagedFinalizeBlockEvent is an event staged during `FinalizeBlock`.
```
jonfung-dydx, line 201: Can we add context to this comment to explain staging more and why it is necessary? + this is not exposed externally, so maybe move it out of query.proto?

protocol/streaming/constants.go
```diff
@@ -0,0 +1,12 @@
+package streaming
+
+const (
```
jonfung-dydx, line 3: nit: can we add some comments indicating this is for full node streaming? streaming is a little generic

protocol/streaming/full_node_streaming_manager.go
```diff
@@ -367,6 +379,84 @@ func (sm *FullNodeStreamingManagerImpl) sendStreamUpdates(
...
+	noGasCtx := ctx.WithGasMeter(ante_types.NewFreeInfiniteGasMeter())
+	store := noGasCtx.TransientStore(sm.streamingManagerTransientStoreKey)
+	return getStagedFinalizeBlockEvents(store)
+}
+
+func (sm *FullNodeStreamingManagerImpl) stageFinalizeBlockEvent(
+	ctx sdk.Context,
+	eventBytes []byte,
+) {
+	noGasCtx := ctx.WithGasMeter(ante_types.NewFreeInfiniteGasMeter())
```
jonfung-dydx, line 448: for my understanding why do we need to add gas meter to ctx? because we write to state?

protocol/streaming/full_node_streaming_manager.go
```diff
@@ -703,6 +793,86 @@ func (sm *FullNodeStreamingManagerImpl) GetSubaccountSnapshotsForInitStreams(
...
+// - Stream all events staged during `FinalizeBlock`.
+// - Stream orderbook updates to sync fills in local ops queue.
+func (sm *FullNodeStreamingManagerImpl) StreamBatchUpdatesAfterFinalizeBlock(
+	ctx sdk.Context,
+	orderBookUpdatesToSyncLocalOpsQueue *clobtypes.OffchainUpdates,
+	perpetualIdToClobPairId map[uint32][]clobtypes.ClobPairId,
+) {
+	// Flush all pending updates, since we want the onchain updates to arrive in a batch.
+	sm.FlushStreamUpdates()
+
```
jonfung-dydx, line 806: Thinking out loud, we don't need to `sm.Lock()` this right? CheckTx transactions won't run in parallel when we are finalizing the block so the cache couldn't possibly get new updates when we are calling this function? It could be the case that we flush on the timer but it should be fine if cache is empty since we flush it first

protocol/streaming/full_node_streaming_manager.go
jonfung-dydx: probably good to just obtain the lock and manually send the updates through in a single batch bypassing the cache

protocol/streaming/full_node_streaming_manager.go
```diff
@@ -703,6 +793,86 @@ func (sm *FullNodeStreamingManagerImpl) GetSubaccountSnapshotsForInitStreams(
...
+	for _, stagedEvent := range stagedEvents {
+		switch event := stagedEvent.Event.(type) {
+		case *clobtypes.StagedFinalizeBlockEvent_OrderFill:
+			finalizedFills = append(finalizedFills, *event.OrderFill)
+		case *clobtypes.StagedFinalizeBlockEvent_SubaccountUpdate:
+			finalizedSubaccountUpdates = append(finalizedSubaccountUpdates, *event.SubaccountUpdate)
+		}
+	}
+
+	telemetry.SetGauge(
```
jonfung-dydx, line 860: no `telemetry` library, use 
```
	metrics.SetGauge(
		metrics.GrpcStreamNumUpdatesBuffered,
		float32(len(sm.streamUpdateCache)),
	)
```

Also add a Count to the end of the metric name to indicate number unit of measurement

protocol/x/clob/keeper/grpc_stream_finalize_block.go
```diff
@@ -0,0 +1,49 @@
...
+	return allUpdates
+}
+
+// Grpc Streaming logic after consensus agrees on a block.
+// - Stream all events staged during `FinalizeBlock`.
+// - Stream orderbook updates to sync fills in local ops queue.
+func (k Keeper) StreamBatchUpdatesAfterFinalizeBlock(
+	ctx sdk.Context,
+) {
+	defer telemetry.MeasureSince(
```
jonfung-dydx, line 36: same, don't use `telemetry` module

protocol/x/clob/keeper/grpc_stream_finalize_block.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Returns the order updates needed to update the fill amount for the orders
+// from local ops queue, according to the latest onchain state (after FinalizeBlock).
+func (k Keeper) getUpdatesToSyncLocalOpsQueue(
+	ctx sdk.Context,
+) *types.OffchainUpdates {
+	localValidatorOperationsQueue, _ := k.MemClob.GetOperationsToReplay(ctx)
+	fetchOrdersInvolvedInOpQueue(localValidatorOperationsQueue)
+	orderIdsFromLocal := fetchOrdersInvolvedInOpQueue(
+		localValidatorOperationsQueue,
```
jonfung-dydx, line 20: Think we need to monitor this to make sure the removal of 
```
		orderIdsFromProposed := fetchOrdersInvolvedInOpQueue(
			operations,
		)
```
will not break logic

protocol/x/clob/keeper/grpc_stream_finalize_block.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib/metrics"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// Returns the order updates needed to update the fill amount for the orders
+// from local ops queue, according to the latest onchain state (after FinalizeBlock).
```
jonfung-dydx, line 13:
```suggestion
// from local ops queue, according to the latest onchain state (after FinalizeBlock).
// Effectively reverts the optimistic fill amounts removed from the CheckTx to DeliverTx state transition.
```

jonfung-dydx (approved):

proto/dydxprotocol/clob/query.proto
teddyding: Will add more documentation in code as opposed to proto file

protocol/streaming/constants.go
teddyding: `protocol/streaming` is the code location we use for FNS specific codes; will add more comments in this file

protocol/streaming/full_node_streaming_manager.go
teddyding: Yes any read/write to store incur the gas meter. We do the same thing for [indexer](https://github.com/dydxprotocol/v4-chain/blob/85222463a9dc5da76e2ee4b6c54fade81f060812/protocol/indexer/indexer_manager/events.go#L111)

protocol/streaming/full_node_streaming_manager.go
teddyding: Yes agreed. In the follow-up PR I will obtain the lock around sending a batch

protocol/streaming/full_node_streaming_manager.go
teddyding: > no telemetry library
What's the context around this?

PR #2258, opened 2024-09-16
[OTE-790]Implement affiliates whitelist logic

affanv14:
### Changelist
Implements whitelisting functionality for affiliates

### Test Plan
Adds unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to update an affiliate's whitelist with new methods for managing affiliate relationships.
	- Added functionality to retrieve and manage affiliate whitelist data dynamically.
	- Enhanced affiliate management with the ability to filter revenue shares based on whitelist status.
	- Expanded internal operations to consider affiliate relationships during match processing.
	- Streamlined key management for affiliate-related data with new naming conventions.
	- Implemented new logic for determining affiliate tiers and fee shares based on whitelist status.

- **Bug Fixes**
	- Improved error handling for duplicate affiliate addresses in the whitelist.

- **Tests**
	- Expanded test coverage for affiliate management, including new test cases for whitelisted affiliates and revenue share retrieval.

- **Documentation**
	- Updated error messages and comments for clarity regarding the new affiliate management features.


protocol/x/affiliates/types/keys.go
```diff
@@ -16,4 +16,6 @@ const (
 	ReferredVolumeKeyPrefix = "ReferredVolume:"
 
 	AffiliateTiersKey = "AffiliateTiers"
+
+	AffiliateWhitelistKey = "AffiliateWhitelist"
```
teddyding, line 20: Let's actually optimize the keys and prefixes a bit (for faster insert/read from IAVL tree)
```
	ReferredByKeyPrefix = "RB:"
	ReferredVolumeKeyPrefix = "RV:"

	AffiliateTiersKey = "AT"

	AffiliateWhitelistKey = "AW"
	```

File: protocol/x/clob/keeper/process_single_match.go [hunk 2]
```diff
@@ -307,6 +309,7 @@ func (k Keeper) persistMatchedOrders(
 	makerFeePpm int32,
 	bigFillQuoteQuantums *big.Int,
 	insuranceFundDelta *big.Int,
+	affiliatesWhitelistMap map[string]uint32, // nolint: unused
```
teddyding, line 312: Why the `nolint`? Isn't `affiliatesWhitelistMap`used below? 

In general `nolint` in non-testing files is red flag

affanv14: this was a remnant from before the clob pr was merged. Should be removed

protocol/x/clob/keeper/process_operations.go
```diff
@@ -144,8 +144,15 @@ func (k Keeper) ProcessInternalOperations(
 
 		switch castedOperation := operation.Operation.(type) {
 		case *types.InternalOperation_Match:
+			affiliatesWhitelistMap, err := k.affiliatesKeeper.GetAffiliateWhitelistMap(ctx)
```
teddyding, line 147: Move this outside of the for loop. The intention is to get the whitelist map once for all operations

teddyding, line 147: Declare `affiliatesWhitelistMap` before the for loop, and query it at most once for all matches in the operations queue.

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -259,6 +267,52 @@ func (k Keeper) GetIndexerEventManager() indexer_manager.IndexerEventManager {
...
+			affiliateWhitelistMap[address] = tier.GetTakerFeeSharePpm()
+		}
+	}
+	return affiliateWhitelistMap, nil
+}
+
+func (k Keeper) SetAffiliateWhitelist(ctx sdk.Context, whitelist types.AffiliateWhitelist) error {
+	store := ctx.KVStore(k.storeKey)
+	addressSet := make(map[string]bool)
+	for _, tier := range whitelist.Tiers {
```
teddyding, line 287: New change: let's also add a protocol constant = 50% as upper bound of the fee share ppm, for fee security

protocol/x/clob/memclob/memclob.go
```diff
@@ -1696,7 +1696,9 @@ func (m *MemClobPriceTimePriority) mustPerformTakerOrderMatching(
 			FillAmount: matchedAmount,
 		}
 
-		success, takerUpdateResult, makerUpdateResult, _, err := m.clobKeeper.ProcessSingleMatch(ctx, &matchWithOrders)
+		affiliatesWhitelistMap := make(map[string]uint32)
+		success, takerUpdateResult, makerUpdateResult, _, err := m.clobKeeper.ProcessSingleMatch(
+			ctx, &matchWithOrders, affiliatesWhitelistMap)
```
teddyding, line 1701: Explicitly pass in an empty map, and document why the input is empty.

protocol/x/affiliates/keeper/keeper_test.go
```diff
@@ -394,6 +394,229 @@ func TestRegisterAffiliateEmitEvent(t *testing.T) {
...
+
+			whitelistMap, err := k.GetAffiliateWhitelistMap(ctx)
+			require.NoError(t, err)
+			require.Equal(t, tc.expectedLength, len(whitelistMap))
+			require.Equal(t, tc.expectedMap, whitelistMap)
+		})
+	}
+}
+
+func TestGetTakerFeeShareViaWhitelist(t *testing.T) {
```
teddyding, line 543: Do we have a unit test where the affiliate qualifies through regular affiliate program and gets overridden to the higher whitelist fee share?

affanv14: Yes, the first unit test here registers an affiliate(which should have revshare of 15%) and we bump it to 40%

protocol/x/affiliates/keeper/keeper.go
affanv14: I think it would be better to add this to `ValidateRevShareSafety`


Now that I think about it, Im gonna make `ValidateRevShareSafety` more accurate - can discuss offline

protocol/x/clob/keeper/process_operations.go
```diff
@@ -136,6 +136,13 @@ func (k Keeper) ProcessInternalOperations(
 	// All short term orders in this map have passed validation.
 	placedShortTermOrders := make(map[types.OrderId]types.Order, 0)
 
+	affiliatesWhitelistMap, err := k.affiliatesKeeper.GetAffiliateWhitelistMap(ctx)
```
teddyding, line 139: Can you do it in a way that make this state read at most once when there are matches, but zero if there's no match: declare `affiliatesWhitelistMap` outside of the scope of for loop, but assign to it only if it's not been initialized within the loop

protocol/x/affiliates/keeper/keeper.go
teddyding: To clarify, we explicitly want a constant `AffiliateWhitelistCap = 50%` for the affiliate whitelist fee share - don't see being added?

protocol/x/clob/memclob/memclob.go
```diff
@@ -1696,9 +1696,10 @@ func (m *MemClobPriceTimePriority) mustPerformTakerOrderMatching(
 			FillAmount: matchedAmount,
 		}
 
-		affiliatesWhitelistMap := make(map[string]uint32)
+		// affiliate revshare doesnt affect subaccount balance, we can pass an empty map
+		// as we dont need to process affiliate revshare during checkTx and only during deliverTx
 		success, takerUpdateResult, makerUpdateResult, _, err := m.clobKeeper.ProcessSingleMatch(
-			ctx, &matchWithOrders, affiliatesWhitelistMap)
+			ctx, &matchWithOrders, make(map[string]uint32))
```
teddyding, line 1702: Nit: pass in literal for readability
```suggestion
			ctx, &matchWithOrders, map[string]uint32{})
```

protocol/x/clob/keeper/process_operations.go
```diff
@@ -151,6 +145,19 @@ func (k Keeper) ProcessInternalOperations(
 
 		switch castedOperation := operation.Operation.(type) {
 		case *types.InternalOperation_Match:
+			// check if affiliate whitelist map is nil and initialize it if it is.
+			// This is done to avoid getting whitelist map on list of operations
+			// where there are no matches.
+			if affiliatesWhitelistMap == nil {
```
teddyding, line 151: If the `AffiliateWhitelist` is empty in state, then this will be `nil` everytime right? Do we need a separate boolean to track whether the state read was done?

protocol/x/clob/memclob/memclob.go
```diff
@@ -1698,8 +1698,12 @@ func (m *MemClobPriceTimePriority) mustPerformTakerOrderMatching(
 
 		// affiliate revshare doesnt affect subaccount balance, we can pass an empty map
```
teddyding, line 1699: Let's use below comment:

```
Pass in empty map to avoid reading `AffiliateWhitelist` from state in every `CheckTx`. This deviates from `DeliverTx` which accounts for affiliate whitelist correctly. 
This deviation is ok becasue rev shares/fees are distributed to the recipient’s bank balance and not settled at the subaccount level, and won’t affect the collateralization of future operations in the operations queue.
```
```

File: protocol/x/clob/keeper/process_operations.go (hunk 9 above)
Comment (affanv14) [reply protocol/x/clob/keeper/process_operations.go:151]: golang dfiferentiates between a nil map and an empty map

Comment (affanv14) [reply protocol/x/clob/keeper/process_operations.go:151]:
```
package main

import "fmt"

func main() {
	fmt.Println("Hello, 世界")
	x := map[string]uint32{}
	if x == nil {
		fmt.Println("x is nil")
	}
	var y map[string]uint32
	if y == nil {
		fmt.Println("y is nil")
	}
}

```

affanv14: try this piece of code(confirmed it myself)

teddyding (approved):

PR #2275, opened 2024-09-17
[CT-1194] add more relevant authenticators

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new authenticators (`AllOf` and `AnyOf`) for flexible authentication strategies.
	- Added a `MessageFilter` to enhance message authentication based on JSON patterns.
	- Implemented a method to retrieve KV Store keys from the `App` struct.

- **Bug Fixes**
	- Improved error handling and tracking for sub-authenticators.

- **Tests**
	- Added comprehensive test suites for the new authenticators and message filtering logic to ensure robust functionality.
	- Included tests for the `MessageFilter` to validate its authentication capabilities.

### DIFF
--- a/protocol/x/accountplus/authenticator/message_filter.go
+++ b/protocol/x/accountplus/authenticator/message_filter.go
@@ -1,18 +1,22 @@
 package authenticator
 
 import (
+	"strings"
+
 	errorsmod "cosmossdk.io/errors"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+const SEPARATOR = ","
+
 var _ Authenticator = &MessageFilter{}
 
 // MessageFilter filters incoming messages based on a predefined JSON pattern.
 // It allows for complex pattern matching to support advanced authentication flows.
 type MessageFilter struct {
-	msgType string
+	msgTypes []string
 }
 
 // NewMessageFilter creates a new MessageFilter with the provided EncodingConfig.
@@ -32,7 +36,7 @@ func (m MessageFilter) StaticGas() uint64 {
 
 // Initialize sets up the authenticator with the given data, which should be a valid JSON pattern for message filtering.
 func (m MessageFilter) Initialize(config []byte) (Authenticator, error) {
-	m.msgType = string(config)
+	m.msgTypes = strings.Split(string(config), SEPARATOR)
 	return m, nil
 }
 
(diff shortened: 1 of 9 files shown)


protocol/x/accountplus/authenticator/message_filter.go
```diff
@@ -0,0 +1,82 @@
+package authenticator
```
jayy04, line 1: Osmosis's implementation of `message_filter` actually matches full JSON patterns ([link](https://github.com/osmosis-labs/osmosis/blob/main/x/smart-account/authenticator/message_filter.go#L79)), which is an overkill for our usecase (might have some perf implications as well). 

For our implementation, I rewrote `message_filter` to only filter on MsgTypeURL for simplicity.

protocol/x/accountplus/authenticator/message_filter.go
```diff
@@ -0,0 +1,82 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var _ Authenticator = &MessageFilter{}
+
+// MessageFilter filters incoming messages based on a predefined JSON pattern.
+// It allows for complex pattern matching to support advanced authentication flows.
+type MessageFilter struct {
+	msgType string
```
teddyding, line 15: Instead of 
```
	AnyOf(
		MessageFilter(MsgPlaceOrder),
		MessageFilter(MsgCancelOrder),
	),
```

Would it be easier/more performant to make a message filter takes a list of messages instead? So
```
MessageFilter(MsgPlaceOrder, MsgCancelOrder)
```

protocol/x/accountplus/authenticator/message_filter.go
```diff
@@ -0,0 +1,82 @@
...
+// Track is a no-op in this implementation but can be used to track message handling.
+func (m MessageFilter) Track(ctx sdk.Context, request AuthenticationRequest) error {
+	return nil
+}
+
+// Authenticate checks if the provided message conforms to the set JSON pattern.
+// It returns an AuthenticationResult based on the evaluation.
+func (m MessageFilter) Authenticate(ctx sdk.Context, request AuthenticationRequest) error {
+	if request.Msg.TypeURL != m.msgType {
+		return errorsmod.Wrapf(
```
teddyding, line 48: I see Osmosis also returns an error here instead of a boolean value. For my understanding, how does the upstream function differentiate between an authentication failure (e.g. invalid signature) or message being filtered out?

jayy04: are you suggesting to add both the error and the boolean value? personally feel like that's unnecessary, let me know if you disagree. 

> how does the upstream function differentiate between an authentication failure (e.g. invalid signature) or message being filtered out?

upstream functions don't really need to differentiate between different authentication failures since this is part of the ante handlers flow, and any transactions that fail ante handlers should just get outright rejected, and no special handling is required. 

for humans looking at debug logs, they can still differentiate between authentication errors using the additional details in the wrapped version of the error. 

We can maybe use specific errors instead of the generic `ErrUnauthorized` error. created [this](https://linear.app/dydx/issue/CT-1207/use-more-specific-errors-when-authenticators-fail-instead-of) ticket to track

protocol/x/accountplus/authenticator/message_filter.go
jayy04: updated!

PR #2298, opened 2024-09-19
[OTE-816] Update Revshare logic for affiliates

affanv14:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced revenue sharing calculations to include affiliate fee impacts, improving accuracy in revenue distribution.
  
- **Bug Fixes**
	- Adjusted expected balances in fee distribution tests to reflect updated revenue share logic.

- **Refactor**
	- Streamlined test cases by removing scenarios related to exceeding net fees, indicating a refined focus in revenue sharing behavior.
	- Simplified the `Keeper` struct by removing unnecessary dependencies related to revenue sharing.
	- Removed the `RevShareKeeper` interface and associated parameters from various components, simplifying interactions within the system.


protocol/x/revshare/keeper/revshare_test.go
```diff
@@ -856,52 +856,6 @@ func TestKeeper_GetAllRevShares_Invalid(t *testing.T) {
 				require.NoError(t, err)
 			},
 		},
-		{
-			name:                              "Total fees shared exceeds net fees from market mapper and affiliates",
```
affanv14, line 860: These tests wont work anymore

protocol/x/revshare/keeper/revshare.go
```diff
@@ -171,17 +171,17 @@ func (k Keeper) GetAllRevShares(
 	makerFees := fill.MakerFeeQuoteQuantums
```
teddyding, line 171: Let's change `REV_SHARE_FEE_SOURCE_NET_FEE` to `REV_SHARE_FEE_SOURCE_NET_PROTOCL_REVENUE` to disambiguate

`NET_FEE` could be understood as taker fee minus rebates

protocol/x/revshare/keeper/revshare.go
```diff
@@ -163,25 +163,25 @@ func (k Keeper) GetAllRevShares(
...
 
 	totalFeesShared := big.NewInt(0)
 	takerFees := fill.TakerFeeQuoteQuantums
 	makerFees := fill.MakerFeeQuoteQuantums
 	netFees := big.NewInt(0).Add(takerFees, makerFees)
 
-	affiliateRevShares, err := k.getAffiliateRevShares(ctx, fill, affiliatesWhitelistMap)
+	affiliateRevShares, affiliateFeesShared, err := k.getAffiliateRevShares(ctx, fill, affiliatesWhitelistMap)
 	if err != nil {
 		return types.RevSharesForFill{}, err
```
teddyding, line 176: Comment on caller function `persistMatchedOrders`:

Instead propagating error returned from `GetAllRevShares` (which would ultimately fail the match transaction), it's better to just pass in empty `RevSharesForFill` to `DistributeFees`. This way if any rev share calculation encounters issue, the chain will continue to be able to process matches (without rev share). 

Order matching + rev share both working > Only order matching working > Order matching not working

protocol/x/revshare/keeper/revshare.go
```diff
@@ -163,25 +163,25 @@ func (k Keeper) GetAllRevShares(
...
 	}
-
-	unconditionalRevShares, err := k.getUnconditionalRevShares(ctx, netFees)
+	netFeesSubAffiliateFeesShared := big.NewInt(0).Sub(netFees, affiliateFeesShared)
```
teddyding, line 178: For sanity check, let's return error if `netFeesSubAffiliateFeesShared` <= 0

teddyding, line 178:
```suggestion
	netFeesSubAffiliateFeesShared := new(big.Int).Sub(
		netFees,
		affiliateFeesShared,
	)
```

protocol/x/revshare/keeper/revshare.go
```diff
@@ -163,25 +163,25 @@ func (k Keeper) GetAllRevShares(
...
+	unconditionalRevShares, err := k.getUnconditionalRevShares(ctx, netFeesSubAffiliateFeesShared)
```
teddyding, line 179: For `getMarketMapperRevShare` and `getUnconditionalRevShares` update input variable name from `netFees` to `netFeesSubAffiliateFeesShared`

protocol/x/subaccounts/keeper/transfer_test.go
```diff
@@ -1500,10 +1500,10 @@ func TestDistributeFees(t *testing.T) {
 				MonthlyRollingTakerVolumeQuantums: 1_000_000,
 			},
 			expectedSubaccountsModuleAccBalance:     big.NewInt(100),  // 600 - 500
-			expectedFeeModuleAccBalance:             big.NewInt(2888), // 2500 + 500 - 50
-			expectedMarketMapperAccBalance:          big.NewInt(50),   // 10% of 500
+			expectedFeeModuleAccBalance:             big.NewInt(2892), // 2500 + 500 - 108
```
teddyding, line 1503: Realized that we aren't purely testing `DistributeFees` function in here, but it's more of an e2e test on affiliate + rev share + distribute fees. This isn't the best practice - should have a function that exclusively tests `DistributeFees`, which doesn't rely on `affiliatesKeeper` or `revshareKeeper.` (We should also remove delete the `revShareKeeper`  on `subaccountsKeeper`

protocol/x/revshare/keeper/revshare_test.go
```diff
@@ -332,23 +332,23 @@ func TestKeeper_GetAllRevShares_Valid(t *testing.T) {
 					},
```
teddyding, line 332: For my understanding, how did we set up the tier for `Bob` in this test case? Didn't find anything related in the `setUp()`

protocol/x/revshare/keeper/revshare_test.go
```diff
@@ -420,33 +420,33 @@ func TestKeeper_GetAllRevShares_Valid(t *testing.T) {
...
```
teddyding, line 420: Comment on line 411: `negative unconditional and market mapper` doesn't make sense - did you mean maker fee is negative

protocol/x/revshare/keeper/revshare_test.go
```diff
@@ -420,33 +420,33 @@ func TestKeeper_GetAllRevShares_Valid(t *testing.T) {
...
 					{
 						Recipient:         constants.BobAccAddress.String(),
-						RevShareFeeSource: types.REV_SHARE_FEE_SOURCE_NET_FEE,
+						RevShareFeeSource: types.REV_SHARE_FEE_SOURCE_NET_PROTOCOL_REVENUE,
 						RevShareType:      types.REV_SHARE_TYPE_UNCONDITIONAL,
-						QuoteQuantums:     big.NewInt(1_600_000),
+						QuoteQuantums:     big.NewInt(1_300_000),
```
teddyding, line 425: At least for one of the test cases, could we add comment on math behind these values? For example, 

```
						QuoteQuantums:     big.NewInt(1_300_000), // (10 - 2 - 10 * 0.15) * 20% = 1.3
```

protocol/x/revshare/keeper/revshare.go
affanv14: sounds good. I'll add some error logs to catch in case it fails

protocol/x/revshare/keeper/revshare_test.go
affanv14: bob address has coins already setup [here](https://github.com/dydxprotocol/v4-chain/blob/affan/update-revshare/protocol/testutil/constants/genesis.go#L771) for staking. This just puts bob at the max tier

protocol/x/revshare/keeper/revshare_test.go
affanv14: yeah.. my b

protocol/x/subaccounts/keeper/transfer_test.go
```diff
@@ -1407,48 +1388,26 @@ func TestDistributeFees(t *testing.T) {
...
 			unconditionalRevShareAcctAddr: "",
-			revshareParams: revsharetypes.MarketMapperRevenueShareParams{
-				Address:         constants.AliceAccAddress.String(),
-				RevenueSharePpm: 100_000, // 10%
-				ValidDays:       240,
-			},
-			setRevenueShare:    true,
-			revShareExpiration: 100,
-		},
-		"success - market mapper rev share expired": {
```
affanv14, line 1418: Removing this since its an e2e test. Added an equivalent test in revshare_test.go

affanv14: The test in revshare_test.go broke - looks like there was already an existing test which was commented out

protocol/x/revshare/keeper/revshare_test.go
teddyding: I see, could we document in the unit test? Not very obvious to reader

protocol/x/revshare/keeper/revshare.go
teddyding: Did we do this in this PR? Or plan to do in another one?

protocol/x/revshare/keeper/revshare.go
```diff
@@ -173,17 +173,25 @@ func (k Keeper) GetAllRevShares(
...
-	netFeesSubAffiliateFeesShared := big.NewInt(0).Sub(netFees, affiliateFeesShared)
+	netFeesSubAffiliateFeesShared := new(big.Int).Sub(netFees, affiliateFeesShared)
 	unconditionalRevShares, err := k.getUnconditionalRevShares(ctx, netFeesSubAffiliateFeesShared)
 	if err != nil {
-		return types.RevSharesForFill{}, err
+		log.ErrorLogWithError(ctx, "error getting unconditional rev shares", err)
+		return types.RevSharesForFill{}, nil
+	}
+
+	if netFeesSubAffiliateFeesShared.Sign() <= 0 {
```
teddyding, line 186: Nit: move to immediately below line 129

protocol/x/revshare/keeper/revshare.go
affanv14: This is done in the PR . Check [here](https://github.com/dydxprotocol/v4-chain/blob/e69bbb77a7e9edcfbd4ec2bf6a848d483187c30f/protocol/x/revshare/keeper/revshare.go#L181-L195). Also updated tests in revshare_test.go

teddyding: Oh, the original comment was suggesting that we keep `GetAllRevShares` behavior as this but handle error differently in `persistMatchedOrders`, since it's the caller responsibility to determine what to do with the returned error from `GetAllRevShares` 

Changing `GetAllRevShares` behavior works too but we are changing the signature of the function. Also it doesn't make sense to have this function return error anymore. 

We should either:
- Keep `GetAllRevShares` as is, and just log out error (don't propogate errors further up) in `persistMatchedOrders`.
- Fully change signature of `GetAllRevShares` so it doesn't return errors anymore, maybe even rename it to `GetAllValidRevShares`

The first seems a bit cleaner to me since we just need one error log instead of multiple

protocol/x/revshare/keeper/revshare.go
```diff
@@ -173,24 +172,20 @@ func (k Keeper) GetAllRevShares(
...
 	if err != nil {
-		log.ErrorLogWithError(ctx, "error getting affiliate rev shares", err)
-		return types.RevSharesForFill{}, nil
+		return types.RevSharesForFill{}, err
 	}
 	netFeesSubAffiliateFeesShared := new(big.Int).Sub(netFees, affiliateFeesShared)
 	if netFeesSubAffiliateFeesShared.Sign() <= 0 {
-		log.ErrorLog(ctx, "net fees sub affiliate fees shared is less than or equal to 0")
-		return types.RevSharesForFill{}, nil
+		return types.RevSharesForFill{}, err
```
teddyding, line 179: What `err` are we returning here?

teddyding (approved):

PR #2325, opened 2024-09-23
[CT-1198] Add smart account and circuit break ante handler

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Release Notes

- **New Features**
	- Introduced new metric constants for enhanced monitoring of the authenticator system.
	- Added the `AuthenticatorDecorator` to manage authentication logic during transaction processing.
	- Implemented the `CircuitBreakerDecorator` for dynamic transaction routing based on circuit breaker status.
	- Added functionality to check for selected authenticator extensions in transactions.
	- Enhanced codec capabilities with a new interface for handling authenticator transaction options.

- **Bug Fixes**
	- Added specific error handling for inactive smart accounts.

- **Tests**
	- Established comprehensive test suites for both `AuthenticatorDecorator` and `CircuitBreakerDecorator`, ensuring robust functionality and error handling. 

- **Documentation**
	- Updated error handling documentation to include new error variables and interfaces.


protocol/x/accountplus/ante/circuit_breaker.go
```diff
@@ -0,0 +1,46 @@
...
+) (newCtx sdk.Context, err error) {
+	// Check that the authenticator flow is active
+	if specified, _ := lib.HasSelectedAuthenticatorTxExtensionSpecified(tx, ad.accountPlusKeeper); specified {
+		// Return and call the AnteHandle function on all the authenticator decorators.
+		return ad.authenticatorAnteHandlerFlow(ctx, tx, simulate)
+	}
+
+	// Return and call the AnteHandle function on all the original decorators.
+	return ad.originalAnteHandlerFlow(ctx, tx, simulate)
+}
```
jayy04, line 46: Note for @teddyding - This is different from Osmosis's logic. 

Osmosis
- If smart account active AND tx extension present -> use new auth flow
- else fallback to default flow 

dYdX
- if tx extension missing -> use original flow 
- else use new flow 
-- *if smart account inactive -> return error* (this is the gist of the diff, no longer falls back to original flow but returns error instead)
-- proceed with new flow

jayy04: Reason to internalize this is for the resting maker order issue. Since the proposed solution in the slack channel doesn't actually run ante handlers, we need to internalize so that we no longer need to worry about 
- order placed using new auth flow 
- gov turns off smart account 
- order now falling back to old flow

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,298 @@
...
+	defer metrics.ModuleMeasureSince(
+		types.ModuleName,
+		metrics.AuthenticatorDecoratorAnteHandleLatency,
+		time.Now(),
+	)
+
+	// Make sure smart account is active.
+	if active := ad.accountPlusKeeper.GetIsSmartAccountActive(ctx); !active {
+		return ctx, types.ErrSmartAccountNotActive
+	}
```
jayy04, line 64: this is internalized. see the other [comment](https://github.com/dydxprotocol/v4-chain/pull/2325/files#r1772014493)

protocol/x/accountplus/ante/circuit_breaker_test.go
```diff
@@ -0,0 +1,194 @@
...
+		accounts = append(
+			accounts,
+			authtypes.NewBaseAccount(accAddress, priv.PubKey(), 0, 0),
+		)
+
+		// Add the test accounts' addresses to an array for later use
+		s.TestAccAddress = append(s.TestAccAddress, accAddress)
+	}
+
+	// Initialize the Osmosis application
```
teddyding, line 77: Comment refers to Osmosis

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+
+	return next(ctx, tx, simulate)
+}
+
+// ValidateAuthenticatorFeePayer enforces that the tx fee payer has not been set manually
+// to an account different to the signer of the first message. This is a requirement
+// for the authenticator module.
+// The only user of a manually set fee payer is with fee grants, which are not
+// available on osmosis
```
teddyding, line 236: Comment refers to Osmosis

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+		return ctx, errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "no messages in transaction")
+	}
+
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, errorsmod.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	// The fee payer is the first signer of the transaction. This should have been enforced by the
+	// LimitFeePayerDecorator
```
teddyding, line 84: Don't see this decorator - does this mean `ValidateAuthenticatorFeePayer`?

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+					"failed to generate authentication data "+
+						"(account = %s, authenticator id = %d, msg index = %d, msg type url = %s)",
+					account,
+					selectedAuthenticator.Id,
+					msgIndex,
+					sdk.MsgTypeURL(msg),
+				)
+		}
+
+		a11r := selectedAuthenticator.Authenticator
```
teddyding, line 163: Why this variable naming?

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+				selectedAuthenticator.Id,
+				selectedAuthenticator.Authenticator.Type(),
+			)
+		}
+	}
+
+	// If the transaction has been authenticated, we call Track(...) on every message
+	// to notify its authenticator so that it can handle any state updates.
+	for _, track := range tracks {
+		if err := track(); err != nil {
```
teddyding, line 224: Question: I see our current authenticators either call `subTrack` recursively or is empty. What's an actual use case of this?

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+					sdk.MsgTypeURL(msg),
+				)
+		}
+
+		a11r := selectedAuthenticator.Authenticator
+		stringId := strconv.FormatUint(selectedAuthenticator.Id, 10)
+		authenticationRequest.AuthenticatorId = stringId
+
+		// Consume the authenticator's static gas
+		ctx.GasMeter().ConsumeGas(a11r.StaticGas(), "authenticator static gas")
```
teddyding, line 168: To confirm, this is just the static portion but not the full gas cost right (inferred from this [comment](https://github.com/dydxprotocol/v4-chain/blob/c34d56d6e395e8311556883c3b3afe67f25d47f7/protocol/x/accountplus/authenticator/signature_authenticator.go#L37))

protocol/x/accountplus/ante/ante.go
```diff
@@ -0,0 +1,299 @@
...
+		signers, _, err := ad.cdc.GetMsgV1Signers(msg)
+		if err != nil {
+			return ctx, errorsmod.Wrap(sdkerrors.ErrUnauthorized, "failed to get signers")
+		}
+		// Enforce only one signer per message
+		if len(signers) != 1 {
+			return sdk.Context{}, errorsmod.Wrap(sdkerrors.ErrUnauthorized, "messages must have exactly one signer")
+		}
+
+		// By default, the first signer is the account that is used
```
teddyding, line 112: Nit: update comment to reflect that the above guarantees there is one and one signer only

teddyding, line 112: Nit: updat to reflect that the above guarantees there is one and one signer only
```suggestion
		// Get the account corresponding to the only signer of this message
```

protocol/x/accountplus/ante/circuit_breaker.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/dydxprotocol/v4-chain/protocol/x/accountplus/lib"
+)
+
+// CircuitBreakerDecorator routes transactions through appropriate ante handlers based on
+// the IsCircuitBreakActive function.
```
teddyding, line 11: Think we removed `IsCircuitBreakActive`

protocol/x/accountplus/ante/ante.go
jayy04: We don't need it right now for our current use case since our authentications are all stateless / static (public key signature verification, message filter, clob pair id, subaccount number).

This persists any state changes that is required by the authenticator. e.g. if we want to support something like `You can place up to X orders`, we would be using the `track` method to increment a counter in state

jayy04: also down to get rid of this for v7.1 for simplicity

protocol/x/accountplus/ante/ante.go
jayy04: yeah - since our validations are mostly stateless, gas consumption (static + gas incurred during `Authenticate`) are going to be 0 for most of the validations. `SignatureVerification` incurs `SigVerifyCostSecp256k1` though

jayy04: thinking about this more, seems like we are double charging gas [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/accountplus/authenticator/signature_authenticator.go#L60) and [here](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/app/ante.go#L137).

removed gas for the new `SignatureAuthenticator`. let me know if you have objections.

protocol/x/accountplus/ante/ante.go
jayy04: I guess this variable name was too long for them so they used something similar to a16z.

I don't have a preference on this, let me know if you do.

protocol/x/accountplus/ante/ante.go
teddyding: Link doesn't work for me - which comment were you referring to/

protocol/x/accountplus/ante/ante.go
teddyding: Nit: let's stick to our naming conventions and use sth like `authenticator` or `selectedAuth`?

protocol/x/accountplus/ante/ante.go
jayy04: oh the big comment https://github.com/dydxprotocol/v4-chain/pull/2325#discussion_r1772014493

protocol/x/accountplus/ante/ante.go
teddyding: It looks like Osmosis way of solving this is to create two separate chain of decorators, and only include `SigGasConsumeDecorator` in the [classic chain](https://github.com/osmosis-labs/osmosis/blob/4ce0dbf17e0fffbecdcf5781b50cb9334acd4b1b/app/ante.go#L73). Did we consider this approach?

Also, is there any other modifications needed for the new route? For instance Osmosis code implies this [decorator](https://github.com/osmosis-labs/osmosis/blob/4ce0dbf17e0fffbecdcf5781b50cb9334acd4b1b/app/ante.go#L71C8-L71C29) is only needed for the classic route.

protocol/x/accountplus/ante/circuit_breaker.go
teddyding: Makes sense! It may be helpful to document this (why we aren't conditioning based on `IsSmartAcccountActive` here - the restring maker order issue is not obvious

protocol/x/accountplus/ante/ante.go
teddyding: I see, no objections to keeping this in -  IIUC it doesn't increase the attack vector since no implemented authenticators implement tracking

protocol/x/accountplus/ante/ante.go
jayy04: I was thinking for simplicity we can just have 
```
sigVerification: accountplusante.NewCircuitBreakerDecorator(
			options.AccountplusKeeper,
			accountplusante.NewAuthenticatorDecorator( // this is the new ante handler
				...
			),
			customante.NewSigVerificationDecorator( // this is the old sigverify (without replay protection)
				...
			),
		),
```
i.e. without the two chained decorators and just share everything else except for sigverify. Agree that things like `NewSetPubKeyDecorator` is only relevant for old approach, but I feel like that's an optimization that we can worry about later. wdyt?

jayy04: as for the gas behavior tests, created [this](https://linear.app/dydx/issue/CT-1238/write-tests-to-verify-gas-behavior-of-the-new-decorator) ticket to track

teddyding (approved):

PR #2403, opened 2024-09-30
Filter out to single tick per interval.

vincentwschau:
### Changelist
It's possible for a vault to be instantiated in the middle of the day/hour, and start having PnL ticks in the middle of the day/hour. This breaks the aggregation logic for PnL ticks as it introduces an extra PnL tick in the middle of the day / hour for the megavault.
This change adds a filtering function that only takes the aggregated PnL ticks with block times closest to the start of the day / hour.

### Test Plan
Unit tests.

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration parameter for vault profit and loss history, allowing for more granular control.
	- Enhanced functionality for fetching historical profit and loss data with improved resolution handling and filtering logic.

- **Bug Fixes**
	- Updated test cases to ensure accurate handling of historical profit and loss data for vault subaccounts, accommodating multiple vaults and undefined values.

### DIFF
--- a/indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
@@ -517,11 +517,7 @@ function filterOutIntervalTicks(
   return Array.from(ticksPerInterval.values());
 }
 
-function getResolution(resolution?: PnlTickInterval) {
-  if (resolution === undefined) {
-    return PnlTickInterval.day;
-  }
-
+function getResolution(resolution: PnlTickInterval = PnlTickInterval.day) {
   return resolution;
 }
 
--- a/indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
+++ b/indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
@@ -474,8 +474,9 @@ function getPnlTicksWithCurrentTick(
 
 /**
  * Takes in a map of block heights to PnlTicks and filters out the closest pnl tick per interval.
- * @param pnlTicksByBlock
- * @returns
+ * @param pnlTicksByBlock Map of block number to pnl tick.
+ * @param resolution Resolution of interval.
+ * @returns Array of PnlTicksFromDatabase, one per interval.
  */
 function filterOutIntervalTicks(
   pnlTicksByBlock: Map<number, PnlTicksFromDatabase>,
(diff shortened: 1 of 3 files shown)


indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
```diff
@@ -294,21 +295,22 @@ router.get(
...
 ): Promise<PnlTicksFromDatabase[]> {
   if (vaultSubaccountIds.length === 0) {
     return [];
   }
-  let pnlTickInterval: PnlTickInterval;
-  if (resolution === undefined) {
-    pnlTickInterval = PnlTickInterval.day;
+
+  let windowSeconds: number;
+  if (resolution === PnlTickInterval.day) {
```
tqin7, line 305: nit: maybe define a function that converts from resolution to seconds? thinking about the possibility of adding more resolution values in the future

indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
```diff
@@ -470,6 +472,55 @@ function getPnlTicksWithCurrentTick(
...
+  // Track start of days to closest block by block time.
+  const blocksPerInterval: Map<string, number> = new Map();
+  // Track start of days to closest Pnl tick.
+  const ticksPerInterval: Map<string, PnlTicksFromDatabase> = new Map();
+  pnlTicksByBlock.forEach((pnlTick: PnlTicksFromDatabase, block: number): void => {
+    const blockTime: DateTime = DateTime.fromISO(pnlTick.blockTime).toUTC();
+    blockToBlockTime.set(block, blockTime);
+
+    const startOfInterval: DateTime = blockTime.toUTC().startOf(resolution);
+    const startOfIntervalBlock: number | undefined = blocksPerInterval.get(startOfInterval.toISO());
```
tqin7, line 495: nit: not sure how much work this call is. if notable, define `startOfInterval.toISO()` as a const so we don't call it many times?

tqin7 (approved):

indexer/services/comlink/src/controllers/api/v4/vault-controller.ts
vincentwschau: This doesn't work with the underlying SQL query.

vincentwschau:
>how are pnl ticks handled with generic subaccounts (what happens if a subaccount gets transferred to for the first time in the middle of a day/hour)?

A pnl tick only gets created at each hour.

tqin7 (approved):

vincentwschau:
@Mergifyio backport release/indexer/v7.x

PR #2412, opened 2024-09-30
[CT-1202] logic to handle unauthorized maker orders when authenticato…

jayy04:
…rs are removed

### Changelist

### Test Plan

### Author/Reviewer Checklist

## Release Notes

- **New Features**
  - Introduced a method for validating transaction authenticators, enhancing transaction security.
  - Added functionality to retrieve transaction bytes for short-term orders.
  - Expanded account management capabilities with a new interface for smart accounts.
  - Integrated `AccountPlusKeeper` for improved account management across modules.

- **Improvements**
  - Refined order handling logic and improved checks for maker orders.
  - Enhanced testing coverage for order placements, liquidations, and state management.
  - Standardized return value handling across multiple methods for better error management.

- **Bug Fixes**
  - Added error handling for transactions with multiple signers to prevent processing errors.
  - Improved logging for daemon registration failures to enhance monitoring.

- **Documentation**
  - Updated interfaces and method signatures to reflect new functionalities and improve clarity.

### DIFF
--- a/protocol/x/clob/keeper/authenticators.go
+++ b/protocol/x/clob/keeper/authenticators.go
@@ -1,0 +1,73 @@
+package keeper
+
+import (
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	accountpluslib "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/lib"
+	aptypes "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// MaybeValidateAuthenticators checks if the transaction has authenticators specified and if so,
+// validates them. It returns an error if the authenticators are not valid or removed from state.
+func (k Keeper) MaybeValidateAuthenticators(ctx sdk.Context, txBytes []byte) error {
+	// Decode the tx from the tx bytes.
+	tx, err := k.txDecoder(txBytes)
+	if err != nil {
+		return err
+	}
+
+	// Check if the tx had authenticator specified.
+	specified, txOptions := accountpluslib.HasSelectedAuthenticatorTxExtensionSpecified(tx, k.cdc)
+	if !specified {
+		return nil
+	}
+
+	// The tx had authenticators specified.
+	// First make sure smart account flow is enabled.
+	if active := k.accountPlusKeeper.GetIsSmartAccountActive(ctx); !active {
+		return aptypes.ErrSmartAccountNotActive
+	}
+
+	// Make sure txn is a SigVerifiableTx and get signers from the tx.
+	sigVerifiableTx, ok := tx.(authsigning.SigVerifiableTx)
+	if !ok {
+		return errorsmod.Wrap(sdkerrors.ErrTxDecode, "invalid transaction type")
+	}
+
+	signers, err := sigVerifiableTx.GetSigners()
+	if err != nil {
+		return err
+	}
+
+	if len(signers) != 1 {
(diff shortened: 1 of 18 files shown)


protocol/x/clob/keeper/authenticators.go
```diff
@@ -0,0 +1,73 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	accountpluslib "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/lib"
+	aptypes "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// MaybeValidateAuthenticators checks if the transaction has authenticators specified and if so,
+// validates them. It returns an error if the authenticators are not valid or removed from state.
+func (k Keeper) MaybeValidateAuthenticators(ctx sdk.Context, txBytes []byte) error {
```
teddyding, line 15: None of the logic here is specific to x/clob - should this function be a method in CLOB or Accountplus?

protocol/x/clob/keeper/authenticators.go
```diff
@@ -0,0 +1,73 @@
+package keeper
```
teddyding, line 1: For my understanding, this PR doesn't need to touch `ProcessOperations` explicitly, because it's already taken care of by calling `AnteHandler` on the orders right?

protocol/x/clob/keeper/authenticators.go
```diff
@@ -0,0 +1,73 @@
...
+	accountpluslib "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/lib"
+	aptypes "github.com/dydxprotocol/v4-chain/protocol/x/accountplus/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+// MaybeValidateAuthenticators checks if the transaction has authenticators specified and if so,
+// validates them. It returns an error if the authenticators are not valid or removed from state.
+func (k Keeper) MaybeValidateAuthenticators(ctx sdk.Context, txBytes []byte) error {
+	// Decode the tx from the tx bytes.
+	tx, err := k.txDecoder(txBytes)
```
teddyding, line 17: We need to decode maker orders for all fills (regardless of smart account) - do we think this is okay because there are not many fews?

protocol/x/clob/keeper/authenticators.go
```diff
@@ -0,0 +1,73 @@
...
+		return errorsmod.Wrap(types.ErrTxnHasMultipleSigners, "only one signer is allowed")
+	}
+
+	account := sdk.AccAddress(signers[0])
+
+	// Retrieve the selected authenticators from the extension and make sure they are valid, i.e. they
+	// are registered and not removed from state.
+	//
+	// Note that we only verify the existence of the authenticators here without actually
+	// runnning them. This is because all current authenticators are stateless and do not read/modify any states.
```
teddyding, line 55: Could we document the we are not running the authenticators due to deadlock concerns

protocol/x/clob/keeper/authenticators.go
jayy04: correct. 

again, the assumption here is that the light weight validation is sufficient because all current validations are stateless

protocol/x/clob/keeper/authenticators.go
jayy04: I was planning to look at frame graphs after this is merged and deployed and optimize only if needed. I feel like this might be okay since like you said, only a very small percentage of the orders get matched.

note that all crossing maker orders get decoded, not just those that generate a fill

protocol/x/clob/keeper/authenticators.go
teddyding: Correct, and `ProcessOperations` do run the anthenticators explicitly since we cannot trust the proposer

PR #2441, opened 2024-10-02
Return undefined from getOrderbookMidPriceMap

adamfraser:
### Changelist
Removes code that hits the OrderbookCache when generating candles. This change will be fixed in a later update


dydxwill (approved):

adamfraser:
@mergify backport release/indexer/v7.x

PR #2461, opened 2024-10-03
[CT-1262] add e2e tests for new auth flow failure cases

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Release Notes

- **New Features**
	- Added a comprehensive test suite for the `PlaceOrder` functionality within the permissioned context of the Central Limit Order Book (CLOB) module.
	- Enhanced authentication capabilities with multiple authenticators for more complex validation scenarios.

- **Bug Fixes**
	- Improved error handling in various authenticator methods to provide more context on failures.

- **Documentation**
	- Updated comments and documentation to reflect changes in the authentication process and error handling.


protocol/x/clob/e2e/permissioned_keys_test.go
```diff
@@ -0,0 +1,561 @@
...
+	Signers        []cryptotypes.PrivKey
+
+	ExpectedRespCode uint32
+	ExpectedLog      string
+}
+
+func TestPlaceOrder_PermissionedKeys_Failures(t *testing.T) {
+	config := []aptypes.SubAuthenticatorInitData{
+		{
+			Type:   "SignatureVerification",
```
teddyding, line 40: Nit (no blocking this PR): did we consider using defining an enum type for the authenticator types? Feels like a lot of wasted payload on these strings. Instead the payload could just be an int

protocol/x/clob/e2e/permissioned_keys_test.go
```diff
@@ -0,0 +1,561 @@
...
+					Block: 2,
+					Msgs: []TestSdkMsg{
+						{
+							Msg: clobtypes.NewMsgPlaceOrder(
+								testapp.MustScaleOrder(
+									constants.Order_Bob_Num0_Id11_Clob1_Buy5_Price40_GTB20,
+									testapp.DefaultGenesis(),
+								),
+							),
+							Authenticators: []uint64{0},
```
teddyding, line 101: Question (tangential to PR): is it correct that `AuthenticatorId` is [global](https://github.com/dydxprotocol/v4-chain/blob/eefff671ea1c29dfcf5ba174f345821089924e5d/protocol/x/accountplus/keeper/authenticators.go#L172), but authenticators are always stored under account prefix. Is the main motivation behind global ID so that we just need one ID counter?

protocol/x/clob/e2e/permissioned_keys_test.go
```diff
@@ -0,0 +1,561 @@
...
+	Msg            sdk.Msg
+	Authenticators []uint64
+	Fees           sdk.Coins
+	Gas            uint64
+	AccountNum     []uint64
+	SeqNum         []uint64
+	Signers        []cryptotypes.PrivKey
+
+	ExpectedRespCode uint32
+	ExpectedLog      string
```
teddyding, line 34: Could we add an `ExpectedGas` field?

protocol/x/clob/e2e/permissioned_keys_test.go
```diff
@@ -0,0 +1,561 @@
...
+							ExpectedRespCode: aptypes.ErrAnyOfVerification.ABCICode(),
+							ExpectedLog:      aptypes.ErrAnyOfVerification.Error(),
+						},
+					},
+				},
+			},
+			expectedOrderIdsInMemclob: map[clobtypes.OrderId]bool{
+				constants.Order_Bob_Num0_Id11_Clob1_Buy5_Price40_GTB20.OrderId: false,
+			},
+		},
```
teddyding, line 495: I think (?) we support multi-msg transaction + authenticator. Should we add a test where one of the messages were not authenticated successfully?

protocol/x/clob/e2e/permissioned_keys_test.go
jayy04: yeah this decision was from x/smart-account, and I think it was mostly just for simplicity (at least from what I've seen in the code)

protocol/x/clob/e2e/permissioned_keys_test.go
jayy04: hmm these are short term orders with infinite gas meters and also IIUC if a transaction fails check tx, it doesn't incur any gas?

I can add expected gas for success cases, for stateful orders

protocol/x/clob/e2e/permissioned_keys_test.go
jayy04: let me address this separately. created [this](https://linear.app/dydx/issue/CT-1287/use-enums-for-authenticator-types) ticket to keep track

protocol/x/clob/e2e/permissioned_keys_test.go
jayy04: Added!

protocol/x/clob/e2e/permissioned_keys_test.go
teddyding: Oh I was referring to failed transactions that are not ST orders (e.g. transfer/send) since those should still incur gas cost even if transaction failed.

teddyding (approved):

PR #2479, opened 2024-10-08
[CT-1262] add e2e tests for permissioned keys success cases

jayy04:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Release Notes

- **New Features**
	- Introduced a new decorator for managing public key events within transactions.
	- Added a gas consumption step during signature verification to optimize transaction processing.
	- Enhanced authentication capabilities with the addition of an `AuthenticatorManager`.
	- New test cases for successful order placements using permissioned keys.

- **Bug Fixes**
	- Improved transaction processing logic by restructuring ante decorators for clarity and maintainability.
	- Enhanced isolation in tests for affiliate registration to ensure accurate results.

- **Tests**
	- Enhanced test coverage for circuit breaker and permissioned key functionalities.
	- Added a new test for successful order placements using permissioned keys.


protocol/x/accountplus/ante/circuit_breaker.go
```diff
@@ -11,15 +11,15 @@ import (
...
+	authenticatorAnteHandlerFlow sdk.AnteHandler
+	originalAnteHandlerFlow      sdk.AnteHandler
 }
 
 // NewCircuitBreakerDecorator creates a new instance of CircuitBreakerDecorator with the provided parameters.
 func NewCircuitBreakerDecorator(
 	cdc codec.BinaryCodec,
-	auth sdk.AnteDecorator,
-	classic sdk.AnteDecorator,
+	auth sdk.AnteHandler,
```
teddyding, line 21: FMI, why is this needed?

teddyding, line 21: nit: 
```suggestion
	accountPlusAuth sdk.AnteHandler,
```

protocol/app/ante.go
```diff
@@ -124,15 +124,21 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 		),
 		sigVerification: accountplusante.NewCircuitBreakerDecorator(
 			options.Codec,
-			accountplusante.NewAuthenticatorDecorator(
-				options.Codec,
-				options.AccountplusKeeper,
-				options.AccountKeeper,
-				options.SignModeHandler,
+			sdk.ChainAnteDecorators(
```
teddyding, line 127: Did we make this ante handler instead of a singleton decorator mostly to be consistent with below (classic chain)

protocol/app/ante.go
```diff
@@ -124,15 +124,21 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
...
+					options.Codec,
+					options.AccountplusKeeper,
+					options.AccountKeeper,
+					options.SignModeHandler,
+				),
 			),
-			customante.NewSigVerificationDecorator(
-				options.AccountKeeper,
-				options.SignModeHandler,
+			sdk.ChainAnteDecorators(
```
teddyding, line 135: For my understanding why was it necessary to move `SetPubKey` and `SigGas` under this chain and thus not applicable to smart account flow?

protocol/x/clob/e2e/permissioned_keys_test.go
```diff
@@ -662,3 +662,758 @@ func TestPlaceOrder_PermissionedKeys_Failures(t *testing.T) {
...
+			expectedOrderIdsInMemclob: map[clobtypes.OrderId]bool{
+				constants.Order_Bob_Num0_Id11_Clob1_Buy5_Price40_GTB20.OrderId:       false,
+				constants.Order_Alice_Num0_Id1_Clob1_Sell5_Price15_GTB20_IOC.OrderId: false,
+			},
+			expectedOrderFillAmounts: map[clobtypes.OrderId]uint64{
+				constants.Order_Bob_Num0_Id11_Clob1_Buy5_Price40_GTB20.OrderId:       0,
+				constants.Order_Alice_Num0_Id1_Clob1_Sell5_Price15_GTB20_IOC.OrderId: 0,
+			},
+		},
+		"Stateful maker order can be matched even if permissioned key is removed": {
```
teddyding, line 1241: Could we provide some context on these behaviors when the order is removed? My understanding (please CWIW): 
- For ST orders, we HAVE to remove the maker order because we can't trust BP on this maker order being valid
- For stateful orders, we don't HAVE TO remove the maker order because we can trust the onchain state. We MAY choose to remove the maker stateful order, but needs additional work (e.g. prune orders when removing authenticators)

protocol/x/accountplus/authenticator/signature_authenticator.go
```diff
@@ -56,6 +56,11 @@ func (sva SignatureVerification) Initialize(config []byte) (types.Authenticator,
 // Authenticate takes a SignaturesVerificationData struct and validates
 // each signer and signature using signature verification
 func (sva SignatureVerification) Authenticate(ctx sdk.Context, request types.AuthenticationRequest) error {
+	// First consume gas for verifying the signature
+	params := sva.ak.GetParams(ctx)
+	ctx.GasMeter().ConsumeGas(params.SigVerifyCostSecp256k1, "secp256k1 signature verification")
```
jayy04, line 61: added gas here @teddyding .

we currently only support secp256k1 signature verification so charging some static gas here.

protocol/app/ante.go
jayy04: SetPubKey expects the signer to match the public key ([link](https://github.com/dydxprotocol/cosmos-sdk/blob/dydx-fork-v0.50.5/x/auth/ante/sigverify.go#L92-L96)), and I was getting some errors here. 

After moving `SetPubKey`, I was getting some `public key = nil` errors in `SigGas` [here](https://github.com/dydxprotocol/cosmos-sdk/blob/dydx-fork-v0.50.5/x/auth/ante/sigverify.go#L455) because public key is not set. So moved this as well.

This is also how osmosis does it ([link](https://github.com/osmosis-labs/osmosis/blob/2d938d0d31ea35c11286a46cc30f9a2e03997332/app/ante.go#L73))

protocol/app/ante.go
```diff
@@ -124,15 +124,21 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
...
+				ante.NewSetPubKeyDecorator(options.AccountKeeper),
+				ante.NewSigGasConsumeDecorator(options.AccountKeeper, options.SigGasConsumer),
```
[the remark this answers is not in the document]
jayy04: Added some gas [here](https://github.com/dydxprotocol/v4-chain/pull/2479#discussion_r1803803369). we only support secp256k1 so it's a static value.

protocol/x/clob/e2e/permissioned_keys_test.go
jayy04: yeah that's correct. for ST orders, it's not necessarily about trust, it's more about maintaining the invariant that operations queue is always valid.

protocol/app/ante/pubkey.go
```diff
@@ -0,0 +1,110 @@
+package ante
```
jayy04, line 1: @teddyding I added a new decorator that looks identical to `SetPubKey` but only emits events 

reference from Osmosis ([link](https://github.com/osmosis-labs/osmosis/blob/2d938d0d31ea35c11286a46cc30f9a2e03997332/app/ante.go#L87))

protocol/x/accountplus/authenticator/signature_authenticator.go
teddyding: Could we documment the fact that we only supports `secp256k1`? For example link to [here](https://github.com/dydxprotocol/v4-chain/blob/301b4385069765c443f1dc3816dd1229bf24ab0f/protocol/x/accountplus/authenticator/signature_authenticator.go#L93)

protocol/x/clob/e2e/permissioned_keys_test.go
teddyding: Thanks, to clarify I was suggesting to comment on the test cases about this difference in expected behavior between ST and stateful order

protocol/app/ante/pubkey.go
teddyding: Nice, how did you figure this was necessary?

jayy04: Was just looking at the `SetPubKey` decorator and also using Osmosis as reference.

teddyding (approved):

PR #2745, opened 2025-03-11
Skip persisting vault orders in ender

tqin7:
### Changelist
Ender
- set vault order events to be of type `skipped_event` and thus SQL scripts simply return dummy json upon processing them
- however, retain handlers as we still need to send vault order kafka events to vulcan

Vulcan
- do not look up vault orders in database

Note
- with this change, Ender and Vulcan each keeps a cache of vault addresses in memory
- Tried using a redis cache shared by Ender and Vulcan but processing latency is much higher than this approach mainly due to network calls

### Test Plan
- unit tests
- tested on staging

## Summary by CodeRabbit

- **New Features**
  - Enhanced vault order management with dedicated processing for order placements, cancellations, and updates, ensuring more robust and reliable order handling.
  - Introduced new models and utility functions to optimize the consistency and accuracy of vault order data.

- **Refactor**
  - Streamlined event processing by conditionally bypassing unnecessary operations for vault orders, improving overall performance and efficiency.


teddyding:
In general, have you tested the new FE behavior after this change?

indexer/services/ender/src/handlers/stateful-order/conditional-order-placement-handler.ts
```diff
@@ -29,6 +29,9 @@ export class ConditionalOrderPlacementHandler extends
 
   // eslint-disable-next-line @typescript-eslint/require-await
   public async internalHandle(resultRow: pg.QueryResultRow): Promise<ConsolidatedKafkaEvent[]> {
+    if (resultRow.order === undefined) {
```
teddyding, line 32: For my understand why is conditional order related to vault?

indexer/packages/postgres/src/loops/vault-refresher.ts
```diff
@@ -0,0 +1,55 @@
...
+  NodeEnv,
+} from '@dydxprotocol-indexer/base';
+import _ from 'lodash';
+
+import config from '../config';
+import * as VaultTable from '../stores/vault-table';
+import { Options, VaultFromDatabase } from '../types';
+import { startUpdateLoop } from './loopHelper';
+
+let vaultAddresses: Set<string> = new Set();
```
teddyding, line 12: Why are we declaring a singleton cache in Postgres? Shouldn't this live in the actual service (`ender` in this case?)

indexer/services/vulcan/src/handlers/order-remove-handler.ts
```diff
@@ -135,7 +136,9 @@ export class OrderRemoveHandler extends Handler {
       });
     }
 
-    if (this.isStatefulOrderCancelation(orderRemove)) {
+    // Skip vault orders as they are not persisted.
+    if (this.isStatefulOrderCancelation(orderRemove) &&
```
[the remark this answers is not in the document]
tqin7: for vault orders, need to skip logic in vulcan that looks up orders in database

indexer/services/ender/src/handlers/stateful-order/conditional-order-placement-handler.ts
tqin7: discussed offline. this is more of a hacky fix to conditional order handlers as conditional orders are also stateful order events.

think a cleaner way would be for example: expose `shouldSkipHandlers()` in `validator`. will include in follow-up commit

indexer/services/vulcan/src/helpers/orders.ts
```diff
@@ -0,0 +1,6 @@
+import { vaultRefresher } from '@dydxprotocol-indexer/postgres';
+import { IndexerOrderId } from '@dydxprotocol-indexer/v4-protos';
+
+export function isVaultOrder(orderId: IndexerOrderId): boolean {
+  return vaultRefresher.isVault(orderId.subaccountId!.owner!);
```
teddyding, line 5: Where is the vault cache initialized for `vulcan`?

indexer/services/ender/src/lib/block-processor.ts
```diff
@@ -232,8 +232,10 @@ export class BlockProcessor {
         ...this.block.events[eventProto.blockEventIndex],
         subtype: SKIPPED_EVENT_SUBTYPE,
       };
-      // Set handlers to empty array if event is to be skipped.
-      handlers = [];
+      // Set handlers to empty array if a non-stateful-order event is to be skipped.
```
teddyding, line 235: Need more context here - this is very hard for reader to follow

indexer/services/ender/src/lib/block-processor.ts
```diff
@@ -232,8 +232,10 @@ export class BlockProcessor {
...
+      if (eventProto.type !== DydxIndexerSubtypes.STATEFUL_ORDER) {
```
teddyding, line 236: This is also an empty pattern - we adding some obscure conditional chains that makes this code hard to reason about.

Seems like the validator should able to differentiate between `shouldExcludeEventForSql` and `shouldExcludEventForHandler`.

indexer/services/vulcan/src/handlers/order-place-handler.ts
```diff
@@ -109,8 +110,10 @@ export class OrderPlaceHandler extends Handler {
       // isn't updated.
       // For stateful and conditional orders, look the order up in the db for the createdAtHeight
       // and send any cached order updates for the stateful or conditional order
+      // Do not execute logic for vault orders as they are not persisted.
```
teddyding, line 113: Document downstream effect and why we think this is okay

indexer/services/vulcan/src/helpers/orders.ts
tqin7: good catch! added under `index.ts`

indexer/services/vulcan/src/handlers/order-place-handler.ts
tqin7: clarified the comment

indexer/packages/postgres/src/config.ts
```diff
@@ -36,7 +36,7 @@ export const postgresConfigSchema = {
   MARKET_REFRESHER_INTERVAL_MS: parseInteger({ default: 30_000 }), // 30 seconds
   LIQUIDITY_TIER_REFRESHER_INTERVAL_MS: parseInteger({ default: 30_000 }), // 30 seconds
   BLOCK_HEIGHT_REFRESHER_INTERVAL_MS: parseInteger({ default: 1_000 }), // 1 second
-  VAULT_REFRESHER_INTERVAL_MS: parseInteger({ default: 300_000 }), // 5 minutes
+  VAULT_REFRESHER_INTERVAL_MS: parseInteger({ default: 30_000 }), // 30 seconds
```
teddyding, line 39: Is this being used?

indexer/packages/postgres/src/lib/helpers.ts
```diff
@@ -1,14 +1,15 @@
...
+// Gets `goodTilBlockTime` from an ISO string.
 export function blockTimeFromIsoString(isoString: IsoString): number {
   const dateTime: DateTime = DateTime.fromISO(isoString, { zone: 'utc' });
   return Math.floor(dateTime.toMillis() / 1000);
 }
 
-export function getVaultAddresses(): string[] {
-  return vaultAddresses;
-}
+// Set of the addresses of vaults that quote on clob pairs 0 to 999.
```
teddyding, line 12: Could we add a TODO item to find a more permanent solution here, so we don't end up forgetting about this and run into issues with vault ID > 1000

indexer/services/ender/src/validators/stateful-order-validator.ts
```diff
@@ -237,9 +239,37 @@ export class StatefulOrderValidator extends Validator<StatefulOrderEventV1> {
...
-   * Skip order uuids in config env var.
+   * Skip SQL processing for
+   * - order uuids to be skipped in config env var.
+   * - vault stateful orders.
    */
-  public shouldExcludeEvent(): boolean {
+  public shouldSkipSql(): boolean {
+    const orderUUIDsToSkip: string[] = config.SKIP_STATEFUL_ORDER_UUIDS.split(',');
+    if (orderUUIDsToSkip.length === 0) {
+      return false;
```
teddyding, line 249: Does this prematurely return `false` regardless of if order is from `vault`, as long as `SKIP_STATEFUL_ORDER_UUIDS` is non-empty?
Additionally, we should reuse code related to `SKIP_STATEFUL_ORDER_UUIDS` between `shouldSkipSql` and `shouldSkipHandlers`

indexer/services/vulcan/src/handlers/order-place-handler.ts
```diff
@@ -109,8 +110,10 @@ export class OrderPlaceHandler extends Handler {
       // isn't updated.
       // For stateful and conditional orders, look the order up in the db for the createdAtHeight
       // and send any cached order updates for the stateful or conditional order
+      // No need to send cached order updates for vault orders as they are not cached or persisted.
```
teddyding, line 113: This isn't accurate right? Should say something like
"Since vault orders are not persisted by ender and therefore not cached here, we are unable to look up vault related orders. This is a conscious tradeoff to improve ender latency, and lead to <effect in Vulcan and downstream UX>". Let's apply this comment to all 3 handlers

indexer/packages/postgres/src/config.ts
tqin7: not used. will delete

indexer/services/vulcan/src/handlers/order-place-handler.ts
tqin7: thought about these more and updated which logic to skip in these handlers for vault orders. overall we should still cache order updates for vault orders (tested on FE too)

tqin7:

PR #2778, opened 2025-04-03
[CT-1354] Extend Order Proto for TWAP

anmolagrawal345:
### Changelist
TWAP orders will be supported through the existing order proto. What needs to be added:
- Twap (and suborder) order flags
- Sequence number in the OrderId
- Twap Order Configs

Let's add these to the proto and implement some order validation checks.

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced support for TWAP orders, enabling trades to be spread over time with configurable duration, interval, and slippage settings.
	- Enhanced order identification to clearly distinguish TWAP orders from standard orders.
- **Bug Fixes**
	- Improved validation logic for placing TWAP orders to ensure compliance with new constraints.
- **Tests**
	- Expanded validation test cases to ensure TWAP orders meet all defined constraints, improving overall order reliability.


proto/dydxprotocol/clob/order.proto
```diff
@@ -95,6 +95,33 @@ message LongTermOrderPlacement {
...
 }
 
+// TwapOrderPlacement represents the placement of a TWAP order in
+// the TWAP Order State. It will store the original parent TWAP order as
+// well as maintain the state of the remaining legs and quantums
+// to be executed.
+message TwapOrderPlacement {
+  Order order = 1 [ (gogoproto.nullable) = false ];
+  // The total number of legs to be executed.
+  uint32 total_legs = 2;
```
shrenujb, line 105: Do we need this? You should just be able to calculate this using duration/interval in the TWAP params

anmolagrawal345: it was more for convenience because I was having to calculate it everywhere

proto/dydxprotocol/clob/order.proto
```diff
@@ -95,6 +95,33 @@ message LongTermOrderPlacement {
...
+  // The number of quantums remaining to be executed.
+  uint64 remaining_quantums = 4;
+  // The block height at which the order was placed.
+  uint32 block_height = 5;
+}
+
+// TwapTriggerPlacement represents the placement of a TWAP suborder in
+// state. It stores the stateful order itself, the `BlockHeight` and
+// `TransactionIndex` at which the order was placed and triggered.
+message TwapTriggerPlacement {
```
shrenujb, line 117: Doesnt the trigger_block_time need to be before the OrderId for ordering purposes?

proto/dydxprotocol/clob/order.proto
```diff
@@ -95,6 +95,33 @@ message LongTermOrderPlacement {
...
+message TwapOrderPlacement {
+  Order order = 1 [ (gogoproto.nullable) = false ];
+  // The total number of legs to be executed.
+  uint32 total_legs = 2;
+  // The number of legs remaining to be executed.
+  uint32 remaining_legs = 3;
+  // The number of quantums remaining to be executed.
+  uint64 remaining_quantums = 4;
+  // The block height at which the order was placed.
+  uint32 block_height = 5;
```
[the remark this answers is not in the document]
anmolagrawal345: I originally was trying to follow the paradigm of the other OrderPlacements that have TransactionOrdering - but I think I agree we could take this out. How would we want to handle TWAPs on chain restart?

proto/dydxprotocol/clob/order.proto
```diff
@@ -215,6 +242,27 @@ message Order {
...
+
+  // Interval in seconds for each suborder to execute. Must be a
+  // whole number, a factor of the duration, greater than 30 seconds,
+  // and less than 3600 seconds (1 hour).
+  uint32 interval = 2;
+
+  // Slippage percentage for each suborder. This will be applied to
+  // the oracle price each time a suborder is triggered. Must be
+  // between 0 and 5000 (0% and 50%).
+  uint32 slippage_percent = 3;
```
shrenujb, line 265: Not sure if slippage is the right word for this. I think something like price_tolerance would be better

protocol/x/clob/types/message_place_order.go
```diff
@@ -84,8 +101,8 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
 		return errorsmod.Wrapf(ErrReduceOnlyDisabled, "reduce only orders must be short term IOC orders")
 	}
 
-	if msg.Order.Subticks == uint64(0) {
-		return errorsmod.Wrapf(ErrInvalidOrderSubticks, "order subticks cannot be 0")
+	if msg.Order.Subticks == uint64(0) && !msg.Order.IsTwapOrder() {
```
shrenujb, line 104: Can you add a comment for why we're excluding TWAP orders here?

proto/dydxprotocol/clob/order.proto
anmolagrawal345: oh I actually wasn't aware protobuf would handle that for us. Currently (in the next PR) I have a method (GetTWAPTriggerKey) that creates the key for us that starts with the timestamp.

protocol/x/clob/types/order_id_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // numOrderIdFlagsTestCases is set to 129 to verify that we run a test case where
 // `OrderFlags` is greater than one byte (proto varints are encoded with 7 bits per byte).
-const numOrderIdFlagsTestCases = 129
+const numOrderIdFlagsTestCases = 512
```
shrenujb, line 17: Not sure why the previous value was 128+1 but should the change be 512+1 too?

protocol/x/clob/types/message_place_order.go
```diff
@@ -109,5 +126,46 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
...
+				MaxTwapOrderDuration,
+			)
+		}
+		if parameters.Duration%parameters.Interval != 0 {
+			return errorsmod.Wrapf(
+				ErrInvalidPlaceOrder,
+				"TWAP order duration must be a multiple of the interval",
+			)
+		}
+		if parameters.SlippagePercent > MaxTwapOrderSlippagePercent {
```
shrenujb, line 161: For the price tolerance, the user can arbitrarily set this as high or low as they want. We just want to make sure it makes sense from a math perspective. 0 < slippage < 100

proto/dydxprotocol/clob/order.proto
shrenujb: I dont think we should add state which we can just calculate from other state variables. In this case, the calculation is also trivial

proto/dydxprotocol/clob/order.proto
shrenujb: > (GetTWAPTriggerKey) that creates the key for us that starts with the timestamp.

This makes sense to me. What is the TwapTriggerPlacement struct used for then?

proto/dydxprotocol/clob/order.proto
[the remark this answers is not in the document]
shrenujb: You already have the triggers in protocol state, so the end blocker would simply look at that to see if any twap suborders need to be triggered

proto/dydxprotocol/clob/order.proto
anmolagrawal345: yea that's fair I can remove it

proto/dydxprotocol/clob/order.proto
anmolagrawal345: actually I suppose we could remove it entirely and only depend on the keys in the store. I originally had twapTriggerPlacements containing the original order itself which was used as the template to create the suborder- but I since moved away to generating it myself (since most of the configs for the suborder are going to be constants). Going to give that a go and make the changes

protocol/x/clob/types/order_id.go
```diff
@@ -10,9 +10,11 @@ import (
...
 
 const (
-	OrderIdFlags_ShortTerm   = uint32(0)
-	OrderIdFlags_Conditional = uint32(32)
-	OrderIdFlags_LongTerm    = uint32(64)
+	OrderIdFlags_ShortTerm    = uint32(0)
+	OrderIdFlags_Conditional  = uint32(32)
+	OrderIdFlags_LongTerm     = uint32(64)
+	OrderIdFlags_Twap         = uint32(128)
+	OrderIdFlags_TwapSuborder = uint32(256)
```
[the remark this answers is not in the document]
anmolagrawal345: this comment seems outdated so I have updated it with a copy of the valid orderid flags

protocol/x/clob/types/message_place_order.go
```diff
@@ -109,5 +128,46 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
...
 
+	if msg.Order.IsTwapOrder() {
+		if msg.Order.TwapParameters == nil {
+			return errorsmod.Wrapf(
+				ErrInvalidPlaceOrder,
+				"TWAP order must have a TWAP config",
+			)
+		}
+		parameters := msg.Order.TwapParameters
+		if parameters.Interval < MinTwapOrderInterval ||
```
shrenujb, line 139: These should be inclusive of the bounds

protocol/x/clob/types/message_place_order.go
```diff
@@ -109,5 +128,46 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
...
+			parameters.Interval > MaxTwapOrderInterval {
+			return errorsmod.Wrapf(
+				ErrInvalidPlaceOrder,
+				"TWAP order interval must be between %d seconds and %d seconds",
+				MinTwapOrderInterval,
+				MaxTwapOrderInterval,
+			)
+		}
+		if parameters.Duration < MinTwapOrderDuration ||
```
shrenujb, line 148: Same here. These should be inclusive of the bounds

protocol/x/clob/types/message_place_order.go
```diff
@@ -109,5 +128,46 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
...
+				MaxTwapOrderDuration,
+			)
+		}
+		if parameters.Duration%parameters.Interval != 0 {
+			return errorsmod.Wrapf(
+				ErrInvalidPlaceOrder,
+				"TWAP order duration must be a multiple of the interval",
+			)
+		}
+		if parameters.PriceTolerance > MaxTwapOrderPriceTolerance {
```
shrenujb, line 163: wrong check

PR #2779, opened 2025-04-04
[CT-1353] Handle TWAP Order Placement

anmolagrawal345:
### Changelist
This PR depends on #2778
Now that the order proto supports twap orders (and required parameters) let's handle the initial twap order placement. Some key desired behaviors are:
- Parent TWAP orders are submitted and stored in the TWAPOrderPlacementStore. They are separate from the LongTermOrderPlacementStore because these orders are not actually placed on the orderbook, but rather just to maintain the state for the suborders.
- The first suborder is added to the TWAPTriggerPlacementStore for processing in the EndBlocker (later PR)
- EquityTier limits are done for parent TWAP orders but not suborders
- Collateral checks are done for both parent and sub-TWAP orders
- Parent TWAP orders are allowed to configure 0 subticks (unique to this order type). This signals that this twap is a market twap and suborders will not enforce a price floor/ceiling. When this is the case, for the parent twap order, let's use the oracle price as price for the collateral check as a best approximation.

### Test Plan
Unit testing for extending the functionality of handling MsgPlaceOrders.

### Author/Reviewer Checklist


protocol/x/clob/keeper/msg_server_place_order.go
```diff
@@ -128,6 +128,7 @@ func (k Keeper) HandleMsgPlaceOrder(
 			ctx,
 			order.OrderId,
 		)
+		// TODO: add indexer event here
 	} else {
```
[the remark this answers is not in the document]
anmolagrawal345: will be done in a later PR

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
```diff
@@ -672,6 +760,58 @@ export interface OrderSDKType {
...
+  interval: number;
+  /**
+   * Slippage percentage for each suborder. This will be applied to
+   * the oracle price each time a suborder is triggered. Must be
+   * between 0 and 5000 (0% and 50%).
+   */
+
+  slippagePercent: number;
+}
+/** TwapParameters represents the necessary configuration for a TWAP order. */
```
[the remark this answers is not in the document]
anmolagrawal345: these are handled in the dependent PR (#2778)

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,117 @@
...
+
+		if triggerPlacement.Order.OrderId.SubaccountId.Owner == orderId.SubaccountId.Owner &&
+			triggerPlacement.Order.OrderId.SubaccountId.Number == orderId.SubaccountId.Number &&
+			triggerPlacement.Order.OrderId.ClientId == orderId.ClientId &&
+			triggerPlacement.Order.OrderId.ClobPairId == orderId.ClobPairId {
+			triggerPlacements = append(triggerPlacements, &triggerPlacement)
+		}
+	}
+	return triggerPlacements, len(triggerPlacements) > 0
+}
```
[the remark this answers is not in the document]
anmolagrawal345: using a prefix iterator here is not possible because the keys are [time, orderId] to enforce time priority of suborder execution. This method is actually only used for convenience though (and could be deleted) but is not used in actual production execution logic.

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
```diff
@@ -495,6 +507,76 @@ export interface LongTermOrderPlacementSDKType {
...
+ */
+
+export interface TwapTriggerPlacement {
+  /** The generated suborder given a TWAP order. */
+  order?: Order;
+  /** The block height at which the order was triggered. */
+
+  triggerBlockTime: Long;
+}
+/**
```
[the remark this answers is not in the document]
anmolagrawal345: it's supposed to be trigger time. I have updated the comment

protocol/x/clob/types/order.go
```diff
@@ -241,3 +259,23 @@ func (o *Order) GetOrderLabels() []gometrics.Label {
 		o.OrderId.GetOrderIdLabels()...,
 	)
 }
+
+func (o *Order) GetTotalLegsTWAPOrder() uint32 {
+	if o.IsTwapOrder() {
+		return o.TwapParameters.Duration / o.TwapParameters.Interval
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
anmolagrawal345: this check isn't necessary considering it comes from places where the order has already passed validation. if an order is a twap, the validation exists in message_place_order

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,96 @@
+package keeper
+
+import (
+	"encoding/binary"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+func (k Keeper) SetTWAPOrderPlacement(ctx sdk.Context,
```
shrenujb, line 10: Would you want to update this to accepting remaining legs and quantums as a param since you'd have to update those with every suborder placement and fill?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,96 @@
...
+
+	k.cdc.MustUnmarshal(b, &val)
+	return val, true
+}
+
+// GetTwapTriggerPlacement gets a TWAP trigger placement for a given orderId.
+// Returns false if no trigger placement exists in store with `orderId`.
+// This iterates over the entire store because the keys in the store are
+// formatted as [timestamp, orderId].
+func (k Keeper) GetTwapTriggerPlacement(
```
shrenujb, line 52: Why do you need this function? for testing?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,96 @@
...
+	ctx sdk.Context,
+	suborderId types.OrderId,
+	triggerOffset int64,
+) {
+	triggerStore := k.GetTWAPTriggerOrderPlacementStore(ctx)
+	triggerTime := ctx.BlockTime().Unix() + triggerOffset
+
+	triggerKey := types.GetTWAPTriggerKey(triggerTime, suborderId)
+
+	// The value in the map is not used, so we can set it to an empty byte slice.
```
shrenujb, line 94: I thought the value in the map would point to the suborderId?

protocol/x/clob/keeper/twap_order_state_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/dydxprotocol/v4-chain/protocol/mocks"
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	keepertest "github.com/dydxprotocol/v4-chain/protocol/testutil/keeper"
+	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTWAPOrderTriggerStoreOrderingSameOrderId(t *testing.T) {
```
shrenujb, line 17: Can we technically run into this case? I would think you would only have 1 entry for a suborder with the same Id

protocol/x/clob/keeper/twap_order_state_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// Create test orders with different timestamps and order IDs
+	suborderIds := []types.OrderId{
+		{
+			SubaccountId: constants.Alice_Num0,
+			ClientId:     0,
+			OrderFlags:   types.OrderIdFlags_TwapSuborder,
+			ClobPairId:   0,
+		},
+		{
```
shrenujb, line 65: this is the same as above?

protocol/x/clob/keeper/orders.go
```diff
@@ -1018,6 +1042,22 @@ func (k Keeper) MustValidateReduceOnlyOrder(
 	return nil
 }
 
+func (k Keeper) GetSubticksForCollatCheck(ctx sdk.Context, order types.Order) (types.Subticks, error) {
```
shrenujb, line 1045: Could you add a function description here describing why we do what we do?

protocol/x/clob/keeper/orders.go
```diff
@@ -1018,6 +1042,22 @@ func (k Keeper) MustValidateReduceOnlyOrder(
...
+	if order.IsTwapOrder() && order.Subticks == uint64(0) {
+		// if twap market order, use the current oracle price as subticks
+		clobPairId := order.GetClobPairId()
+		clobPair, found := k.GetClobPair(ctx, clobPairId)
+		if !found {
+			return 0, types.ErrInvalidClob
+		}
+		oraclePriceSubticksRat := k.GetOraclePriceSubticksRat(ctx, clobPair)
+		orderSubticks := lib.BigRatRound(oraclePriceSubticksRat, false).Uint64()
```
shrenujb, line 1054: is this the lib we use in other places for oracle subticks as well?

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: I have methods in the next PR to handle those cases. Since we have to decrement the legs after every trigger, but only update the quantums by fill, I split them up which made it easier to use.

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: Yea actually ended up being primarily for testing to ensure the store is exactly what we expect. I could take it out if we prefer though and handle it in the test fixtures

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: This was based off our earlier discussion where we had the TWAPTriggerPlacement struct, but then realized there was no need because I am generating all the order fields on placement. And the key itself needs to be timestamp x suborderId to avoid clashes. So i just store an empty value in the map to maintain ordering, and decode the key value on suborder generation.

protocol/x/clob/keeper/twap_order_state_test.go
anmolagrawal345: Technically not actually - I was mainly testing that the timestamp based ordering was working as intended all else being equal. The test name is a bit misleading though I can change that

protocol/x/clob/keeper/twap_order_state_test.go
anmolagrawal345: No this one is given the same timestamp but subtly different orderIds (clobPair IDs are diff) the ordering is what we expect. The first 2 are the same orderId but different timestamps. The other 2 are diff clobPair and diff subaccountId

protocol/x/clob/keeper/orders.go
anmolagrawal345: yep I based this off the usage of this method in other places - ie in clob_pair.go:495 and mev.go:363

protocol/x/clob/keeper/twap_order_state.go
shrenujb: Fine to keep it here. Would be nice to add a comment mentioning this is only being used for testing

protocol/x/clob/keeper/orders.go
shrenujb: func description still missing?

protocol/x/clob/keeper/twap_order_state_test.go
shrenujb: I guess this is the same thing as the above case where we shouldn't have the same orderId multiple times in this store?

anmolagrawal345: Yea we won't this was just testing the timestamp + orderId ordering in the keystore. I added a comment clarifying that though

shrenujb (approved):

PR #2780, opened 2025-04-08
[CT-1355] Handle Suborders in End Blocker

anmolagrawal345:
### Changelist
This PR is dependent on #2778 and #2779.

Now that we have twap orders being stored upon placement, let's add logic to the end blocker to manage them. Using the TWAPTriggerStore, we can loop over the upcoming suborders (in order of timestamp) to determine if they should be placed. For each such suborder, we generate a new order based off the state stored for the parent twap, market price, and current block time.

Currently, this does not handle TWAP order cancellations and expects the parent twap order to be present in the state when looking for it (otherwise it panics). This will be handled in the next change.

### Test Plan
Unit testing and new end-to-end integration tests to test the end blocker functionality.

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive support for Time-Weighted Average Price (TWAP) orders, including automated suborder generation, placement, and lifecycle management.
  - Added new error messages for invalid TWAP order placements.
  - Enhanced order placement commands to explicitly flag short-term orders.
  - Added functionality to adjust oracle prices by configurable price tolerance expressed in parts per million (ppm).
  - Added new rounding functions to improve numerical precision in order calculations.

- **Bug Fixes**
  - Improved validation and state management for TWAP orders and suborders, ensuring correct triggering, placement, and matching behavior.

- **Documentation**
  - Clarified documentation for price tolerance fields, specifying units as parts per million (ppm) and updating valid ranges for increased precision.
  - Updated constant values and comments for better clarity and accuracy.

- **Tests**
  - Added extensive end-to-end and unit tests for TWAP order placement, suborder generation, and matching scenarios to ensure robust functionality.

- **Refactor**
  - Modularized order trigger key encoding/decoding and improved code readability with helper functions and digit separators in constants.

### DIFF
--- a/indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
+++ b/indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
@@ -731,9 +731,9 @@ export interface TwapParameters {
 
   interval: number;
   /**
-   * Price tolerance for each suborder. This will be applied to
+   * Price tolerance in ppm for each suborder. This will be applied to
    * the oracle price each time a suborder is triggered. Must be
-   * be in the range [0, 10000).
+   * be in the range [0, 1_000_000).
    */
 
   priceTolerance: number;
@@ -754,9 +754,9 @@ export interface TwapParametersSDKType {
 
   interval: number;
   /**
-   * Price tolerance for each suborder. This will be applied to
+   * Price tolerance in ppm for each suborder. This will be applied to
    * the oracle price each time a suborder is triggered. Must be
-   * be in the range [0, 10000).
+   * be in the range [0, 1_000_000).
    */
 
   price_tolerance: number;
--- a/protocol/lib/quantums.go
+++ b/protocol/lib/quantums.go
@@ -88,3 +88,25 @@ func QuoteToBaseQuantums(
 
 	return result
 }
+
+// BigRatRoundToMultiple rounds a big.Rat to the nearest multiple of the given number.
+// If roundUp is true, it rounds up to the nearest multiple, otherwise it rounds down.
+func BigRatRoundToMultiple(
+	value *big.Rat,
+	multiple *big.Int,
+	roundUp bool,
+) *big.Int {
+	// Convert the value to a big.Int
+	valueInt := new(big.Int).Div(value.Num(), value.Denom())
+
+	// Calculate the remainder
+	remainder := new(big.Int).Mod(valueInt, multiple)
+
+	if roundUp && remainder.Sign() != 0 {
+		valueInt.Add(valueInt, new(big.Int).Sub(multiple, remainder))
+	} else {
+		valueInt.Sub(valueInt, remainder)
+	}
+
+	return valueInt
+}
--- a/protocol/x/clob/keeper/orders.go
+++ b/protocol/x/clob/keeper/orders.go
@@ -971,15 +971,11 @@ func (k Keeper) PerformStatefulOrderValidation(
 	}
 
 	if order.IsTwapOrder() {
-		totalLegs := order.GetTotalLegsTWAPOrder()
-		minOrderSize := uint64(totalLegs) * clobPair.StepBaseQuantums
-		if order.Quantums < minOrderSize {
+		num_suborders := uint64(order.TwapParameters.Duration / order.TwapParameters.Interval)
+		if order.Quantums/num_suborders < clobPair.StepBaseQuantums {
 			return errorsmod.Wrapf(
 				types.ErrInvalidPlaceOrder,
-				"Generated TWAP suborder sizes (%v/%v) must be greater than min size for clob pair %v",
-				order.Quantums,
-				totalLegs,
-				clobPair.StepBaseQuantums,
+				"TWAP suborder sizes must be greater than the minimum order size for the market",
 			)
 		}
 	}
@@ -1238,6 +1234,33 @@ func (k Keeper) InitStatefulOrders(
 			telemetry.IncrCounter(1, types.ModuleName, metrics.PlaceOrder, metrics.Hydrate, metrics.Matched)
 		}
 	}
+}
+
+// GetOraclePriceAdjustedByPercentageSubticks returns the oracle price in subticks
+// adjusted by a given directional price tolerance in ppm, rounded to the nearest multiple
+// of SubticksPerTick. A positive price tolerance increases the price, while a negative price
+// tolerance decreases it.
+//
+// For example:
+//   - price tolerance = 500_000 means 50% higher than oracle price
+//   - price tolerance = -500_000 means 50% lower than oracle price
+func (k Keeper) GetOraclePriceAdjustedByPercentageSubticks(
+	ctx sdk.Context,
+	clobPair types.ClobPair,
+	directionalPriceTolerancePpm int32,
+) uint64 {
+	oraclePriceSubticksRat := k.GetOraclePriceSubticksRat(ctx, clobPair)
+	adjustment := int32(1_000_000) + directionalPriceTolerancePpm
+
+	adjustedPrice := lib.BigRatMulPpm(oraclePriceSubticksRat, uint32(adjustment))
+	// Round to the nearest multiple of SubticksPerTick
+	roundedSubticks := lib.BigRatRoundToMultiple(
+		adjustedPrice,
+		new(big.Int).SetUint64(uint64(clobPair.SubticksPerTick)),
+		directionalPriceTolerancePpm >= 0, // round up for positive adjustments, down for negative
+	)
+
+	return roundedSubticks.Uint64()
 }
 
 // sendOffchainMessagesWithTxHash sends all the `Message` in the offchainUpdates passed in along with
--- a/protocol/x/clob/keeper/process_single_match.go
+++ b/protocol/x/clob/keeper/process_single_match.go
@@ -293,6 +293,29 @@ func (k Keeper) ProcessSingleMatch(
 		curMakerPruneableBlockHeight,
 	)
 
+	if !matchWithOrders.TakerOrder.IsLiquidation() {
+		taker_order := matchWithOrders.TakerOrder.MustGetOrder()
+		if taker_order.OrderId.IsTwapSuborder() {
+			log.ErrorLog(ctx, "handling twap order", fmt.Sprintf("%+v", taker_order))
+			// Get the parent order ID by removing the TWAP suborder flag
+			parentOrderId := types.OrderId{
+				SubaccountId: taker_order.OrderId.SubaccountId,
+				ClientId:     taker_order.OrderId.ClientId,
+				OrderFlags:   types.OrderIdFlags_Twap, // Set directly to TWAP
+				ClobPairId:   taker_order.OrderId.ClobPairId,
+			}
+
+			// Update the parent TWAP order state
+			if err := k.UpdateTWAPOrderRemainingQuantityOnFill(
+				ctx,
+				parentOrderId,
+				fillAmount.ToUint64(),
+			); err != nil {
+				return false, takerUpdateResult, makerUpdateResult, affiliateRevSharesQuoteQuantums, err
+			}
+		}
+	}
+
 	return true, takerUpdateResult, makerUpdateResult, affiliateRevSharesQuoteQuantums, nil
 }
 
--- a/protocol/x/clob/keeper/process_single_match.go
+++ b/protocol/x/clob/keeper/process_single_match.go
@@ -294,14 +294,14 @@ func (k Keeper) ProcessSingleMatch(
 	)
 
 	if !matchWithOrders.TakerOrder.IsLiquidation() {
-		taker_order := matchWithOrders.TakerOrder.MustGetOrder()
-		if taker_order.OrderId.IsTwapSuborder() {
+		takerOrder := matchWithOrders.TakerOrder.MustGetOrder()
+		if takerOrder.OrderId.IsTwapSuborder() {
 			// Get the parent order ID by removing the TWAP suborder flag
 			parentOrderId := types.OrderId{
-				SubaccountId: taker_order.OrderId.SubaccountId,
-				ClientId:     taker_order.OrderId.ClientId,
+				SubaccountId: takerOrder.OrderId.SubaccountId,
+				ClientId:     takerOrder.OrderId.ClientId,
 				OrderFlags:   types.OrderIdFlags_Twap, // Set directly to TWAP
-				ClobPairId:   taker_order.OrderId.ClobPairId,
+				ClobPairId:   takerOrder.OrderId.ClobPairId,
 			}
 
 			// Update the parent TWAP order state
--- a/protocol/x/clob/keeper/process_single_match.go
+++ b/protocol/x/clob/keeper/process_single_match.go
@@ -296,7 +296,6 @@ func (k Keeper) ProcessSingleMatch(
 	if !matchWithOrders.TakerOrder.IsLiquidation() {
 		taker_order := matchWithOrders.TakerOrder.MustGetOrder()
 		if taker_order.OrderId.IsTwapSuborder() {
-			log.ErrorLog(ctx, "handling twap order", fmt.Sprintf("%+v", taker_order))
 			// Get the parent order ID by removing the TWAP suborder flag
 			parentOrderId := types.OrderId{
 				SubaccountId: taker_order.OrderId.SubaccountId,
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -1,10 +1,20 @@
 package keeper
 
 import (
-	"encoding/binary"
+	"math/big"
 
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+const (
+	TWAP_SUBORDER_GOOD_TIL_BLOCK_TIME_OFFSET = 3
+)
+
+var (
+	TWAP_MAX_SUBORDER_CATCHUP_MULTIPLE = big.NewInt(3)
 )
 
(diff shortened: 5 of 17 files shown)


protocol/lib/quantums.go
```diff
@@ -88,3 +88,25 @@ func QuoteToBaseQuantums(
...
+	remainder := new(big.Int).Mod(valueInt, multiple)
+
+	if roundUp && remainder.Sign() != 0 {
+		valueInt.Add(valueInt, new(big.Int).Sub(multiple, remainder))
+	} else {
+		valueInt.Sub(valueInt, remainder)
+	}
+
+	return valueInt
+}
```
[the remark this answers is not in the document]
anmolagrawal345: this will never be negative because it is using the clobPair.SubticksPerTick which is always positive

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
```diff
@@ -672,6 +708,58 @@ export interface OrderSDKType {
...
+
+  interval: number;
+  /**
+   * Price tolerance for each suborder. This will be applied to
+   * the oracle price each time a suborder is triggered. Must be
+   * between 0 and 10000 (0% and 100%).
+   */
+
+  price_tolerance: number;
 }
```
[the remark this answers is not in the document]
anmolagrawal345: this is handled via message_place_order in the ValidateBasic method

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,285 @@
...
+
+func (k Keeper) SetTWAPOrderPlacement(ctx sdk.Context,
+	order types.Order,
+	blockHeight uint32,
+) {
+	store := k.GetTWAPOrderPlacementStore(ctx)
+	orderKey := order.OrderId.ToStateKey()
+
+	total_legs := order.TwapParameters.Duration / order.TwapParameters.Interval
+
```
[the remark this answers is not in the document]
anmolagrawal345: message validations are done in message_place_order

protocol/x/clob/keeper/twap_order_state_test.go
```diff
@@ -0,0 +1,120 @@
...
+				Quantums: 1000,
+				TwapParameters: &types.TwapParameters{
+					Duration: 300, // 5 minutes
+					Interval: 60,  // 1 minute
+				},
+			},
+			blockHeight:       100,
+			expectedTotalLegs: 5,
+			expectedQuantums:  1000,
+		},
```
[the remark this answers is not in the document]
anmolagrawal345: these values are invalid and will be handled by the message validations in message_place_order

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,286 @@
...
+			SubaccountId:   orderId.SubaccountId,
+			ClientId:       orderId.ClientId,
+			OrderFlags:     types.OrderIdFlags_Twap, // Set directly to TWAP
+			ClobPairId:     orderId.ClobPairId,
+		}
+
+		twapOrderPlacement, found := k.GetTwapOrderPlacement(ctx, parentOrderId)
+		if !found {
+			panic("parent twap order not found") // TODO: (anmol) handle order cancellation
+		}
```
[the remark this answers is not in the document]
anmolagrawal345: this will be handled in a follow up PR dedicated to cancellations and expiries of TWAP orders

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,287 @@
...
+		parentOrderId := types.OrderId{
+			SubaccountId: orderId.SubaccountId,
+			ClientId:     orderId.ClientId,
+			OrderFlags:   types.OrderIdFlags_Twap, // Set directly to TWAP
+			ClobPairId:   orderId.ClobPairId,
+		}
+
+		twapOrderPlacement, found := k.GetTwapOrderPlacement(ctx, parentOrderId)
+		if !found {
+			panic("parent twap order not found") // TODO: (anmol) handle order cancellation
```
shrenujb, line 137: We never want to panic in end blocker. This will lead to a chain halt

anmolagrawal345: yep - this specific in the next PR which deals with cancellations/expiries, but I can remove the panic here for safety (and will check the cancels/expiry PR for the same concern)

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,287 @@
...
+	)
+
+	// Round down to nearest multiple of StepBaseQuantums
+	quantumsByStepBaseQuantums := uint64(math.Floor(suborderQuantums / float64(clobPair.StepBaseQuantums)))
+	suborderQuantumsRounded := quantumsByStepBaseQuantums * clobPair.StepBaseQuantums
+
+	return suborderQuantumsRounded
+}
+
+func (k Keeper) GenerateSuborder(
```
shrenujb, line 255: func description

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,287 @@
...
+	twapOrderPlacement.RemainingLegs--
+
+	// Store updated state
+	store := k.GetTWAPOrderPlacementStore(ctx)
+	orderKey := twapOrderPlacement.Order.OrderId.ToStateKey()
+	twapOrderPlacementBytes := k.cdc.MustMarshal(twapOrderPlacement)
+	store.Set(orderKey, twapOrderPlacementBytes)
+}
+
+func (k Keeper) UpdateTWAPOrderStateOnSuborderFill(
```
shrenujb, line 203: func description

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,287 @@
...
+	triggerStore := k.GetTWAPTriggerOrderPlacementStore(ctx)
+	triggerTime := ctx.BlockTime().Unix() + triggerOffset
+
+	triggerKey := types.GetTWAPTriggerKey(triggerTime, suborderId)
+
+	// The value in the map is not used, so we can set it to an empty byte slice.
+	triggerStore.Set(triggerKey, []byte{})
+}
+
+func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
```
shrenujb, line 106: func description

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,287 @@
...
+	iterator.Close()
+
+	for _, op := range operationsToProcess {
+		// Delete from trigger store
+		triggerStore.Delete(op.keyToDelete)
+
+		// decrement remaining legs
+		k.DecrementTwapOrderRemainingLegs(ctx, &op.twapOrderPlacement)
+
+		if op.twapOrderPlacement.RemainingLegs == 0 {
```
shrenujb, line 161: Should we do this operation at the bottom of the loop? Does it matter?

anmolagrawal345: you're right ya - I actually rewrote this logic in the subsequent PR which explicitly handles cancellations/expiries. But agreed this should definitely be after the suborder is placed to make sure we place the last one!

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +102,202 @@
...
+			OrderFlags:   types.OrderIdFlags_Twap, // Set directly to TWAP
+			ClobPairId:   orderId.ClobPairId,
+		}
+
+		twapOrderPlacement, found := k.GetTwapOrderPlacement(ctx, parentOrderId)
+		if !found {
+			// TODO: (anmol) handle order cancellation
+			return
+		}
+
```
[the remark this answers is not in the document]
anmolagrawal345: this is just a placeholder until the next PR which is going to explicitly handle this case

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +102,202 @@
...
+		k.DecrementTwapOrderRemainingLegs(ctx, &op.twapOrderPlacement)
+
+		// add the next suborder to the trigger store
+		k.AddSuborderToTriggerStore(
+			ctx,
+			op.suborderToPlace.OrderId,
+			int64(op.twapOrderPlacement.Order.TwapParameters.Interval),
+		)
+
+		if op.twapOrderPlacement.RemainingLegs == 0 {
```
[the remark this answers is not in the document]
anmolagrawal345: we want to fire off the final suborder. in this part of the code, the last suborder has been generated and needs to be released

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +102,202 @@ func (k Keeper) AddSuborderToTriggerStore(
...
+	if !found {
+		return errorsmod.Wrapf(
+			types.ErrInvalidTwapOrderPlacement,
+			"TWAP order %+v does not exist",
+			parentOrderId,
+		)
+	}
+
+	// Update remaining quantums
+	twapOrderPlacement.RemainingQuantums -= filledQuantums
```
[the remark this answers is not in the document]
anmolagrawal345: this case is not possible because the quantums that are set for every suborder will be capped at the remaining quantums, so we will never get filled for more than the configured amount

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +104,210 @@ func (k Keeper) AddSuborderToTriggerStore(
...
+			SubaccountId: orderId.SubaccountId,
+			ClientId:     orderId.ClientId,
+			OrderFlags:   types.OrderIdFlags_Twap, // Set directly to TWAP
+			ClobPairId:   orderId.ClobPairId,
+		}
+
+		twapOrderPlacement, found := k.GetTwapOrderPlacement(ctx, parentOrderId)
+		if !found {
+			// TODO: (anmol) handle order cancellation
+			return
```
shrenujb, line 147: If the orig twap order was not found, you should continue to the other triggers

anmolagrawal345: yea its actually continue in the subsequent PR when cancels are actually handled. I put a return here as a sort of fail softly mechanism but can change it to also be continue

protocol/x/clob/keeper/twap_order_state.go
shrenujb: Are you going to fix it in this PR or should I review the next for this?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +104,210 @@ func (k Keeper) AddSuborderToTriggerStore(
...
+	}
+
+	priceTolerancePpm := int32(parentOrder.TwapParameters.PriceTolerance)
+	if parentOrder.Side == types.Order_SIDE_SELL {
+		// for sell orders, we want to adjust the price down
+		priceTolerancePpm = -priceTolerancePpm
+	}
+
+	// calculate the suborder price with slippage adjustment
+	clobPair := k.mustGetClobPair(ctx, parentOrder.GetClobPairId())
```
shrenujb, line 301: I think this panics as well. For the stuff thats going to panic, I think we should err log it instead and alert on it
@teddyding whats the usual practice for things like this in the endblocker?

protocol/x/clob/keeper/orders.go
```diff
@@ -1240,6 +1236,33 @@ func (k Keeper) InitStatefulOrders(
...
+// tolerance decreases it.
+//
+// For example:
+//   - price tolerance = 500_000 means 50% higher than oracle price
+//   - price tolerance = -500_000 means 50% lower than oracle price
+func (k Keeper) GetOraclePriceAdjustedByPercentageSubticks(
+	ctx sdk.Context,
+	clobPair types.ClobPair,
+	directionalPriceTolerancePpm int32,
+) uint64 {
```
shrenujb, line 1251: I'm a bit confused. We convert subticks to uint64 here but the assignment from this func assigns to the subticks field?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +104,210 @@ func (k Keeper) AddSuborderToTriggerStore(
...
+func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
+	triggerStore := k.GetTWAPTriggerOrderPlacementStore(ctx)
+	blockTime := ctx.BlockTime().Unix()
+	iterator := triggerStore.Iterator(nil, nil)
+	defer iterator.Close()
+
+	var operationsToProcess []struct {
+		keyToDelete        []byte
+		suborderToPlace    types.Order
+		twapOrderPlacement types.TwapOrderPlacement
```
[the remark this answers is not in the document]
anmolagrawal345: Yea this is just the metadata we need to both update the parent twap and fire off the suborder itself. I think operations makes more sense here esp in the context of the next PR where I add a operation type which tells us if we need to delete a parent twap or update - which made the invariants we impose here a bit more readable imo.

protocol/x/clob/keeper/orders.go
anmolagrawal345: yea from my understanding setting the subticks field requires a uint64 that is in subticks-space, so it was easiest to get the price directly from the subticks lib and then adjust it

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: hm interesting yea fair I can definitely change it to log. I think my assumption with something like this one was we would have validated this earlier on in the tx cycle. For orders, we check the clob pair is valid in `PerformStatefulOrderValidation` right when the server sees the message?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -94,3 +104,210 @@ func (k Keeper) AddSuborderToTriggerStore(
...
+	suborderId types.OrderId,
+	twapOrderPlacement types.TwapOrderPlacement,
+	blockTime int64,
+) types.Order {
+	parentOrder := twapOrderPlacement.Order
+	order := types.Order{
+		OrderId:    suborderId,
+		Side:       twapOrderPlacement.Order.Side,
+		ReduceOnly: twapOrderPlacement.Order.ReduceOnly,
+	}
```
shrenujb, line 292: This should also have the order flag set to TWAP_SUBORDER? Or does that happen elsewhere?

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -293,6 +293,29 @@ func (k Keeper) ProcessSingleMatch(
 		curMakerPruneableBlockHeight,
 	)
 
+	if !matchWithOrders.TakerOrder.IsLiquidation() {
```
shrenujb, line 296: why do we have this check here?

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: Yea! The suborderId already has the orderFlag set to suborderId. The flow is:
1. TWAP Order is submitted
2. We place the initial suborderId in the trigger store (and flip the flag) - cf SetTWAPOrderPlacement
3. Everytime we iterate over the trigger store, we grab the suborderId from the key and reuse it for subsequent suborders we place

This way we can leave out any orderId logic from the end blocker itself and its handled entirely by the trigger store.

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -293,6 +293,29 @@ func (k Keeper) ProcessSingleMatch(
...
+		taker_order := matchWithOrders.TakerOrder.MustGetOrder()
+		if taker_order.OrderId.IsTwapSuborder() {
+			log.ErrorLog(ctx, "handling twap order", fmt.Sprintf("%+v", taker_order))
```
shrenujb, line 299: Remove this or make this a debug log

anmolagrawal345: ah this was for local testing it creeped in during a rebase probably apologies

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -1,12 +1,22 @@
...
+	"math/big"
 
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
 )
 
+const (
+	TWAP_SUBORDER_GOOD_TIL_BLOCK_TIME_OFFSET = 3
```
shrenujb, line 13: Can you add a comment stating what this is used for?

protocol/x/clob/keeper/process_single_match.go
anmolagrawal345: because MustGetOrder panics on getting called for a liquidation order

protocol/x/clob/keeper/twap_order_state.go
shrenujb: usually we do perform the validations ahead. I just want to be sure we follow best practices from a security standpoint.
Not sure if this is possible, but you could have a case where the validation succeeds for the tx, but another tx modifies state which causes the clob pair to not exist which would then result in this panicing

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: I think this fixes it actually - because now the only thing conditioned on remaining legs is added the next trigger to the store. We will still place the order even if remaining legs goes to 0. So in the final suborder case, we will decrement remaining legs from 1 -> 0, not place another in the trigger store, and fire off the last suborder.

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: Right makes sense and if the downside is an actual chain halt then better to err on the side of caution. I think it would actually be easier to implement in the next PR because I add other error handling for GenerateOrder there. But happy to add it here if you think its necessary though

shrenujb (approved):

PR #2785, opened 2025-04-15
[CT-1356] Handle TWAP Cancels and Expiry

anmolagrawal345:
### Changelist
This PR is dependent on #2780 and #2779.

Now that we can fire off suborders from the end blocker, we need to handle the cases on which we will stop processing suborders for a twap order. The cases are:
- Parent TWAP is cancelled
- Parent TWAP is complete
- Suborder fails to be placed

The state for all these cases is already stored in the TWAP Placement Store - so when processing the suborders in the end blocker we can simply check it and act accordingly. When one of these cases is true, we need to remove both the parent twap from the placement store as well as delete the key in the trigger store. Since there is only one key in the trigger store per parent twap at any time, we should just be able to delete the existing key and not need to iterate over the whole store to find others.

### Test Plan
Add e2e test to ensure that twap order placement prevents any further suborders from being placed.

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved handling and tracking of TWAP (Time-Weighted Average Price) order lifecycle, including better management of suborders, cancellations, and stateful order counts.
  * Enhanced error handling and state management for TWAP orders, ensuring accurate updates when collateral is depleted or orders are cancelled.

* **Bug Fixes**
  * Resolved issues with suborder triggering and removal, preventing creation of new suborders when collateral is insufficient or after cancellation.

* **Tests**
  * Added comprehensive end-to-end tests for TWAP order lifecycle, covering collateral depletion, cancellation, multi-suborder scenarios, and stateful order count accuracy.

* **Other Improvements**
  * Improved internal logic for order state management, resulting in more robust and reliable order processing for users.

### DIFF
--- a/protocol/x/clob/abci.go
+++ b/protocol/x/clob/abci.go
@@ -78,9 +78,6 @@ func EndBlocker(
 	// Prune any fill amounts from state which are now past their `pruneableBlockHeight`.
 	keeper.PruneStateFillAmountsForShortTermOrders(ctx)
 
-	// Place any TWAP suborders that are due
-	keeper.GenerateAndPlaceTriggeredTwapSuborders(ctx)
-
 	// Prune expired stateful orders completely from state.
 	expiredStatefulOrderIds := keeper.RemoveExpiredStatefulOrders(ctx, ctx.BlockTime())
 	for _, orderId := range expiredStatefulOrderIds {
@@ -113,6 +110,9 @@ func EndBlocker(
 	// Update the memstore with expired order ids.
 	// These expired stateful order ids will be purged from the memclob in `Commit`.
 	processProposerMatchesEvents.ExpiredStatefulOrderIds = expiredStatefulOrderIds
+
+	// Place any TWAP suborders that are due
+	keeper.GenerateAndPlaceTriggeredTwapSuborders(ctx)
 
 	// Poll out all triggered conditional orders from `UntriggeredConditionalOrders` and update state.
 	triggeredConditionalOrderIds := keeper.MaybeTriggerConditionalOrders(ctx)
--- a/protocol/x/clob/e2e/twap_orders_test.go
+++ b/protocol/x/clob/e2e/twap_orders_test.go
@@ -382,7 +382,11 @@ func TestTwapOrderStopsPlacingSubordersWhenCollateralIsDepleted(t *testing.T) {
 		Quantums: 25_000_000_000,
 		Subticks: 100_000_000,
 	}
-	for _, checkTx := range testapp.MustMakeCheckTxsWithClobMsg(ctx, tApp.App, *clobtypes.NewMsgPlaceOrder(matchingOrder1)) {
+	for _, checkTx := range testapp.MustMakeCheckTxsWithClobMsg(
+		ctx,
+		tApp.App,
+		*clobtypes.NewMsgPlaceOrder(matchingOrder1),
+	) {
 		resp := tApp.CheckTx(checkTx)
 		require.True(t, resp.IsOK(), "Expected CheckTx to succeed. Response: %+v", resp)
 	}
@@ -432,7 +436,11 @@ func TestTwapOrderStopsPlacingSubordersWhenCollateralIsDepleted(t *testing.T) {
 		Quantums: 25_000_000_000, // 1 BTC
 		Subticks: 100_000_000,
 	}
-	for _, checkTx := range testapp.MustMakeCheckTxsWithClobMsg(ctx, tApp.App, *clobtypes.NewMsgPlaceOrder(matchingOrder2)) {
+	for _, checkTx := range testapp.MustMakeCheckTxsWithClobMsg(
+		ctx,
+		tApp.App,
+		*clobtypes.NewMsgPlaceOrder(matchingOrder2),
+	) {
 		resp := tApp.CheckTx(checkTx)
 		require.True(t, resp.IsOK(), "Expected CheckTx to succeed. Response: %+v", resp)
 	}
@@ -468,17 +476,14 @@ func TestTwapOrderStopsPlacingSubordersWhenCollateralIsDepleted(t *testing.T) {
 		},
 	})
 
-	prev_alice_subaccount := tApp.App.SubaccountsKeeper.GetSubaccount(ctx, constants.Alice_Num0)
-	require.NotNil(t, prev_alice_subaccount)
-
 	// tApp.App.SubaccountsKeeper.SetSubaccount(ctx, constants.Alice_Num0_0USD) // doesnt work
 	withdrawal := &sendingtypes.MsgWithdrawFromSubaccount{
 		Sender:    constants.Alice_Num0,
 		Recipient: constants.BobAccAddress.String(), // send to bob
 		AssetId:   constants.Usdc.Id,
-		Quantums:  99_999_900_011_000_000, // remaining balance
+		Quantums:  99_999_995_000_000_000, // remaining balance + 95B to be below collat requirements for next suborder
 	}
-	
+
 	CheckTx_MsgWithdrawFromSubaccount := testapp.MustMakeCheckTx(
 		ctx,
 		tApp.App,
@@ -490,14 +495,7 @@ func TestTwapOrderStopsPlacingSubordersWhenCollateralIsDepleted(t *testing.T) {
 		withdrawal,
 	)
 	tApp.CheckTx(CheckTx_MsgWithdrawFromSubaccount)
-	
-	post_alice_subaccount := tApp.App.SubaccountsKeeper.GetSubaccount(ctx, constants.Alice_Num0)
-	require.NotNil(t, post_alice_subaccount)
-
 	ctx = tApp.AdvanceToBlock(6, testapp.AdvanceToBlockOptions{})
-
-	alice_subaccount := tApp.App.SubaccountsKeeper.GetSubaccount(ctx, constants.Alice_Num0)
-	require.NotNil(t, alice_subaccount)
 
 	// --- Third suborder trigger (should fail to place due to insufficient collateral) ---
 	ctx = tApp.AdvanceToBlock(7, testapp.AdvanceToBlockOptions{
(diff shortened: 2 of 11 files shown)


protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -0,0 +1,312 @@
...
+
+func (k Keeper) DecrementTwapOrderRemainingLegs(
+	ctx sdk.Context,
+	twapOrderPlacement types.TwapOrderPlacement,
+) {
+	store := k.GetTWAPOrderPlacementStore(ctx)
+	orderKey := twapOrderPlacement.Order.OrderId.ToStateKey()
+	if twapOrderPlacement.IsCompleted() {
+		panic("twap order has already been completed")
+	}
```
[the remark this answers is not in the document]
anmolagrawal345: this condition should never be met because GenerateSuborder returns false if the twapOrder is complete (aka does not generate another order). This method is thus only called when new suborders are being generated and remaining legs != 0

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -455,3 +465,34 @@ func (k Keeper) SetStatefulOrderCount(
 		)
 	}
 }
+
+func (k Keeper) CheckAndIncrementStatefulOrderCount(
+	ctx sdk.Context,
+	orderId types.OrderId,
+) {
+	if orderId.IsTwapSuborder() {
```
shrenujb, line 473: Can you add a comment for why TwapSuborders don't get included here

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -165,53 +187,59 @@ func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
...
 	ctx sdk.Context,
-	twapOrderPlacement *types.TwapOrderPlacement,
+	twapOrderPlacement types.TwapOrderPlacement,
 ) {
-	if twapOrderPlacement.RemainingLegs == 0 {
-		return // TODO: (anmol) handle end of twap order case
+	store := k.GetTWAPOrderPlacementStore(ctx)
+	orderKey := twapOrderPlacement.Order.OrderId.ToStateKey()
+	if twapOrderPlacement.IsCompleted() {
+		panic("twap order has already been completed")
```
shrenujb, line 238: this panic would be in Endblocker so we should probably substitute this for an error and alert on it instead

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -143,20 +161,24 @@ func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
 
 		twapOrderPlacement, found := k.GetTwapOrderPlacement(ctx, parentOrderId)
 		if !found {
-			// TODO: (anmol) handle order cancellation
+			// If parent TWAP was cancelled/found, do not place any pending suborders.
```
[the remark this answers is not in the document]
anmolagrawal345: yep

protocol/x/clob/types/order.go
```diff
@@ -288,3 +288,8 @@ func TriggerTimeToBytes(triggerTime int64) []byte {
 func TimeFromTriggerKey(triggerKey []byte) int64 {
 	return int64(binary.BigEndian.Uint64(triggerKey[0:8]))
 }
+
+// IsCompleted returns true if the TWAP order has no remaining legs to execute.
+func (t *TwapOrderPlacement) IsCompleted() bool {
```
shrenujb, line 293: dont think we need this function since its just a single line

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: sounds good! this should in theory be impossible because the condition to end up here is the same check done earlier in the suborder generation route but you're right (per our earlier discussion) to avoid panicking in the end blocker. changed to an error log!

protocol/x/clob/keeper/stateful_order_state.go
anmolagrawal345: Yep - suborders are essentially normal stateful orders so in theory they could be cancelled the same way. However, I'd expect someone to cancel via the TWAP parent order directly as opposed to trying to cancel a specific suborder.

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -455,3 +465,34 @@ func (k Keeper) SetStatefulOrderCount(
...
+		return
+	}
+	subaccountId := orderId.SubaccountId
```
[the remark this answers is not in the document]
anmolagrawal345: hm thats a good catch actually - I was originally trying to have all order types handled here but parent TWAPs do not enter this path at all. I added the increment to twap_order_state. apologies for the oversight there!

protocol/x/clob/keeper/stateful_order_state.go
shrenujb: I meant to add a comment in code itself

anmolagrawal345: added - think I had misplaced it in an earlier commit

protocol/x/clob/abci.go
```diff
@@ -114,6 +111,9 @@ func EndBlocker(
 	// These expired stateful order ids will be purged from the memclob in `Commit`.
 	processProposerMatchesEvents.ExpiredStatefulOrderIds = expiredStatefulOrderIds
 
+	// Place any TWAP suborders that are due
+	keeper.GenerateAndPlaceTriggeredTwapSuborders(ctx)
```
shrenujb, line 115: Curious. was there a reason this had to be moved down here?

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -165,53 +189,66 @@ func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
...
 
-	// Store updated state
+func (k Keeper) DecrementTwapOrderRemainingLegs(
+	ctx sdk.Context,
+	twapOrderPlacement types.TwapOrderPlacement,
+) {
 	store := k.GetTWAPOrderPlacementStore(ctx)
 	orderKey := twapOrderPlacement.Order.OrderId.ToStateKey()
-	twapOrderPlacementBytes := k.cdc.MustMarshal(twapOrderPlacement)
+	if twapOrderPlacement.IsCompleted() {
```
shrenujb, line 242: How could we run into this condition?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -319,19 +308,43 @@ func (k Keeper) MustRemoveStatefulOrder(
...
+		// Delete the Stateful order placement from state.
+	}
 
-	// Delete the Stateful order placement from state.
 	k.DeleteLongTermOrderPlacement(ctx, orderId)
+
+	// Cancelling a TWAP parent order will attempt to cancel the in-flight suborder.
+	// Since a TWAP suborder is maintained as a normal stateful order, cancelling a
+	// suborder follows the same flow as other stateful orders.
+	if order.OrderId.IsTwapOrder() {
```
shrenujb, line 331: Where would this be triggered from besides the user themselves?
I imagine it should be triggered if there's an issue with suborder generation as well?

protocol/x/clob/e2e/twap_orders_test.go
```diff
@@ -295,3 +295,268 @@ func TestTWAPOrderWithMatchingOrders(t *testing.T) {
 	require.True(t, found2, "Third suborder should exist in trigger store")
 	require.Equal(t, ctx.BlockTime().Unix()+int64(twapOrder.TwapParameters.Interval), triggerTime)
 }
+
```
shrenujb, line 299: Is there a way to make these tests parametrized or are they testing different things?

protocol/x/clob/e2e/twap_orders_test.go
```diff
@@ -295,3 +295,268 @@ func TestTWAPOrderWithMatchingOrders(t *testing.T) {
...
+	// Advance block time by 100 seconds to complete TWAP order
+	ctx = tApp.AdvanceToBlock(5, testapp.AdvanceToBlockOptions{
+		BlockTime: ctx.BlockTime().Add(time.Second * 100),
+	})
+
+	// Verify stateful order count is 0 after completion
+	statefulOrderCount = tApp.App.ClobKeeper.GetStatefulOrderCount(ctx, constants.Alice_Num0)
+	require.Equal(t, uint32(0), statefulOrderCount, "Stateful order count should be 0 after completion")
+}
```
shrenujb, line 755: Could you add a test where the user subaccount runs out of collateral midway through suborders and the parent order has to be cancelled?

protocol/x/clob/abci.go
anmolagrawal345: Ya I think I was running into a case where I was trying to place a suborder with an existing ID on the orderbook but the previous suborder hadn't yet been removed

protocol/x/clob/e2e/twap_orders_test.go
anmolagrawal345: They're testing different situations and the structures of the tests themselves are all a bit different as well so it was easier to keep them separate

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: this is just a safety check - we shouldn't run into this condition

protocol/x/clob/keeper/stateful_order_state.go
anmolagrawal345: this flow is just for explicit user cancels. If there is an error when placing a suborder, we directly delete both the parent and suborder entries from the stores

protocol/x/clob/e2e/twap_orders_test.go
anmolagrawal345: yep added

protocol/x/clob/keeper/twap_order_state.go
shrenujb: ok could you add alerting in DD on this log then

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -289,12 +329,16 @@ func (k Keeper) GenerateSuborder(
...
+		return nil, false
+	}
+
 	parentOrder := twapOrderPlacement.Order
 	order := types.Order{
 		OrderId:    suborderId,
 		Side:       twapOrderPlacement.Order.Side,
-		ReduceOnly: twapOrderPlacement.Order.ReduceOnly,
+		ReduceOnly: twapOrderPlacement.Order.ReduceOnly, // TODO: (anmol) client metadata?
```
shrenujb: what is this TODO here?

protocol/x/clob/e2e/twap_orders_test.go
```diff
@@ -296,6 +297,230 @@ func TestTWAPOrderWithMatchingOrders(t *testing.T) {
...
+				},
+			}),
+		},
+	})
+
+	prev_alice_subaccount := tApp.App.SubaccountsKeeper.GetSubaccount(ctx, constants.Alice_Num0)
+	require.NotNil(t, prev_alice_subaccount)
+
+	// tApp.App.SubaccountsKeeper.SetSubaccount(ctx, constants.Alice_Num0_0USD) // doesnt work
+	withdrawal := &sendingtypes.MsgWithdrawFromSubaccount{
```
shrenujb, line 452: doesn't work? what doesn't work?

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: discussed offline with Tyler - the client metadata is used by the FE to determine the order type (to distinguish between market and limit order). We landed on using a new value, 2, to be able to distinguish a twap suborder here.

protocol/x/clob/e2e/twap_orders_test.go
anmolagrawal345: removed comment - basically SetSubaccount was not working as I had imagined

shrenujb (approved):

PR #2787, opened 2025-04-15
[CT-1358] Support Limit Price for TWAP

anmolagrawal345:
### Changelist
This PR is dependent on #2779, #2780, and #2785.

This change floors/caps the subticks of the generated suborders for a twap order based on the subticks of passed in for the parent twap order, ensuring a limit price is maintained.

### Test Plan
Unit tests to ensure the order generation logic.

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -1,10 +1,18 @@
 package keeper
 
 import (
-	"encoding/binary"
+	"math"
 
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dydxprotocol/v4-chain/protocol/lib"
 	"github.com/dydxprotocol/v4-chain/protocol/x/clob/types"
+)
+
+const (
+	TWAP_SUBORDER_GOOD_TIL_BLOCK_TIME_OFFSET = 3
+
+	TWAP_MAX_SUBORDER_CATCHUP_MULTIPLE = 3
 )
 
 func (k Keeper) SetTWAPOrderPlacement(ctx sdk.Context,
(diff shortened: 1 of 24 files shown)


protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -318,6 +318,16 @@ func (k Keeper) calculateSuborderQuantums(
 	return suborderQuantumsRounded.Uint64()
 }
 
+func chooseLimitOrAdjustedPrice(
```
shrenujb: Doesnt seem like we need a separate function for this?

protocol/x/clob/keeper/twap_order_state_test.go
```diff
@@ -544,6 +544,92 @@ func TestGenerateSuborder(t *testing.T) {
 				},
 			},
 		},
+		"limit twap buy order with 50% price tolerance": {
```
shrenujb, line 547: Does price tolerance matter here anymore?

anmolagrawal345: no but I left it to still show that the limit price is used

protocol/x/clob/keeper/twap_order_state.go
anmolagrawal345: just for clarity that we choose one over the other if the subticks is specified

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -350,8 +360,11 @@ func (k Keeper) GenerateSuborder(
 
 	// calculate the suborder price with slippage adjustment
 	clobPair := k.mustGetClobPair(ctx, parentOrder.GetClobPairId())
-	order.Subticks = k.GetOraclePriceAdjustedByPercentageSubticks(ctx, clobPair, priceTolerancePpm)
+	suborderAdjustedPrice := k.GetOraclePriceAdjustedByPercentageSubticks(ctx, clobPair, priceTolerancePpm)
 
+	// set the subticks based on the adjusted price and the limit price (if configured)
+	// by the parent twap order
+	order.Subticks = chooseLimitOrAdjustedPrice(parentOrder, suborderAdjustedPrice)
```
shrenujb: I think it makes sense to have the above 2 calls in a single function which sets the order subticks
We can add a comment on the function mentioning which situations we use choose the oracle adjusted price vs provided limit price

anmolagrawal345: sure agreed. I created a new calculateSuborderSubticks in the twap_order_state to handle all cases in one place

PR #2798, opened 2025-04-22
(bug) don't remove CLOSE event handler on websocket send error

UnbornAztecKing:
#### Don't remove CLOSE event handler on websocket send error
 bug: close event handler was removed on ping/pong failure disconnect
 this prevented removal of connection from subscription and connection maps

#### Skip logging error if stream was destroyed
 avoid excessive logging during socket disconnection
 mitigate misbehaving clients

#### Stop publishing batched messages on error and log
 do not attempt to send remaining batch messages on publishing errors

#### Add bufferutil and utf-8-validate for performance
 recommended by ws and automatically enabled
 binary polyfills for handling and validating strings

#### Use node:crypto for v4 uuid
 since we only use v4's, we don't need a third-party dep

#### Raise semi space and old space size in start script
 a node application with these components can utilize 1-2cpu effectively
 typical memory provisioned by vendors is 2-8Gib/cpu
 based on a 4Gib host:
  256 Mib datadog sidecar container reservation
  960 Mib application container system reservation
  256 Mib semi space to handle object generation rate at load
  2624 Mib old heap to utilize minimum expected available resources

### Test Plan

Verify websocket is disconnected and no more than one send to closed websocket error is generated following failure of client to respond to heartbeat.

Run in testnet and internal environments and validated behavior under heavy misbehaving load

### DIFF
--- a/indexer/pnpm-lock.yaml
+++ b/indexer/pnpm-lock.yaml
@@ -756,10 +756,10 @@ importers:
       '@types/lodash': ^4.14.182
       '@types/node': ^18.19.31
       '@types/response-time': ^2.3.5
-      '@types/uuid': ^8.3.4
       '@types/ws': ^8.5.10
       axios: ^1.2.1
       body-parser: ^1.20.0
+      bufferutil: ^4.0.9
       cors: ^2.8.5
       dd-trace: ^3.32.1
       dotenv-flow: ^3.2.0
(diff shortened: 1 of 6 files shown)


tqin7 (approved):

UnbornAztecKing:

PR #2818, opened 2025-04-30
(perf) use luxon.DateTime.fromSQL vs moment()

UnbornAztecKing:
### Changelist
Comlink spends much of its time parsing postgres timestamp values and rendering them to ISO strings.
This was performed with the top-level moment() function, that needs to sniff the format for each value.
luxon's DateTime.fromSQL method is intended for database timestamp parsing.

### Test Plan
Unit Test

Run in internal environments and public testnet.

## Summary by CodeRabbit

- **Chores**
  - Replaced the timestamp parsing library to improve date handling consistency.
  - Removed an unused dependency to streamline the package.

### DIFF
--- a/indexer/packages/postgres/package.json
+++ b/indexer/packages/postgres/package.json
@@ -47,7 +47,6 @@
     "lodash": "^4.17.21",
     "long": "^5.2.1",
     "luxon": "^3.0.1",
-    "moment": "^2.29.4",
     "objection": "^2.2.16",
     "objection-unique": "^1.2.2",
     "pg": "^8.7.3",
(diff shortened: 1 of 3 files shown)


UnbornAztecKing:
It's not clear that Luxon is faster.
I'm investigating getting the datetime strings consistently formatted out of the pg-types interface, as the performance issue is due to the need to sniff the variable format of the time.
The interface is truncating milliseconds such as 0.690 to 0.69, and I'd like to avoid a kludge to detect this, and avoid detection entirely and move to an explicit format.

shrenujb (approved):

tqin7 (approved):

UnbornAztecKing:

PR #2835, opened 2025-05-06
(perf) patch mangled timestamps from pg for from ISO

UnbornAztecKing:
### Changelist
The indexer database adapter is pg, which produces mangled/truncated milliseconds in timestamps and prints spaces between the date and time.

Currently, luxon.DateTime.fromSQL is used to minimize format sniffing.
luxon.DateTime.fromISO wasn't chosen as the timestamps are not ISO8601 compliant.

With this change, the timestamps are patched to fix the separator and then parsed with .fromISO, resulting in a 10x speedup on SQL timestamp microbenchmarks.

### Test Plan
Unit Test

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of PostgreSQL TIMESTAMPTZ values to ensure accurate parsing and conversion of timestamps.

### DIFF
--- a/indexer/packages/postgres/src/db/pg-config.ts
+++ b/indexer/packages/postgres/src/db/pg-config.ts
@@ -13,5 +13,5 @@ const utcZone = {
 
 pg.types.setTypeParser(
   pg.types.builtins.TIMESTAMPTZ,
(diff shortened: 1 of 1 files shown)


shrenujb (approved):

UnbornAztecKing:

PR #2837, opened 2025-05-07
[CT-1362] Support Builder Codes in the Protocol

anmolagrawal345:
### Changelist
This is the implementation for builder codes which will act as a top up to the existing revenue sharing functionality we have for affiliates. Currently, revenue sharing works as a percent of fees accrued by the protocol (which is based off the market that the trade is happening on). Broker fees should act independently, where we should be able to specify some fee that is charged to the owner of the trade.

Current implementation here does 2 things:
1. Extends the order proto to include a builder code. This builder code will include an address to which the fee will be paid to as well as the fee in ppm.
2. We calculate the broker fee that will be imposed on order placement and order matching. The fee should exist on both maker and taker orders. This calculated fee will then be part of the collateral checks that happen and will also be reduced from the subaccount when the order is filled.

### Test Plan
Localnet testing successful. Also extended existing e2e tests.

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying builder codes and associated fees on orders, allowing partners or builders to receive a fee for matched orders.
  - Builder fees are now paid out to designated builder addresses when orders with builder codes are filled.
  - Order fill events and related UI components now include builder code information and builder fee details.

- **Bug Fixes**
  - Improved validation for builder code parameters, ensuring builder addresses are valid and fee rates are within allowed limits.

- **Tests**
  - Introduced comprehensive tests to verify correct handling, validation, and payout of builder fees in various order placement and matching scenarios.

- **Documentation**
  - Enhanced documentation for new order fields and event attributes related to builder codes and fees.


protocol/x/revshare/keeper/revshare.go
```diff
@@ -165,18 +165,28 @@ func (k Keeper) GetAllRevShares(
...
+	makerBuilderFee := k.affiliatesKeeper.GetBuilderFee(ctx, fill.MakerBuilderCode, fill.FillQuoteQuantums)
 	netFees := big.NewInt(0).Add(takerFees, makerFees)
+	totalBuilderFees := big.NewInt(0).Add(takerBuilderFee, makerBuilderFee)
+	totalFees := big.NewInt(0).Add(netFees, totalBuilderFees)
 
 	// when net fee is zero, no rev share is generated from the fill
-	if netFees.Sign() == 0 {
+	if totalFees.Sign() == 0 {
 		return types.RevSharesForFill{}, nil
 	}
```
anmolagrawal345, line 191: do we have maker rebates? in which case totalFees could potentially be 0 if maker rebate was the same magnitude as builder fee?

protocol/x/revshare/types/types.go
```diff
@@ -52,6 +53,7 @@ const (
 	REV_SHARE_TYPE_MARKET_MAPPER
 	REV_SHARE_TYPE_UNCONDITIONAL
 	REV_SHARE_TYPE_AFFILIATE
+	REV_SHARE_TYPE_BROKER
```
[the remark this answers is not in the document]
anmolagrawal345: done

protocol/x/clob/keeper/orders.go
```diff
@@ -1061,6 +1062,7 @@ func (k Keeper) GetSubticksForCollatCheck(ctx sdk.Context, order types.Order) (t
 // for orders to determine whether or not they may be added to the orderbook.
 func (k Keeper) AddOrderToOrderbookSubaccountUpdatesCheck(
 	ctx sdk.Context,
+	builderCode *types.BuilderCode,
```
shrenujb, line 1065: nit: Why add this in the middle of the params and not at the end?

protocol/x/clob/types/message_place_order.go
```diff
@@ -128,6 +128,24 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
...
+	if msg.Order.BuilderCode != nil {
+		if _, err := sdk.AccAddressFromBech32(msg.Order.BuilderCode.BuilderAddress); err != nil {
+			return errorsmod.Wrapf(
+				ErrInvalidBuilderCode,
+				"builder code address '%s' must be a valid bech32 address, but got error '%v'",
+				msg.Order.BuilderCode.BuilderAddress,
+				err.Error(),
+			)
+		}
+		if msg.Order.BuilderCode.FeePpm <= 0 || msg.Order.BuilderCode.FeePpm > 10_000 {
```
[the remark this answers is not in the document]
anmolagrawal345: yep - only wanted this field to be used to actually apply fees

protocol/x/clob/keeper/orders.go
anmolagrawal345: Felt more natural to group the subaccountId and builder code params together as they are part of the order proto and leave the final param to be the pending order. When modifying the usages of this method elsewhere it also felt cleaner. Do we have a strong preference to adding it at the end?

protocol/x/revshare/keeper/revshare.go
```diff
@@ -165,18 +165,28 @@ func (k Keeper) GetAllRevShares(
...
+
+	takerBuilderFee := k.affiliatesKeeper.GetBuilderFee(ctx, fill.TakerBuilderCode, fill.FillQuoteQuantums)
+	makerBuilderFee := k.affiliatesKeeper.GetBuilderFee(ctx, fill.MakerBuilderCode, fill.FillQuoteQuantums)
 	netFees := big.NewInt(0).Add(takerFees, makerFees)
+	totalBuilderFees := big.NewInt(0).Add(takerBuilderFee, makerBuilderFee)
+	totalFees := big.NewInt(0).Add(netFees, totalBuilderFees)
 
 	// when net fee is zero, no rev share is generated from the fill
-	if netFees.Sign() == 0 {
+	if totalFees.Sign() == 0 {
```
shrenujb, line 189: I dont think this is accurate. We should not be considering builder fees to determine the other revshares

protocol/x/revshare/types/types.go
```diff
@@ -71,6 +73,8 @@ func (r RevShareType) String() string {
 		return "REV_SHARE_TYPE_UNCONDITIONAL"
 	case REV_SHARE_TYPE_AFFILIATE:
 		return "REV_SHARE_TYPE_AFFILIATE"
+	case REV_SHARE_TYPE_BUILDER:
```
shrenujb, line 76: I wonder if we should structure this as a revshare because its technically not. Also when we add actual builder revshares that will have to be a different category

proto/dydxprotocol/clob/order.proto
```diff
@@ -249,6 +252,17 @@ message TwapParameters {
   uint32 price_tolerance = 3;
 }
 
+// BuilderCode represents the metadata for the partner or builder of an
+// order. This allows them to specify a fee for providing there service
+// which will be paid out in the event of an order fill.
+message BuilderCode {
```
shrenujb, line 258: I think we should call this BuilderCodeParams since these are params to the order

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: discussed offline - changed this to check if any revShares are created for both the builder and protocol fees. If there are no rev shares for either, then we can short circuit out.

protocol/x/revshare/types/types.go
anmolagrawal345: I actually think it makes sense to keep this here. With the paradigm that we have currently, when a match happens we offload all the rev sharing complexity to our revShareKeeper. Builder codes are essentially a revenue feature with our partners - just not "sharing" our protocol revenue. 

I reformatted the GetAllRevShares function to have a clear delineation between builderRevShare and protocolRevShare to make this even clearer. Lmk your thoughts!

proto/dydxprotocol/clob/order.proto
shrenujb: The params helps because it is a parameter of the order.proto itself. I think this is the convention we apply to most of our rpc calls

protocol/x/revshare/keeper/revshare.go
```diff
@@ -311,3 +374,21 @@ func (k Keeper) getMarketMapperRevShare(
...
+) ([]types.RevShare, *big.Int) {
+	revShares := []types.RevShare{}
+
+	builderFeeQuoteQuantums := builderCode.GetBuilderFee(fill.FillQuoteQuantums)
+	revShares = append(revShares, types.RevShare{
+		Recipient:         builderCode.BuilderAddress,
+		RevShareFeeSource: types.REV_SHARE_FEE_SOURCE_BUILDER_FEE,
+		RevShareType:      types.REV_SHARE_TYPE_BUILDER,
+		QuoteQuantums:     builderFeeQuoteQuantums,
+		RevSharePpm:       builderCode.FeePpm,
```
shrenujb, line 390: Will this revSharePpm field be actually used anywhere for the builder code revenue distribution? Just asking to understand

protocol/x/revshare/keeper/revshare.go
```diff
@@ -163,39 +163,122 @@ func (k Keeper) GetAllRevShares(
...
+		makerBuilderRevShares, makerBuilderFeeQuoteQuantums := getBuilderRevShare(*fill.MakerBuilderCode, fill)
+		revShares = append(revShares, makerBuilderRevShares...)
+		totalBuilderFees.Add(totalBuilderFees, makerBuilderFeeQuoteQuantums)
+	}
+
+	return revShares, totalBuilderFees
+}
 
-	// when net fee is zero, no rev share is generated from the fill
+func (k Keeper) getProtocolRevShares(
```
shrenujb, line 254: Add a comment explaining what these revshares consist of

protocol/x/revshare/keeper/revshare.go
```diff
@@ -204,31 +287,11 @@ func (k Keeper) GetAllRevShares(
...
-
-		// Add the rev share to the total for the fee source
-		feeSourceToQuoteQuantums[revShare.RevShareFeeSource].Add(
-			feeSourceToQuoteQuantums[revShare.RevShareFeeSource], revShare.QuoteQuantums)
-
-		// Add the rev share ppm to the total for the fee source
-		feeSourceToRevSharePpm[revShare.RevShareFeeSource] += revShare.RevSharePpm
-	}
-	//check total fees shared is less than or equal to net fees
-	if totalFeesShared.Cmp(netFees) > 0 {
```
shrenujb, line 222: I think we should still have this comparison for the protocol revshares by itself rather than it folded into the all revshares

protocol/x/revshare/keeper/revshare.go
```diff
@@ -163,39 +163,122 @@ func (k Keeper) GetAllRevShares(
...
 ) (types.RevSharesForFill, error) {
 	revShares := []types.RevShare{}
+	feeSourceToQuoteQuantums, feeSourceToRevSharePpm := buildRevShareToFeeSourceMaps()
+
+	// get the builder rev shares
+	revShares, builderFees := getBuilderRevShares(fill, revShares)
+
+	// get the protocol rev shares
+	netFees := big.NewInt(0).Add(fill.TakerFeeQuoteQuantums, fill.MakerFeeQuoteQuantums)
+	revShares, affiliateRevShare, err := k.getProtocolRevShares(
```
shrenujb, line 173: This is a confusing pattern where you send a similarly named object ("revShares") to another function to be mutated and returned as an object with the same name

I would suggest you have 2 separate functions which return separate revShare arrays and then concat them here on the top level function to better clarity

shrenujb: Something like

builderShares, builderfees := getBuilderRevShares()

protocolRevShares, affiliateRevShare, err := getProtocolRevShares()

allRevShares = append(builderShares, protocolRevShares)

protocol/x/clob/keeper/orders.go
shrenujb: > Felt more natural to group the subaccountId and builder code params together

What do you mean by this?

Best practices in most langs and also Go will tell you that you should always be adding new parameters at the end, especially when they are optional like in this case

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: No not actually - the quoteQuantums is what's actually used to distribute the fees. I think the ppm is only primarily currently used in `AddRewardSharesForFill` only uses the NET_PROTOCOL_REVENUE revShareType

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: I don't think so actually because now total fees includes builder fees + protocol fees and totalFeesShared iterates over the array of all rev shares which also includes both types. So the total fees incurred by this transaction should never exceed protocol fees + builder fees.

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: yep did exactly that!

protocol/x/clob/keeper/orders.go
anmolagrawal345: fair I changed it - didnt have a strong preference I subconsciously just put it second

proto/dydxprotocol/clob/order.proto
anmolagrawal345: updated. I called it builderCodeParameters to keep in line with how we named the TWAP fields

proto/dydxprotocol/clob/order.proto
```diff
@@ -249,6 +253,17 @@ message TwapParameters {
   uint32 price_tolerance = 3;
 }
 
+// BuilderCodeParameters represents the metadata for the partner or builder of
+// an order. This allows them to specify a fee for providing there service which
+// will be paid out in the event of an order fill.
+message BuilderCodeParameters {
+  // The address of the builder to which the fee will be paid.
+  string builder_address = 1;
```
northstar456, line 261: Add the address string option `[(cosmos_proto.scalar) = "cosmos.AddressString"]`. [Example](https://github.com/dydxprotocol/v4-chain/blob/999c6ece656bfbd60a6f56d4a6887723dd9af9e6/proto/dydxprotocol/clob/tx.proto#L145)

protocol/x/clob/keeper/process_operations.go
```diff
@@ -527,6 +527,7 @@ func (k Keeper) PersistMatchOrdersToState(
 				),
 			)
 		}
+		// TODO: (anmol) update fill event to include builder codes
```
northstar456, line 530: Associate TODO items with linear ticket

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -388,6 +388,21 @@ func (k Keeper) persistMatchedOrders(
...
+	var makerBuilderCodeParams *types.BuilderCodeParameters
+	var takerBuilderCodeParams *types.BuilderCodeParameters
+	// apply broker fees for taker and maker separately
+	if !matchWithOrders.MakerOrder.IsLiquidation() {
+		makerBuilderCodeParams = matchWithOrders.MakerOrder.MustGetOrder().BuilderCodeParameters
+		builderFee := makerBuilderCodeParams.GetBuilderFee(matchWithOrders.FillAmount.ToBigInt())
+		bigMakerQuoteBalanceDelta.Sub(bigMakerQuoteBalanceDelta, builderFee)
+	}
+
+	if !matchWithOrders.TakerOrder.IsLiquidation() {
```
northstar456, line 400: Liquidation taker order are protocol generated so should nit have `BuilderCodeParameters` right?

Maybe log error here if we ever see such a case

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -388,6 +388,21 @@ func (k Keeper) persistMatchedOrders(
 		bigTakerQuoteBalanceDelta.Sub(bigTakerQuoteBalanceDelta, insuranceFundDelta)
 	}
 
+	var makerBuilderCodeParams *types.BuilderCodeParameters
+	var takerBuilderCodeParams *types.BuilderCodeParameters
+	// apply broker fees for taker and maker separately
+	if !matchWithOrders.MakerOrder.IsLiquidation() {
```
northstar456, line 394: What's the thinking on not having builder code share apply to liquidation maker?

protocol/x/revshare/keeper/revshare.go
```diff
@@ -162,40 +162,134 @@ func (k Keeper) GetAllRevShares(
 	fill clobtypes.FillForProcess,
 	affiliatesWhitelistMap map[string]uint32,
 ) (types.RevSharesForFill, error) {
-	revShares := []types.RevShare{}
+	allRevShares := []types.RevShare{}
+	feeSourceToQuoteQuantums, feeSourceToRevSharePpm := buildRevShareToFeeSourceMaps()
```
northstar456, line 166: I agree with Shrenuj that we should structure builder fee outside of x/revshare logic. It’s as simple as a transfer from the trader’s subaccount to the builder’s bank address during PersistSingleMatch. There’s no need to touch x/revshare at all and confuse them together which also allows for potential error.
Mental model is pretty clear:
RevShare: handle where the trading fees (maker + taker) should go
Builder fee: has nothing to do with trading fees or protocol revenue - just a separate tranfer from trader to builder

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -388,6 +388,21 @@ func (k Keeper) persistMatchedOrders(
 		bigTakerQuoteBalanceDelta.Sub(bigTakerQuoteBalanceDelta, insuranceFundDelta)
 	}
 
+	var makerBuilderCodeParams *types.BuilderCodeParameters
```
northstar456, line 391: Could you add a test case [in this file](https://github.com/dydxprotocol/v4-chain/blob/999c6ece656bfbd60a6f56d4a6887723dd9af9e6/protocol/x/clob/memclob/memclob_place_order_test.go#L2203), where a taker order (with just enough to collateralize initial position) would successfully match with a maker order, but would fail collateral check if there's taker builder code?

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -471,13 +486,15 @@ func (k Keeper) persistMatchedOrders(
...
-		MarketId:              perpetual.Params.MarketId,
+		TakerAddr:              matchWithOrders.TakerOrder.GetSubaccountId().Owner,
+		TakerFeeQuoteQuantums:  bigTakerFeeQuoteQuantums,
+		MakerAddr:              matchWithOrders.MakerOrder.GetSubaccountId().Owner,
+		MakerFeeQuoteQuantums:  bigMakerFeeQuoteQuantums,
+		FillQuoteQuantums:      bigFillQuoteQuantums,
+		ProductId:              perpetualId,
+		MarketId:               perpetual.Params.MarketId,
+		MakerBuilderCodeParams: makerBuilderCodeParams,
+		TakerBuilderCodeParams: takerBuilderCodeParams,
```
northstar456, line 497: Below, let's emit a Cosmos SDK event that allows generic Cosmos indexers to index builder code fees

protocol/x/clob/keeper/process_single_match.go
anmolagrawal345: It was my understanding is liquidation orders come from the protocol itself which should not implement a builder code. Additionally, there's check ensures safety because if we call MustGetOrder on a liquidation order we panic.

proto/dydxprotocol/clob/order.proto
```diff
@@ -1,6 +1,7 @@
 syntax = "proto3";
 package dydxprotocol.clob;
 
+import "cosmos_proto/cosmos.proto";
```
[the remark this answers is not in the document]
anmolagrawal345: I believe the cosmos_proto dependency is already being pulled in correctly

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: moved it outside and into the process_single_match logic (similar to the transfer insurance pattern)!

protocol/x/revshare/keeper/revshare.go
anmolagrawal345: after the change to move the builder codes outside the revshare logic this is back to the status quo - it only reflects the protocol fees

protocol/x/revshare/keeper/revshare_test.go
```diff
@@ -699,7 +699,7 @@ func TestKeeper_GetAllRevShares_Valid(t *testing.T) {
 			},
 		},
 		{
-			name: "No rev shares",
+			name:                     "No rev shares",
```
shrenujb, line 702: need to remove this change

protocol/x/subaccounts/keeper/transfer.go
```diff
@@ -424,6 +424,41 @@ func (k Keeper) TransferInsuranceFundPayments(
...
+	productId uint32,
+	builderCodeParams clobtypes.BuilderCodeParameters,
+	fillQuantums *big.Int,
+) error {
+	collateralPoolAddr, err := k.GetCollateralPoolFromPerpetualId(ctx, productId)
+	if err != nil {
+		return err
+	}
+
+	builderFeeQuantums := builderCodeParams.GetBuilderFee(fillQuantums)
```
shrenujb, line 438: There's a bit of a circular dependency here
Clob module calls into this (subaccounts) which then calls into clob again

It'd be better to just calculate the builder fees above and call a more generic subaccountToBank function here

protocol/x/clob/types/message_place_order.go
```diff
@@ -128,6 +130,25 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
 		}
 	}
 
+	if msg.Order.BuilderCodeParameters != nil {
```
shrenujb, line 133: Need to add tests for these in message_place_order_test.go

protocol/x/clob/types/message_place_order_test.go
```diff
@@ -601,6 +601,78 @@ func TestMsgPlaceOrder_ValidateBasic(t *testing.T) {
...
+						GoodTilBlockTime: uint32(100),
+					},
+					BuilderCodeParameters: &BuilderCodeParameters{
+						BuilderAddress: sample.AccAddress(),
+						FeePpm:         MaxBuilderCodeFeePpm + 1,
+					},
+				},
+			},
+			err: ErrInvalidBuilderCode,
+		},
```
shrenujb, line 675: just need to add another case for success with builder code params

protocol/x/clob/keeper/process_single_match.go
northstar456: Similar to above: if a condition is never true, should just check and panic, and the remaining logic should be outside of the `if` block. The current code causes confusion

protocol/x/subaccounts/keeper/transfer.go
```diff
@@ -424,6 +424,39 @@ func (k Keeper) TransferInsuranceFundPayments(
...
+) error {
+	collateralPoolAddr, err := k.GetCollateralPoolFromPerpetualId(ctx, productId)
+	if err != nil {
+		return err
+	}
+
+	_, coinToTransfer, err := k.assetsKeeper.ConvertAssetToCoin(
+		ctx,
+		assettypes.AssetUsdc.Id,
+		new(big.Int).Abs(builderFeeQuantums),
```
northstar456, line 441: Is there any valid case where `builderFeeQuantums` should be negative? We should probably panic if `builderFeeQuantums.Sign() <= 0`, and remove the `.Abs()` here

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -470,6 +492,30 @@ func (k Keeper) persistMatchedOrders(
 		return takerUpdateResult, makerUpdateResult, affiliateRevSharesQuoteQuantums, err
 	}
 
+	// Transfer builder fees for taker and maker builders if they exist
+	if containsBuilderParams(matchWithOrders.TakerOrder) {
```
northstar456, line 496: Let's clearly document what's happening under the hood and why we are just sending a transfer between collateral pool and the builder bank account (because the builder fee is always taken out from trader's subaccount quote balance). 

We need to be extremely careful about the flow of funds here, to make sure there's loss of protocol funds.

protocol/x/subaccounts/keeper/transfer.go
```diff
@@ -424,6 +424,39 @@ func (k Keeper) TransferInsuranceFundPayments(
 	)
 }
 
+func (k Keeper) TransferBuilderFees(
```
northstar456, line 427: Add docstring

protocol/x/subaccounts/keeper/transfer.go
```diff
@@ -424,6 +424,46 @@ func (k Keeper) TransferInsuranceFundPayments(
...
+	builderFeeQuantums *big.Int,
+	builderAddress string,
+) error {
+	collateralPoolAddr, err := k.GetCollateralPoolFromPerpetualId(ctx, productId)
+	if err != nil {
+		return err
+	}
+
+	if builderFeeQuantums.Sign() < 0 {
+		panic("builder fee quantums is negative")
```
northstar456: Nit: add `builderAddress`, `builderFeeQuantums` to panic message

protocol/x/clob/keeper/process_single_match.go
northstar456: Thoughts on this comment?

protocol/x/clob/keeper/process_single_match.go
anmolagrawal345: MustGetOrder already panics if called on a liquidation order so I think this case actually makes sense. It also follows the pattern of the rest of the method for special casing against liquidation orders - so I'd argue this one is okay for now

protocol/x/clob/keeper/process_single_match.go
anmolagrawal345: agreed! added the panic and moved the maker order logic out of the if block

teddyding (approved):

anmolagrawal345:
@Mergifyio backport release/protocol/v8.1.x

anmolagrawal345:

PR #2855, opened 2025-05-12
Fix(orderbook): Fix to-indexer order replacement logic

teddyding:
### Changelist
This previous [change](https://github.com/dydxprotocol/v4-chain/pull/1593/files) refactored the way order replacement is handled, and moved the responsibility from indexer to indexer full node. Full node needs to send a `remove` message before `place` in the case of an order replacement. However, this logic wasn't applied to the case where an order is replaced at the same price level. This issue results in incorrect orderbook on indexer, and is the root cause of abnormally deep orderbook for majors where order replacement is heavily used.

[Context thread](https://dydx-team.slack.com/archives/C06EMK1746T/p1746565019449449?thread_ts=1742850384.243169&cid=C06EMK1746T)

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved order replacement handling to ensure removal messages are always generated for existing orders, enhancing consistency in off-chain updates.

### DIFF
--- a/protocol/x/clob/memclob/memclob.go
+++ b/protocol/x/clob/memclob/memclob.go
@@ -501,16 +501,14 @@ func (m *MemClobPriceTimePriority) PlaceOrder(
 		// create an order removal message first so we can remove the original price level from the orderbook.
 		// TODO (CT-884): send OrderReplaceV1 message for replacement orders and add order-replace-handler to Vulcan
 		orderId := order.OrderId
-		if existingOrder, found := orderbook.getOrder(orderId); found {
-			if order.Subticks != existingOrder.Subticks {
-				if message, success := off_chain_updates.CreateOrderRemoveMessageWithReason(
-					ctx,
-					orderId,
-					indexersharedtypes.OrderRemovalReason_ORDER_REMOVAL_REASON_REPLACED,
-					ocutypes.OrderRemoveV1_ORDER_REMOVAL_STATUS_BEST_EFFORT_CANCELED,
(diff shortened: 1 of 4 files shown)


protocol/x/clob/memclob/memclob.go
```diff
@@ -501,16 +501,14 @@ func (m *MemClobPriceTimePriority) PlaceOrder(
 		// create an order removal message first so we can remove the original price level from the orderbook.
 		// TODO (CT-884): send OrderReplaceV1 message for replacement orders and add order-replace-handler to Vulcan
 		orderId := order.OrderId
-		if existingOrder, found := orderbook.getOrder(orderId); found {
-			if order.Subticks != existingOrder.Subticks {
```
shrenujb, line 505: Pretty much all we've done here is remove this check. Does this mean that only the order replacement where size was changed but price was not was resulting in this issue?

teddyding: The issue impacted any order replacement where price is the same regardless of if the order size was changed or not (eg trader could be sending same price, same prize but a larger GTG)

shrenujb (approved):

teddyding:
@Mergifyio backport release/protocol/v8.x

teddyding:
@Mergifyio release/protocol/v8.1.x

tqin7:
@mergifyio backport release/protocol/v8.1.x

PR #2870, opened 2025-05-30
feat(permissioned key): require sig verification in authenticator tree

northstar456:
### Changelist

### Test Plan

### Author/Reviewer Checklist

### DIFF
--- a/protocol/x/accountplus/authenticator/signature_authenticator.go
+++ b/protocol/x/accountplus/authenticator/signature_authenticator.go
@@ -94,16 +94,16 @@ func (sva SignatureVerification) OnAuthenticatorAdded(
 	account sdk.AccAddress,
 	config []byte,
 	authenticatorId string,
-) error {
+) (requireSigVerification bool, err error) {
 	// We allow users to pass no data or a valid public key for signature verification.
 	if len(config) != secp256k1.PubKeySize {
-		return fmt.Errorf(
+		return false, fmt.Errorf(
 			"invalid secp256k1 public key size, expected %d, got %d",
 			secp256k1.PubKeySize,
 			len(config),
 		)
 	}
-	return nil
+	return true, nil
 }
 
 func (sva SignatureVerification) OnAuthenticatorRemoved(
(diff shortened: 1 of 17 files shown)


protocol/x/accountplus/authenticator/signature_authenticator.go
```diff
@@ -94,16 +94,16 @@ func (sva SignatureVerification) OnAuthenticatorAdded(
 	account sdk.AccAddress,
 	config []byte,
 	authenticatorId string,
-) error {
+) (requireSigVerification bool, err error) {
 	// We allow users to pass no data or a valid public key for signature verification.
 	if len(config) != secp256k1.PubKeySize {
-		return fmt.Errorf(
+		return false, fmt.Errorf(
```
jyzhe, line 100: mega nit: think we should still return true here to make it easier to reason about

protocol/x/accountplus/keeper/authenticators.go
```diff
@@ -90,11 +90,19 @@ func (k Keeper) AddAuthenticator(
...
 	}
 
+	if !requireSigVerification {
+		return 0, fmt.Errorf(
+			"unsafe: authenticator tree does not require signature verification all possible paths, type = %v, config = %v",
+			authenticatorType,
+			config,
+		)
+	}
+
```
jyzhe, line 105: for my understanding - this means that we now disallow standalone authenticators, right?

like users can't register a naked MessageFilter, for example

teddyding: Yes exactly, since a single `MessageFilter` is unsafe and any one would be able to send that message for it.

protocol/x/accountplus/authenticator/signature_authenticator.go
teddyding: Hmm might make sense to stick to the convention that function returns an error, then return all other values in their default value (e.g. 0, "", false) since it's the caller's job to check for error and not use the other return values. Thoughts?

shrenujb (approved):

northstar456:
@Mergifyio backport release/protocol/v8.1.x

PR #2974, opened 2025-07-11
perf: optimize postgres findFundingIndexMap(s)

UnbornAztecKing:
roundtable PnL task is using excessive memory and heap profiling shows significant heap accumulation in funding-index-updates-table

### Changelist

findFundingIndexMap
- execute db query promptly and defer await
- use lightweight object vs reduce

findFundingIndexMaps
- call toFixed on both heights
- use lightweight object vs reduce

### Test Plan
Unit test and deployment and validation in internal networks and testnet.

### DIFF
--- a/indexer/packages/postgres/src/stores/funding-index-updates-table.ts
+++ b/indexer/packages/postgres/src/stores/funding-index-updates-table.ts
@@ -1,5 +1,4 @@
 import Big from 'big.js';
-import _ from 'lodash';
 import { QueryBuilder } from 'objection';
 
 import { BUFFER_ENCODING_UTF_8, DEFAULT_POSTGRES_OPTIONS } from '../constants';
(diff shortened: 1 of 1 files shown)


indexer/packages/postgres/src/stores/funding-index-updates-table.ts
```diff
@@ -198,48 +197,45 @@ export async function findFundingIndexMap(
...
-
-  const fundingIndexUpdates: FundingIndexUpdatesFromDatabase[] = await baseQuery
+    )
     .distinctOn(FundingIndexUpdatesColumns.perpetualId)
     .where(FundingIndexUpdatesColumns.effectiveAtHeight, '<=', effectiveBeforeOrAtHeight)
     // Optimization to reduce number of rows needed to scan
     .where(
       FundingIndexUpdatesColumns.effectiveAtHeight,
       '>',
+      // Why do we scan more than one hour of blocks?
```
tqin7, line 211: probably some buffer?

indexer/packages/postgres/src/stores/funding-index-updates-table.ts
```diff
@@ -198,48 +197,45 @@ export async function findFundingIndexMap(
...
+    options,
   );
+
+  const bigZero = Big(0);
+  const fundingIndexMap: FundingIndexMap = Object.create(null);
+  for (const perpetualMarket of perpetualMarkets) {
+    fundingIndexMap[perpetualMarket.id] = bigZero;
+  }
+
+  const fundingIndexUpdates: FundingIndexUpdatesFromDatabase[] = await fundingIndexUpdatesQuery;
```
tqin7, line 233: why does deferring this await help? bc it can de-allocate heap more quickly?

tqin7 (approved):

indexer/packages/postgres/src/stores/funding-index-updates-table.ts
UnbornAztecKing: old: exec, wait, read, exec, wait, read
new: exec, exec, wait, read, wait (maybe 0 wait), read

the idea is to execute both queries before awaiting results so that the waits may overlap and the total wait is reduced

indexer/packages/postgres/src/stores/funding-index-updates-table.ts
UnbornAztecKing: yeah i'm wondering if it can be reduced significantly below 4? if this is part of the quadratic cost, reducing it to 2 hours could help a lot

tqin7: yeah 2 should work too i think

UnbornAztecKing:

PR #2981, opened 2025-07-15
Add logic for ORRS

jusbar23:
### Changelist
* Add logic to attribute revenue to order routers during ProcessSingleMatch

### Test Plan
* Unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for order router addresses in orders, including new fields and validation.
  * Order match events now include order router addresses and fee details.
  * Integrated order router revenue share calculations into the revenue sharing system.
  * Introduced maximum allowed revenue share limit for order routers with enforcement.

* **Bug Fixes**
  * Enhanced validation to prevent invalid order router addresses.

* **Tests**
  * Expanded test coverage for order router address validation, revenue share limits, revenue share interactions, and fee distribution involving order routers.

* **Chores**
  * Removed deprecated message type for setting order router revenue shares.


proto/dydxprotocol/clob/order.proto
```diff
@@ -235,6 +235,11 @@ message Order {
   // builder_code_parameters is the metadata for the
   // partner or builder of an order specifying the fees charged.
   BuilderCodeParameters builder_code_parameters = 13;
+
+  // order_router_address is the address of the order router that placed the
+  // order.
+  string order_router_address = 14
+      [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
```
shrenujb, line 242: is this nullable?

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
```diff
@@ -644,6 +644,12 @@ export interface Order {
    */
 
   builderCodeParameters?: BuilderCodeParameters;
+  /**
+   * order_router_address is the address of the order router that placed the
+   * order.
+   */
+
+  orderRouterAddress: string;
```
anmolagrawal345, line 652: optional?

proto/dydxprotocol/clob/order.proto
shrenujb: basically if someone doesn't specify this in their proto construction, we dont want to break them

jusbar23: Added the nullable attribute. I created a few orders in test cases without the order router addr and it looks like it can be created without the field set

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
jusbar23: Adding optional leads to compiler errors...added a gogoproto.nullable = true flag to the proto file

proto/dydxprotocol/clob/order.proto
jusbar23: After digging further, it appears that this would be set to empty string if the field is not set in the proto. Using optional causes compiler errors in the ast library

indexer/packages/v4-protos/src/codegen/dydxprotocol/clob/order.ts
jusbar23: See previous comment. After digging further, it appears that this would be set to empty string if the field is not set in the proto. Using optional causes compiler errors in the ast library

protocol/x/revshare/types/types.go
```diff
@@ -52,12 +53,14 @@ const (
...
 	REV_SHARE_TYPE_UNCONDITIONAL
 	REV_SHARE_TYPE_AFFILIATE
+	REV_SHARE_TYPE_ORDER_ROUTER
 )
 
 type RevSharesForFill struct {
 	AffiliateRevShare        *RevShare
 	FeeSourceToQuoteQuantums map[RevShareFeeSource]*big.Int
 	FeeSourceToRevSharePpm   map[RevShareFeeSource]uint32
+	OrderRouterRevShares     []RevShare
```
shrenujb: Dumb q but do you need this if you can add the revshare in the fee source map?

jusbar23: Technically no, since the affilliate and order router is mutually exclusive. So you can infer the taker = order router or affiliate and maker = order router. This would change if we ever add more fees into the taker & maker that are not mutually exclusive

protocol/x/clob/keeper/process_single_match.go
```diff
@@ -575,6 +577,25 @@ func (k Keeper) persistMatchedOrders(
 		bigFillQuoteQuantums,
 	)
 
+	takerOrderRouterFeeQuoteQuantums := big.NewInt(0)
```
shrenujb, line 580: Is this logic only used to extrapolate info for the indexer event?

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -586,6 +586,10 @@ func (k Keeper) ProcessDeleveraging(
 			"",
 			big.NewInt(0),
 			big.NewInt(0),
+			"",
```
shrenujb, line 591: maybe add a comment above this mentioning these are order router rev share fields

protocol/x/clob/types/message_place_order_test.go
```diff
@@ -696,6 +696,27 @@ func TestMsgPlaceOrder_ValidateBasic(t *testing.T) {
...
+					Quantums: uint64(42),
+					Subticks: uint64(10),
+					GoodTilOneof: &Order_GoodTilBlockTime{
+						GoodTilBlockTime: uint32(100),
+					},
+					OrderRouterAddress: "invalid_order_router_address",
+				},
+			},
+			err: ErrInvalidOrderRouterAddress,
+		},
```
shrenujb, line 719: Add a test for valid order router address

protocol/x/clob/keeper/process_single_match.go
jusbar23: Yes, since the specific fields aren't explicitly saved in the FillForProcess struct

protocol/x/clob/keeper/deleveraging.go
jusbar23: Done

protocol/x/revshare/types/message_set_order_router_rev_share.go
```diff
@@ -7,14 +7,16 @@ import (
...
-	kMaxOrderRouterRevSharePpm = lib.OneHundredThousand * 5
+	// 80% of the revenue share is the max allowed. Realistically this will be lower,
+	// set it high so we don't need a protocol upgrade to change this
+	kMaxOrderRouterRevSharePpm = lib.OneHundredThousand * 8
 )
 
 var _ types.Msg = &MsgSetOrderRouterRevShare{}
 
-// ValidateBasic performs validation to check that order router rev share is under 50%
+// ValidateBasic performs validation to check that order router rev share is under 80%
```
shrenujb: We should not be adding logic into ValidateBasic anymore. This pattern is deprecated.
https://docs.cosmos.network/main/learn/beginner/tx-lifecycle 
All validation should go into the message handling itself

protocol/x/revshare/types/message_set_order_router_rev_share_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/dydxprotocol/v4-chain/protocol/testutil/constants"
+	types "github.com/dydxprotocol/v4-chain/protocol/x/revshare/types"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	validAuthority = constants.AliceAccAddress.String()
+)
+
+func TestMsgSetOrderRouterRevShare_ValidateBasic(t *testing.T) {
```
shrenujb: I would remove these tests and add the validation logic testing into the main MsgRouter tests

protocol/x/revshare/types/types.go
shrenujb: Hmm so the affiliate rev share has both the AffiliateRevShare structure in this struct and a mapping in the FeeSourceTo* maps?

shrenujb: I'd be in favor of not having to add a new struct here every time

protocol/x/revshare/types/message_set_order_router_rev_share_test.go
jusbar23: Done

protocol/x/revshare/types/message_set_order_router_rev_share.go
jusbar23: Done

protocol/x/clob/types/message_place_order_test.go
jusbar23: Done

protocol/x/revshare/keeper/revshare.go
```diff
@@ -292,6 +311,67 @@ func (k Keeper) getAffiliateRevShares(
...
+	}
+
+	orderRouterRevShares := []types.RevShare{}
+	takerOrderRouterRevSharePpm, err := k.GetOrderRouterRevShare(ctx, fill.TakerOrderRouterAddr)
+	if err != nil {
+		// We can ignore invalid order router addresses for takers, assume rev share is 0
+		if errors.Is(err, types.ErrOrderRouterRevShareNotFound) {
+			k.Logger(ctx).Error("order router rev share not found for taker: " + fill.TakerOrderRouterAddr)
+		} else {
+			return nil, err
```
shrenujb: What would errors here mean for the actual fill itself? Does the fill fail?

shrenujb: What kind of other error would you get here?

jusbar23: This won't fail if the rev share doesn't exist, instead it will just log an Error

It will fail if the order router rev share is not a valid address..this state should never happen because the address validations should happen when an order is made.

jusbar23: Errors here are either

Address not valid
Rev Share not found

protocol/x/clob/keeper/deleveraging.go
```diff
@@ -582,10 +582,14 @@ func (k Keeper) ProcessDeleveraging(
 			false, // IsLiquidation is false since this isn't a liquidation match.
 			true,  // IsDeleverage is true since this is a deleveraging match.
 			perpetualId,
-			"",
-			"",
-			big.NewInt(0),
-			big.NewInt(0),
+			/** taker_builder_address=*/ "",
```
shrenujb: Sorry I meant to say one comment preceding all the fields considering they were all zero. Ex
// Order router revshare fields
"",
"",
big.NewInt(0),
big.NewInt(0),

protocol/x/clob/types/message_place_order.go
```diff
@@ -78,6 +78,17 @@ func (msg *MsgPlaceOrder) ValidateBasic() (err error) {
 		return errorsmod.Wrapf(ErrInvalidOrderQuantums, "order size quantums cannot be 0")
 	}
 
+	if msg.Order.OrderRouterAddress != "" {
```
shrenujb, line 81: As we talked, it'd be good to add validation to check if this order router address even exists in state before proceeding forward with the order

protocol/x/clob/keeper/deleveraging.go
jusbar23: done

protocol/x/clob/types/message_place_order.go
jusbar23: Moved to place order

protocol/x/clob/keeper/process_single_match.go
shrenujb: Could you add a small comment above this block specifying that?

protocol/x/revshare/keeper/revshare.go
```diff
@@ -213,21 +218,39 @@ func (k Keeper) GetAllRevShares(
 	if err != nil {
 		return types.RevSharesForFill{}, err
 	}
-	netFeesSubAffiliateFeesShared := new(big.Int).Sub(netFees, affiliateFeesShared)
-	if netFeesSubAffiliateFeesShared.Sign() <= 0 {
+
+	var orderRouterRevShares []types.RevShare
+	netFeesSubRevenueShare := netFees
+	// No affiliate fees shared, so we can generate order router rev shares
```
anmolagrawal345, line 224: So if a user was referred by an affiliate, and then that address uses a telegram bot with ORR, the bot wouldn't get anything?

jusbar23: Yeah, the priority for rev shares will be to the affiliate, then ORRS

anmolagrawal345: It still feels weird to me that we're essentially allowing some users to use 3rd party services for free. Not a great loophole for builders. If we allow both fees - do we run into issues?

anmolagrawal345 (approved):

protocol/x/revshare/keeper/revshare.go
```diff
@@ -213,21 +218,39 @@ func (k Keeper) GetAllRevShares(
...
+	if affiliateRevShares == nil {
```
northstar456, line 225: Switch to `if len(affiliateRevShares) == 0 {` which is true for both nil slices and non-nil zero-length slices. Current code is technically correct since `getAffiliateRevShares` never returns non-nil zero-length slice, but we shouldn't rely on that.

protocol/x/clob/keeper/orders.go
```diff
@@ -187,6 +187,13 @@ func (k Keeper) PlaceShortTermOrder(
 	order.OrderId.MustBeShortTermOrder()
 	orderLabels := order.GetOrderLabels()
 
+	if _, err := k.revshareKeeper.GetOrderRouterRevShare(
```
northstar456, line 190: Curious what's the goal of this? A stateless schema check? 

This is the "hot" code path of CLOB so we should be careful with any additional operations, including state reads like [this](https://github.com/dydxprotocol/v4-chain/blob/b1b7629e7561eb0dcb0ed8d895f08fdbe820e395/protocol/x/revshare/keeper/revshare.go#L119)

protocol/x/revshare/keeper/revshare.go
northstar456, line 225: This means that in the below case:

- Maker order specified router, maker has positive trading fees
- Taker order has an affiliate

The maker order would not be eligible for router rev share. Is this intended? If so we should clearly document

protocol/x/revshare/keeper/revshare.go
```diff
@@ -292,6 +315,65 @@ func (k Keeper) getAffiliateRevShares(
...
+) ([]types.RevShare, error) {
+	if fill.TakerOrderRouterAddr == "" && fill.MakerOrderRouterAddr == "" {
+		return nil, nil
+	}
+
+	orderRouterRevShares := []types.RevShare{}
+	takerOrderRouterRevSharePpm, err := k.GetOrderRouterRevShare(ctx, fill.TakerOrderRouterAddr)
+	if err != nil {
+		// This should never happen
+		k.Logger(ctx).Error("order router rev share not found for taker: " + fill.TakerOrderRouterAddr)
```
northstar456, line 332: nit: log out error. Also message should say "invalid" instead of "not found" - the latter makes assumption about implementation detail of `GetOrderRouterRevShare`. 

Same for below

protocol/x/clob/keeper/orders.go
shrenujb: Should we remove this from the short term order placement and just keep it for the stateful order placement?

protocol/x/revshare/keeper/revshare.go
jusbar23: done

protocol/x/clob/keeper/orders.go
jusbar23: This validates that the order router address is valid, or rejects the order. The single match assumes that it exists, so we will need to move the validation to process_single_match if we take this out

protocol/x/revshare/keeper/revshare.go
jusbar23: Done

protocol/x/revshare/keeper/revshare.go
jusbar23: Added a comment, this is intended for now

protocol/x/clob/keeper/orders.go
jusbar23: done

northstar456: It's probably still cleaner to do the stateless check (without the state read). @jayy04 Wdyt? Could you remind us the place where we do stateless validations on place order message?

jyzhe: Validating the order router address can be done in ValidateBasic ([link](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/x/clob/types/message_place_order.go#L39)). ValidateBasic IIRC is deprecated or soon to be deprecated (?), but we were still using it for stateless checks. Invalid addresses will get rejected right away. 

If you want to keep the stateful checks, slightly cleaner to move this check to PerformStatefulOrderValidation ([link](https://github.com/dydxprotocol/v4-chain/blob/2b72c662d71fa5b91067b82fdf0d0bc31504e251/protocol/x/clob/keeper/orders.go#L828)), which is shared between short term and stateful.

teddyding: Thanks^
I see that we already have router address validation [here](https://github.com/dydxprotocol/v4-chain/blob/2b72c662d71fa5b91067b82fdf0d0bc31504e251/protocol/x/clob/types/message_place_order.go#L81) - so believe we are covered for stateless check

PR #2989, opened 2025-07-17
[ENG-239] Add support for updating referral codes

lukabuz:
### Changelist
- Add new route to allow for updating afilliate/referral codes
- Add tests for the above

Some notes
- The naming here is a bit confusing, within the db these codes are called 'usernames'. Within the controller code they are 'affiliate codes', and elsewhere 'referral codes'. This PR does not have any opinion on which one should be the de-facto name, but rather just tries to follow what is already there.
- Technically, there can be many rows in the subaccount-usernames table, but at the moment the uniqueness of subaccount id is enforced by the code(having no code path that allows for the creation of >1 row for a subaccount id). Therefore, some of the code in this PR makes the assumption that there is only 1 per subaccount id. To allow for more, there would have to be more structural frontend changes which was deemed out of scope of this specific PR.
- There are now new tests added for the signature verification function as it is already used in other places

### Test Plan
Unit testing(already done) and deploy to staging(or another dev env?) and test

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced a new API endpoint to update referral codes associated with user addresses.
  * Added support for updating usernames for subaccounts via the new endpoint.
  * Extended API documentation and schema to include the new referral code update functionality.
  * Added asynchronous functions to update and delete subaccount usernames by subaccount ID.

* **Documentation**
  * Updated API documentation with details, usage examples, and response schema for the new referral code update endpoint.

* **Tests**
  * Added comprehensive tests for the new referral code update endpoint, covering success and various failure scenarios.
  * Streamlined test setups by removing redundant username creation and updating tests to use username updates.
  * Ensured test data cleanup by deleting usernames before subaccount deletions in relevant tests.
  * Cleaned username data before username generator tests for consistent test state.

* **Bug Fixes**
  * Improved test reliability by cleaning up username data before subaccount deletions.
  * Removed redundant username creation in multiple test suites to streamline setup.

* **Chores**
  * Renamed internal constants for clarity related to request TTL.
  * Reorganized database table references for subaccount usernames and subaccounts.


lukabuz:
@coderabbitai summary

indexer/services/comlink/public/api-documentation.md
```diff
@@ -581,6 +581,96 @@ fetch(`${baseURL}/affiliates/metadata?address=string`,
...
+
+# For the deployment by DYDX token holders, use
+# baseURL = 'https://indexer.dydx.trade/v4'
+baseURL = 'https://indexer.v4testnet.dydx.exchange/v4'
+
+r = requests.post(f'{baseURL}/affiliates/code', headers = headers)
+
+print(r.json())
+
+```
```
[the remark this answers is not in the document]
lukabuz: Auto generated

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@ class AffiliatesController extends Controller {
...
+    // 2. There is already an entry for the subaccountId
+    await SubaccountUsernamesTable.update({
+      username: newCode,
+      subaccountId,
+    });
+
+    return {
+      referralCode: newCode,
+    };
+  }
```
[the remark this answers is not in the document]
lukabuz: The uniqueness check here is simply to provide an error message that is readable - the actual uniqueness is handled on the db level

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@
...
+
+    const subaccountId = subAccount[0].id;
+
+    // there are assumptions here that
+    // 1. There is only one entry per subaccountId
+    // 2. There is already an entry for the subaccountId
+    await SubaccountUsernamesTable.update({
+      username: newCode,
+      subaccountId,
+    });
```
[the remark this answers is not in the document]
lukabuz: I considered using the Transaction abstraction exported from the postgres package within comlink(doing transaction.get() to get a new one, passing it into all db calls, and committing it at the end of the controller) - but ran into some unstable behaviour there(other db calls started failing because the default transaction id was being picked up for some reason).

I looked at other uses of the Transaction class and saw that none of them were in comlink, they were all within the postgres package itself - this led me to conclude that it was not intended to be used within comlink. I made the decision not to use it because there are no real benefits to using transaction isolation here - the only write call is an update at the end, on a column that already has an unique constraint on it.

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@ class AffiliatesController extends Controller {
     };
   }
 
+  @Post('/code')
```
shrenujb: I'd favor a name like "referral" or "referralCode" over just 'code' for this route

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@ class AffiliatesController extends Controller {
...
+      address,
+      newCode,
+    }: {
+      address: string,
+      newCode: string,
+    } = body;
+
+    // Check if the referral code already exists.
+    // There is a unique constraint but doing this allows us to have a better error message.
+    const existingUsernameRow = await SubaccountUsernamesTable.findByUsername(
```
shrenujb, line 149: this might be a question for @teddyding but why are the affiliate names tied to subaccountId rather than just address?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@ class AffiliatesController extends Controller {
...
+        address,
+        subaccountNumber: 0,
+      },
+      [],
+    );
+    // There is a code-level restriction, but it is possible to
+    // have more than one subaccount for an address
+    // It is also possible for there to be no username, if the task to create it has not run yet
+    if (subAccount.length !== 1) {
+      throw new InvalidParamError('Referral code does not exist');
```
shrenujb, line 166: Are all subaccountIds already guaranteed to have "username" associated with them?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +128,59 @@ class AffiliatesController extends Controller {
...
+    }
+
+    const subaccountId = subAccount[0].id;
+
+    // there are assumptions here that
+    // 1. There is only one entry per subaccountId
+    // 2. There is already an entry for the subaccountId
```
shrenujb: Is assumption #2 a guarantee? All subaccountIds already have a affiliate name associated with them?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -287,6 +345,88 @@ router.get(
   },
 );
 
+router.post(
+  '/code',
+  ...checkSchema({
+    address: {
```
shrenujb: there are already some predefined checkSchema functions which you may want to use here instead of defining your own

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -287,6 +345,88 @@ router.get(
...
+      in: ['body'],
+      isString: true,
+      errorMessage: 'address must be a valid string',
+    },
+    newCode: {
+      in: ['body'],
+      isString: true,
```
shrenujb: is this sufficient validation for the referralCode? I assume we wouldn't want any special characters in this

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
shrenujb: would also change the name to be more reflective in the rest of the PR

lukabuz: Its not clear in this context but the full route here would be `POST /affiliates/code` which I figured was descriptive enough. Can change it to either though, I don't have a strong opinion on it

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
lukabuz: The way entries in this table are generated are through the subaccount-username-generator task in roundtable - which runs at a default of every 30 seconds. Also, the UI of this is set up in a way where this route can not be called without there already being a code(the edit button appears next to the displayed code).

So there is a condition where if you call this route within 30s of creating a subaccount this route will fail. The way to fix this would be an upsert instead of an update, which I tried but ran into [this issue](https://github.com/dydxprotocol/v4-chain/pull/2989#discussion_r2213544608) with transaction isolation.

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
lukabuz: Changes pushed

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
lukabuz: Changes pushed

indexer/services/comlink/__tests__/controllers/api/v4/affiliates-controller.test.ts
```diff
@@ -177,6 +208,135 @@ describe('affiliates-controller#V4', () => {
...
+      verifySignatureMock.mockResolvedValue(true);
+      fromFixedLengthMock.mockResolvedValue({} as ExtendedSecp256k1Signature);
+      verifyADR36AminoMock.mockReturnValue(true);
+      toBech32Mock.mockReturnValue(testConstants.defaultAddress);
+      jest.spyOn(DateTime, 'now').mockReturnValue(DateTime.fromSeconds(1620000000)); // Mock current time
+      jest.spyOn(stats, 'increment');
+    });
+
+    afterEach(async () => {
+      await SubaccountUsernamesTable.deleteBySubaccountId(defaultSubaccountId);
```
shrenujb: why do you have to do this and not just include it as part of clearData which happens next?

indexer/services/comlink/__tests__/controllers/api/v4/affiliates-controller.test.ts
```diff
@@ -177,6 +208,135 @@ describe('affiliates-controller#V4', () => {
...
 
+  describe('POST /code', () => {
+    const defaultSubaccountId = SubaccountTable.uuid(
+      testConstants.defaultSubaccount.address,
+      testConstants.defaultSubaccount.subaccountNumber,
+    );
+
+    beforeEach(async () => {
+      await testMocks.seedData();
+      await SubaccountUsernamesTable.create({
```
shrenujb: This could be part of the seedData?

indexer/services/comlink/src/lib/validation/schemas.ts
```diff
@@ -286,6 +286,47 @@ export const RegisterTokenValidationSchema = [
     }),
 ];
 
+export const UpdateReferralCodeSchema = checkSchema({
+  address: {
```
shrenujb, line 290: probably makes sense to create a const for this if it doesn't exist already given how common the address as a param should be

indexer/services/comlink/src/lib/validation/schemas.ts
```diff
@@ -286,6 +286,47 @@ export const RegisterTokenValidationSchema = [
...
+    in: ['body'],
+    isString: true,
+    errorMessage: 'newCode must be a valid string',
+    custom: {
+      options: validateReferralCode,
+    },
+  },
+  signedMessage: {
+    in: ['body'],
+    isString: true,
```
shrenujb, line 304: are there any length constraints that can be put on both signedMessage and pubKey? How does the other API validate these?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
shrenujb: Are the subaccount usernames only used for affiliates or something else as well?
Seems kinda unnecessary to have this generator-task running and instead just ask affiliates to always create their own link.

Understand this would be out of the scope of this specific task but wanted to discuss this somewhere

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
shrenujb: I think "referralCode" would still be a more clear name

indexer/services/comlink/src/lib/validation/schemas.ts
lukabuz: The other API does not have any validations on this - Given that ultimately this message and pubKey are used in cryptographic 'validation' anyway I figured it was not worth doing anything extra here

indexer/services/comlink/__tests__/controllers/api/v4/affiliates-controller.test.ts
lukabuz: The context here is that the deletion script had some fk constraints between the tables that was causing the Promise.all delete to fail, so I figured I would just handle it here instead of changing the core script. Updated the PR with changes the core clear script, its much cleaner

lukabuz:
@shrenujb pushed fixes for 2nd round of comments

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
northstar456: Slight preference for `referralCode` to be consistent with https://github.com/dydxprotocol/v4-chain/blob/5924462a042276929052782d9a2f2d9c2f52f40b/indexer/services/comlink/public/api-documentation.md?plain=1#L4408

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
northstar456: Yeah this is an unfortunate relic from the social work stream back in early 2024. It might have to do with different social profiles for different subaccounts. Didn't make sense to me but we already went ahead with it at the time

indexer/packages/postgres/src/helpers/db-helpers.ts
```diff
@@ -8,13 +8,13 @@ import { rawQuery } from './stores-helpers';
 const layer2Tables = [
   'perpetual_positions',
   'fills',
-  'subaccount_usernames',
```
northstar456, line 11: For my understanding, why is `subaccounts` -> `subaccount_usernames` here needed?

lukabuz: subaccount_usernames has a foreign key reference to subaccounts, so first we have to delete subaccount_usernames and then delete subaccounts

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +129,59 @@ class AffiliatesController extends Controller {
...
+        address,
+        subaccountNumber: 0,
+      },
+      [],
+    );
+    // There is a code-level restriction, but it is possible to
+    // have more than one subaccount for an address
+    // It is also possible for there to be no username, if the task to create it has not run yet
+    if (subAccount.length !== 1) {
+      throw new InvalidParamError('Referral code does not exist');
```
northstar456: This error message seems inaccurate - we don't expect returned `subaccounts.length > 1` for now so we only see this error if `subaccounts.length == 0`? Which is a weird edge case and we are basically telling user that they should wait until a default is generated. Perhaps use "referral code update not available yet - please try again later"?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -123,6 +129,59 @@ class AffiliatesController extends Controller {
...
+      address: string,
+      newCode: string,
+    } = body;
+
+    // Check if the referral code already exists.
+    // There is a unique constraint but doing this allows us to have a better error message.
+    const existingUsernameRow = await SubaccountUsernamesTable.findByUsername(
+      newCode,
+    );
+    if (existingUsernameRow) {
```
northstar456, line 152: Reasoning through a potential race condition here:

- Alice and Bob both request to use the same name. 
- Since the check here and the update below are not atomic, the second request (say Bob) to write to DB get will error with some weird error like `unique constraint violated`. 

I think this leads to Bob receiving an 500 internal error with the database error (?). If we don't do the check-update atomically, we could probably handle this case better

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -287,6 +346,59 @@ router.get(
...
+      newCode,
+      signedMessage,
+      pubKey,
+      timestamp,
+    }: CreateReferralCodeRequest = req.body;
+
+    try {
+      const failedValidationResponse = await validateSignature(
+        res,
+        AccountVerificationRequiredAction.UPDATE_CODE,
```
northstar456, line 374: Nit: now that we use `validateSignature` more generically, could we update this [constant](https://github.com/dydxprotocol/v4-chain/blob/5924462a042276929052782d9a2f2d9c2f52f40b/indexer/services/comlink/src/helpers/compliance/compliance-utils.ts#L80) used by `validateSignature` to something like `WRITE_REQUETS_TTL_SECONDS`?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
northstar456: Some context on subaccount username [here](https://dydx-team.slack.com/archives/C035Y4EBM7F/p1752847625090779?thread_ts=1752763256.149379&cid=C035Y4EBM7F). For now I think we should keep the default username behavior, since the current FE design assumes there's always an existing code

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
lukabuz: I considered this, but ran into this [this](https://github.com/dydxprotocol/v4-chain/pull/2989#discussion_r2213544608). If this is high priority, can you point me to whoever has the most context on the Transaction class so I can get in touch with them?

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
lukabuz: Change pushed

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
```diff
@@ -386,8 +386,58 @@ router.post(
...
+    } finally {
+      stats.timing(
+        `${config.SERVICE_NAME}.${controllerName}.create_code.timing`,
+        Date.now() - start,
+      );
+    }
+  },
+);
+
+router.post(
```
northstar456, line 410: Nit: could we add brief documentation on why separate endpoint is needed

lukabuz: Done

indexer/services/comlink/src/controllers/api/v4/affiliates-controller.ts
td-apriori: This is definitely an edge case and we just have to make sure is error is propagated correctly. Maybe catch the error from `await SubaccountUsernamesTable.update({` and just throw some sane error msg for now?

northstar456: This is an edge case and we just have to make sure is error is propagated correctly. Maybe catch the error from `await SubaccountUsernamesTable.update({` and just throw some sane error msg for now?

indexer/services/comlink/__tests__/controllers/api/v4/transfers-controller.test.ts
```diff
@@ -1104,6 +1105,7 @@ describe('transfers-controller#V4', () => {
     });
 
     it('Returns successfully when source subaccount does not exist', async () => {
+      await SubaccountUsernamesTable.deleteBySubaccountId(testConstants.defaultSubaccountId);
```
shrenujb, line 1108: Why does the transfers controller need this?

shrenujb: It has nothing to do with subaccount usernames

lukabuz: Right after this there is existing code that does a SubaccountTable.deleteById, and the fk constraint on the subaccount usernames table prevents this. Another way to fix this is to add an ON DELETE CASCADE to that constraint, but I figured since this only comes up in this specific test scenario(and afaik subaccount deletions dont generally happen outside of tests), its best not to add a migration for this case

shrenujb: Is the fk constraint newly added with this PR?

lukabuz: No, it was already there

lukabuz:
Original commits were not signed and I was unaware of the signature requirement - Had to retroactively sign commits and force push onto here @northstar456

lukabuz:
@mergifyio backport release/indexer/v8.1.7

lukabuz:
@mergify backport release/indexer/v8.1.x

PR #3000, opened 2025-07-24
[ENG-250] Add API endpoint for Turnkey Integration

Kefancao:
### Changelist
Add capabilities for sign-in and account creation via turnkey. 

### Test Plan

### Author/Reviewer Checklist


indexer/packages/postgres/src/stores/turnkey-users-table.ts
```diff
@@ -0,0 +1,86 @@
...
+  options: Options = DEFAULT_POSTGRES_OPTIONS,
+): Promise<TurnkeyUserFromDatabase | undefined> {
+  const baseQuery: QueryBuilder<TurnkeyUserModel> = setupBaseQuery<TurnkeyUserModel>(
+    TurnkeyUserModel,
+    options,
+  );
+  return baseQuery
+    .where(TurnkeyUserColumns.email, email)
+    .first();
+}
```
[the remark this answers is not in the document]
Kefancao: Will address

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,534 @@
...
+      rootUsers: [
+        {
+          userName: 'End User',
+          userEmail: params.email,
+          apiKeys: [],
+          authenticators,
+          oauthProviders,
+        },
+        {
+          userName: 'API User',
```
lukabuz, line 217: Is having the API user set as the root users the correct way to go here? Especially since the root quorum is set to 1 below. AFAIK this makes our API user can create arbitrary policies and even sign arbitrary transactions(through policies it can create)

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,534 @@
...
+      svm_address: svmAddress,
+      evm_address: evmAddress,
+      salt,
+      created_at: new Date().toISOString(),
+    });
+
+    // Best efforts to check that the subOrg.rootUserIds[0] is the end user
+    const user = await this.bridgeSenderApiClient.getUser({
+      organizationId: subOrg.subOrganizationId,
+      userId: subOrg.rootUserIds?.[0] as string,
```
lukabuz: Instead of using rootUserIds[0] here, just do a call to getUsers and find the user that matches the name "End user" that you set above - right now the order of the rootUserIds of the createSubOrganization does respect the order in which you input the users, but thats implicit behaviour and not guaranteed to stay the same.

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,534 @@
...
+      authenticators.push({
+        authenticatorName: params.authenticatorName,
+        challenge: params.challenge,
+        attestation: params.attestation,
+      });
+
+    }
+    const subOrg = await this.parentApiClient.createSubOrganization({
+      subOrganizationName: this.getUUID(),
+      rootUsers: [
```
lukabuz, line 208: This is a nit, but it may be a good idea to use turnkey's tag feature here. Just tag each end user as "end-user" or something, and same with the API user. It makes stuff like migrations and platform-wide maintenance scripts much easier to write in the future(easier to identify types of users)

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    return {
+      organizationId: suborg.subOrgId,
+      salt: suborg.salt,
+    };
+  }
+
+  // Utility methods
+  private isValidEmail(email: string): boolean {
```
lukabuz: Best to move this to a util function in the helpers folder. Also, email validation is a [fun rabbithole](https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression) but this regex should probably be good enough for now

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Yeah we unfortunately need this API user here until we can configure policies on it. We don't know the user's dydx address is until the front end gets a signature of the onboarding message :(

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Good point yeah

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Nice, didn't know they had this

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Sure can do!

PR #3004, opened 2025-07-29
Add market as query parameter (alias to ticker) to v4/PerpetualMarkets

davidli1997:
### Changelist
Adding `market` as a valid parameter, ex. `v4/perpetualMarkets?market=BTC-US`D.
It should not be used in conjunction with ticker, but otherwise it behaves the exact same way.

Motivation for this PR:

> this works: https://indexer.dydx.trade/v4/perpetualMarkets?ticker=BTC-USD
> 
> but this does not: [https://indexer.dydx.trade/v4/perpetualMarkets?market=BTC-USD](https://indexer.dydx.trade/v4/perpetualMarkets?ticker=BTC-USD)
> 
> Given that we use 'market' everywhere else in the API, please add support for 'market' as alias for 'ticker' on this and any other endpoint that only uses ticker. This change will make our API consistent with the other endpoints that only support 'market'.

### Test Plan
- `market` should only accepts valid value
- `market` should not be defined alongside ticker
- `market` should act as a filter

### Author/Reviewer Checklist

### DIFF
--- a/indexer/services/comlink/__tests__/controllers/api/v4/perpetual-markets-controller.test.ts
+++ b/indexer/services/comlink/__tests__/controllers/api/v4/perpetual-markets-controller.test.ts
@@ -162,20 +162,20 @@ describe('perpetual-markets-controller#V4', () => {
       );
     });
 
-    it('Returns 404 with unknown ticker or market', async () => {
+    it('Returns 400 with unknown ticker or market', async () => {
       let response: request.Response = await sendRequest({
         type: RequestMethod.GET,
         path: `/v4/perpetualMarkets?${getQueryString({ ticker: invalidTicker })}`,
-        expectedStatus: 404,
+        expectedStatus: 400,
       });
-      expect(response.body.error).toContain('Not Found');
+      expect(response.body.errors[0].msg).toContain('ticker must be a valid ticker');
 
       response = await sendRequest({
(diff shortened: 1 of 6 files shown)


indexer/services/comlink/__tests__/controllers/api/v4/perpetual-markets-controller.test.ts
```diff
@@ -162,14 +162,77 @@ describe('perpetual-markets-controller#V4', () => {
...
+        path: `/v4/perpetualMarkets?${getQueryString({ market: invalidTicker })}`,
+        expectedStatus: 400,
+      });
+      expect(response.body.errors[0].msg).toContain('ticker must be a valid ticker');
+    });
+
+    it('Returns 400 when both ticker and market are provided', async () => {
       const response: request.Response = await sendRequest({
         type: RequestMethod.GET,
-        path: `/v4/perpetualMarkets/${invalidTicker}`,
```
davidli1997, line 168: The old test was wrong, it was not checking for invalid ticker, it's returning `404` because of bad input format (using `/` instead of `?`)
Invalid `ticker` or `market` should return `400`.

davidli1997:
Tested on staging & works as expected.

tqin7 (approved):

PR #2815, opened 2025-04-28
[CT-1361] Add Support for TWAP Events in the Indexer

anmolagrawal345:
### Changelist
This PR is dependent on #2780, #2785, and #2787.

This change focuses on the indexer side of the changes for the TWAP order project. Now that the protocol can handle TWAP orders, we need emit and handle new events to add support to the indexer. The general workflow is going to be:
1. TWAP Order emits an order placement message and added to orders table
2. Suborders fills emit messages but placement does not (TBD)
3. Add a handler on the indexer side to update the quantity and average price on suborder fill messages

### Test Plan
Extend existing order placement handler tests to support twap orders and do e2e testing in dev/staging.

### Author/Reviewer Checklist
- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [ ] If the PR has breaking postgres changes to the indexer add the `indexer-postgres-breaking` label.
- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [ ] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [ ] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
- [ ] Manually add any of the following labels: `refactor`, `chore`, `bug`.

## Summary by CodeRabbit

* **New Features**
  * Added TWAP order support (TWAP orders and TWAP suborders) across placement, fills, removals, and public APIs; responses expose optional duration, interval, priceTolerance.
* **Database**
  * Added orders columns for duration, interval, priceTolerance and extended order/fill enums to include TWAP types; migration included.
* **Documentation**
  * OpenAPI and public API docs updated to describe TWAP types, fields, and query filtering.
* **Tests**
  * Extended test suites and helpers to cover TWAP workflows and edge cases.
* **Chores**
  * Added DB helper constants and SQL mapping for TWAP flag handling.

PR #3039, opened 2025-08-11
[CT-1365] Safely Handle Exceptions to TWAP Suborders

anmolagrawal345:
### Changelist
In TWAP order execution, we call HandleMsgPlaceOrder for suborders. While unexpected, we should ensure that the code does not panic since this is executed during the EndBlocker. To do so, let's wrap the function call with a recover, log the error, and simply proceed with the EndBlocker if we do run into an error.

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improves stability of TWAP-triggered order placement by recovering from unexpected panics during end-block processing, preventing rare node crashes.
  * Converts placement panics into handled errors so failed TWAP orders and associated trigger entries are reliably cleaned up, avoiding stuck or inconsistent states.
  * Adds richer contextual logging for placement failures to aid troubleshooting while preserving normal success behavior.

### DIFF
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -1,8 +1,8 @@
 package keeper
 
 import (
-	"runtime/debug"
 	"math/big"
+	"runtime/debug"
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -2,7 +2,6 @@ package keeper
 
 import (
 	"math/big"
-	"runtime/debug"
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"runtime/debug"
 	"math/big"
 
 	errorsmod "cosmossdk.io/errors"
@@ -411,11 +412,16 @@ func (k Keeper) safeHandleMsgPlaceOrder(
 		// recover stops a panicking goroutine and returns the error if present
 		// this will help us catch any exceptions resulting from order placement
 		if r := recover(); r != nil {
+			var orderId any
+			if msg != nil {
+				orderId = msg.Order.OrderId
+			}
 			k.Logger(ctx).Error(
 				"panic recovered in HandleMsgPlaceOrder for twap suborder",
 				"panic", r,
-				"orderId", msg.Order.OrderId,
+				"orderId", orderId,
 				"isStateful", isStateful,
+				"stack", string(debug.Stack()),
 			)
 			err = errorsmod.Wrapf(
 				types.ErrInvalidPlaceOrder,
--- a/protocol/x/clob/keeper/twap_order_state.go
+++ b/protocol/x/clob/keeper/twap_order_state.go
@@ -419,7 +419,7 @@ func (k Keeper) safeHandleMsgPlaceOrder(
 			"isStateful", isStateful,
 			"stack", fmt.Sprintf("%+v", err),
 		)
-		
+
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -398,3 +399,37 @@ func (k Keeper) GenerateSuborder(
...
+
+// safeHandleMsgPlaceOrder safely calls HandleMsgPlaceOrder with panic recovery.
+// This is used in end blockers where panics should be caught and logged rather than
+// causing the entire block to fail.
+func (k Keeper) safeHandleMsgPlaceOrder(
+	ctx sdk.Context,
+	msg *types.MsgPlaceOrder,
+	isStateful bool,
+) (err error) {
+	defer func() {
```
northstar456: Instead of the current approach, we should use the [RunCached](https://github.com/dydxprotocol/v4-chain/blob/57a3d2f001ce17cac92ef5fc8ecae72d2fa219ea/protocol/lib/abci/util.go#L11) primitive to make sure execution is atomic (discard state if we panic). Otherwise we could end up in some weird dirty state

Search `abci.RunCached` for example usage

anmolagrawal345: updated it to use RunCached

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -398,3 +400,37 @@ func (k Keeper) GenerateSuborder(
 
 	return &order, true
 }
+
+// safeHandleMsgPlaceOrder safely calls HandleMsgPlaceOrder with panic recovery.
+// This is used in end blockers where panics should be caught and logged rather than
+// causing the entire block to fail.
+func (k Keeper) safeHandleMsgPlaceOrder(
```
northstar456, line 407: This function doesn't do a ton now besides calling run cached and propagating error. I'd recommend just getting rid of it and calling `RunCached` directly for readability.

anmolagrawal345: I actually prefer to keep it for readability because the function its being called from is already quite complex

northstar456: In that case, consider changing the function below which reduces nested blocks. 

Also, do we really need to wrap the `err` with `ErrInvalidPlaceOrder`? Seems like the error object is not actually used upstream besides checking against `nil` 

```
	err = abci.RunCached(ctx, func(ctx sdk.Context) error {
		return k.HandleMsgPlaceOrder(ctx, msg, isStateful)
	})
	if err == nil {
		return nil
	}

	k.Logger(ctx).Error(
		"failed to handle TWAP suborder placement via HandleMsgPlaceOrder (panic recovered or error)",
		"cause", err,
		"orderId", msg.Order.OrderId,
		"isStateful", isStateful,
		"stack", fmt.Sprintf("%+v", err),
	)
	return errorsmod.Wrapf(
		types.ErrInvalidPlaceOrder,
		"error in HandleMsgPlaceOrder: %v",
		err,
	)
```

protocol/x/clob/keeper/twap_order_state.go
```diff
@@ -206,7 +208,7 @@ func (k Keeper) GenerateAndPlaceTriggeredTwapSuborders(ctx sdk.Context) {
 			)
 
 			// place triggered suborder
-			err := k.HandleMsgPlaceOrder(ctx, &types.MsgPlaceOrder{Order: *op.suborderToPlace}, true)
+			err := k.safeHandleMsgPlaceOrder(ctx, &types.MsgPlaceOrder{Order: *op.suborderToPlace}, true)
```
northstar456, line 211: Confirming the downstream effect when see an error: we'd delete the entire TWAP order plus the sub order so it ceases to exit? Since no event is emitted the user will just see a lingering order right? Is that intended for now given we don't expect this to happen?

anmolagrawal345: Yep this is actually addressed in the indexer PR where we emit an cancelled event if the parent TWAP is deleted
https://github.com/dydxprotocol/v4-chain/pull/2815/files#diff-17fddf1d4e953f76005580c654e355c04130e2422507f541d0da4dbc1231e3e8

jusbar23 (approved):

anmolagrawal345:
@Mergifyio backport release/protocol/9.0.x

anmolagrawal345:
@Mergifyio backport release/protocol/v9.0.x

anmolagrawal345:

PR #3050, opened 2025-08-21
Kcao/skip go bridge ctlr

Kefancao:
### Changelist

### Test Plan

### Author/Reviewer Checklist


indexer/packages/postgres/src/db/migrations/migration_files/20250809162119_add_turnkey_users_smart_account_address.ts
```diff
@@ -0,0 +1,18 @@
+import * as Knex from "knex";
```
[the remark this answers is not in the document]
Kefancao: How does this prevent creating a duplicate index on the same column?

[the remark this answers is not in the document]
Kefancao: Done

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,549 @@
...
+    const evmAddressChecksum = checksumAddress(recovered);
+    let user = await TurnkeyUsersTable.findByEvmAddress(evmAddressChecksum);
+    if (!user) {
+      throw new TurnkeyError('No user found for recovered EVM address');
+    }
+
+    await TurnkeyUsersTable.updateDydxAddressByEvmAddress(user.evm_address, dydxAddress);
+
+    return { success: true };
+  }
```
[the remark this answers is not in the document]
Kefancao: We always use checkSummed

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,567 @@
...
+          session: resp.session,
+          salt: resp.salt,
+          dydxAddress: resp.dydxAddress || '',
+        };
+      } catch (error) {
+        throw this.wrapTurnkeyError(error, 'Social signin failed');
+      }
+    } else if (signinMethod === SigninMethod.PASSKEY) {
+      if (!challenge || !attestation) {
+        throw new Error('challenge and attestation are required for passkey signin');
```
shrenujb: change error to "Passkeys are not supported"

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,567 @@
...
+    let svmAddress = '';
+    // smart account address can be derived offchain before we send any user ops.
+    // smart account address is needed by the frontend to display the correct
+    // deposit address for the avalanche chain since it does not support eip7702.
+    let smartAccountAddress = '';
+    for (const address of subOrg.wallet?.addresses || []) {
+      if (address.startsWith('0x')) {
+        // evm always starts with 0x
+        evmAddress = address;
+        smartAccountAddress = await getSmartAccountAddress(evmAddress);
```
shrenujb: This is a remote call. What happens when this times out or fails?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,567 @@
...
+      smart_account_address: smartAccountAddress,
+      salt,
+      created_at: new Date().toISOString(),
+    });
+
+    // need to also add the svm and evm addresses to the alchemy hook
+    if (evmAddress && svmAddress) {
+      // We don't need to wait for it since
+      // frontend doesn't really neeed the results???
+      addAddressesToAlchemyWebhook(evmAddress, svmAddress).catch((error) => {
```
shrenujb: this should probably wait for completion

indexer/services/comlink/src/helpers/alchemy-helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+}
+
+export function getRPCEndpoint(chainId: string): string {
+  if (!isSupportedEVMChainId(chainId)) {
+    throw new Error(`Unsupported chainId: ${chainId}`);
+  }
+  return `${config.ZERODEV_API_BASE_URL}/${config.ZERODEV_API_KEY}/chain/${chainId}`;
+}
+
+// TODO: Verify that this function is 1000% correct. @RUI and @TYLER and @JARED
```
Kefancao, line 252: NEED TO BE REVIEWED. SUPER DUPER IMPORTANT @ruixhuang @tyleroooo @jaredvu

indexer/services/comlink/src/helpers/alchemy-helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+  sourceAddress: string,
+  dydxAddress: string,
+): string {
+  if (isSupportedEVMChainId(chainId) || chainId === 'solana') {
+    return sourceAddress;
+  }
+  switch (chainId) {
+    case 'noble-1':
+      return toClientAddressWithPrefix(CosmosPrefix.NOBLE, dydxAddress) || '';
+    case 'osmosis-1':
```
shrenujb: Will Skip still route through osmosis or neutron post the only cctp change?
@tyleroooo

indexer/services/comlink/src/helpers/alchemy-helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+      return toClientAddressWithPrefix(CosmosPrefix.OSMO, dydxAddress) || '';
```
anmolagrawal345: lets throw an error here instead of falling back onto an empty string. we dont expect this so an error is appropriate

indexer/services/comlink/src/helpers/alchemy-helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+  chainId: string,
+  sourceAddress: string,
+  dydxAddress: string,
+): string {
+  if (isSupportedEVMChainId(chainId) || chainId === 'solana') {
+    return sourceAddress;
+  }
+  switch (chainId) {
+    case 'noble-1':
+      return toClientAddressWithPrefix(CosmosPrefix.NOBLE, dydxAddress) || '';
```
shrenujb: throw error if toClientAddressWithPrefix fails

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,567 @@
...
+    } else {
+      this.bridgeSenderApiClient = new TurnkeyServerSDK({
+        apiBaseUrl: config.TURNKEY_API_BASE_URL,
+        apiPrivateKey: config.TURNKEY_API_SENDER_PRIVATE_KEY,
+        apiPublicKey: config.TURNKEY_API_SENDER_PUBLIC_KEY,
+        defaultOrganizationId: config.TURNKEY_ORGANIZATION_ID,
+      }).apiClient();
+    }
+  }
+
```
Kefancao, line 96: Explain why

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,629 @@
...
+  chainId: string,
+}> {
+  const { event: { transaction, activity, network } } = e.body;
+  let chainId = '';
+  chainId = alchemyNetworkToChainIdMap[network];
+  const addressesToProcess = new Map<string, string>();
+  // for evm parsing only
+  if (activity) {
+    for (const act of activity) {
+      const fromAddress = act.toAddress;
```
Kefancao: change name

shrenujb: this should be called bridgeOriginAddress or something that explains that

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,629 @@
...
+    dydxAddress = record?.dydx_address || '';
+  } else {
+    throw new Error(`Unsupported chainId: ${chainId}`);
+  }
+  return dydxAddress;
+}
+
+@Route('bridging')
+class BridgeController extends Controller {
+  @Post('/sweep')
```
Kefancao: Remove

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,629 @@
...
+        },
+      });
+      logger.info({
+        at: `${controllerName}#sweep->startEvmBridge`,
+        message: 'Assets found',
+        assets,
+      });
+
+      for (let asset of assetsToSearch) {
+        asset = asset && asset.startsWith('0x') ? checksumAddress(asset as Address) : asset;
```
Kefancao: Check on this more

shrenujb: Add parens around "asset && asset.startsWith('0x')"

Kefancao: Remove the ternary and only checksum it

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,629 @@
...
+        const balance = assets?.chains?.[chainId]?.denoms?.[asset]?.amount;
+        if (balance && parseInt(balance, 10) > 0) {
```
Kefancao: Replace with THRESHOLD

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,629 @@
...
+  }
+
+  async startSolanaBridge(
+    fromAddress: string,
+    amount: string,
+    sourceAssetDenom: string,
+    chainId: string,
+  ): Promise<BridgeResponse> {
+    const path = await route({
+      goFast: true,
```
Kefancao, line 235: add allow_unsafe = false

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,228 @@
...
+        at: `${controllerName}#getSkipCallData`,
+        message: `Message ${index + 1} SVM Transaction`,
+        svmTx: msg.svmTx,
+      });
+    } else if ('multiChainMsg' in msg) {
+      logger.info({
+        at: `${controllerName}#getSkipCallData`,
+        message: `Message ${index + 1} Multi-Chain Message`,
+        multiChainMsg: msg.multiChainMsg,
+      });
```
Kefancao: remove

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: actually this is needed because sometimes asset doesn't start with 0x, in which case we do not want to check-sum it.

indexer/packages/postgres/src/db/migrations/migration_files/20250709170936_create_turnkey_users.ts
```diff
@@ -0,0 +1,25 @@
...
+  return knex
+    .schema
+    .createTable('turnkey_users', (table) => {
+      table.text('suborg_id').primary(); // Primary key
+      table.text('email').nullable().unique(); // optional but unique
+      table.text('svm_address').notNullable().unique(); // indexed
+      table.text('evm_address').notNullable().unique(); // indexed
+      table.text('salt').notNullable(); // used to generate dydx keypair when combining the onboarding signature
+      table.text('dydx_address').nullable().unique();
+      table.timestamp('created_at').notNullable();
```
shrenujb, line 13: what is this used for?

indexer/packages/postgres/src/db/migrations/migration_files/20250809162119_add_turnkey_users_smart_account_address.ts
```diff
@@ -0,0 +1,13 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  return knex.schema.alterTable('turnkey_users', (table) => {
+    table.text('smart_account_address').nullable().unique();
```
shrenujb: why do you have a separate migration for adding a column for a table you're newly adding in the above migration?
It makes more sense to me to have it all in one migration file

indexer/packages/postgres/src/stores/turnkey-users-table.ts
```diff
@@ -0,0 +1,134 @@
...
+export async function create(
+  turnkeyUserToCreate: TurnkeyUserCreateObject,
+  options: Options = { txId: undefined },
+): Promise<TurnkeyUserFromDatabase> {
+  return TurnkeyUserModel.query(
+    Transaction.get(options.txId),
+  ).insert(turnkeyUserToCreate).returning('*');
+}
+
+export async function upsert(
```
shrenujb, line 101: when would you use this?

indexer/packages/postgres/src/models/turnkey-user-model.ts
```diff
@@ -0,0 +1,67 @@
...
+        'suborg_id',
+        'svm_address',
+        'evm_address',
+        'salt',
+        'created_at',
+      ],
+      properties: {
+        suborg_id: { type: 'string' },
+        email: { type: ['string', 'null'] },
+        svm_address: { type: 'string' },
```
shrenujb, line 25: each of email, svm_address, evm_address, smart_account_address and dydx_address should have unique constraints on them. Is there a way to specify that when creating the table?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+} from '../../../types';
+
+// Polyfill fetch globally as it's needed by the turnkey sdk.
+/* eslint-disable @typescript-eslint/no-explicit-any */
+(global as any).fetch = fetch;
+
+export const router: express.Router = express.Router();
+const controllerName: string = 'turnkey-controller';
+
+interface SignInRequest {
```
shrenujb, line 37: Can you add some struct documentation explaining what each of these fields does?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
shrenujb: Unresolving since I don't see the docs. We need some documentation explaining why this function exists and what we expect the client flow to look like

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+      provider,
+      oidcToken,
+      challenge,
+      attestation,
+      magicLink,
+    } = body;
+    // Determine authentication method
+    if (signinMethod === SigninMethod.EMAIL) {
+      if (!userEmail || !targetPublicKey) {
+        throw new Error('userEmail is required for email signin');
```
shrenujb: error should be userEmail and targetPublicKey required for signin?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // email signin creates a suborg if it doesn't already exist.
+  private async emailSignin(
+    userEmail: string,
+    targetPublicKey: string,
+    magicLink?: string,
+  ): Promise<TurnkeyCreateSuborgResponse> {
+    // Validate email format
+    if (!isValidEmail(userEmail)) {
```
shrenujb: Ideally this should be on the API input validation itself

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+      return undefined;
+    }
+
+    // if we don't have an email, we need to find the suborg id by oidc token or credential id.
+    let suborgId: string;
+    if (p.oidcToken) {
+      suborgId = await this.getSuborgByOIDCToken(p.oidcToken);
+    } else if (p.credentialId) {
+      suborgId = await this.getSuborgByCredentialId(p.credentialId);
+    } else if (p.email) {
```
shrenujb: Why are you checking for email again?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+    };
+  }
+
+  // default 32 bytes.
+  private generateSalt(bytes: number = 32): string {
+    return randomBytes(bytes).toString('hex');
+  }
+
+  // this function assumes every user has only one suborg if they have an account with us.
+  private async getSuborgByOIDCToken(oidcToken: string): Promise<string> {
```
shrenujb: Are these remote calls? What exactly is happening here to retrieve the subOrg?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+        email: userEmail,
+      });
+    }
+
+    // Validate magic link template if provided
+    const magicLinkTemplate = config.TURNKEY_MAGIC_LINK_TEMPLATE || magicLink;
+    if (magicLinkTemplate) {
+      try {
+        // eslint-disable-next-line no-new
+        new URL(magicLinkTemplate.replace('%s', 'test'));
```
shrenujb: Similar to email, ideally this should be on the API input validation

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+          dydxAddress: user?.dydx_address || '',
+        };
+      }
+    }
+    return undefined;
+  }
+
+  // returns the suborgId plus salt and adds the user to the turnkey users table store.
+  private async createSuborg(params: CreateSuborgParams): Promise<TurnkeyCreateSuborgResponse> {
+    const oauthProviders: TurnkeyApiTypes['v1OauthProviderParams'][] = [];
```
shrenujb: where does 'v1OauthProviderParams' and 'v1AuthenticatorParamsV2' come from?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+
+      // serialize the attestation object.
+      authenticators.push({
+        authenticatorName: params.authenticatorName,
+        challenge: params.challenge,
+        attestation: params.attestation,
+      });
+
+    }
+    const subOrg = await this.parentApiClient.createSubOrganization({
```
shrenujb: As we spoke please rename parentApiClient to turnkeyApiClient or something similar which describes that this is a remote call to turnkey

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+      this.parentApiClient = turnkeyClient;
+    } else {
+      this.parentApiClient = new TurnkeyServerSDK({
+        apiBaseUrl: config.TURNKEY_API_BASE_URL,
+        apiPrivateKey: config.TURNKEY_API_PRIVATE_KEY,
+        apiPublicKey: config.TURNKEY_API_PUBLIC_KEY,
+        defaultOrganizationId: config.TURNKEY_ORGANIZATION_ID,
+      }).apiClient();
+    }
+    // bridge sender client is used to kick off bridging transactions.
```
shrenujb: Why would the parentApiClient and bridgeSenderApiClient be different?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+    // smart account address is needed by the frontend to display the correct
+    // deposit address for the avalanche chain since it does not support eip7702.
+    let smartAccountAddress = '';
+    for (const address of subOrg.wallet?.addresses || []) {
+      if (address.startsWith('0x')) {
+        // evm always starts with 0x
+        evmAddress = address;
+        smartAccountAddress = await getSmartAccountAddress(evmAddress);
+        smartAccountAddress = checksumAddress(smartAccountAddress as Address);
+      } else {
```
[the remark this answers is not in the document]
shrenujb: probably makes sense to take this coderrabbit suggestion

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+          {
+            curve: 'CURVE_ED25519',
+            pathFormat: 'PATH_FORMAT_BIP32',
+            path: "m/44'/501'/0'/0'", // Standard Solana derivation path
+            addressFormat: 'ADDRESS_FORMAT_SOLANA',
+          },
+        ],
+      },
+    });
+
```
shrenujb: Why does the below logic need to be part of the createSuborg function? It feels like these should be separated out, especially as this is a massive function doing remote calls do multiple different places

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
shrenujb: How do we handle any failures with these remote calls to turnkey today?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+      } catch {
+        throw new Error('Invalid magic link template URL');
+      }
+    }
+    const emailAuthResponse = await this.parentApiClient.emailAuth({
+      email: userEmail,
+      targetPublicKey,
+      emailCustomization: {
+        appName: 'dydx',
+        logoUrl: 'https://cdn.prod.website-files.com/649ca755d082f1dfc4ed62a4/6870a124cba22652a69c409d_icon%20(1).png',
```
shrenujb: These should be constants defined in a constants file

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+      suborg = await this.createSuborg({
+        providerName: provider,
+        oidcToken,
+      });
+    }
+
+    const oauthLoginResponse = await this.parentApiClient.oauthLogin({
+      oidcToken,
+      publicKey: targetPublicKey,
+      invalidateExisting: true,
```
shrenujb: what does this param mean?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+          session: resp.session,
+          salt: resp.salt,
+          dydxAddress: resp.dydxAddress || '',
+        };
+      } catch (error) {
+        throw this.wrapTurnkeyError(error, 'Social signin failed');
+      }
+    } else if (signinMethod === SigninMethod.PASSKEY) {
+      if (!challenge || !attestation) {
+        throw new Error('Passkey signin requires challenge and attestation. Passkey signing is currently not supported.');
```
shrenujb: "Passkey signing is currently not supported" should be thrown regardless of any condition?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+  private getUUID(): string {
+    return randomBytes(16).toString('hex');
+  }
+
+  /*
+   * Returns the suborgId plus salt if the user exists.
+   * Additionally will include the dydxAddress if the user has one uploaded already.
+   *
+   */
+  private async getSuborg(p: GetSuborgParams): Promise<TurnkeyCreateSuborgResponse | undefined> {
```
shrenujb: I think we should move some of these turnkey functions to a different turnkey specific lib file rather than having everything in the controller

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+          dydxAddress: resp.dydxAddress || '',
+        };
+      } catch (error) {
+        throw this.wrapTurnkeyError(error, 'Passkey signin failed');
+      }
+    }
+    throw new Error(`Invalid signin method. Must be one of: ${SigninMethod.EMAIL}, ${SigninMethod.SOCIAL}, ${SigninMethod.PASSKEY}`);
+  }
+
+  private getUUID(): string {
```
shrenujb: why do we need this? Why not just use the built in JS uuid?

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,558 @@
...
+
+    return {
+      organizationId: suborg.subOrgId,
+      salt: suborg.salt,
+      dydxAddress: suborg.dydxAddress || '',
+    };
+  }
+
+  // default 32 bytes.
+  private generateSalt(bytes: number = 32): string {
```
shrenujb: you have exactly the same function to generate salt and uuid. can be combined.
Also this should probably belong in a different file

indexer/packages/postgres/src/stores/turnkey-users-table.ts
Kefancao: I'll take this out

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: No we want to eventually add it. FE just won't call it rn.

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: It invalidates the old session.

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Okayyy

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
Kefancao: Throw an error and expects fe to retry.

indexer/packages/postgres/src/models/permission-approval-model.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  static get jsonSchema() {
+    return {
+      type: 'object',
+      required: [
+        'suborg_id',
+      ],
+      properties: {
+        suborg_id: { type: 'string' },
```
shrenujb, line 21: I dont think this should be the schema

schema should be
suborg_id
chain_id
approval

This would scale to N number of chains rather than having to add a new column for every new chain added

indexer/services/comlink/src/controllers/api/v4/test.py
```diff
@@ -0,0 +1,9 @@
+import os
```
shrenujb: think you added this local file by mistake

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
```diff
@@ -0,0 +1,276 @@
...
+    const user = await TurnkeyUsersTable.findByEvmAddress(evmAddressChecksum);
+    if (!user) {
+      throw new TurnkeyError('No user found for recovered EVM address');
+    }
+
+    await TurnkeyUsersTable.updateDydxAddressByEvmAddress(user.evm_address, dydxAddress);
+
+    // configure the policies now
+    await this.policyEngine.configureSolanaPolicy(dydxAddress, user.suborg_id);
+    // TODO: Add the evm policy configuration here too.
```
shrenujb: for any TODOs coming in the future. create a linear ticket and mention that in the TODO comment

indexer/services/comlink/src/helpers/alchemy-helpers.ts
```diff
@@ -0,0 +1,290 @@
...
+  PublicClient,
+} from 'viem';
+import {
+  arbitrum, avalanche, base, mainnet, optimism,
+} from 'viem/chains';
+
+import config from '../config';
+import { create4xxResponse } from '../lib/helpers';
+
+const evmChainIdToAlchemyWebhookId: Record<string, string> = {
```
shrenujb, line 22: These should ideally be defined in a constants file

indexer/packages/postgres/src/models/permission-approval-model.ts
Kefancao: DOne

indexer/services/comlink/src/helpers/alchemy-helpers.ts
Kefancao: okay

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,803 @@
...
+  endpointOptions: {
+    endpoints: {
+      solana: {
+        rpc: 'https://go.skip.build/api/rpc/solana',
+      },
+    },
+  },
+});
+
+// need to add this so that the address that triggered the activity is ignored.
```
shrenujb: this is commented out. do you still need this?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,803 @@
...
+async function getDydxAddress(address: string, chainId: string): Promise<string> {
+  let dydxAddress = '';
+  if (isSupportedEVMChainId(chainId)) {
+    // look up in turnkey table
+    logger.info({
+      at: `${controllerName}#getDydxAddress`,
+      message: 'Looking up in turnkey table for evm address',
+      address,
+    });
+    const normalizedAddress = (address && address.startsWith('0x'))
```
shrenujb: normalizedEvmAddress would be a better name here

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,793 @@
...
+} from '@zerodev/sdk';
+import { KERNEL_V3_3, KERNEL_V3_1 } from '@zerodev/sdk/constants';
+import express from 'express';
+import {
+  Controller, Post, Query, Route,
+} from 'tsoa';
+import {
+  Address,
+  checksumAddress, http,
+} from 'viem';
```
shrenujb, line 23: what is viem?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Took it out!

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: Still needs to be removed?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: it's an evm library

Kefancao: Has a lot of utility Fns

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: okok

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      solana: {
+        rpc: 'https://go.skip.build/api/rpc/solana',
+      },
+    },
+  },
+});
+
+const turnkeySenderClient = new Turnkey({
+  apiBaseUrl: config.TURNKEY_API_BASE_URL as string,
+  apiPublicKey: config.TURNKEY_API_SENDER_PUBLIC_KEY as string,
```
shrenujb: How're the public and private keys added to the config?
Is there any risk in exposing these as they are stored as strings in memory?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      const assets = await balances({
+        chains: {
+          [chainId]: {
+            address: fromAddress,
+            denoms: [usdcAddressByChainId.solana],
+          },
+        },
+      });
+      // get usdc amount and check that the usd amount is greater than the threshold.
+      const usdAmt = assets?.chains?.[chainId]?.denoms?.[usdcAddressByChainId.solana]?.valueUsd;
```
shrenujb: There's a lot of places in this chain where things can fail and end up having usdAmt as undefined. Seems like you need another check to see if this is undefined and fail

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      if (!amount) {
```
shrenujb, line 135: This validation should be in the param validation

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+        throw new Error('Amount is required for solana');
+      }
+      const amountNum = parseInt(amount, 10);
+      if (!Number.isFinite(amountNum) || amountNum <= 0) {
```
shrenujb: This validation should also live in the param validation

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+        throw new Error('Amount must be a positive integer for solana');
+      }
+      if (usdAmt && parseFloat(usdAmt) >= config.BRIDGE_THRESHOLD_USDC) {
```
shrenujb: add another parens on the 2nd conditional to make it explicit

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+          await this.startSolanaBridge(fromAddress, amount, usdcAddressByChainId.solana, chainId);
+        } catch (error) {
+          logger.error({
+            at: `${controllerName}#sweep->startSolanaBridge`,
+            message: `Failed to bridge token ${usdcAddressByChainId.solana}`,
+            error,
+          });
+        }
+      } else {
+        throw new Error(`Amount must be greater than ${config.BRIDGE_THRESHOLD_USDC} to start auto bridge`);
```
shrenujb: What does this mean then? If the user moves less than $20 into their solana wallet, that won't be bridged over?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+@Route('bridging')
+class BridgeController extends Controller {
+  @Post('/sweep')
+  async sweep(
+    @Query() fromAddress: string,
+      @Query() chainId: string,
+      // optionally provide the contract and amount, primarily used for solana.
+      @Query() amount?: string,
+  ): Promise<BridgeResponse> {
+    const normalizedFromAddress = (isSupportedEVMChainId(chainId) && fromAddress?.startsWith('0x'))
```
shrenujb: move this assignment into the evm if block

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      },
+      amountIn: amount,
+      sourceAssetDenom,
+      sourceAssetChainId: chainId,
+      destAssetChainId: 'dydx-mainnet-1',
+      cumulativeAffiliateFeeBps: '0',
+      destAssetDenom: usdcAddressByChainId['dydx-mainnet-1'],
+    });
+
+    const dydxAddress = await getDydxAddress(fromAddress, chainId);
```
shrenujb: this should probably go before you call the route function given thats a remote call and this one does an internal DB lookup

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+    if (!dydxAddress) {
+      throw new Error('Failed to derive dYdX address');
+    }
+
+    if (!path) {
```
shrenujb, line 248: Each error path should be right after the computation of the path

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      userAddresses,
+      getSvmSigner: getSvmSigner(record?.suborg_id || '', fromAddress),
+      svmFeePayer: {
+        address: solanaSponsorPublicKey, // Replace with the fee payer's Solana address
+        signTransaction: (dataToSign: Buffer) => {
+          const data = new Uint8Array(dataToSign);
+          return turnkeyGasFeePayer.signMessage(data, solanaSponsorPublicKey);
+        },
+      },
+      onTransactionBroadcast: async (
```
shrenujb, line 283: bit confused with this syntax. whats going on here?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      throw new Error(`Failed to find a turnkey user for svm address: ${fromAddress}`);
+    }
+
+    const solanaSponsorPublicKey = config.SOLANA_SPONSOR_PUBLIC_KEY;
+    if (!solanaSponsorPublicKey) {
+      throw new Error(
+        'Missing required environment variable: SOLANA_SPONSOR_PUBLIC_KEY',
+      );
+    }
+    await executeRoute({
```
shrenujb, line 271: Does executeRoute not return anything based on the success or failure of the execution?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+      logger.info({
+        at: `${controllerName}#sweep->startEvmBridge`,
+        message: 'Assets found',
+        assets,
+      });
+
+      for (const asset of assetsToSearch) {
+        const normalizedAsset = (asset && asset.startsWith('0x')) ? checksumAddress(asset as Address) : asset;
+        const balance = assets?.chains?.[chainId]?.denoms?.[normalizedAsset]?.amount;
+        const usdAmount = assets?.chains?.[chainId]?.denoms?.[normalizedAsset]?.valueUsd;
```
shrenujb: same comment as above. lots of possible places where this becomes undefined

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+        // To sweep and asset, user needs to have at least BRIDGE_THRESHOLD_USDC in it.
+        if (balance && parseInt(balance, 10) > 0 &&
```
shrenujb, line 182: The balance also has to be greater than the bridge threshold to initiate the bridge?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+    } else {
+      throw new Error(`Unsupported chainId: ${chainId}`);
+    }
+
+    return {
+      success: true,
+    };
+  }
+
+  async startSolanaBridge(
```
shrenujb, line 223: This function should be in a helper file - something like bridging-helper

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: They are injected via secret manager.

No way around storing it as a string unfortunately.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Yup it's deeper down there!

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Can't be done in param validation as the usdAmt is an api call!

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+          at: `${controllerName}#startEvmBridgePre7702`,
+          message: 'Failed to get EOA address from smart account address',
+          error,
+        });
+        throw new Error(
+          `Cannot proceed with bridge, cannot get EOA address from smart account address : ${formatError(error)}`,
+        );
+      }
+    }
+    const normalizedSrcAddress = (srcAddress && srcAddress.startsWith('0x'))
```
shrenujb: I've seen this normalization line in several places now. There should be a helper function for this

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Yes.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: They are callbacks.

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+  // support for hex amounts.
+  let amountToUse = amount;
+  if (amount.startsWith('0x')) {
+    amountToUse = parseInt(amount, 16).toString();
+  }
+  const routeResult = await route({
+    amountIn: amountToUse, // Desired amount in smallest denomination (e.g., uatom)
+    sourceAssetDenom,
+    sourceAssetChainId: chainId,
+    destAssetDenom: usdcAddressByChainId['dydx-mainnet-1'],
```
shrenujb: "dydx-mainnet-1" should be a constant

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: callbacks give u info! 

This doesn't wait until onchain execution is success/fail.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Yup, always checking for undefined!

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: This is just to make sure that the balance is a positive number. Can't compare balance to bridge threshold since they're not on the same quantum/price.

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  logger.info({
+    at: 'skip-helper#getSkipCallData',
+    message: 'Route result obtained',
+    routeResult,
+    dydxAddress,
+  });
+
+  const userAddresses = await Promise.all(
```
shrenujb: I believe you have this duplicate logic in the startSolanaBridge as well. Some opportunity to consolidate it

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+    timeoutSeconds,
+    amountIn: routeResult?.amountIn,
+    amountOut: routeResult.estimatedAmountOut || '0',
+    sourceAssetChainId: routeResult?.sourceAssetChainId,
+    sourceAssetDenom: routeResult?.sourceAssetDenom,
+    destAssetChainId: routeResult?.destAssetChainId,
+    destAssetDenom: routeResult?.destAssetDenom,
+    operations: routeResult?.operations,
+    addressList,
+    slippageTolerancePercent: '1',
```
shrenujb: Why is this hardcoded to 1? This should be a config somewhere

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  const validLength = addressList.length === routeResult.requiredChainAddresses.length ||
+    addressList.length === routeResult.chainIds?.length;
+
+  if (!validLength) {
+    throw new Error('executeRoute error: invalid address list');
+  }
+
+  const timeoutSeconds = '60'; // Set a timeout for the messages request
```
shrenujb: Make this a constant or config

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+  if (Object.values(ethDenomByChainId).map(
+    (x) => x.toLowerCase(),
+  ).includes(sourceAssetDenom.toLowerCase())) {
+    value = BigInt(amount);
+  }
+
+  // this is the actual call data that is responsible for the bridge.
+  const callData = [
+    {
+      to: (toAddress.startsWith('0x') ? toAddress : (`0x${toAddress}`)) as Hex,
```
shrenujb, line 160: why do we have to do this? The response from Skip is unpredictable?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: took out the normalization stuff!

indexer/services/comlink/src/helpers/skip-helper.ts
```diff
@@ -0,0 +1,232 @@
...
+      value,
+      data: data.startsWith('0x') ? data as Hex : (`0x${data}`) as Hex, // "0x",
+    },
+  ];
+  // if it's usdc, we need a swap to eth first, this approves a swap.
```
shrenujb: wat?
why swap from usdc to eth?

indexer/services/comlink/src/helpers/skip-helper.ts
Kefancao: this is the bridging logic. We need them to give us the call data.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,753 @@
...
+  }
+
+  async getKernelAccount(
+    chainId: string,
+    fromAddress: string,
+    suborgId: string,
+  ): Promise<CreateKernelAccountReturnType<EntryPointVersion>> {
+    // Initialize a Turnkey-powered Viem Account
+    // needs to sign with eoa address.
+    const turnkeyAccount = await createAccount({
```
shrenujb: this is a general question but are there timeouts configured in all of these remote calls?

indexer/services/comlink/src/helpers/skip-helper.ts
Kefancao: This is just what skip requires.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: No timeouts configured if it's not required to be configurable.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: This is "amount" which is a param in your REST call

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: This is not resolved

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: Sure, but is there any risk of exposing this to the outside world?

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: But what does it mean when their API hangs? Will this thread just be stuck at that point?

indexer/packages/postgres/src/db/migrations/migration_files/20250709170936_create_turnkey_users.ts
shrenujb: this was unanswered @Kefancao 
whats the purpose of the created_at field?

Kefancao: This is just some metadata around the turnkey user. Could be used for us to know when accounts were created etc.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: 👍, left it in previously because balance was not necessarily the user's most up to date record of the true balance on chain. But will take out

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: hmm this file pretty much already contains all the core logic around bridging. The controller class holds the all the key bridging logic. Do you want me to move this class out to a separate file all together? Would prefer to not break down the bridging code into separate files.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: Yeah most of these requests are hitting Turnkey or zerodev api. I'm not sure if it'll be that likely for it to hang as opposed to just returning a 500 error when it's unavailable. 

What would you recommend here?

indexer/services/comlink/src/controllers/api/v4/test.py
Kefancao: Yup!

indexer/packages/postgres/src/stores/turnkey-users-table.ts
Kefancao: Oh nvm, this is because we want to be able to update the user to associate them with other auth methods in the future.

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
Kefancao: I don't think so. At least not obvious how that would happen to me!

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,768 @@
...
+  } else {
+    throw new Error(`Unsupported chainId: ${chainId}`);
+  }
+  return dydxAddress;
+}
+
+@Route('bridging')
+class BridgeController extends Controller {
+  async sweep(
+    @Query() fromAddress: string,
```
shrenujb: if sweep is not an external API, then these params don't need the Query() prefix

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: here is fine based on what you mentioned

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
shrenujb: We can't rely on external APIs to work properly so should have appropriate safeguards. Do most of these libs not provide a way to configure a timeout? If not, then maybe there's nothing we can do there

indexer/services/comlink/src/controllers/api/v4/skip-bridge-controller.ts
```diff
@@ -0,0 +1,768 @@
...
+  },
+);
+
+router.post(
+  '/startBridge',
+  ...CheckBridgeSchema,
+  handleValidationErrors,
+  ExportResponseCodeStats({ controllerName }),
+  async (req: express.Request, res: express.Response) => {
+    // await dbHelpers.clearData();
```
shrenujb: these comments need to be removed

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
shrenujb: unresolving this based on the action above

indexer/services/comlink/src/controllers/api/v4/turnkey-controller.ts
shrenujb: Does the current indexer code support passkeys? If so, we should remove the line "Passkey is not currently supported". If not and we don't want it to go into the next code block then it should always return an error if signinMethod is passkey

indexer/services/comlink/src/helpers/skip-helper.ts
shrenujb: The question is specifically on the ternary operator usage where it seems like you dont know the format of toAddress and data

indexer/services/comlink/src/helpers/skip-helper.ts
shrenujb: It requires a swap from usdc to eth first to bridge usdc? why?

PR #3079, opened 2025-09-16
Add Leverage To the Protocol

anmolagrawal345:
### Changelist

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

- **New Features**
  - Per-subaccount leverage management: submit UpdateLeverage transactions (CLI tx command, Msg RPC) and fetch leverage via new Leverage query (gRPC/LCD/CLI).

- **Rate Limiting**
  - Configurable per‑N‑block leverage‑update limits added to genesis/config and validated; enforced during CheckTx with a dedicated leverage rate limiter.

- **Tests**
  - E2E and rate‑limit tests added to validate leverage flows and enforcement.


protocol/x/clob/rate_limit/order_rate_limiter.go
```diff
@@ -188,3 +188,38 @@ func (r *placeAndCancelOrderRateLimiter) PruneRateLimits(ctx sdk.Context) {
...
+// supports limiting based upon how many leverage updates per account per N blocks.
+//
+// The rate limiting must only be used during `CheckTx` because the rate limiting information is not recovered
+// on application restart preventing it from being deterministic during `DeliverTx`.
+//
+// Depending upon the provided types.BlockRateLimitConfiguration, the returned RateLimiter may rely on:
+//   - `ctx.BlockHeight()` in RateLimit to track which block the rate limit should apply to.
+//   - `ctx.BlockHeight()` in PruneRateLimits and should be invoked during `EndBlocker`.
+func NewUpdateLeverageRateLimiter(config types.BlockRateLimitConfiguration) RateLimiter[string] {
+	if err := config.Validate(); err != nil {
```
shrenujb, line 203: Isn't this config going to be empty when the upgrade happens? Will this result in validation failure here?

protocol/x/clob/types/message_update_leverage.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var _ sdk.Msg = &MsgUpdateLeverage{}
+
+// ValidateBasic validates the message's subaccount ID and leverage data.
+func (msg *MsgUpdateLeverage) ValidateBasic() error {
```
shrenujb: ValidateBasic is deprecated so we shouldn't use that. Any message validation should be in the message handler itself

protocol/x/clob/keeper/leverage.go
```diff
@@ -0,0 +1,117 @@
...
+func (k Keeper) GetMaxLeverageForPerpetual(ctx sdk.Context, perpetualId uint32) (uint32, error) {
+	// Get the perpetual and its liquidity tier in one call
+	_, _, liquidityTier, err := k.perpetualsKeeper.GetPerpetualAndMarketPriceAndLiquidityTier(ctx, perpetualId)
+	if err != nil {
+		return 0, err
+	}
+
+	// Calculate max leverage: 1,000,000 / initial_margin_ppm
+	if liquidityTier.InitialMarginPpm == 0 {
+		return 0, types.ErrInvalidLeverage
```
shrenujb: is this the right error returned here? Seems like the error should be InitialMarginPpm set to 0 for this market

proto/dydxprotocol/clob/tx.proto
```diff
@@ -211,3 +214,26 @@ message MsgUpdateLiquidationsConfig {
...
+// MsgUpdateLeverage is a request type used for updating leverage for
+// perpetuals.
+message MsgUpdateLeverage {
+  option (cosmos.msg.v1.signer) = "subaccount_id";
+
+  // The subaccount that is updating leverage.
+  dydxprotocol.subaccounts.SubaccountId subaccount_id = 1;
+
+  // List of perpetual leverage settings.
+  repeated LeverageEntry perpetual_leverage = 2;
```
shrenujb: Do we ever expect to set leverage for multiple pairs at once? I would imagine it should only be required for one?
cc @tyleroooo

tyleroooo: Yeah the workflows I envision never include a bulk-set unless we're setting like a global default in which case we'd try to set it for all markets or maybe have a separate thing stored per user that's the default leverage

protocol/x/clob/types/message_update_leverage.go
```diff
@@ -0,0 +1,62 @@
...
+	if _, err := sdk.AccAddressFromBech32(msg.SubaccountId.Owner); err != nil {
+		return errorsmod.Wrap(
+			ErrInvalidAddress,
+			fmt.Sprintf(
+				"subaccount owner '%s' must be a valid bech32 address, but got error '%v'",
+				msg.SubaccountId.Owner,
+				err.Error(),
+			),
+		)
+	}
```
shrenujb: Do we have anything that wraps all these above validations in a single func? We use subaccountId in so many places so I'd expect this same validation logic to exist in multiple places

protocol/x/clob/types/message_update_leverage.go
```diff
@@ -0,0 +1,62 @@
...
+				fmt.Sprintf("leverage for perpetual %d cannot be zero", entry.PerpetualId),
+			)
+		}
+
+		if perpetualIds[entry.PerpetualId] {
+			return errorsmod.Wrap(
+				ErrInvalidLeverage,
+				fmt.Sprintf("duplicate perpetual ID %d", entry.PerpetualId),
+			)
+		}
```
shrenujb: Do we also need to validate that the perpetual Id entered is a valid perpetual id?

proto/dydxprotocol/clob/tx.proto
anmolagrawal345: do we expect API traders to use this? we could just keep the API flexible

protocol/x/clob/rate_limit/order_rate_limiter.go
anmolagrawal345: yep it will be empty to start but the validation loops through all the settings, so if it's empty the validation is a no-op. I added validation for our new config here as well

protocol/x/clob/types/message_update_leverage.go
anmolagrawal345: added validation that its a valid clobPairID and a mapping to a perpID exists

protocol/x/clob/types/message_update_leverage.go
anmolagrawal345: used the subaccountId validation

protocol/testutil/keeper/clob.go
```diff
@@ -278,6 +278,7 @@ func createClobKeeper(
 		constants.TestEncodingCfg.TxConfig.TxDecoder(),
 		flags.GetDefaultClobFlags(),
 		rate_limit.NewNoOpRateLimiter[sdk.Msg](),
+		rate_limit.NewNoOpRateLimiter[string](),
```
shrenujb: Add a side comment mentioning which rate limiter this is

anmolagrawal345: done

protocol/x/clob/types/block_rate_limit_config.go
```diff
@@ -9,6 +9,8 @@ const (
 	MaxShortTermOrdersAndCancelsPerNBlocksLimit     = 10_000_000
 	MaxStatefulOrdersPerNBlocksNumBlocks            = 10_000
 	MaxStatefulOrdersPerNBlocksLimit                = 1_000_000
+	MaxLeverageUpdatesPerNBlocksNumBlocks           = 2_000
+	MaxLeverageUpdatesPerNBlocksLimit               = 10_000
```
shrenujb, line 13: What does this mean? We allow 10k msgs per 2k blocks?

protocol/x/clob/types/leverage.go
```diff
@@ -0,0 +1,80 @@
...
+			)
+		}
+
+		// Validate that the clob pair ID is a valid clob pair ID
+		if clob, found := clobKeeper.GetClobPair(ctx, ClobPairId(entry.ClobPairId)); !found {
+			return errorsmod.Wrap(
+				ErrInvalidClob,
+				fmt.Sprintf("clob pair ID %d does not exist", entry.ClobPairId),
+			)
+		} else if _, err := clob.GetPerpetualId(); err != nil {
```
shrenujb: Why do we need this check? The protocol guarantees that a clob pair should always be mapped to a perpetual id

protocol/x/clob/types/leverage.go
```diff
@@ -0,0 +1,80 @@
...
+) (map[uint32]uint32, error) {
+	if err := ValidateUpdateLeverageMsg(ctx, msg, clobKeeper); err != nil {
+		return nil, err
+	}
+
+	perpetualLeverageMap := make(map[uint32]uint32)
+	for _, entry := range msg.ClobPairLeverage {
+		clob, _ := clobKeeper.GetClobPair(ctx, ClobPairId(entry.ClobPairId))
+		perpetualId := clob.MustGetPerpetualId()
+		perpetualLeverageMap[perpetualId] = entry.Leverage
```
shrenujb, line 71: Why construct this map based on the perpetual and not the clob pair? I imagine when we do collateral checks for placing orders, we actually use the clob pair more

anmolagrawal345: We actually use the perpetualID - cf AddOrderToOrderbookSubaccountUpdatesCheck in orders.go so I opted to just store the perpetualID directly

protocol/x/clob/types/leverage.go
anmolagrawal345: just a safety check

protocol/x/clob/types/block_rate_limit_config.go
anmolagrawal345: Ya that's the upper limit we will allow from the rate configs. Considering the rate limiting is a protected transaction we can still set tight rate limiting but this gives us similar flexibility to the other rate limited messages

protocol/x/clob/types/leverage.go
```diff
@@ -0,0 +1,75 @@
...
+			return errorsmod.Wrap(
+				ErrInvalidClob,
+				fmt.Sprintf("clob pair ID %d does not exist", entry.ClobPairId),
+			)
+		} 
+		clobPairIds[entry.ClobPairId] = true
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
anmolagrawal345: this is fine because this is handled in the UpdateLeverage function when the transaction is let through

jusbar23 (approved):

anmolagrawal345:

PR #3082, opened 2025-09-18
upgrade cometbft and v9.1

tqin7:
### Changelist
upgrade cometbft where validator set hash is reverted to the standard computation (it was modified during the introduction of reduced proposer set feature and ibc client errors were observed due to mismatched validator set hashes)

also add v9.1 upgrade handler

### Test Plan
- reproduced IBC error on localnet + staging without this cometbft upgrade
- verified successful IBC on localnet + staging with this cometbft upgrade
- will conduct upgrade test on staging once merged


tqin7:

PR #3094, opened 2025-09-23
Update TWAP Parent Store when Suborder is a Maker

anmolagrawal345:
### Changelist
There was a bug in the order matching logic where if a twap suborder is actually a maker order (which it can be for 3 seconds), and is matched, the parent twap order state is not updated (RemainingQuantums is not decremented). This can result in a user getting filled for more than they requested, because the next suborder will assume that there was no fill and increase in size in an attempt to catch up. 

This fix is simple - it just checks both the maker and taker order in the matching logic for if it is a suborder and updates the respective parent twap store. This also handles the case for if both maker and taker are twap suborders.

### Test Plan
Added a unit test to simulate this situation and verified that the RemainingQuantums was updated as expected.

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures TWAP remaining quantities update correctly when TWAP suborders act as maker or taker during matching, improving accuracy of state transitions, fills, and trigger timing across blocks.

* **Refactor**
  * Centralized TWAP remaining-quantity updates into a single helper flow to reduce inconsistency and edge-case risk.

* **Tests**
  * Added end-to-end coverage validating TWAP orders functioning as maker in matching scenarios, including triggers, suborder placements, and remaining legs/quantums.

### DIFF
--- a/protocol/x/clob/e2e/twap_orders_test.go
+++ b/protocol/x/clob/e2e/twap_orders_test.go
@@ -385,7 +385,7 @@ func TestTWAPOrderWithMatchingOrdersWhereTWAPOrderIsMaker(t *testing.T) {
 		},
 	}
 
-	// Place the TWAP order
+	// Place the TWAP order so the first suborder is resting on the book
 	for _, checkTx := range testapp.MustMakeCheckTxsWithClobMsg(
 		ctx,
 		tApp.App,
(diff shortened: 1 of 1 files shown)


anmolagrawal345:

PR #3109, opened 2025-09-29
Dont attribute new revenue if user exceeds 30d max volume and deprecate AffiliateWhitelist

jusbar23:
### Changelist
* Update protocol to use overrides instead of whitelist
* Don't attribute new revenue if user exceeds 30d max revenue 
* Add overrides to automatically set tier to max

### Test Plan
* Test cases

### Side Effects:
* Affilliate whitelist will no longer be recognized, only affiliate overrides

### Author/Reviewer Checklist

## Summary by CodeRabbit

- New Features
  - Stats now record per-fill affiliate fees and per-user affiliate revenue; stats expiration hooks enforce a 30-day rolling window and adjust attributed volumes.
  - Affiliate overrides (per-address) replace the previous whitelist behavior.
  - New API to retrieve user-specific stats.

- Changes
  - Top affiliate tier taker fee share increased to 25% (from 20%).
  - Rev-share and match processing now propagate affiliate revenue/override data; attribution caps applied and updated on expiration.

- Documentation
  - Clarified that global and user stats represent a 30-day rolling window.

- Chores
  - Affiliate CLI reads client config and uses renamed parameters.


proto/dydxprotocol/stats/stats.proto
```diff
@@ -60,6 +66,9 @@ message UserStats {
 
   // Maker USDC in quantums
   uint64 maker_notional = 2;
+
+  // Affiliate revenue generated in quantums
+  uint64 affiliate_revenue_generated_quantums = 3;
```
northstar456, line 72: Is this referring to the user as an referee? eg the revenue they generate to their affiliate? If so could we make this more clear in comments

protocol/x/affiliates/types/constants.go
```diff
@@ -26,7 +26,7 @@ var (
 			{
 				ReqReferredVolumeQuoteQuantums: 50_000_000_000_000, // 50 million USDC
 				ReqStakedWholeCoins:            100_000_000,        // 100m whole coins
-				TakerFeeSharePpm:               200_000,            // 20%
+				TakerFeeSharePpm:               250_000,            // 25%
```
northstar456, line 29: What's the context for this change?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -144,6 +144,42 @@ func (k Keeper) AddReferredVolume(
 	return nil
 }
 
+// AddReferredCommission adds the referred commission from a block to the commission a referred user generated.
+func (k Keeper) AddReferredCommission(
```
northstar456: Confused about the concept of "referred commission". Here we increment this value for a referee, so the value refer to "the affiliate revenue generated from one specific referee". However `GetReferredCommission` below considers the stored value as the total commission for an affiliate user? The fact that we are accessing the same prefix key with two different sets is concerning

Note we use `ReferredVolume` throughout this repo to refer to total referred volume from an affiliate.

protocol/x/revshare/keeper/revshare.go
```diff
@@ -192,7 +192,8 @@ func (k Keeper) ValidateRevShareSafety(
 func (k Keeper) GetAllRevShares(
 	ctx sdk.Context,
 	fill clobtypes.FillForProcess,
-	affiliatesWhitelistMap map[string]uint32,
+	affiliateOverrides map[string]bool,
+	affiliateParameters *affiliatetypes.AffiliateParameters,
```
northstar456: Prefer `pass-by-value` rather than `pass-by-pointer` especially this is a read-only input.

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -468,8 +541,24 @@ func (k Keeper) AggregateAffiliateReferredVolumeForFills(
...
 		}
 		if referredByAddrMaker != "" {
-			if err := k.AddReferredVolume(ctx, referredByAddrMaker, lib.BigU(fill.Notional)); err != nil {
-				return err
+			attributableVolume := fill.Notional
+			makerUserStats := k.statsKeeper.GetUserStats(ctx, fill.Maker)
+			previousVolume := (makerUserStats.TakerNotional + makerUserStats.MakerNotional +
+				previouslyAttributedVolume[fill.Maker])
+			// If parameter is 0 then no limit is applied
+			if affiliateParams.Maximum_30DAttributableVolumePerReferredUserNotional != 0 {
```
northstar456: Code repeat between make and taker. Let's add a helper function and re-use as much as possible.

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -453,8 +509,25 @@ func (k Keeper) AggregateAffiliateReferredVolumeForFills(
...
+				cap := affiliateParams.Maximum_30DAttributableVolumePerReferredUserNotional
+				if previousVolume >= cap {
+					attributableVolume = 0
+				} else if previousVolume+attributableVolume > cap {
+					attributableVolume = cap - previousVolume
+				}
+			}
+			previouslyAttributedVolume[fill.Taker] += attributableVolume
+			if attributableVolume > 0 {
+				if err := k.AddReferredVolume(ctx, referredByAddrTaker, lib.BigU(attributableVolume)); err != nil {
```
northstar456: To confirm my understanding, the net effect of this is so that an affiliate cannot gain tier solely by a few users? Isn't that easily sybil-able?

protocol/x/affiliates/types/constants.go
jusbar23: I added another affiliate tier at the top end for 25% rev share PPM, with the override this will automatically be selected

protocol/x/affiliates/keeper/keeper.go
jusbar23: Good call out, done

proto/dydxprotocol/stats/stats.proto
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
jusbar23: I don't think we actually need this, this is just so we can surface it from the affiliate-info RPC. However this data is already stored in UserStats so we can just reference this when we need to decide whether the user has reached the max attributable revenue in a 30d window

protocol/x/revshare/keeper/revshare.go
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
jusbar23: Yes agreed, the current assumption is that by doing this we will prevent users from having a single affiliate route all traffic and then moving up to the max fee tier (4bps to 2.5bps) By gating the attributable volume, we can keep all affiliates at 4bps (or whatever number seems appropriate)

protocol/scripts/affiliates/update_affiliate_parameters.py
```diff
@@ -71,7 +91,7 @@ def load_yml(file_path) -> Dict[str, Any]:
 
 def get_proposal_id(node, chain):
     cmd = [
-        "dydxprotocold",
+        "/Users/justinbarnett/projects/v4-chain/protocol/build/dydxprotocold",
```
northstar456: Fix this?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -270,7 +245,7 @@ func (k Keeper) GetTakerFeeShare(
 func (k Keeper) GetTierForAffiliate(
 	ctx sdk.Context,
 	affiliateAddr string,
-	affiliateOverrides map[string]bool,
+	affiliateOverrides *map[string]bool,
```
northstar456: You almost never have to pass map is a pointer since it already is reference type. Note `map[string]bool` can be `nil` itself

protocol/scripts/affiliates/update_affiliate_parameters.py
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -487,6 +464,35 @@ func (k Keeper) GetAffiliateOverridesMap(ctx sdk.Context) (map[string]bool, erro
 	return affiliateOverridesMap, nil
 }
 
+func (k Keeper) addReferredVolumeIfQualified(ctx sdk.Context, referee string,
```
northstar456: Nit: put on separate lines

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -487,6 +464,35 @@ func (k Keeper) GetAffiliateOverridesMap(ctx sdk.Context) (map[string]bool, erro
...
+	referrer string, volume uint64, affiliateParams *types.AffiliateParameters,
+	previouslyAttributedVolume *map[string]uint64) error {
```
northstar456: Nit: don't take `map` as pointer

protocol/x/revshare/keeper/revshare.go
northstar456: We are still passing `	affiliateParameters *affiliatetypes.AffiliateParameters` here?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -436,12 +452,60 @@ func (k Keeper) GetAffiliateOverrides(ctx sdk.Context) (types.AffiliateOverrides
...
+	// Get the user stats from the referee
+	refereeUserStats := k.statsKeeper.GetUserStats(ctx, referee)
+
+	// If parameter is 0 then no limit is applied
+	previousVolume := (refereeUserStats.TakerNotional + refereeUserStats.MakerNotional +
+		(*previouslyAttributedVolume)[referee])
+
+	if affiliateParams.Maximum_30DAttributableVolumePerReferredUserNotional != 0 {
```
northstar456: For readability,  `affiliateParams.Maximum_30DAttributableVolumePerReferredUserNotional` and use `cap` in formulas. Or, take in an additional `volumeCapPerReferee` in this function

protocol/x/affiliates/types/keys.go
```diff
@@ -13,7 +13,9 @@ const (
 const (
 	ReferredByKeyPrefix = "RB:"
 
-	ReferredVolumeKeyPrefix = "RV:"
+	ReferredVolumeInWindowKeyPrefix = "RVW:"
+
+	ReferredCommissionKeyPrefix = "RC:"
```
northstar456, line 18: Is this still used?

protocol/x/affiliates/keeper/keeper.go
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
jusbar23: Removed

protocol/x/affiliates/keeper/keeper.go
jusbar23: Added a temporary cap variable for readability

protocol/x/affiliates/types/keys.go
jusbar23: Removed Referred Commission

protocol/x/affiliates/keeper/keeper.go
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -197,7 +222,7 @@ func (k Keeper) GetAllAffilliateOverrides(ctx sdk.Context) (types.AffiliateOverr
 func (k Keeper) GetTakerFeeShare(
 	ctx sdk.Context,
 	address string,
-	affiliatesWhitelistMap map[string]uint32,
+	affiliateOverrides *map[string]bool,
```
teddyding: Take in map directly without a pointer

teddyding: Pls do a search of `*map[` in this PR and make sure we get to all of those

jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
jusbar23: Thanks, didn't know this

protocol/x/revshare/keeper/revshare.go
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
northstar456: Did you commit this change? Still seeing the same as before

jusbar23: Sorry I think it was committed to a different branch

protocol/scripts/affiliates/update_affiliate_parameters.py
```diff
@@ -31,7 +33,7 @@
 def vote_for(node, chain, proposal_id, person):
     print("voting as " + person)
     cmd = [
-        "dydxprotocold",
+        "dydxprotocol",
```
northstar456: Is this intended? Isn't the binary `dydxprotocold`

jusbar23: Yeah, this is just a helper script so can be different per user. I'll update to dydxprotocold

jusbar23:
https://github.com/Mergifyio backport release/indexer/v9.x

jusbar23:

jusbar23:
https://github.com/Mergifyio backport release/indexer/v9.x

PR #3116, opened 2025-10-02
Add no fee holiday to protocol

davidli1997:
### Changelist

### Test Plan

### Author/Reviewer Checklist


protocol/x/feetiers/types/fee_holiday.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+const (
+	// Maximum duration for a fee holiday (30 days in seconds)
+	MaxFeeHolidayDuration = 30 * 24 * 60 * 60
+)
+
+func (m *FeeHolidayParams) Validate(currentTime time.Time) error {
+	// Validate time range (start < end)
+	if m.StartTimeUnix >= m.EndTimeUnix {
```
shrenujb, line 14: this is reversed?

protocol/x/feetiers/types/fee_holiday.go
```diff
@@ -0,0 +1,30 @@
...
+		return ErrInvalidTimeRange
+	}
+
+	// Validate reasonable time range (max 30 days)
+	duration := m.EndTimeUnix - m.StartTimeUnix
```
shrenujb, line 19: I dont think we need to add a max duration. This is something that can be checked on governance. If we have to set a time, something like 90 days is potentially better

protocol/x/feetiers/types/fee_holiday.go
shrenujb: sorry this is correct. didnt see that this was the failure case

PR #3117, opened 2025-10-02
Integrate commission and overrides to fee tier calculation

jusbar23:
### Changelist
* Don't attribute any more revenue if user has exceeded the 30d cap
* Use the new 30d affiliate volume window to calculate tier
* Interpret overrides as being the highest tier

### Test Plan
* Unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

- New Features
  - Affiliate info now includes 30-day rolling referred volume.
  - Fees respect a referee minimum fee tier override from affiliate parameters.
  - Enforced 30-day caps on affiliate revenue/volume attribution per referred user.
  - Affiliate parameters are included in genesis and exposed via queries.

- Chores
  - Renamed affiliate parameter fields to quote-quantum terminology across APIs and schemas; updated testnet defaults.
  - Updated protobufs and indexer codegen to match new fields.
  - Revised affiliate-parameters CLI script: new args/defaults and payload keys.

### DIFF
--- a/.github/workflows/protocol-build-and-push-snapshot.yml
+++ b/.github/workflows/protocol-build-and-push-snapshot.yml
@@ -6,7 +6,6 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/protocol/v[0-9]+.[0-9]+.x'  # e.g. release/protocol/v0.1.x
       - 'release/protocol/v[0-9]+.x'  # e.g. release/protocol/v1.x
-      - 'add_user_stats'
 
 jobs:
   build-and-push-snapshot-dev:
--- a/.github/workflows/indexer-build-and-push-dev-staging.yml
+++ b/.github/workflows/indexer-build-and-push-dev-staging.yml
@@ -7,7 +7,6 @@ on: # yamllint disable-line rule:truthy
       - 'release/indexer/v[0-9]+.[0-9]+.x'  # e.g. release/indexer/v0.1.x
       - 'release/indexer/v[0-9]+.x'  # e.g. release/indexer/v1.x
     # TODO(DEC-837): Customize github build and push to ECR by service with paths
-      - 'add_user_stats'
 
 jobs:
   # Build and push to dev
--- a/.github/workflows/protocol-build-and-push.yml
+++ b/.github/workflows/protocol-build-and-push.yml
@@ -6,7 +6,6 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/protocol/v[0-9]+.[0-9]+.x'  # e.g. release/protocol/v0.1.x
       - 'release/protocol/v[0-9]+.x'  # e.g. release/protocol/v1.x
-      - 'add_user_stats'
 
 jobs:
   build-and-push-dev:
--- a/protocol/x/affiliates/keeper/grpc_query_test.go
+++ b/protocol/x/affiliates/keeper/grpc_query_test.go
@@ -29,11 +29,15 @@ func TestAffiliateInfo(t *testing.T) {
 				Address: constants.AliceAccAddress.String(),
 			},
 			res: &types.AffiliateInfoResponse{
-				IsWhitelisted:             false,
-				Tier:                      0,
-				FeeSharePpm:               types.DefaultAffiliateTiers.Tiers[0].TakerFeeSharePpm,
-				ReferredVolume_30DRolling: dtypes.NewIntFromUint64(types.DefaultAffiliateTiers.Tiers[0].ReqReferredVolumeQuoteQuantums),
-				StakedAmount:              dtypes.NewIntFromUint64(uint64(types.DefaultAffiliateTiers.Tiers[0].ReqStakedWholeCoins) * 1e18),
+				IsWhitelisted: false,
+				Tier:          0,
+				FeeSharePpm:   types.DefaultAffiliateTiers.Tiers[0].TakerFeeSharePpm,
+				ReferredVolume_30DRolling: dtypes.NewIntFromUint64(
+					types.DefaultAffiliateTiers.Tiers[0].ReqReferredVolumeQuoteQuantums,
+				),
+				StakedAmount: dtypes.NewIntFromUint64(
+					uint64(types.DefaultAffiliateTiers.Tiers[0].ReqStakedWholeCoins) * 1e18,
+				),
 			},
 			setup: func(ctx sdk.Context, k keeper.Keeper, tApp *testapp.TestApp) {
 				err := k.RegisterAffiliate(ctx, constants.BobAccAddress.String(), constants.AliceAccAddress.String())
@@ -59,11 +63,15 @@ func TestAffiliateInfo(t *testing.T) {
 				Address: constants.AliceAccAddress.String(),
 			},
 			res: &types.AffiliateInfoResponse{
-				IsWhitelisted:             false,
-				Tier:                      0,
-				FeeSharePpm:               types.DefaultAffiliateTiers.Tiers[0].TakerFeeSharePpm,
-				ReferredVolume_30DRolling: dtypes.NewIntFromUint64(types.DefaultAffiliateTiers.Tiers[0].ReqReferredVolumeQuoteQuantums),
-				StakedAmount:              dtypes.NewIntFromUint64(uint64(types.DefaultAffiliateTiers.Tiers[0].ReqStakedWholeCoins) * 1e18),
+				IsWhitelisted: false,
+				Tier:          0,
+				FeeSharePpm:   types.DefaultAffiliateTiers.Tiers[0].TakerFeeSharePpm,
+				ReferredVolume_30DRolling: dtypes.NewIntFromUint64(
+					types.DefaultAffiliateTiers.Tiers[0].ReqReferredVolumeQuoteQuantums,
+				),
+				StakedAmount: dtypes.NewIntFromUint64(
+					uint64(types.DefaultAffiliateTiers.Tiers[0].ReqStakedWholeCoins) * 1e18,
+				),
 			},
 			setup: func(ctx sdk.Context, k keeper.Keeper, tApp *testapp.TestApp) {
 				stakingKeeper := tApp.App.StakingKeeper
--- a/protocol/x/affiliates/keeper/keeper_test.go
+++ b/protocol/x/affiliates/keeper/keeper_test.go
@@ -1095,7 +1095,7 @@ func TestUpdateAffiliateParameters(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(
 		t,
-		uint64(100_000_000_000),
+		uint64(100_000_000_000_000),
 		affiliateParameters.GetMaximum_30DAttributableVolumePerReferredUserQuoteQuantums(),
 	)
 	require.Equal(t, uint32(2), affiliateParameters.GetRefereeMinimumFeeTierIdx())


protocol/x/affiliates/types/constants.go
```diff
@@ -32,9 +32,9 @@ var (
 	}
 
 	DefaultAffiliateParameters = AffiliateParameters{
-		Maximum_30DAttributableVolumePerReferredUserNotional:       100,
-		RefereeMinimumFeeTierIdx:                                   1,
-		Maximum_30DAttributableRevenuePerReferredUserQuoteQuantums: 100,
+		Maximum_30DAttributableVolumePerReferredUserNotional:       100_000_000_000, // 100 million USDC
```
northstar456: Just realized we are storing whole USDC notional here vs. quote quantums below. Seems like these should be consistent. 

In ether unit, `100_000_000_000` is not 100 million USDC

protocol/x/affiliates/types/constants.go
```diff
@@ -32,9 +32,9 @@ var (
...
+		RefereeMinimumFeeTierIdx:                                   2,
```
northstar456: Are we intentionally bumping this? Note you'll need to do state migration (or gov vote)

protocol/x/feetiers/keeper/keeper.go
```diff
@@ -60,7 +60,11 @@ func (k *Keeper) SetVaultKeeper(vk types.VaultKeeper) {
 	k.vaultKeeper = vk
 }
 
-func (k Keeper) getUserFeeTier(ctx sdk.Context, address string) (uint32, *types.PerpetualFeeTier) {
+func (k Keeper) getUserFeeTier(
+	ctx sdk.Context,
+	address string,
+	feeTierOverride uint32,
```
northstar456: nit: `feeTierOverrideIdx`. `feeTierOverride`reads a like a fee amount.

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -582,20 +594,19 @@ func (k Keeper) OnStatsExpired(
...
 
-	// Subtract the expired volume (use taker volume for consistency with how it's added)
-	expiredVolume := lib.BigU(deltaAttributedVolume)
-	newVolume := new(big.Int).Sub(currentVolume, expiredVolume)
+	// Get the min of cap and resultingVolume
+	capInt := big.NewInt(int64(affiliateParams.Maximum_30DAttributableVolumePerReferredUserNotional))
+
+	// New volume should be the minimum of the resulting user volume, current referred volume,
+	// and below the cap
+	newVolume := lib.BigMin(capInt, lib.BigMin(currentVolume, big.NewInt(int64(resultingVolume))))
```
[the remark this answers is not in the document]
jusbar23: I thought I was being clever but it turns out I am the fool

.github/workflows/protocol-build-and-push-snapshot.yml
```diff
@@ -6,6 +6,7 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/protocol/v[0-9]+.[0-9]+.x'  # e.g. release/protocol/v0.1.x
       - 'release/protocol/v[0-9]+.x'  # e.g. release/protocol/v1.x
+      - 'add_user_stats'
```
northstar456: (Remember to remove)

proto/dydxprotocol/affiliates/query.proto
```diff
@@ -63,7 +63,7 @@ message AffiliateInfoResponse {
   // The affiliate's taker fee share in parts-per-million (for both VIP and
   // regular affiliate).
   uint32 fee_share_ppm = 3;
-  // The affiliate's all-time referred volume in quote quantums.
+  // The affiliate's 30d referred volume in quote quantums.
   bytes referred_volume = 4 [
```
northstar456, line 67: One thing I realized is that, we completely just updated `ReferredVolumeKeyPrefix` to `ReferredVolumeInWindowKeyPrefix` in the previous PR. 

Some concerns around this - 

- For existing affiliate that has already reached the top tier, are we okay with resetting their progress from zero? 
- We left the dangling `RV:` in state - should do a proper cleanup

protocol/x/revshare/keeper/revshare.go
```diff
@@ -302,8 +302,24 @@ func (k Keeper) getAffiliateRevShares(
...
 	}
 
+	userStats := k.statsKeeper.GetUserStats(ctx, takerAddr)
+	if userStats != nil {
+		// If the affiliate revenue generated is greater than the maximum 30d attributable volume
+		// per referred user notional, then no affiliate rev share is generated
+		// Disable this check if it is 0
+		cap := affiliateParams.Maximum_30DAttributableRevenuePerReferredUserQuoteQuantums
+		if cap != 0 &&
+			userStats.AffiliateRevenueGeneratedQuantums >= cap {
```
northstar456: Nit: these two names should be read similar. Since `Maximum_30DAttributableRevenuePerReferredUserQuoteQuantums` is the cap value on `AffiliateRevenueGeneratedQuantums`

proto/dydxprotocol/affiliates/query.proto
jusbar23: Talked offline, we are going to move everyone to 15% base, this will give everyone an automatic boost and for the 30d window they will get some extra commission.

We will clean up the state in the upgrade handler

protocol/x/revshare/keeper/revshare.go
jusbar23: Done

protocol/testing/testnet/genesis.json
```diff
@@ -40,7 +40,7 @@
       "affiliate_parameters": {
         "maximum_30d_attributable_volume_per_referred_user_notional": "100000000000",
```
northstar456: notional (indicates currency = USDC) and quote quantums (indicate unit = 1e-6 USDC) kinda means the same thing here. Just stick to `quote_quantums`?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -110,11 +109,17 @@ func (k Keeper) GetReferredBy(ctx sdk.Context, referee string) (string, bool) {
 	return string(referredByPrefixStore.Get([]byte(referee))), true
 }
 
+// SetReferredVolume sets the referred volume for an affiliate.
 func (k Keeper) SetReferredVolume(
```
northstar456: This function isn't used anymore?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -273,10 +278,11 @@ func (k Keeper) GetTierForAffiliate(
...
 
-	// If not then set it normally
-	referredVolume, err := k.GetReferredVolume(ctx, affiliateAddr)
-	if err != nil {
-		return 0, 0, err
+	// Get the affiliate revenue generated in the last 30d
+	userStats := k.statsKeeper.GetUserStats(ctx, affiliateAddr)
+	var referredVolume *big.Int
+	if userStats != nil {
+		referredVolume = big.NewInt(int64(userStats.AffiliateReferredVolumeQuoteQuantums))
```
northstar456:
```suggestion
		referredVolume = new(big.Int).SetUint64(userStats.AffiliateReferredVolumeQuoteQuantums)
```
don't do unsafe casting

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -474,11 +483,15 @@ func (k Keeper) addReferredVolumeIfQualified(
 ) error {
 	// Get the user stats from the referee
 	refereeUserStats := k.statsKeeper.GetUserStats(ctx, referee)
+	if refereeUserStats == nil {
```
northstar456, line 407: Do we always expect a non-nil stats for the referee here?

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -383,6 +391,7 @@ func (k Keeper) SetAffiliateWhitelist(ctx sdk.Context, whitelist types.Affiliate
 	return nil
 }
 
+// DO NOT USE: This will be deprecated soon.
```
northstar456, line 315:
```
// Deprecated: xxxxxxxxx
```

Use proper godoc for deprecation

protocol/x/affiliates/types/constants.go
```diff
@@ -32,9 +32,9 @@ var (
 	}
 
 	DefaultAffiliateParameters = AffiliateParameters{
-		Maximum_30DAttributableVolumePerReferredUserNotional:       100,
-		RefereeMinimumFeeTierIdx:                                   1,
-		Maximum_30DAttributableRevenuePerReferredUserQuoteQuantums: 100,
+		Maximum_30DAttributableVolumePerReferredUserNotional:    100_000_000_000, // 100 million USDC
```
northstar456: This is not 100k rather than 100 million USDC (assuming you mean quote quantums here as well)

protocol/x/affiliates/keeper/keeper.go
jusbar23: Yes, based on my understanding the UserStats should be set in the stats keeper. I'm not sure the ordering but I think this would cover the case if this is the user's first fill and they don't have stats yet. Please let me know if this is incorrect, I'm not super familiar

protocol/x/affiliates/keeper/keeper.go
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
jusbar23: Removed

protocol/x/affiliates/keeper/keeper.go
jusbar23: Done

protocol/x/affiliates/types/constants.go
jusbar23: Done

proto/dydxprotocol/affiliates/query.proto
northstar456, line 67: For best practice, if we are deprecating this field and creating another, suggest we create `referred_volume_30d_rolling` and mark the old field as deprecated ([deprecated = true])

proto/dydxprotocol/stats/stats.proto
```diff
@@ -70,6 +70,9 @@ message UserStats {
 
   // Affiliate revenue generated in quantums
   uint64 affiliate_revenue_generated_quantums = 3;
+
+  // Affiliate referred volume in quote quantums
```
northstar456:
```suggestion
  // Referred volume in quote quantums with this user being an affiliate
```

proto/dydxprotocol/stats/stats.proto
```diff
@@ -70,6 +70,9 @@ message UserStats {
...
+  uint64 affiliate_referred_volume_quote_quantums = 4;
```
northstar456: This 30d rolling right? Update var name to clarify. Same for var above

proto/dydxprotocol/stats/stats.proto
```diff
@@ -70,6 +70,9 @@ message UserStats {
 
   // Affiliate revenue generated in quantums
```
northstar456:
```suggestion
  // Affiliate revenue generated in quantums with this user being a referee
```

protocol/x/affiliates/types/constants.go
```diff
@@ -32,9 +32,9 @@ var (
 	}
 
 	DefaultAffiliateParameters = AffiliateParameters{
-		Maximum_30DAttributableVolumePerReferredUserNotional:       100,
-		RefereeMinimumFeeTierIdx:                                   1,
-		Maximum_30DAttributableRevenuePerReferredUserQuoteQuantums: 100,
+		Maximum_30DAttributableVolumePerReferredUserNotional:    100_000_000_000_000, // 100 million USDC
```
northstar456: Keep it consistent
```suggestion
		Maximum_30DAttributableVolumePerReferredUserQuoteQuantums    100_000_000_000_000, // 100 million USDC
```

protocol/x/clob/types/expected_keepers.go
```diff
@@ -116,7 +116,7 @@ type BlockTimeKeeper interface {
 }
 
 type FeeTiersKeeper interface {
-	GetPerpetualFeePpm(ctx sdk.Context, address string, isTaker bool) int32
+	GetPerpetualFeePpm(ctx sdk.Context, address string, isTaker bool, feeTierOverride uint32) int32
```
northstar456: Think I left a comment on the earlier PR to rename all `feeTierOverride` to `feeTierOverrideIdx`. The former can be misunderstood as raw fee amount

protocol/x/stats/types/expected_keepers.go
```diff
@@ -20,5 +20,5 @@ type StakingKeeper interface {
 
 // StatsExpirationHook is called when stats are expired from the rolling window
 type StatsExpirationHook interface {
-	OnStatsExpired(ctx sdk.Context, userAddress string, resultingUserStats *UserStats) error
+	OnStatsExpired(ctx sdk.Context, userAddress string, previousUserStats *UserStats, resultingUserStats *UserStats) error
```
northstar456: Delete this - not used?

jusbar23: Done

protocol/x/clob/types/expected_keepers.go
jusbar23: Done

protocol/x/affiliates/types/constants.go
jusbar23: Done

proto/dydxprotocol/stats/stats.proto
jusbar23: Done

proto/dydxprotocol/stats/stats.proto
jusbar23: Done

proto/dydxprotocol/stats/stats.proto
jusbar23: Done

proto/dydxprotocol/affiliates/query.proto
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
```diff
@@ -273,10 +199,11 @@ func (k Keeper) GetTierForAffiliate(
...
-	// If not then set it normally
-	referredVolume, err := k.GetReferredVolume(ctx, affiliateAddr)
-	if err != nil {
-		return 0, 0, err
+	// Get the affiliate revenue generated in the last 30d
+	userStats := k.statsKeeper.GetUserStats(ctx, affiliateAddr)
+	var referredVolume *big.Int
+	if userStats != nil {
+		referredVolume = new(big.Int).SetUint64(userStats.AffiliateReferredVolumeQuoteQuantums)
 	}
```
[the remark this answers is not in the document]
jusbar23: Done

.github/workflows/protocol-build-and-push-snapshot.yml
jusbar23: Done

protocol/x/affiliates/keeper/keeper.go
northstar456: If this is a valid scenario do we want to return error? Looks like we propagate this error all the way block and fail `AggregateAffiliateReferredVolumeForFills` as a whole. 

Could we make sure we test this case

jusbar23: It looks like it should never return a nil response https://github.com/dydxprotocol/v4-chain/blob/a4df7afa0137a3f3ce88b311ce6cfec08d827c16/protocol/x/stats/keeper/keeper.go#L152-L154

jusbar23: So i think this should never happen

jusbar23:
https://github.com/Mergifyio backport release/indexer/v9.x

jusbar23:
https://github.com/Mergifyio backport release/indexer/v9.x

jusbar23:

jusbar23:
https://github.com/Mergifyio backport release/indexer/v9.x

PR #3144, opened 2025-10-14
fix(skip-deposits): Fix rounding error introduced by parseInt

Kefancao:
### Changelist
Balances is returning a number that is large in precision and parseInt parses up to 16/17 digits, so extra digits just get rounded off which is causing bridges to fail since you can't transfer more than you have if the amount rounds upwards.

https://linear.app/dydx/issue/ENG-1078/fix-bug-in-rounding-on-auto-bridges

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric accuracy and consistency for bridge/deposit limit calculations (avoids rounding and type-related edge cases) and updated validation to accept numeric values as strings.
* **Tests**
  * Added extensive tests for limit logic, address encoding/conversion, error handling, and many edge cases to boost reliability and prevent regressions.

### DIFF
--- a/indexer/services/comlink/src/lib/validation/schemas.ts
+++ b/indexer/services/comlink/src/lib/validation/schemas.ts
@@ -272,7 +272,7 @@ const checkBridgeSchema: Record<string, ParamSchema> = {
   },
   'event.activity.*.value': {
     in: 'body',
-    isNumeric: true,
+    isString: true,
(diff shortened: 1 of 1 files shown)


UnbornAztecKing:

Kefancao:
@mergify backport release/indexer/v9.x

PR #3153, opened 2025-10-15
Check Leverage on Withdrawals and Updates

anmolagrawal345:
### Changelist
Now that we use leverage in our margining calculations - we should also ensure the same collateral checks are being applied in other flows. The 2 main ones here are validating that withdrawals and incoming leverage updates respect this.

For withdrawals, we get the check for free since it lives in the core risk calculation. The test creates 2 identical positions in 2 identical subaccounts, except one has a lower leverage set, and ensure that the withdrawable balance is lower than one without leverage set.

For incoming leverage updates, we take the new configuration and run it against the same risk checks as order placement, but with empty position updates. This will use the new leverage map to check if the position passes initial collateralization checks. If it does, we update the leverage map with the new configuration.

### Test Plan
Added tests in leverage_e2e_test to test both these cases.

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced leverage update validation to verify margin requirements before applying changes, preventing operations that would violate collateralization constraints.
  * Added comprehensive error handling for leverage updates that conflict with margin requirements.

* **Tests**
  * Added end-to-end tests validating withdrawal scenarios with leverage positions and margin constraints.
  * Added tests verifying leverage update behavior with existing positions.

### DIFF
--- a/protocol/x/clob/keeper/leverage_e2e_test.go
+++ b/protocol/x/clob/keeper/leverage_e2e_test.go
@@ -290,7 +290,12 @@ func TestWithdrawalWithLeverage(t *testing.T) {
 	// Verify leverage was set immediately (ante handler sets it)
 	aliceLeverageMap, found := tApp.App.SubaccountsKeeper.GetLeverage(ctx, &constants.Alice_Num0)
 	require.True(t, found, "Alice's leverage should be set")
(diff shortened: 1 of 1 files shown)


anmolagrawal345:

PR #3234, opened 2025-10-30
[ENG-1188] Move parent subaccount transfer filtering to SQL to respect limit parameter

davidli1997:
#### Bug Fix: `/transfers/parentSubaccountNumber` returning only ~10 results instead of respecting limit

**Problem:**
The `/transfers/parentSubaccountNumber` endpoint was incorrectly returning only 10-11 transfers even when no limit was specified (default should be 1000). 

**Root Cause:**
The endpoint had a two-step process:
1. Query database for up to 1000 transfers involving all child subaccounts of a parent
2. Filter out same-parent transfers (e.g., transfers between subaccount 0 and subaccount 128, both parent 0) in memory

When a parent subaccount had many internal transfers between its children, the database would return 1000 transfers (mostly same-parent), filter them out, and leave only ~10 cross-parent transfers.

**Solution:**
- Created new database function `findAllToOrFromParentSubaccount()` that filters same-parent transfers at the SQL level
- Uses LEFT JOINs with the `subaccounts` table to filter transfers where both sender and recipient have the same address AND same parent subaccount number (`subaccountNumber % 128`)
- The SQL filter runs BEFORE the LIMIT is applied, ensuring the correct number of cross-parent transfers are returned

**Changes:**
1. **New Database Function** (`packages/postgres/src/stores/transfer-table.ts`)
   - Added `findAllToOrFromParentSubaccount()` with SQL-level filtering
   - Added `ParentSubaccountTransferQueryConfig` interface
   - Filters same-parent transfers using: `(sender_sa.subaccountNumber % 128) = (recipient_sa.subaccountNumber % 128)`

2. **Updated Controller** (`indexer/services/comlink/src/controllers/api/v4/transfers-controller.ts`)
   - Modified `getTransfersForParentSubaccount()` to use new database function
   - Removed in-memory filtering logic (now handled in SQL)

#### Key Test Cases
1. **Same-parent filtering**: Verified transfers between child subaccounts of same parent (e.g., 0→128) are excluded
2. **Cross-parent transfers**: Verified transfers between different parents (e.g., parent 0→parent 1) are included
3. **Deposits/Withdrawals**: Verified wallet↔subaccount transfers are included
4. **Limit validation**: Created 35 same-parent + 15 cross-parent transfers, verified all 15 cross-parent transfers returned (not just 10)
5. **Pagination**: Verified pagination works correctly with SQL-level filtering

#### Manul Testing on Staging
curl "https://indexer.v4staging.dydx.exchange/v4/transfers/parentSubaccountNumber?address=dydx1nzuttarf5k2j0nug5yzhr6p74t9avehn9hlh8m&parentSubaccountNumber=0" | jq '{transferCount: (.transfers | length), totalResults: .totalResults, pageSize: .pageSize, offset: .offset}'
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  427k  100  427k    0     0   278k      0  0:00:01  0:00:01 --:--:--  278k
{
  "transferCount": 1000,
  "totalResults": null,
  "pageSize": null,
  "offset": null
}
```

Now it retrieves up the API limit - 1,000

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * Transfer query updated to exclude transfers between child subaccounts of the same parent and to support pagination, height/time filters, and deposits/withdrawals.

* **Behavior Changes**
  * API now returns the raw transformed transfer results from the parent-aware query (client-side parent filtering removed).

* **Tests**
  * Comprehensive tests added/updated to cover cross-parent vs same-parent transfers, pagination, limits, and edge cases.

### DIFF
--- a/indexer/packages/postgres/src/types/query-types.ts
+++ b/indexer/packages/postgres/src/types/query-types.ts
@@ -105,7 +105,6 @@ export enum QueryableField {
   SVM_ADDRESS = 'svm_address',
   EVM_ADDRESS = 'evm_address',
   DYDX_ADDRESS = 'dydx_address',
-  PARENT_SUBACCOUNT_NUMBER = 'parentSubaccountNumber',
 }
 
 export interface QueryConfig {


indexer/packages/postgres/src/stores/transfer-table.ts
```diff
@@ -330,6 +331,122 @@ export async function findAllToOrFromSubaccountId(
...
+    // Exclude transfers where both sender and recipient are child subaccounts of the same parent
+    .whereRaw(`
+      NOT (
+        transfers."senderSubaccountId" IS NOT NULL 
+        AND transfers."recipientSubaccountId" IS NOT NULL
+        AND sender_sa.address = recipient_sa.address 
+        AND (sender_sa."subaccountNumber" % 128) = (recipient_sa."subaccountNumber" % 128)
+      )
+    `)
+    .select('transfers.*');
```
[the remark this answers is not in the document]
davidli1997: The indexes already exist as composite indexes (e.g., transfers_sendersubaccountid_createdat_index). Postgres can use the first column of a composite index for JOIN operations, so no additional indexes are needed.

jusbar23 (approved):

indexer/packages/postgres/src/stores/transfer-table.ts
```diff
@@ -330,6 +331,126 @@ export async function findAllToOrFromSubaccountId(
...
+    createdBeforeOrAtHeight,
+    createdBeforeOrAt,
+    page,
+  }: ParentSubaccountTransferQueryConfig,
+  requiredFields: QueryableField[],
+  options: Options = DEFAULT_POSTGRES_OPTIONS,
+): Promise<PaginationFromDatabase<TransferFromDatabase>> {
+  verifyAllRequiredFields(
+    {
+      [QueryableField.LIMIT]: limit,
```
jusbar23, line 359: Should we set an upper limit for this? and validate that limit is greater than 0

davidli1997: We are already doing this on the API level
`{"errors":[{"value":-1,"msg":"limit must be a positive integer that is not greater than max: 1000","param":"limit","location":"query"}]}`
`{"errors":[{"value":0.5,"msg":"limit must be a positive integer that is not greater than max: 1000","param":"limit","location":"query"}]}`
`{"errors":[{"value":1001,"msg":"limit must be a positive integer that is not greater than max: 1000","param":"limit","location":"query"}]}`

jusbar23:
@mergify backport release/indexer/v9.5.x

PR #3243, opened 2025-11-05
[ENG-1296] set cap on expired orders processed per block

anmolagrawal345:
### Changelist
This change builds off the same pattern for setting caps on deleveraging and liquidations per block. We need to cap the expired orders removal per block to prevent the situation in which the end blocker is stuck removing a large number of orders (a potential attack vector).

### Test Plan

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * Added block limits configuration system to control the rate of expired stateful order removal per block.
  * Added new gRPC query endpoint and CLI command to retrieve block limits configuration.
  * Added governance message to update block limits configuration.

* **Improvements**
  * Enhanced telemetry to track the count of expired orders removed during block processing.


.github/workflows/protocol-build-and-push.yml
```diff
@@ -6,6 +6,7 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/protocol/v[0-9]+.[0-9]+.x'  # e.g. release/protocol/v0.1.x
       - 'release/protocol/v[0-9]+.x'  # e.g. release/protocol/v1.x
+      - 'anmol/eng-1296'
```
jusbar23, line 9: reminder to remove

protocol/x/clob/types/errors.go
```diff
@@ -348,6 +348,11 @@ var (
 		1022,
 		"Liquidation conflicts with ClobPair status",
 	)
+	ErrInvalidBlockLimitsConfig = errorsmod.Register(
```
jusbar23: Do we use this? Doesn't seem like we have any validation logic

protocol/x/clob/types/block_limits_config.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+// Validate checks that the BlockLimitsConfig is valid.
+// Note: MaxStatefulOrderRemovalsPerBlock can be 0, which means no cap (process all expired orders).
+func (config *BlockLimitsConfig) Validate() error {
+	// No validation needed - 0 is a valid value meaning "no cap"
```
jusbar23: Do we want to set an upper bound for his?

anmolagrawal345: dont think so since its governance protected. if for some reason we need to bump it past any threshold, we'll have to do an upgrade to support it

protocol/x/clob/types/errors.go
anmolagrawal345: deleted - the only validation logic lives in the server itself which verifies the message came from an authorized address. otherwise, think we can just use the value. dont need to check negative values either since its a uint

proto/dydxprotocol/clob/tx.proto
```diff
@@ -215,6 +219,23 @@ message MsgUpdateLiquidationsConfig {
 // MsgUpdateLiquidationsConfig is the Msg/LiquidationsConfig response type.
 message MsgUpdateLiquidationsConfigResponse {}
 
+// MsgUpdateBlockLimitsConfig is a request type for updating the block limits
+// configuration.
+message MsgUpdateBlockLimitsConfig {
```
shrenujb, line 224: Would this be a gov msg? Are there other limits we already have which we use gov msgs to update?

anmolagrawal345: Yep this follows the same pattern as MsgUpdateLiquidationConfig which also seems gov protected.

proto/dydxprotocol/clob/block_limits_config.proto
```diff
@@ -0,0 +1,12 @@
...
+package dydxprotocol.clob;
+
+option go_package = "github.com/dydxprotocol/v4-chain/protocol/x/clob/types";
+
+// BlockLimitsConfig stores global per-block limits for the CLOB module.
+message BlockLimitsConfig {
+  // The maximum number of expired stateful orders that can be removed from
+  // state in a single block. This prevents performance degradation when
+  // processing a large number of expired orders.
+  uint32 max_stateful_order_removals_per_block = 1;
```
northstar456, line 11: Is there a strong reason not to put this under `BlockRateLimitConfiguration`, which is also stored in state? 

Not too concerned about the differences between this field (end blocker) and existing block rate limits (checktx)

Doing that would save us from so many stubs/new messages added in this PR

northstar456, line 11: For my context, why did we decide to add this recently? Is this an AI from a security report?

protocol/x/clob/keeper/stateful_order_state.go
```diff
@@ -220,11 +223,26 @@ func (k Keeper) RemoveExpiredStatefulOrders(ctx sdk.Context, blockTime time.Time
...
+	// Check if there's a cap on order removals per block.
+	// A value of 0 means no cap (process all expired orders).
+	// This is retrieved from on-chain governance-controlled config.
+	blockLimitsConfig := k.GetBlockLimitsConfig(ctx)
+	maxRemovals := blockLimitsConfig.MaxStatefulOrderRemovalsPerBlock
+	hasCap := maxRemovals > 0
+	numRemoved := uint32(0)
+
+	// Process orders up to the cap (if set), otherwise process all
 	for ; it.Valid(); it.Next() {
```
northstar456, line 236: A bit cleaner to put condition in loop guard 
```
	maxRemovals := blockLimitsConfig.MaxStatefulOrderRemovalsPerBlock
	numRemoved := uint32(0)

	// Process orders up to the cap (if set), otherwise process all
	for ; it.Valid() && (maxRemovals == 0 || numRemoved < maxRemovals); it.Next() {
		var orderId types.OrderId
		k.cdc.MustUnmarshal(it.Value(), &orderId)
		expiredOrderIds = append(expiredOrderIds, orderId)
		store.Delete(it.Key())
		numRemoved++
	}
	return expiredOrderIds
}
```

PR #3250, opened 2025-11-19
Fix issue with 30d volume not being removed, and commission not being capped

jusbar23:
### Changelist
* Fix issue with 30d volume not being decremented
* Fix issue with commissions about to go over the cap

### Test Plan
* Test cases

### Author/Reviewer Checklist

## Release Notes

* **New Features**
  * Added affiliate attribution tracking with support for referrer and referee volume attribution in 30-day rolling periods.
  * Enhanced affiliate statistics to include attributed volume metrics alongside existing revenue tracking.
  * Affiliate info queries now return 30-day rolling attributed volume data.

* **Bug Fixes**
  * Fixed affiliate revenue share calculations to properly enforce configured per-user caps across all scenarios.

* **Improvements**
  * Deprecated staked coins requirement in affiliate tier configuration; volume-based metrics now primary focus.

### DIFF
--- a/.github/workflows/indexer-build-and-push-dev-staging.yml
+++ b/.github/workflows/indexer-build-and-push-dev-staging.yml
@@ -6,7 +6,6 @@ on: # yamllint disable-line rule:truthy
       - main
       - 'release/indexer/v[0-9]+.[0-9]+.x'  # e.g. release/indexer/v0.1.x
       - 'release/indexer/v[0-9]+.x'  # e.g. release/indexer/v1.x
-      - 'justin/bug_fix_95'
     # TODO(DEC-837): Customize github build and push to ECR by service with paths
 
 jobs:
--- a/.github/workflows/indexer-lint.yml
+++ b/.github/workflows/indexer-lint.yml
@@ -8,7 +8,6 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/indexer/v[0-9]+.[0-9]+.x'  # e.g. release/indexer/v0.1.x
       - 'release/indexer/v[0-9]+.x'  # e.g. release/indexer/v1.x
-      - 'justin/*'
     paths:
       - 'indexer/**'
 
--- a/.github/workflows/protocol-build-and-push-snapshot.yml
+++ b/.github/workflows/protocol-build-and-push-snapshot.yml
@@ -6,7 +6,6 @@ on:  # yamllint disable-line rule:truthy
       - main
       - 'release/protocol/v[0-9]+.[0-9]+.x'  # e.g. release/protocol/v0.1.x
       - 'release/protocol/v[0-9]+.x'  # e.g. release/protocol/v1.x
-      - 'justin/bug_fix_95'
 
 jobs:
   build-and-push-snapshot-dev:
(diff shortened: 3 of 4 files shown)


proto/dydxprotocol/stats/stats.proto
```diff
@@ -6,6 +6,17 @@ option go_package = "github.com/dydxprotocol/v4-chain/protocol/x/stats/types";
...
 
+// AffiliateAttribution stores the affiliate revenue attributed to a user
+// during match.
+message AffiliateAttribution {
+  // Referrer address (the affiliate receiving the fee)
+  string referrer_address = 1;
+  // Referred volume to attribute in quote quantums
+  uint64 referred_volume_quote_quantums = 2;
+  // Referee address (the trader whose volume is being attributed)
+  string referee_address = 3;
```
northstar456: Remind me - is there affiliate fee share on maker now?

northstar456: This is only ever gonna be taker (and maker?) right. Any way to avoid storing another address?

northstar456: Note we don't need to use exactly this proto (which is for onchain storage) for the applicaiton like `buildAttributableVolumeAttributions`

protocol/x/stats/keeper/keeper.go
```diff
@@ -283,6 +326,8 @@ func (k Keeper) ExpireOldStats(ctx sdk.Context) {
 		stats.TakerNotional -= removedStats.Stats.TakerNotional
 		stats.MakerNotional -= removedStats.Stats.MakerNotional
 		stats.Affiliate_30DRevenueGeneratedQuantums -= removedStats.Stats.Affiliate_30DRevenueGeneratedQuantums
+		stats.Affiliate_30DReferredVolumeQuoteQuantums -= removedStats.Stats.Affiliate_30DReferredVolumeQuoteQuantums
```
northstar456, line 335: Is anything needed to fix the current mainnet state? Eg is anyone's 30d rolling stats consisting of any stale data?

proto/dydxprotocol/stats/stats.proto
jusbar23: No, affiliate fee isn't given on maker

protocol/x/stats/keeper/keeper.go
jusbar23: Theres a few issues

1) Referrees with consistenty high volume before the upgrade aren't getting new volume attributed
2) New referrees after the update will get new volume attributed. After the update this will get moved to the new `Affiliate_30DReferredVolumeQuoteQuantums` the previously attributed volume won't fall of without a forced reset

I think the impact will be minimal for most users, however there will be some corrupted data where some affiliates will have a perpetually higher affiliate tier

proto/dydxprotocol/stats/stats.proto
jusbar23: We can do this if we call GetReferredBy in the `buildAttributableVolumeAttributions` function

We need to do 2 operations

Referred volume (need original address)
Referee attributed volume (need referee address)

I opted to pass it just to avoid the extra state retrieval

protocol/x/stats/keeper/keeper.go
```diff
@@ -238,7 +238,17 @@ func (k Keeper) ProcessBlockStats(ctx sdk.Context) {
...
-				referee := attribution.RefereeAddress
+
+				// Determine referee address based on role
+				var referee string
+				if attribution.Role == types.AffiliateAttribution_ROLE_TAKER {
+					referee = fill.Taker
+				} else if attribution.Role == types.AffiliateAttribution_ROLE_MAKER {
+					referee = fill.Maker
+				} else {
+					// Skip invalid role
```
northstar456: Nit: log error here?

jusbar23: Done

jusbar23:
@mergifyio backport release/protocol/v9.5.x

PR #3261, opened 2025-11-25
9.5 upgrade & add underflow protection for referred volume

jusbar23:
### Changelist
* Add 9.5 upgrade and underflow protection for ExpireOldStats

### Test Plan
* Unit tests

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **New Features**
  * v9.5 upgrade: migrates 30-day referred volumes into epoch-level statistics and registers a v9.5 upgrade handler.

* **Bug Fixes**
  * Expiration logic now guards against numeric underflow when trimming old stats.

* **Tests**
  * Added unit and container-style tests for the v9.5 migration (empty and multi-user scenarios); removed legacy v9.4 container test.

* **Chores**
  * Bumped test/version files to v9.5.

### DIFF
--- a/protocol/app/upgrades/v9.5/upgrade.go
+++ b/protocol/app/upgrades/v9.5/upgrade.go
@@ -3,6 +3,7 @@ package v_9_5
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	upgradetypes "cosmossdk.io/x/upgrade/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -77,16 +78,23 @@ func Migrate30dReferredVolumeToEpochStats(
 		migratedCount++
 
 		ctx.Logger().Info(fmt.Sprintf(
-			"Migrated referred volume for address %s: %d",
+			"Migrated referred volume for address %s (%d of %d): Affiliate_30DReferredVolumeQuoteQuantums=%d",
 			address,
+			migratedCount,
+			len(allAddressesWithReferredVolume),
 			referredVolume,
 		))
 	}
 
-	// Convert map back to slice
+	// Convert map back to slice - must be deterministic to avoid state hash mismatch
+	keys := make([]string, 0, len(userStatsMap))
+	for k := range userStatsMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
 	epochStats.Stats = make([]*statstypes.EpochStats_UserWithStats, 0, len(userStatsMap))
-	for _, userStats := range userStatsMap {
-		epochStats.Stats = append(epochStats.Stats, userStats)
+	for _, k := range keys {
+		epochStats.Stats = append(epochStats.Stats, userStatsMap[k])
 	}
 
 	// Save the updated epoch stats


protocol/x/stats/keeper/keeper.go
```diff
@@ -332,8 +332,20 @@ func (k Keeper) ExpireOldStats(ctx sdk.Context) {
...
+		} else {
+			stats.Affiliate_30DReferredVolumeQuoteQuantums = 0
+		}
+
+		if stats.Affiliate_30DAttributedVolumeQuoteQuantums > removedStats.Stats.Affiliate_30DAttributedVolumeQuoteQuantums {
+			stats.Affiliate_30DAttributedVolumeQuoteQuantums -= removedStats.Stats.Affiliate_30DAttributedVolumeQuoteQuantums
+		} else {
+			stats.Affiliate_30DAttributedVolumeQuoteQuantums = 0
+		}
+
```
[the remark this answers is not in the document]
UnbornAztecKing: @jusbar23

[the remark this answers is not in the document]
jusbar23: done

protocol/app/upgrades/v9.5/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+			// User already in epoch stats, add the referred volume
+			epochUserStats.Stats.Affiliate_30DReferredVolumeQuoteQuantums += referredVolume
+		}
+
+		migratedCount++
+
+		ctx.Logger().Info(fmt.Sprintf(
+			"Migrated referred volume for address %s: %d",
+			address,
+			referredVolume,
```
teddyding, line 85: Nit: also log out `Affiliate_30DReferredVolumeQuoteQuantums`

protocol/app/upgrades/v9.5/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+			userStatsMap[address] = epochUserStats
+		} else {
+			// User already in epoch stats, add the referred volume
+			epochUserStats.Stats.Affiliate_30DReferredVolumeQuoteQuantums += referredVolume
+		}
+
+		migratedCount++
+
+		ctx.Logger().Info(fmt.Sprintf(
+			"Migrated referred volume for address %s: %d",
```
teddyding: Nit: add a string like `${migratedCount} of ${len(allAddressesWithReferredVolume)` as part of string to track progress?

protocol/app/upgrades/v9.5/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+			address,
+			referredVolume,
+		))
+	}
+
+	// Convert map back to slice
+	epochStats.Stats = make([]*statstypes.EpochStats_UserWithStats, 0, len(userStatsMap))
+	for _, userStats := range userStatsMap {
```
teddyding: Critical: do not iterate over map. This results in non-determinism and potentially state hash mismatch.

protocol/app/upgrades/v9.5/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+			// User not in epoch stats yet, create new entry
+			epochUserStats = &statstypes.EpochStats_UserWithStats{
+				User: address,
+				Stats: &statstypes.UserStats{
+					Affiliate_30DReferredVolumeQuoteQuantums: referredVolume,
+				},
+			}
+			userStatsMap[address] = epochUserStats
+		} else {
+			// User already in epoch stats, add the referred volume
```
teddyding, line 74: For my understanding, when does this happen? Do we expect this to happen?

protocol/app/upgrades/v9.5/upgrade.go
jusbar23: Done

protocol/app/upgrades/v9.5/upgrade.go
jusbar23: Done

protocol/app/upgrades/v9.5/upgrade.go
jusbar23: I don't think it should, but just to be safe

protocol/app/upgrades/v9.5/upgrade.go
jusbar23: Done

protocol/app/upgrades/v9.5/upgrade.go
teddyding: If we don't expect this to happening, then shouldn't we:
- Log error explicitly
- Avoid incrementing `referredVolume` since this is unexpected territory and may lead to further state corruption?

jusbar23: No we need to handle this case because the user might have already had stats processed in this epoch due to this value being modified in the stats end blocker. Since the user map is built from the current epoch stat this case is possible and the user should be increment.

jusbar23:
@mergify backport release/protocol/v9.5.x

PR #3293, opened 2025-12-22
fix: fix collision handling in comlink username task

lukabuz:
### Changelist
- Remove transactions from subaccount username generator task to prevent username collisions breaking non-collided inserts
- Fix small test case for subaccount usernames(we now seed 2 subaccount usernames already)

### Test Plan
This was tested locally using the test suite

### Author/Reviewer Checklist

## Summary by CodeRabbit

* **Tests**
  * Added tests for fallback username selection when conflicts occur, mixed-batch scenarios, and switched username lookups to read-replica assertions.

* **Refactor**
  * Simplified subaccount username generation flow with clearer success/failure handling, improved retry/fallback behavior, adjusted transaction handling, and updated logging.

* **Chores**
  * Added an insert-with-conflict-feedback operation to detect whether a username was actually created.

### DIFF
--- a/indexer/services/roundtable/__tests__/tasks/subaccount-username-generator.test.ts
+++ b/indexer/services/roundtable/__tests__/tasks/subaccount-username-generator.test.ts
@@ -120,4 +120,72 @@ describe('subaccount-username-generator', () => {
     expect(conflict.length).toBe(1);
     expect(conflict[0].subaccountId).not.toEqual(targetSubaccount.id);
   });
+
+  it('Handles batch where one username succeeds and the other needs a fallback', async () => {
+    const subaccounts: SubaccountFromDatabase[] = await SubaccountTable.findAll(
+      {
+        subaccountNumber: 0,
+      },
+      [QueryableField.SUBACCOUNT_NUMBER],
+      {},
+    );
+    expect(subaccounts.length).toBeGreaterThanOrEqual(2);
+
+    const sub0 = subaccounts[0];
+    const sub1 = subaccounts[1];
+
+    const { generateUsernameForSubaccount } = require('../../src/helpers/usernames-helper');
+
+    const sub0Attempt0 = generateUsernameForSubaccount(sub0.address, 0, 0);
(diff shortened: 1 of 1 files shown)


indexer/services/roundtable/src/tasks/subaccount-username-generator.ts
```diff
@@ -41,7 +36,7 @@ export default async function runTask(): Promise<void> {
           await SubaccountUsernamesTable.create({
```
jusbar23: I think having a separate transaction per insert can cause a bottleneck since this is alot slower having to do a round trip each time. There is another way we could change the username to do nothing on a conflict

```
INSERT INTO subaccount_usernames (username, subaccount_id) VALUES (...) ON CONFLICT (username) DO NOTHING RETURNING username, subaccount_id;
```

This should return the number of rows inserted. If the number doesnt equal the amount trying to be inserted we know we need to retry. However the rows that do work will be written so we won't need to abort the entire batch

lukabuz: Updated, added a test case for this too

indexer/services/roundtable/__tests__/tasks/subaccount-username-generator.test.ts
```diff
@@ -66,11 +66,58 @@ describe('subaccount-username-generator', () => {
...
+      1,
+    );
+    expect(created[0].username).toEqual(fallbackUsername);
+
+    const conflict = await SubaccountUsernamesTable.findAll(
+      { username: [usernameAttempt0] }, [QueryableField.USERNAME], {},
+    );
+    expect(conflict.length).toBe(1);
+    expect(conflict[0].subaccountId).not.toEqual(targetSubaccount.id);
+  });
```
jusbar23, line 122: Can you add a test case if there is 2 usernames in a batch, one works and the other doesnt?

jusbar23 (approved):

jusbar23:
@mergifyio backport release/indexer/v9.x

PR #3305, opened 2026-01-09
fix(postgres): query master for latest block to avoid replication lag

UnbornAztecKing:
# fix: query master DB for latest block to avoid replication lag

## Summary

- Changed `getLatest()` in `block-table.ts` to default to reading from the master database instead of read replica
- Introduced new `USE_MASTER_POSTGRES_OPTIONS` constant that explicitly disables read replica routing
- Prevents stale data issues when querying for the most recent block in the presence of replication lag

## Details

**Rationale**
- Latest block queries are time-sensitive and particularly vulnerable to replication lag
- Reading stale block data from a replica can cause incorrect indexing behavior or API responses
- Other block queries (e.g., by height) retain the existing replica routing behavior

## Risk & Impact

**Low-to-medium risk:**
- Increases load on master database for `getLatest()` calls; callers relying on this method will now hit master
- If `getLatest()` is called frequently (e.g., in tight loops or high-volume API paths), this could degrade master performance
- No breaking API changes; existing callers can still override the default by passing explicit options
- No data model or migration changes

## Testing

No new tests added; existing coverage reused. The change is a default parameter value swap with no new logic.

Deployed to testnet and internal-mainnet.

### DIFF
--- a/indexer/packages/postgres/src/constants.ts
+++ b/indexer/packages/postgres/src/constants.ts
@@ -125,6 +125,9 @@ export const DEFAULT_POSTGRES_OPTIONS : Options = config.USE_READ_REPLICA
   ? {
     readReplica: true,
   } : {};
+export const USE_MASTER_POSTGRES_OPTIONS : Options = {
+  readReplica: false,
+};
 
 // The maximum number of parent subaccounts per address.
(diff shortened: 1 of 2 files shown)


jusbar23 (approved):

UnbornAztecKing:

PR #3308, opened 2026-01-12
Add safety check for sending builder codes to blocked address

anmolagrawal345:
### Changelist
Add a check in transferBuilderFees which throws an error on blocked address

### Test Plan
Added test to show that a match with a blocked address will fail

### Author/Reviewer Checklist

## Bug Fixes

* Builder fee transfers now validate and reject blocked recipient addresses, preventing transfers to prohibited accounts and strengthening transaction security.

### DIFF
--- a/protocol/app/upgrades/v9.6/upgrade_container_test.go
+++ b/protocol/app/upgrades/v9.6/upgrade_container_test.go
@@ -1,4 +1,5 @@
 //go:build all || container_test
+
 package v_9_6_test
 
 import (


protocol/app/upgrades/v9.6/upgrade_container_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build all || container_test
```
shrenujb, line 1: why is this here?

anmolagrawal345: this was actually causing some test suites to fail. this annotation means this test is run only in certain workflows. This annotation was there before the 9.5 upgrade and then got taken out so just putting it back in

jusbar23 (approved):

anmolagrawal345:

PR #3312, opened 2026-01-15
perf(pnl): add index to oracle_prices for latest prices query used for PnL ticks creation

UnbornAztecKing:
## Summary

- Rewrites the `findLatestPricesBeforeOrAtHeight` query from a subquery pattern to a `LATERAL JOIN` for better performance.
- Adds a covering index on `oracle_prices(marketId, effectiveAtHeight DESC) INCLUDE (price)` to support efficient index-only scans.
- Eliminates the `GROUP BY` / `MAX()` subquery pattern that required full table scans per market.

## Details

**Database Migration**
- New migration `20260115131341_create_index_oracle_prices_market_height_desc_inc_price.ts` creates composite index `oracle_prices_market_height_desc_inc_price`.
- Index covers `marketId`, `effectiveAtHeight DESC` (for ordering), and includes `price` to avoid heap lookups.
- Runs outside a transaction (`transaction: false`) for non-blocking index creation.

**Query Rewrite (`oracle-price-table.ts`)**
- Old approach requires aggregating all rows per market.
- New approach allows the optimizer to use the index and short-circuit after the first matching row per market.

## Risk & Impact

**Medium risk; requires careful rollout:**
- **Performance impact potential**: The new index must be created *before* deploying the query rewrite to avoid performance degradation on large `oracle_prices` tables.
- **Migration timing**: Index creation is blocking and may take time on large datasets.
- **Query semantics**: Behavior should be identical (latest price per market at or before height), but the `LATERAL JOIN` returns all markets from the `markets` table even if no price exists, whereas the old query only returned markets with prices. **Verify this does not change application logic expectations.**
- **Performance**: Expected significant improvement on large datasets; validate with query plan analysis (`EXPLAIN ANALYZE`) in staging.

## Testing

- No new tests added.
- Existing test coverage in `oracle-price-table.ts` implicitly validates correctness of the refactored query.
- Migration includes a `down()` rollback path.
- Deployed to testnet and internal mainnet.

## Summary by CodeRabbit

* **Chores**
  * Enhanced database query performance for oracle price lookups through optimized indexing and improved data retrieval patterns.

### DIFF
--- a/indexer/packages/postgres/src/stores/oracle-price-table.ts
+++ b/indexer/packages/postgres/src/stores/oracle-price-table.ts
@@ -230,10 +230,10 @@ export async function findLatestPricesBeforeOrAtHeight(
     FROM "markets" m
     JOIN LATERAL (
       SELECT "price"
-      FROM "oracle_prices" op
-      WHERE op."marketId" = m."id"
-      AND op."effectiveAtHeight" <= $1
-      ORDER BY op."effectiveAtHeight" DESC
+      FROM "oracle_prices" p
+      WHERE p."marketId" = m."id"
+      AND p."effectiveAtHeight" <= ?::bigint
+      ORDER BY p."effectiveAtHeight" DESC
       LIMIT 1
     ) op ON true;
   `;


UnbornAztecKing:

PR #3378, opened 2026-08-18
fix(lib): keep the fraction when rounding a big.Rat to a multiple

pucedoteth:
### Changelist

`lib.BigRatRoundToMultiple` rounds its `big.Rat` down to a `big.Int` *before* looking at the remainder:

```go
valueInt := new(big.Int).Div(value.Num(), value.Denom())
remainder := new(big.Int).Mod(valueInt, multiple)

if roundUp && remainder.Sign() != 0 {
    valueInt.Add(valueInt, new(big.Int).Sub(multiple, remainder))
} else {
    valueInt.Sub(valueInt, remainder)
}
```

The fractional part is discarded before it can affect the rounding decision. When the floor of the value already sits on a multiple, the remainder is zero, the round-up branch is skipped, and the function returns the floor — a full multiple below the correct answer:

| call | returns | expected |
|---|---|---|
| `BigRatRoundToMultiple(1000.5, 10, roundUp=true)` | `1000` | `1010` |
| `BigRatRoundToMultiple(0.5, 10, roundUp=true)` | `0` | `10` |
| `BigRatRoundToMultiple(10.5, 1, roundUp=true)` | `10` | `11` |

The `multiple = 1` case is the clearest statement of the problem: every integer is a multiple of 1, so the round-up branch is unreachable and the function is just `floor`.

Rounding **down** was already correct — the `else` branch floors to the multiple, and negative values work out because `Div`/`Mod` are Euclidean. Only round-up is affected.

The fix rounds `value / multiple` in rational space and scales back up, reusing the existing `BigRatRound`, so the fraction takes part in the decision.

#### Impact

The only caller is `calculateSuborderSubticks` in `x/clob/keeper/twap_order_state.go`, which rounds a TWAP suborder's limit price with:

```go
priceTolerancePpm >= 0, // round up for positive adjustments, down for negative
```

Buys round up and sells round down so the suborder stays marketable. When a buy's adjusted oracle price lands just above a tick boundary — i.e. `floor(adjustedPrice)` is a multiple of `SubticksPerTick` — the price is rounded *down* instead, one full tick below the intended limit, making the suborder correspondingly less likely to fill. Sells are unaffected.

I suspect this survived because `BigRatRoundToMultiple` had no test coverage at all, unlike its sibling `BigIntRoundToMultiple`.

### Test Plan

Added `TestBigRatRoundToMultiple` to `protocol/lib/quantums_test.go` — 13 cases in the same table-driven style as `TestBigIntRoundToMultiple`, covering round-up/round-down, exact multiples, `multiple = 1`, negatives, and zero.

Three of them fail on `main`:

```
--- FAIL: TestBigRatRoundToMultiple/Rounds_up_when_the_value's_floor_is_already_a_multiple
        expected: 1010   actual: 1000
--- FAIL: TestBigRatRoundToMultiple/Rounds_up_to_the_next_integer_when_the_multiple_is_one
        expected: 11     actual: 10
--- FAIL: TestBigRatRoundToMultiple/Rounds_a_value_below_the_multiple_up_to_the_multiple
        expected: 10     actual: 0
```

Run with the pinned toolchain (`GOTOOLCHAIN=go1.23.3`):

| check | result |
|---|---|
| `go test ./lib/...` | all packages **ok** |
| `go test ./x/clob/keeper/` | **ok** (22.7s) — no existing test encoded the old behaviour |
| `go vet ./lib/` | clean |
| `gofmt -l lib/` | clean |

One note in case it comes up: on Go 1.26 the pre-existing `TestBigDivCeil` fails two subtests with `expected: 0, actual: 0`, from a change in `big.Int`'s internal representation (`abs: nil` vs `[]Word{}`). It reproduces on a clean checkout of `main` and disappears under the repo's pinned Go 1.23.3, so it is unrelated to this change.

### Author/Reviewer Checklist
  - **This is state-breaking.** TWAP buy suborders that hit the affected case will be placed at a different limit price than before, so it needs the label and to go out with an upgrade rather than as a patch to running nodes. I can't add labels on this repo — flagging it here.
- [x] Manually add any of the following labels: `refactor`, `chore`, `bug`. → `bug`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rounding of fractional values to the nearest configured multiple.
  * Corrected rounding behavior for positive, negative, zero, and exact values.
  * Added validation to reject zero or negative rounding multiples.

* **Tests**
  * Added comprehensive coverage for multiple-based rounding scenarios.
  * Added tests confirming invalid multiples are rejected.

### DIFF
--- a/protocol/lib/quantums.go
+++ b/protocol/lib/quantums.go
@@ -91,22 +91,17 @@ func QuoteToBaseQuantums(
 
 // BigRatRoundToMultiple rounds a big.Rat to the nearest multiple of the given number.
 // If roundUp is true, it rounds up to the nearest multiple, otherwise it rounds down.
+// This function expects `multiple` to be positive.
 func BigRatRoundToMultiple(
 	value *big.Rat,
 	multiple *big.Int,
 	roundUp bool,
 ) *big.Int {
-	// Convert the value to a big.Int
-	valueInt := new(big.Int).Div(value.Num(), value.Denom())
-
-	// Calculate the remainder
-	remainder := new(big.Int).Mod(valueInt, multiple)
-
-	if roundUp && remainder.Sign() != 0 {
-		valueInt.Add(valueInt, new(big.Int).Sub(multiple, remainder))
-	} else {
-		valueInt.Sub(valueInt, remainder)
-	}
-
-	return valueInt
+	// Round `value / multiple` to an integer and scale it back up, so that the
+	// fractional part of `value` takes part in the rounding decision. Rounding
+	// `value` to an integer first would discard it, and a value such as `10.5`
+	// would then never round up past the multiple its floor already sits on.
+	quotient := new(big.Rat).Quo(value, new(big.Rat).SetInt(multiple))
+	rounded := BigRatRound(quotient, roundUp)
+	return rounded.Mul(rounded, multiple)
 }
--- a/protocol/lib/quantums.go
+++ b/protocol/lib/quantums.go
@@ -91,12 +91,18 @@ func QuoteToBaseQuantums(
 
 // BigRatRoundToMultiple rounds a big.Rat to the nearest multiple of the given number.
 // If roundUp is true, it rounds up to the nearest multiple, otherwise it rounds down.
-// This function expects `multiple` to be positive.
+// This function always expects the `multiple` parameter to be positive, otherwise it will panic.
 func BigRatRoundToMultiple(
 	value *big.Rat,
 	multiple *big.Int,
 	roundUp bool,
 ) *big.Int {
+	// A zero multiple divides by zero and a negative one silently reverses the
+	// rounding direction, so reject both up front like BigIntRoundToMultiple does.
+	if multiple.Sign() <= 0 {
+		panic("BigRatRoundToMultiple: multiple must be positive")
+	}
+
 	// Round `value / multiple` to an integer and scale it back up, so that the
 	// fractional part of `value` takes part in the rounding decision. Rounding
 	// `value` to an integer first would discard it, and a value such as `10.5`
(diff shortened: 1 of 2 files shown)


pucedoteth:
Good catch — confirmed and fixed in 2ecc95c.

I verified the claim before acting on it. A negative multiple really does reverse the rounding direction:

```
BigRatRoundToMultiple(10.5, -10, roundUp=true)  = 10   // rounded down
BigRatRoundToMultiple(10.5, -10, roundUp=false) = 20   // rounded up
```

and a zero multiple panics with a bare `division by zero` from inside `big.Rat.Quo`.

The fix mirrors `BigIntRoundToMultiple`, which already guards exactly this contract:

```go
if multiple.Sign() <= 0 {
    panic("BigRatRoundToMultiple: multiple must be positive")
}
```

Added the two table cases as suggested (zero and negative), with a `shouldPanic` field matching the style `TestBigIntRoundToMultiple` already uses.

`GOTOOLCHAIN=go1.23.3 go test ./lib/` — **ok**, 15 subtests in `TestBigRatRoundToMultiple` passing. `go vet ./lib/` and `gofmt -l lib/` clean.

I left `nil` alone: `multiple.Sign()` nil-derefs the same way `BigIntRoundToMultiple` does, so guarding it here only would make the two diverge.

pucedoteth:
This was addressed in 2ecc95c: `BigRatRoundToMultiple` now rejects a non-positive `multiple` up front

```go
if multiple.Sign() <= 0 {
    panic("BigRatRoundToMultiple: multiple must be positive")
}
```

and `TestBigRatRoundToMultiple` has table cases for `multiple: 0` and `multiple: -10` asserting the panic. This matches `BigIntRoundToMultiple`, which panics on the same input.

pucedoteth:
CI has never actually run on this PR.

The workflow runs from my fork have been sitting at the "Approve and run workflows" step since I opened it, and GitHub expired them at the 30-day mark. The red ✗ marks dated Aug 18 and Aug 19 are that timeout, not test failures — every one of those runs created zero jobs, so there's no build, lint or test result anywhere on this PR to review.

Could a maintainer approve the workflows so `Protocol Build & Test`, `Protocol Lint` and `Protocol Sims` can report? Happy to rebase onto current `main` first if that's easier.

PR #1318, opened 2024-04-02
Skip equity tier limit check in PlaceShortTermOrder

roy-dydx:
### Changelist
These checks are not effective for their original purpose and take 30% of the CheckTx global lock time in testnet loads.

### Test Plan
image

### Author/Reviewer Checklist
- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label.
- [ ] If the PR has breaking postgres changes to the indexer add the `indexer-postgres-breaking` label.
- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`.
- [ ] If this PR is one of many that implement a specific feature, manually label them all `feature:[feature-name]`.
- [ ] If you wish to for mergify-bot to automatically create a PR to backport your change to a release branch, manually add the label `backport/[branch-name]`.
- [ ] Manually add any of the following labels: `refactor`, `chore`, `bug`.

PR #3391, opened 2026-09-11
fix(clob): enforce short-term order equity tier in CheckTx

tqin7:
### Changelist
- Enforce `ShortTermOrderEquityTiers` in `PlaceShortTermOrder` (CheckTx only) before the memclob insert. A subaccount whose net collateral falls in a tier with `Limit == 0` is rejected, so the tx fails CheckTx and is not gossiped. Short-term placement has had no collateral requirement since #1318 removed the check for CheckTx cost.
- Reduce-only orders are exempt so a subaccount below the floor can still close its position. The memclob already rejects reduce-only orders with no opposing position before matching, so this is not a bypass for position-less subaccounts.
- The result is memoized per subaccount for the block in the keeper and cleared in `PrepareCheckState`. Within a block only optimistic fills can change net collateral, so the first order per subaccount per block pays the net-collateral read and the rest pay a map lookup.
- New counter `clob.place_order.rejected.equity_tier_limit`.
- Activates on upgrade: mainnet, testnet, and pregenesis genesis already configure short-term tiers with a $20 floor. Tunable via `MsgUpdateEquityTierLimitConfiguration`.
- CheckTx-only, so not consensus-breaking and no upgrade handler; placement only, cancels are never gated.

### Test Plan
- New e2e cases in `TestPlaceShortTermOrder_EquityTierLimit`: resting and IOC orders rejected below the floor, reduce-only order passes the gate and fails the reduce-only check instead, order allowed across a multi-tier boundary, unset ladder is a no-op.
- Keeper unit test: a memoized result holds across a same-block balance change and resets after `ClearShortTermEquityTierResults`.
- `go test ./x/clob/... ./app/...` green, including the full `x/clob/e2e` suite.
- Keeper benchmark (`BenchmarkPlaceShortTermOrder`, `BenchmarkValidateSubaccountEquityTierLimitForShortTermOrder`), non-crossing order, in-memory store: placement ~14us with the gate off or warm; cold check ~3us with no positions, ~9us with two, ~53us with twenty; warm check ~60ns.

### Author/Reviewer Checklist
- [ ] If this PR has changes that result in a different app state given the same prior state and transaction list, manually add the `state-breaking` label. (Not state-breaking: CheckTx-only.)
- [ ] If this PR isn't state-breaking but has changes that modify behavior in `PrepareProposal` or `ProcessProposal`, manually add the label `proposal-breaking`. (Neither path is touched.)


tqin7:
Superseded by #3392, which carries the identical patch against `main` and will be backported to `release/protocol/v9.x` via Mergify after merge. The canary image built from this PR's merge ref (`mainnet-full-node:e242ad8`) stays as-is.

PR #3392, opened 2026-09-14
fix(clob): enforce short-term order equity tier in CheckTx

tqin7:
- Enforce `ShortTermOrderEquityTiers` in `PlaceShortTermOrder` (CheckTx only) before the memclob insert. A subaccount whose net collateral falls in a tier with `Limit == 0` is rejected
- The result is memoized per subaccount for the block and cleared in `PrepareCheckState`
- New counter `clob.place_order.rejected.equity_tier_limit`.
- Activates on upgrade: mainnet, testnet, and pregenesis genesis already configure short-term tiers with a $20 floor. Tunable via `MsgUpdateEquityTierLimitConfiguration`.
- CheckTx-only

### Testing
- unit and e2e
- benchmark (`BenchmarkPlaceShortTermOrder`, `BenchmarkValidateSubaccountEquityTierLimitForShortTermOrder`), non-crossing order, in-memory store: placement ~14us with the gate off or warm; cold check ~3us with no positions, ~9us with two, ~53us with twenty; warm check ~60ns.
- been running this on a full node for the past few days

## Summary by CodeRabbit

- **New Features**
  - Short-term orders now respect configured equity-tier limits based on a subaccount’s net collateral.
  - Orders below the applicable minimum may be rejected before placement.
  - Reduce-only orders remain eligible even when equity-tier limits would otherwise prevent new exposure.
  - Orders at or above the configured threshold continue to be accepted, and short-term orders remain available when no equity-tier limits are configured.
- **Bug Fixes**
  - Equity-tier checks are refreshed between blocks to ensure decisions reflect current account conditions.


Kefancao (approved):

tqin7:

